PR #42, opened 2019-11-14
Fix/peter.somogyvari/quorum contracts syntax

petermetz:


denis-yu-glotov:
Peter, did you moved 2 contracts here? Or what was it?

petermetz:
> Peter, did you moved 2 contracts here? Or what was it?

Yes, also made the import file paths relative, wouldn't compile without it.

petermetz:
Original issue: https://github.com/hyperledger-labs/blockchain-integration-framework/issues/37

denis-yu-glotov:
LGTM then, I wonder how it was working without them..

petermetz:
> LGTM then, I wonder how it was working without them..

Thank you! I agree, I'm pretty something else is at play as well. Also would prefer to not duplicate the contracts between the main package's sources and the examples sources. The example - whenever possible - should re-use the code of the main package instead of copy pasting it. 
For the latter, I left a note for ourselves for a refactor:
https://github.com/hyperledger-labs/blockchain-integration-framework/pull/42/commits/842a25a00362973bff4bac33dd732729f7dca31a

PR #50, opened 2019-11-15
Fix CI Flakiness/build issues

petermetz:
Fixes the CI.
test:bc is still not working but the flakiness is gone now.


jonathan-m-hamilton:
LGTM, will let @denis-yu-glotov review before merging (believe he has a bank holiday today)

examples/simple-asset-transfer/quorum/api/package.json
```diff
@@ -11,7 +11,7 @@
     "start": "node index.js",
     "upgrade-interactive": "npm-check --update",
     "test": "cross-env NODE_ENV=test ./node_modules/.bin/mocha --ui bdd --reporter spec --colors routes/**/*.test.js --recursive --exit",
-    "postinstall": "patch ./node_modules/web3-core-helpers/src/formatters.js ./web3_timestamp_fix.patch"
+    "postinstall": "patch --verbose --force ./node_modules/web3-core-helpers/src/formatters.js ./web3_timestamp_fix.patch; echo Line 266 of formatters.js below:; sed '266!d' ./node_modules/web3-core-helpers/src/formatters.js"
```
denis-yu-glotov: Do we want to scare users with these messages?

petermetz: Could argue for and against it as well. (awareness > scared, UX > awareness)
The `--force` is necessary because without it the patch application fails if patch was already applied (didn't want to force a zero exit code to avoid confusing silent failures).
`--verbose` and the dumping of the source code line are there to make sure it's easy to notice/investigate issues, BUT I can refactor it in a way that it's part of a checklist in the static documentation that whenever a dependency is updated the developer must ensure the monkey patching is still working. Will update it in a bit

denis-yu-glotov: No, I meant only about what you `echo`.

petermetz: I'll keep `--verbose` then, will document how to debug instead of echoing the code.

PR #66, opened 2019-11-26
fix(quorum/api): use Web3 HTTP Provider by default

petermetz:
It is deprecated because it does not support
subscriptions, but we do not use that feature anyway
and just need the examples to work so it seems like
an okay trade-off to use HTTP instead of WS.


petermetz:
@denis-yu-glotov  FYI: CI won't pass on this until we merge the other fixes in.

examples/simple-asset-transfer/quorum/api/utils/web3.js
```diff
@@ -1,9 +1,8 @@
 const Web3 = require('web3');
 const config = require('../config/config');
 
-const option = `ws://${config.web3.host[`${process.platform}`]}:${config.web3.port}`;
-const web3 = new Web3(option);
-
+const rpcApiHost = `http://${config.web3.host[`${process.platform}`]}:${config.web3.rpcPort}`;
```
denis-yu-glotov: need to add this `rpcPort` to config. Or why not just leave it as `port`?

petermetz: > need to add this `rpcPort` to config. Or why not just leave it as `port`?

I wanted to keep it separate from the WS API / port. Naming is a little confusing because in the end both HTTP and the WS are using the same JSON RPC API under the hood just with different transport protocols (HTTP, WS).

examples/simple-asset-transfer/quorum/platform/docker-compose.yml
```diff
@@ -54,6 +54,12 @@ x-quorum-def:
         --nodiscover \
         --verbosity 5 \
         --networkid $${NETWORK_ID} \
+        --rpc \
+        --rpcaddr=0.0.0.0 \
+        --rpcvhosts=node1 \
```
denis-yu-glotov: I use "*" here, same as in cors. Will it work with `node1` value?

examples/simple-asset-transfer/quorum/api/utils/web3.js
denis-yu-glotov: ok, then you need to add this `rpcPort` to `config/config`, no?

petermetz:
> ok, then you need to add this rpcPort to config/config, no?

Ooops, yes. :-) 
https://github.com/hyperledger-labs/blockchain-integration-framework/pull/66/commits/fac4060712eaabd075be9bfe61fb3074baeac283

examples/simple-asset-transfer/quorum/platform/docker-compose.yml
petermetz: Thank you!
`node1`: docker-compose guarantees that the host will resolve to the container's IP based on what you name the container in the compose .yml file (also tested it so I know for a fact that it works).
Could've put `*` but I figured if I can make it work with tighter controls then might as well use the least privilege best practice.

PR #69, opened 2019-11-28
Quorum platform upgrade to v2.3.0

denis-yu-glotov:
This upgrade fixes the problem where we could not make transaction in Quorum network.
The upgraded files are taken as-is from https://github.com/jpmorganchase/quorum-examples/tree/master/examples/7nodes by the first commit (except path `examples/7nodes` became `.`). The second commit reduced the number of nodes to 4, as we had before, to reduce resources footprint and run faster on small VMs.
I decided to remove all unnecessary files because we have not tested them and cannot guarantee their work, but please share your opinions on this.


petermetz:
LGTM

PR #70, opened 2019-12-02
Feat/peter.somogyvari/ci e2e tests

petermetz:
This is the narrowed down, compacted version of the draft PR achieving the same thing: https://github.com/hyperledger-labs/blockchain-integration-framework/pull/62

Needs #31 to be merged first!


tools/ci.sh
```diff
@@ -2,41 +2,153 @@
...
+  while docker ps | grep "starting\|unhealthy"; do
+    if [ "$iterationCount" -gt 20 ]; then
+      false;
+    fi
+    iterationCount=$[$iterationCount +1]
+    sleep 15; echo; date;
+  done
+
+  docker ps -a
+  sleep 120
```
denis-yu-glotov: 12 looks enough :)

tools/ci.sh
```diff
@@ -2,41 +2,153 @@
...
+
+  dumpAllLogs
+
+  endedAt=`date +%s`
+  runtime=$((endedAt-startedAt))
+  echo "$(date -Iseconds) [CI] FAILURE - runtime=$runtime seconds."
+  exit 1
+}
+
+function dumpAllLogs()
```
denis-yu-glotov: you can use `after_script` step in .travis.yml for this. 
In fact, I believe this script should stay simple, easy to read and modify. Otherwise this script will soon need tests for itself. This dump logic could be moved to another script and be called from `after_script` step that runs if both script succeeded or failed.

tools/ci.sh
```diff
@@ -2,41 +2,153 @@
...
+  npm run test
+
+  cd examples/simple-asset-transfer
+  npm run fed:quorum:down
+  npm run fabric:down
+  npm run quorum:down
+  npm run quorum:api:down
+
+  npm install
+  rm -rf ./node_modules/websocket/.git/
```
denis-yu-glotov: do you still need this? same below

tools/ci.sh
```diff
@@ -2,41 +2,153 @@
...
+
+  ### FABRIC
+  cd ./fabric/api/
+  rm -rf ./fabric-client-kv-org*
```
denis-yu-glotov: this command exists in `runAll.sh`. Why have its copy here?

examples/simple-asset-transfer/quorum/platform/docker-compose.yml
```diff
@@ -209,7 +209,7 @@ services:
       - PRIVATE_CONFIG=/qdata/tm/tm.ipc
       - NODE_ID=1
     networks:
-      quorum-examples-net:
+      quorum_quorum-examples-net:
```
denis-yu-glotov: why change it here? After upgrading Quorum, one will need to re-do this change again. I would leave this config as close to original as possible. Instead you can update this network name in 'quorum/api/docker-compose.yaml`.

tools/ci.sh
petermetz: > 12 looks enough :)

It did to me as well, until it wasn't when I tested it on weaker hardware. There's probably a slightly lower value for delay that still works most of the time, but this was much quicker for now and given that it is a band-aid regardless of how long the wait is, I figured it's not worth doing a binary search for the "optimal" delay time. Instead what I focused on is having the integration points (mostly the REST API calls and the Web3 transactions) implemented in a more resilient fashion which - if done right - then eliminates the need for any hacky delays completely. This person says explains it much better: https://medium.com/@kelseyhightower/12-fractured-apps-1080c73d481c

tools/ci.sh
petermetz: I wouldn't mind putting it in a different file, but using the travis-only features breaks down the idea that I can run the exact same CI locally that will run remotely on my branch once I open the PR. Tried looking for some tooling from Travis that allows me to "execute" the travis.yml the same way they do, but this is not something they opened up to the public (I guess it makes sense, a chunk of their value add would be given away that way).

tools/ci.sh
petermetz: ✅ 32864944bed098a1573df188376095f1067941e8

tools/ci.sh
petermetz: ✅ 32864944bed098a1573df188376095f1067941e8

examples/simple-asset-transfer/quorum/platform/docker-compose.yml
petermetz: ✅ f5fb752887e03ba3b29c9364bc4dfad577f5f594

tools/ci.sh
petermetz: ✅ 32864944bed098a1573df188376095f1067941e8

PR #72, opened 2019-12-03
Add scenarios to CI

denis-yu-glotov:
Work in progress


packages/core/tools/ci.sh
```diff
@@ -116,10 +132,10 @@ function onTaskFailure()
 function dumpAllLogs()
 {
   set +e # do not crash process upon individual command failures
-  ORIGINAL_PWD=$PWD
-  cd "$PKG_ROOT_DIR" # switch back to the original root dir because we don't know where exactly the script crashed
-  ./tools/dump-all-logs.sh $CI_ROOT_DIR
-  cd "$ORIGINAL_PWD"
+  cd "$PKG_ROOT_DIR" # switch back to the original root dir because we don't
+                     # know where exactly the script crashed
+  [ "$CI_NO_DUMP_ALL_LOGS" ] || ./tools/dump-all-logs.sh $CI_ROOT_DIR
```
petermetz: Can we add back the call to dumpAllLogs now that this check is in place?
Right now the function is not being called at all.

denis-yu-glotov: sure thing :)

PR #75, opened 2019-12-05
fix(core): http resiliency

petermetz:
# WORK IN PROGRESS

The end goal is to not need the CI script to synchronize infrastructure provisioning with intentional delays (`sleep 120`) because we made the code resilient enough so that if a connection to a server is down then it just retries with an exponential back off for a while before really failing.


petermetz:
This is no longer relevant with the new CI script and folder structure so closing.

PR #96, opened 2020-01-02
Script to add missing OS deps to run ci

### LINKED ISSUE #67: Undocumented OS Level Dependencies
### Description
As a user I want to see all the OS level dependencies in the documentation up front so that I don't have to do trial and error when running the examples.

For exmaple: While building Corda I came upon the following undocumented dependencies:

1. Corda needs Java 8 (===): https://github.com/corda/cordapp-example/issues/142

2. Also needed `sudo apt install python-minimal build-essential`

### Acceptance Criteria
1. Documentation is updated to have dependencies set up.
2. Added `./tools/configure.sh` script that turns a completely fresh Ubuntu 18 installation into a machine that can execute `./tools/ci.sh` without having any issues. Could also call the configure script `boostrap.sh` instead, up to the author.

denis-yu-glotov:
This fixes #67.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,11 @@
-language: node_js
-node_js:
-  - 8
+language: minimal
 
 os: linux
 
 dist: bionic
 
-services:
-  - docker
+install:
+  - ./packages/core/tools/configure.sh
 
 script:
   - CI_CONTAINERS_WAIT_TIME=12 CI_NO_CORDA=true ./packages/core/tools/ci.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
-language: minimal
+language: node_js
+node_js:
+  - 8
 
 os: linux
 
 dist: bionic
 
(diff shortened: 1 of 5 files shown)


denis-yu-glotov:
In this PR I modified `.travis.yml` file so it simulates minimal Ubuntu installation and uses supplied `configure.sh` script to install prerequisites. I doubt we should really modify it this way: this adds additional time delay for every CI run.

petermetz:
> I doubt we should really modify it this way: this adds additional time delay for every CI run.

@denis-yu-glotov Maybe we should avoid using the configure script in the CI environment to make it go faster? LGTM either way.

My intention with the `configure.sh` script is to make it easy for first time users to set up a virtual machine for their own development environment so it's okay if the CI does not use the configure.sh script as far as I'm concerned.
Later I'm also planning on adding a Vagrantfile that will pull up a VM and execute the configure.sh script or the equivalent of it through Ansible.

PR #104, opened 2020-02-05
fix(ci): lock mkdirp to 1.0.2

petermetz:
Version 1.0.3 broke the build because it insists on NodeJS 10 or higher.

The part in the lock file is what forces the transitive dependencies (web3)
to use the older version which still allows 1.0.2

    "mkdirp": {
      "version": "1.0.2",
      "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-1.0.2.tgz",
      "integrity": "sha512-N2REVrJ/X/jGPfit2d7zea2J1pf7EAR5chIUcfHffAZ7gmlam5U65sAm76+o4ntQbSRdTjYf7qZz3chuHlwXEA=="
    },


jonathan-m-hamilton:
LGTM

PR #115, opened 2020-02-21
fix(npm-deps): adds lock files where not an issue

petermetz:
There's one lock file that's an exception for now:
packages/core/examples/simple-asset-transfer/package-lock.json
And the reason for it is that we install a dependency from a file packaged
by npm locally so that it's always the latest code directly from the
project folder that gets installed. The latter is necessary so that
developers can easily test code changes they made by running
the CI script.


jonathan-m-hamilton:
LGTM!

PR #117, opened 2020-02-28
Containerize Fabric API Server

### LINKED ISSUE #107: Test suite fails on OSX
### Description
Due to networking differences under OSX (as described [here](https://docs.docker.com/docker-for-mac/networking/)) the Fabric containers are unable to communicate with the API server running on the host via the bridge ip `172.20.0.1`.

The solution is to use the `host.docker.internal` hostname which resolves to the host ip, however linux support for this has only just landed in https://github.com/moby/moby/pull/40007.

Once a new version containing the above PR has been released we should use the above hostname instead of `172.20.0.1` when accessing the host.

### Acceptance Criteria
Test suite runs successfully on OSX and Linux

blefevre:
This PR moves the Fabric API server into a container.

This resolves #107 as the Fabric containers no longer need access to the server running on the host.


blefevre:
CircleCI is passing: https://circleci.com/gh/hyperledger-labs/blockchain-integration-framework/54

PR #118, opened 2020-03-05
docs(examples): added Fabric API image build

petermetz:
This was missing from the documentation since we containerized
the Fabric API server last week.


jonathan-m-hamilton:
LGTM

PR #121, opened 2020-03-16
Fix typo in White Paper

etaroid:


etaroid:
> @etaroid Looks good, but please make the commit message compliant with the linter's requirements. Something like this should do as a commit message and then we are good to go:
> `docs(whitepaper): fix typo in use case`

Thanks for the kind comment! 
I've fixed it.

petermetz:
@etaroid Awesome, one last thing before I can approve is to sign off on your commit. It's a convention that we must keep for everyone who commits code for legal reasons.

If you already have your git signature set up then you can just run the commands shown here: https://github.com/hyperledger-labs/blockchain-integration-framework/pull/121/checks?check_run_id=512630773

If you have never done git signing before then start with these first to provision the PKI necessary for it:
https://help.github.com/en/github/authenticating-to-github/telling-git-about-your-signing-key
https://help.github.com/en/github/authenticating-to-github/signing-commits

etaroid:
> @etaroid Awesome, one last thing before I can approve is to sign off on your commit. It's a convention that we must keep for everyone who commits code for legal reasons.
> 
> If you already have your git signature set up then you can just run the commands shown here: https://github.com/hyperledger-labs/blockchain-integration-framework/pull/121/checks?check_run_id=512630773
> 
> If you have never done git signing before then start with these first to provision the PKI necessary for it:
> https://help.github.com/en/github/authenticating-to-github/telling-git-about-your-signing-key
> https://help.github.com/en/github/authenticating-to-github/signing-commits

Thanks! I've signed my commit off.

PR #130, opened 2020-05-13
docs: whitepaper update

RafaelAPB:
This pull request adds the blockchain migration feature to the whitepaper


RafaelAPB:
@petermetz Thank you for your input, Peter.
@sfuji822 Shingo, thank you for your feedback.

I kept the description general, similarly to the rest of the use cases presented on the whitepaper, so we can adapt our use case accordingly to the evolution of the implementation. 

I believe it is important to include this use case, as it is a novel use case that provides blockchain portability, thus reducing costs and fostering blockchain adoption. Thus, I've incorporated your feedback into the whitepaper, and provided a sequence diagram to clarify the procedure.

I'm welcoming any feedback and suggestions to improve this use case and include it on the whitepaper.

RafaelAPB:
@sfuji822 Thank you for your input. Generally, I agree with your commit.

However, as a note: I believe one could artificially recreate the history of an asset (for example, by attaching metadata to a certain state). For instance, if we were migrating Fabric-Fabric (an hypothetical, unlikely scenario), that would mean that for each key in the world state, the migrated information corresponding to each key would have an additional "previous_history" field. As the migration is supposed to be conducted by a consortium, there is less incentive for collusion regarding the history of an asset, although one cannot be sure. Such doubts are also taken regarding the assets that are migrated: collusion can always happen if the consortium members agree, although unlikely.

RafaelAPB:
@sfuji822 @petermetz @jonathan-m-hamilton any updates on this PR?

petermetz:
> @sfuji822 @petermetz @jonathan-m-hamilton any updates on this PR?

Nothing further from me.

petermetz:
> OK. It is more important to go forward. We can figure out how such history records can be migrate securely in later stage.

@sfuji822 Agreed! Let's hope @RafaelAPB will help out with the detailed design as well (which must include the details about ensuring security and how to achieve it by default as per our design principle to be `secure by default`). :-)

petermetz:
Thank you for the contribution @RafaelAPB !

RafaelAPB:
@sfuji822 and @petermetz surely! It will be much more clear as soon as we develop the consortium management module.

PR #145, opened 2020-05-22
Feature/open id connect web service plugin

petermetz:
Note: there's also smaller improvements to CI/build speed, developer experience, not just the OIDC integration.

- [ ] Persistence layer for users/tokens/sessions/etc. - Knex config parameters and factory
- [x] HTTPS support without having to run a reverse proxy, load balancer (this is a chore, but necessary if we want to keep it simple to try out the software)


petermetz:
Will open another one a little later today/tomorrow

Issue #139, opened 2020-05-19
perf: enable Webpack Persistent Caching for Builds


petermetz:
**Is your feature request related to a problem? Please describe.**
Builds are slow. Performance is a feature. ;-)

**Describe the solution you'd like**
Faster builds. (`npm run build`)

**Describe alternatives you've considered**
Also looking into tools that monitor the file system for changes, but it's hard to integrate those into our build pipeline directly because they aren't specific to Javascript/NodeJS ES modules.

**Additional context**

These would be really nice to leverage:
1. https://github.com/webpack/changelog-v5/blob/master/guides/persistent-caching.md
2. Only generate bundle size analytics report when flag is present (e.g. make it opt-in vs. the current default ON)

takeutak:
I'm sure that development milestones (such as "v0.2.0") will be discussed at the Cactus meeting.  Could you mind if you delete such commits from this PR?
 * https://github.com/hyperledger/cactus/projects/1#card-38540731

petermetz:
> I'm sure that development milestones (such as "v0.2.0") will be discussed at the Cactus meeting. Could you mind if you delete such commits from this PR?
> 
> * https://github.com/hyperledger/cactus/projects/1#card-38540731

@takeutak Sure, removed it. 👍

petermetz:
We plan to switch out webpack to either esbuild or ncc or a combination of them so this is no longer relevant.

PR #148, opened 2020-05-24
Chore/housekeeping prettier ignore ci and build fixes

petermetz:
This is just a bunch of house keeping chores around the build system, the CI and the prettier ignore config which had to be updated to include looking the other way for the contribs/* folder which used to be under the packages/core and packages/connection-chain directories.

Also improves the build speed and stability, adds more build scripts which may seem overwhelming but it allows faster developer iteration by way of providing the choice to only build what you need to/modified the source of.

All of these commits were made as side effects/necessary fixes while me working on the Open ID Connect authentication layer but I didn't want to conflate the eventual PR of that (OIDC) with the housekeeping chores which should make it into master sooner rather than later so everyone else can also benefit from the faster builds for example.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,3 +6,4 @@
 **/generated/
 packages/core
 packages/connection-chain
+contribs/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
   - 10
+  - 12
+  - 14
 
 os: linux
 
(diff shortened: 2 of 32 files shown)


petermetz:
Additional note: merging this will take care of #139

petermetz:
Thanks @sfuji822 just one issue: you squashed the commits with an aggregate message that doesn't conform to the conventional commits: https://github.com/hyperledger/cactus/commit/82cd7c8c3199cbbaab75df1f65d391bcb3cd23a1

Please either a) don't squash the commits or provide an appropriately formatted commit message at the time of the squashing through the github UI when merging in the PR.

Sorry again for being a stickler about the commit log.

PR #156, opened 2020-05-28
Whitepaper update

RafaelAPB:
Adds a sequence diagram supporting the 2.8 use case - blockchain migration; 
Minor improvements on the description of the use case.


whitepaper/whitepaper.md
```diff
@@ -284,32 +284,31 @@ gets replaced by a traditional bank account holding USD.
...
-Nonetheless, arguably a more interesting approach would be to consider cross-smart contract execution functionality or another automatic way of migration.
 
-References: <br>
-E Scheid and Burkhard Rodrigues, B Stiller. 2019. Toward a policy-based blockchain agnostic framework. 16th IFIP/IEEE International Symposium on Integrated Network Management (IM 2019) (2019)
-Philipp Frauenthaler, Michael Borkowski, and Stefan Schulte. 2019. A Framework for Blockchain Interoperability and Runtime Selection.
+Currently, when a consortium wants to migrate their blockchain (e.g., the source blockchain became obsolete, cryptographic algorithms no longer secure, etc), the solution is to re-implement business logic using a different platform, yielding great effort. 
+Data migrations have been performed before on public blockchains [2,3], both recent endeavors to render flexibility to blockchain-based solutions.
```
petermetz, line 301: Can create links that navigate to the references section upon a click when previewing the markdown.
Example links for the 2 and 3 could look like this:

`<sup>[2](#7-references)</sup>`

`<sup>[3](#7-references)</sup>`

whitepaper/whitepaper.md
```diff
@@ -319,7 +318,9 @@ Automatic smart contract migration yields risks for enterprise blockchain system
 
 By expressing my preferences in terms of functional and non-functional requirements, Hyperledger Cactus can recommend a set of suitable blockchains, as the target of the migration.
 Firstly, I could know in real-time the characteristics of the target blockchain that would influence my decision.
-For instance, the platform can analyze see the cost of writing information to Ethereum, the exchange rate US dollar - Ether, the average time to mine a block, the transaction throughput, and the network hash rate ((Frauenthaler et al., 2019).
+
+For instance, the platform can analyze see the cost of writing information to Ethereum, the exchange rate US dollar - Ether, the average time to mine a block, the transaction throughput, and the network hash rate [3].
```
petermetz, line 322: Same as above regarding the reference link, example below:

`<sup>[3](#7-references)</sup>`

RafaelAPB: Sure thing!

whitepaper/use-case-sequuence-diagram-blockchain-migration.puml
```diff
@@ -0,0 +1,78 @@
+@startuml Sequence Diagram - End User Wallet Authentication Authorization
```
petermetz, line 1: @RafaelAPB is that the correct title? ;-)

whitepaper/use-case-sequuence-diagram-blockchain-migration.puml
```diff
@@ -0,0 +1,78 @@
...
+
+skinparam ArrowFontStyle italic
+
+title Hyperledger Cactus\nSequence Diagram - End User Wallet Authentication Authorization
```
petermetz, line 5: Same as above

whitepaper/use-case-sequuence-diagram-blockchain-migration.puml
RafaelAPB: That was not the correct file. Fixed!

RafaelAPB:
> Hmm, not that I'm looking at it more, it looks like the whole puml file is not what the attached picture is being generated from.
> 
> Another thing that's unrelated: can we call them `consortium members` instead of `stakeholders` . I know it's longer and stakeholder sounds technically accurate as well, but what I'm concerned about is it being associated in any kind of way with `proof of stake` consensus which is something we definitely want to avoid because it implies that we are running our own blockchain for consortium management.



> Hmm, not that I'm looking at it more, it looks like the whole puml file is not what the attached picture is being generated from.
> 
> Another thing that's unrelated: can we call them `consortium members` instead of `stakeholders` . I know it's longer and stakeholder sounds technically accurate as well, but what I'm concerned about is it being associated in any kind of way with `proof of stake` consensus which is something we definitely want to avoid because it implies that we are running our own blockchain for consortium management.

Thanks for your feedback, Peter. I think that's a valid point - during the development of the feature I'll take into consideration the naming. I'm not changing it right now because it is very likely that the sequence diagram will be changed. This initial sequence diagram aims at providing better insights on this feature.

petermetz:
> Thanks for your feedback, Peter. I think that's a valid point - during the development of the feature I'll take into consideration the naming. I'm not changing it right now because it is very likely that the sequence diagram will be changed. This initial sequence diagram aims at providing better insights on this feature.

Fair enough. Do you mind if I change it? Merging in something and later discovering that it needs to be changed is cool, but if I already know about something then it bothers me if I leave it the way it is (just a personality trait of mine that sometimes I wish I could turn off :-) )

petermetz:
@RafaelAPB  Depending on your plans/work schedule could also just go with converting this to a draft PR as well. Totally flexible with the working mode, but I also want to be respectful of the other maintainer's time as much as possible: 

Properly reviewing a PR requires everybody to expend their resources (it is hard mental work) so there is a cost to it in that sense for each PR individually. In a way, opening a PR is like assigning a piece of work (the review) to every maintainer.

For example if you expect to drop another PR after this one maybe a few days or a week later, updating the same section/feature, then it's probably easier to keep this one open as a draft while the details get hammered out and *then* engage everyone at full force so they don't have to work on it twice/thrice etc. 
Keeping it in a draft is the best of both worlds in this sense because people who want to nit pick at stuff at their own expense can do so (yours truly) but at the same time it's opt-in to look at it so there's no "work assigned" to maintainers with a draft PR.

Just trying to figure it all out as we go along. Working together as a community in several different time zones is always going to be a challenge, but I'm hoping that small things like this with PR etiquette can go a long way or at least spark a discussion.

**Not meant this as a rant or any sort of complaint at all. I very much appreciate the hard work and the pull requests. :-)**

hartm:
To continue on Peter's comments:  we do need to codify some rules (or more likely, suggestions) for how to handle this kind of thing.  I'm going to try to put together a loose set of project guidelines for PRs and other stuff based on some of the other popular OSS projects.  Hopefully we can talk about them on Monday at the meeting.

But please, @RafaelAPB , don't view this as a complaint at all.  We appreciate all of your hard work on this!

RafaelAPB:
Completely agree! Not taken as a complaint at all!

PR #168, opened 2020-06-08
docs(whitepaper): fix use case 2.7 diagram typo

petermetz:


takeutak:
LGTM because it is only fixing typos

Issue #46, opened 2019-11-14
docs: add Checklists for Dev, Release, Hot-fix


petermetz:
# Checklist of checklists

- [x] New developer's checklist - steps from cloning to running example scenario
- [x] Release checklist - npm version, git tagging, npm publish, github release, code freeze(?)
- [ ] Hot-fix release checklist - branch from tag, backport fix, rebase, versioning scheme, etc
- [ ] PR merge checklist - make sure branch was rebased or rebase it right there and then

PR #172, opened 2020-06-11
docs(build-md): add file to build project for contributors (not users)

petermetz:

#46 (partial)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,0 +1,113 @@
+# Hyperledger Cactus SDK Build Instructions
+
+This is the place to start if you want to give Cactus a spin on your local
+machine or if you are planning on contributing.
+
+> This is not a guide for `using` Cactus for your projects that have business logic
+> but rather a guide for people who want to make changes to the code of Cactus.
+> If you are just planning on using Cactus as an npm dependency for your project,
+> then you might not need this guide at all.
+
+The project uses Typescript for both back-end and front-end components.
+
+## Getting Started
+
+* Install OS level dependencies:
+  * Windows Only
+    * WSL1 or WSL2 or any virtual machine running Ubuntu LTS
+  * Git
+  * NodeJS 12 or newer LTS (we recommend using nvm if available for your OS)
+  * OpenJDK 11
+  * Docker Engine
+  * Docker Compose
+
+* Clone the repository
+
+```sh
+git clone https://github.com/hyperledger/cactus.git
+```
+
+
+Windows specific gotcha: `File paths too long` error when cloning. To fix:
+Open PowerShell with administrative rights and then run the following:
+
+```sh
+git config --system core.longpaths true
+```
+
+* Change directories to the project root
+
+```sh
+cd cactus
+```
+
+* Run the CI script (takes a long time, 10+ minutes on an average laptop)
+
+```sh
+./tools/ci.sh
+```
+
+At this point you should have all packages built and verified with the full
+test suite including unit and integration tests that leverage docker containers
+to run ledgers, contract deployments, etc.
+
+You can start making your changes (use your own fork and a feature branch)
+or just run existing tests and debug them to see how things fit together.
+
+For example you can *run a ledger contract deployment test* via the
+REST API with this command:
+
+```sh
+npx tap --timeout=600 packages/cactus-test-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/deploy-contract-via-web-service.ts
+```
+
+*You can also start the API server* and verify more complex scenarios with an
+arbitrary list of plugins loaded into Cactus. This is useful for when you intend
+to develop your plugin either as a Cactus maintained plugin or one on your own.
+
+```sh
+npm run generate-api-server-config
+```
+
+Notice how this task created a .config.json file in the project root with an
+example configuration that can be used a good starting point for you to make
+changes to it specific to your needs or wants.
+
+The most interesting part of the `.config.json` file is the plugins array which
+takes a list of plugin package names and their options (which can be anything
+that you can fit into a generic JSON object).
+
+Notice that to include a plugin, all you need is specify it's npm package name
+(and ensure that said package is actually installed). This is important since
+it allows you to have your own plugins in their respective, independent Github
+repositories and npm packages where you do not have to seek explicit approval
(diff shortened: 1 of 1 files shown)


sfuji822:
I am leaving additional messages in text what I mentioned in the meeting. 

IMHO, Cactus will not be an independent application at the time of out from the box. 

Thus, when Cactus user wanted to try Cactus with demo application, he/she should go to example directory to find one to run.

I believe each demo application may require further configurations as prerequisites, like starting specific `Validator`  server with pre-configured Ledger with smart contract deployment. We should provide such application specific instruction should be placed at README for each demo applications. 

Additionally, you may describe some instructions for creating reusable docker-image,  if you will provide the way to build docker-image which contains specific Ledger Plugin as `Validator` server runtime.

BUILD.md
```diff
@@ -0,0 +1,137 @@
+# Hyperledger Cactus Build Instructions
+
+This is the place to start if you want to give Cactus a spin on your local
+machine or if you are planning on contributing.
+
+> This is not a guide for `using` Cactus for your projects that have business logic
```
petermetz, line 6: @sfuji822 

1. I added some clarification here so that users won't confuse this with a user-focused guide (which we don't yet have but should have one for sure).
2. Also updated the commit message and the PR title to avoid the same confusion as above mentioned. I can totally see how "getting started with the project" can be misleading without proper context (e.g. that it's getting started from a contributor's perspective, not a user's)

BUILD.md
```diff
@@ -0,0 +1,137 @@
...
+```sh
+git config --system core.longpaths true
+```
+
+#### Random Windows specific issues not covered here
+
+We recommend you use WSL or WSL2 or any Linux VM. We test most frequently on
+Ubuntu LTS which at the time of this writing means 18.04 and/or 20.04.
+
+## Plugin Development
```
sfuji822, line 119: I suggest you remove this section for now.
Instead, we better to instruct those in README.md at specific plugin directory like "ledger-plugins".

BUILD.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### File paths too long error when cloning
+
+Open PowerShell with administrative rights and then run the following:
+
+```sh
+git config --system core.longpaths true
+```
+
+#### Random Windows specific issues not covered here
```
sfuji822, line 114: I recommend removing this from the reasoning mentioned at pre-requiteies.

BUILD.md
```diff
@@ -0,0 +1,137 @@
...
+* Test the REST API directly with tools like cURL or Postman
+* Develop your own applications against it with the `Cactus SDK`
+* Create and test your own plugins
+
+### Windows Gotchas
+
+These are Windows specific problems that you might encounter during an initial
+setup.
+
+#### File paths too long error when cloning
```
sfuji822, line 106: This should be moved to "clone the repository section"

BUILD.md
```diff
@@ -0,0 +1,137 @@
...
+> This is not a guide for `using` Cactus for your projects that have business logic
+> but rather a guide for people who want to make changes to the code of Cactus.
+> If you are just planning on using Cactus as an npm dependency for your project,
+> then you might not need this guide at all.
+
+The project uses Typescript for both back-end and front-end components.
+
+## Getting Started
+
+* Install OS level dependencies:
```
sfuji822, line 15: Are you sure all of them are required to build SDK ?
I believe you can just say "we need docker" and provide URL like for further instruction https://docs.docker.com/docker-for-windows/install-windows-home/

I thought OpenJDK is required only when you build Ledger Plugin for Besu. This condition should be mentioned here.

BUILD.md
```diff
@@ -0,0 +1,137 @@
...
+
+This is the place to start if you want to give Cactus a spin on your local
+machine or if you are planning on contributing.
+
+> This is not a guide for `using` Cactus for your projects that have business logic
+> but rather a guide for people who want to make changes to the code of Cactus.
+> If you are just planning on using Cactus as an npm dependency for your project,
+> then you might not need this guide at all.
+
+The project uses Typescript for both back-end and front-end components.
```
sfuji822, line 11: I thought frontend (Web-application for issuing REST operations) was not in the scope of Cactus.

This statement should be revised with some text like:
"The code written in Typescript in this project requires developers to compile them into Javascript for runtime."

BUILD.md
sfuji822: @petermetz 
Can you modify the title with "Hyperledger Cactus SDK Building Instructions" ?

petermetz: @sfuji822 
Done.

BUILD.md
petermetz: @sfuji822 
How about putting it in the root directory's README.md instead?

That "ledger-plugins" directory will hopefully go away once we reconciled the two code bases.

BUILD.md
petermetz: @sfuji822 Done.

BUILD.md
petermetz: @sfuji822 Done.

BUILD.md
petermetz: > Are you sure all of them are required to build SDK ?
> I believe you can just say "we need docker" and provide URL like for further instruction https://docs.docker.com/docker-for-windows/install-windows-home/

Yes, we don't yet have a build image for docker that would contain all of this, but I agree it's definitely something we should have as an option.

> I thought OpenJDK is required only when you build Ledger Plugin for Besu. This condition should be mentioned here.

It does not, the current besu plugin interacts with the ledger entirely through HTTP so no JDK is necessary for that.
The JDK is needed for the OpenAPI (Swagger) generator which is a part of the build where the declarative API definitions get converted into source code and documentation.

BUILD.md
petermetz: > I thought frontend (Web-application for issuing REST operations) was not in the scope of Cactus.

We've been saying the opposite of that for months. A great deal of effort went into the new arch providing this ability to have the sdk package run the same way on both NodeJS and in the browser and for both to have type assist from the compiler to enhance developer productivity. Also the whole build system as it is, provides the ability for the package authors to cherry pick what modules they want to export on the different platforms (browser or NodeJS) so that the packages do not contain the bloat of one platform on the other (for example the NodeJS file system API is not something that should be present in the bundle that will be running in a front end web application).

PR #190, opened 2020-06-16
feat(corda all in one docker image): runs a Corda Network in one dock…

jweate:
…er container

Provides a base docker image that can be used to spin up a Corda Network all inside one container.
Provides a testing framework to interact with the docker container.

Signed-off-by: Jacob Weate <jacob.weate@accenture.com>


sfuji822:
I can be volunteer for one of reviewers on this pull request. But can I ask you to implement some changes ?

First of all, can you check your PR against [contribution guide](https://github.com/hyperledger/cactus/blob/master/CONTRIBUTING.md) since it seems several problem reported above. 

Second, I would suggest you to create dedicated docker network for testing environment,  ensuring isolation this testing environment from main application network. It might be named like 'testnet-corda' for example.

Last one is some sort of question for you. Is this container can host multiple application ? If it not, I might suggest you to provide the way to create generic container image first, and you also give some instructions for users in README, how to customize docker-compose file to enable multiple containers running in parallel.

PR #196, opened 2020-06-22
test(test-tooling): parametrize besu-all-in-one container

RafaelAPB:
Allows passing input arguments to the docker container running the Besu test ledger. This is useful, for example, to run Besu in different ledgers than the default one.

I've included tests that check if the ledger was successfully initialized, based on the configurations.

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts
@@ -11,12 +11,14 @@ export interface IBesuTestLedgerConstructorOptions {
   containerImageVersion?: string;
   containerImageName?: string;
   rpcApiHttpPort?: number;
+  envVars?: string[]
 }
 
 export const BESU_TEST_LEDGER_DEFAULT_OPTIONS = Object.freeze({
   containerImageVersion: "latest",
   containerImageName: "hyperledger/cactus-besu-all-in-one",
   rpcApiHttpPort: 8545,
+  envVars: ["BESU_NETWORK=dev"]
 });
 
 export const BESU_TEST_LEDGER_OPTIONS_JOI_SCHEMA: Joi.Schema = Joi.object().keys(
@@ -29,6 +31,7 @@ export const BESU_TEST_LEDGER_OPTIONS_JOI_SCHEMA: Joi.Schema = Joi.object().keys
       .min(1024)
       .max(65535)
       .required(),
+    envVars: Joi.array().allow(null).required(),
   }
 );
 
@@ -36,6 +39,7 @@ export class BesuTestLedger implements ITestLedger {
   public readonly containerImageVersion: string;
   public readonly containerImageName: string;
   public readonly rpcApiHttpPort: number;
+  public readonly envVars: string[];
 
   private container: Container | undefined;
 
@@ -51,6 +55,8 @@ export class BesuTestLedger implements ITestLedger {
       BESU_TEST_LEDGER_DEFAULT_OPTIONS.containerImageName;
     this.rpcApiHttpPort =
       options.rpcApiHttpPort || BESU_TEST_LEDGER_DEFAULT_OPTIONS.rpcApiHttpPort;
+    this.envVars =
+        options.envVars || BESU_TEST_LEDGER_DEFAULT_OPTIONS.envVars;
 
     this.validateConstructorOptions();
   }
@@ -141,6 +147,7 @@ export class BesuTestLedger implements ITestLedger {
           // to docker container's IP addresses directly...
           // https://stackoverflow.com/a/39217691
           PublishAllPorts: true,
+          Env: this.envVars,
         },
         {},
         (err: any) => {
@@ -293,6 +300,7 @@ export class BesuTestLedger implements ITestLedger {
         containerImageVersion: this.containerImageVersion,
         containerImageName: this.containerImageName,
         rpcApiHttpPort: this.rpcApiHttpPort,
+        envVars: this.envVars,
       },
       BESU_TEST_LEDGER_OPTIONS_JOI_SCHEMA
     );
--- a/packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/besu-test-ledger-parameters.test.ts
+++ b/packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/besu-test-ledger-parameters.test.ts
@@ -1,0 +1,85 @@
+// tslint:disable-next-line: no-var-requires
+const tap = require("tap");
+import { BesuTestLedger } from "@hyperledger/cactus-test-tooling";
+tap.test("deploys a Besu node with the default configurations", async (assert: any) => {
+    assert.plan(1);
+
+    // No options
+    const besuTestLedger = new BesuTestLedger();
+
+    assert.ok(besuTestLedger);
+    assert.end();
+});
+
+
+tap.test("checks if simple Besu's environment variables are correctly passed", async (assert: any) => {
+    assert.plan(2);
+    const simpleEnvVars = ["BESU_MINER_ENABLED","BESU_NETWORK=dev","BESU_MIN_GAS_PRICE=0"]
+
+    const besuOptions = {
+        envVars: simpleEnvVars
+    }
+    const besuTestLedger = new BesuTestLedger(besuOptions);
+
+    assert.equal(besuTestLedger.envVars,simpleEnvVars);
+    assert.ok(besuTestLedger);
+    assert.end();
+});
+
+tap.test("deploys a Besu Node on the Rinkeby network", async (assert: any) => {
+    assert.plan(2);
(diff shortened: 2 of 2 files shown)


packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/testLedgerParameters.ts
```diff
@@ -0,0 +1,76 @@
...
+        await besuTestLedger.destroy();
+    });
+
+    assert.ok(besuTestLedger);
+    assert.end();
+});
+
+tap.test("deploys a Besu Node on the Rinkeby network", async (assert: any) => {
+    assert.plan(2);
+    let RinkebyNetworkParameters = "--mount type=bind,source=/<myvolume/besu/testnode>,target=/var/lib/besu hyperledger/besu:latest --miner-enabled --miner-coinbase fe3b557e8fb62b89f4916b721be55ceb828dbd73--network=dev --min-gas-price=0";
```
petermetz: Not sure if the linter catches this at present or not, but variables should be camel-case, e.g. starting with lowercase.
I'll double check if the linter complains and if not, will make sure that it does so in the future.

packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts
```diff
@@ -11,12 +11,14 @@ export interface IBesuTestLedgerConstructorOptions {
   containerImageVersion?: string;
   containerImageName?: string;
   rpcApiHttpPort?: number;
+  cmdArgs?: string
```
petermetz, line 14: Could this be an array of strings instead of one string?
Makes it easier to deal with some exotic/special characters on a per argument basis if/when necessary.
Also recommend an array of strings instead of just one big string to keep with the conventions of the related software such as NodeJS child_process/spawn APIs and dockerode as well.

The default value can be an empty array and when calling docker.run(...) it's easy to concatenate the additional cmd args coming from the constructor and the ones prescribed by the logic of `BesuTestLedger` itself.

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/testLedgerParameters.ts
RafaelAPB: It doesn't complain about that, maybe that's why it went unnoticed.

petermetz: > It doesn't complain about that, maybe that's why it went unnoticed.

That's fair. It isn't documented anywhere that it should be camel case so if the linter let's it through then it is to be assumed that everything is as per the style guide (which we don't have explicitly, just through the linter configuration, basically the linter configuration is the style guide)
I'll check on the config.

packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts
```diff
@@ -29,13 +31,15 @@ export const BESU_TEST_LEDGER_OPTIONS_JOI_SCHEMA: Joi.Schema = Joi.object().keys
...
       .required(),
+    cmdArgs: Joi.string().allow(null, '', ' ').required(),
   }
 );
 
 export class BesuTestLedger implements ITestLedger {
   public readonly containerImageVersion: string;
   public readonly containerImageName: string;
   public readonly rpcApiHttpPort: number;
+  public readonly cmdArgs: string;
```
petermetz, line 42: Where do the cmdArgs actually get passed in to the docker.run parameter list?

RafaelAPB:
updated

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/testLedgerParameters.ts
```diff
@@ -0,0 +1,115 @@
...
+        envVars: simpleEnvVars
+    }
+    const besuTestLedger = new BesuTestLedger(besuOptions);
+    await besuTestLedger.start();
+
+    assert.tearDown(async () => {
+        await besuTestLedger.stop();
+        await besuTestLedger.destroy();
+    });
+    assert.equal(besuTestLedger.envVars,simpleEnvVars);
```
petermetz, line 34: @RafaelAPB 
This assertion might not need the backing container to be launched because the class constructor already assigns the `envVars` property regardless of `.start()` having been called or not.

Would it make sense to either (?)

a) Remove the `.start()`, `stop()` and `destroy()` calls (which speeds up the test cases 1000x fold but asserts the same thing) OR
b) Add assertions that verify that the env vars are also present within the container itself and are indeed in effect (I've encountered a lot of bugs in the past where I thought I set some configuration up for some software and later it turned out to be non-effective because of some bug in between).

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/testLedgerParameters.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    const besuTestLedger = new BesuTestLedger(besuOptions);
+    await besuTestLedger.start();
+
+    assert.tearDown(async () => {
+        await besuTestLedger.stop();
+        await besuTestLedger.destroy();
+    });
+    assert.equal(besuTestLedger.envVars,rinkebyNetworkEnvVars);
```
petermetz, line 53: @RafaelAPB 
This assertion might not need the backing container to be launched because the class constructor already assigns the `envVars` property regardless of `.start()` having been called or not.

Would it make sense to either (?)

a) Remove the `.start()`, `stop()` and `destroy()` calls (which speeds up the test cases 1000x fold but asserts the same thing) OR
b) Add assertions that verify that the env vars are also present within the container itself and are indeed in effect (I've encountered a lot of bugs in the past where I thought I set some configuration up for some software and later it turned out to be non-effective because of some bug in between).

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/testLedgerParameters.ts
```diff
@@ -0,0 +1,115 @@
...
```
petermetz, line 73: @RafaelAPB 
This assertion might not need the backing container to be launched because the class constructor already assigns the `envVars` property regardless of `.start()` having been called or not.

Would it make sense to either (?)

a) Remove the `.start()`, `stop()` and `destroy()` calls (which speeds up the test cases 1000x fold but asserts the same thing) OR
b) Add assertions that verify that the env vars are also present within the container itself and are indeed in effect (I've encountered a lot of bugs in the past where I thought I set some configuration up for some software and later it turned out to be non-effective because of some bug in between).

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/testLedgerParameters.ts
```diff
@@ -0,0 +1,115 @@
...
```
petermetz, line 93: @RafaelAPB 
This assertion might not need the backing container to be launched because the class constructor already assigns the `envVars` property regardless of `.start()` having been called or not.

Would it make sense to either (?)

a) Remove the `.start()`, `stop()` and `destroy()` calls (which speeds up the test cases 1000x fold but asserts the same thing) OR
b) Add assertions that verify that the env vars are also present within the container itself and are indeed in effect (I've encountered a lot of bugs in the past where I thought I set some configuration up for some software and later it turned out to be non-effective because of some bug in between).

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/testLedgerParameters.ts
```diff
@@ -0,0 +1,115 @@
...
+        envVars: ethereumEnvVars
+    }
+    const besuTestLedger = new BesuTestLedger(besuOptions);
+    await besuTestLedger.start();
+
+    assert.tearDown(async () => {
+        await besuTestLedger.stop();
+        await besuTestLedger.destroy();
+    });
+    assert.equal(besuTestLedger.envVars,ethereumEnvVars);
```
petermetz, line 112: @RafaelAPB 
This assertion might not need the backing container to be launched because the class constructor already assigns the `envVars` property regardless of `.start()` having been called or not.

Would it make sense to either (?)

a) Remove the `.start()`, `stop()` and `destroy()` calls (which speeds up the test cases 1000x fold but asserts the same thing) OR
b) Add assertions that verify that the env vars are also present within the container itself and are indeed in effect (I've encountered a lot of bugs in the past where I thought I set some configuration up for some software and later it turned out to be non-effective because of some bug in between).

RafaelAPB:
@petermetz  thank you for your comments.
I think the most straightforward option would be A. B would be necessary, and I'm planning to clarify this while I develop the blockchain migrator feature (as I use a parametrized besu test ledger).

petermetz:
@RafaelAPB 
Ooops, one more thing, if you could add the `.test` in the test file's file name that would be great ;-)
Right now it's `packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts` but should be `packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.test.ts` for IDE support

RafaelAPB:
@petermetz, updated!

PR #202, opened 2020-06-25
feat(api-server): TLS, mTLS support

petermetz:
This is not really something that we expect to be used in
production due to the prevalence of load balancers and
reverse proxies in most production scale web application
deployments, but it is a must have for testing and also
for us to be able to claim that we take the secure by
default design principle seriously.


petermetz:
> but I would like to implement this with external server such as nginx.

@sfuji822 This change does not restrict you from doing that in any way. It's all opt-in by way of configuration flags that can turn TLS/mTLS on and off.

petermetz:
FYI: Did a rebase to resolve newly arisen conflicts after the other PRs got merged (then force pushed to keep the branch/PR tidy)

takeutak:
LGTM.  We may fix later as another item in terms of the above discussion points.

petermetz:
> OK. We may fix it later. I will create a new issue as action item to myself.

> LGTM. We may fix later as another item in terms of the above discussion points.

Thank you. Let's continue that discussion in #220 then.

PR #203, opened 2020-06-25
style(tslint): ban PascalCase variables from the code

### LINKED ISSUE #197: style(linter): ban PascalCase variables in TSLint config
### Description
As a maintainer/contributor/developer I want to specify in our style guide (which is the linter config, implicitly describing the style we need) that PascalCase variables are banned because they make it very confusing very quickly when there's class names and variables names next to each other with the same casing so that I am not confused (too much) when reading the code.

The default `recommended` tslint config allows for PascalCase and I usually advocate for staying with defaults as much as possible but this is definitely one of those cases where diverging is justified IMO.
My guess is that they allow PascalCase because in React code you often see it, though not sure at all if this is really it. It doesn't really matter why either to be honest, my opinion is the same regardless.

Just tested it and this can be achieved with this patch:

```sh
diff --git a/tslint.json b/tslint.json
index d4f35c1..9a56a1d 100644
--- a/tslint.json
+++ b/tslint.json
@@ -4,7 +4,15 @@
         "tslint:recommended"
     ],
     "jsRules": {},
-    "rules": {},
+    "rules": {
+        "variable-name": {
+            "options": [
+                "allow-leading-underscore",
+                "ban-keywords",
+                "check-format"
+            ]
+        }
+    },
     "rulesDirectory": [],
     "linterOptions": {
         "exclude": [
```

The end result from this is that the following will no longer be allowed by the linter:

```typescript
let SomeVariableOfMine: string = "hello world";
```

### Acceptance Criteria
1. Update tslint.json to not allow PascalCase variable names
2. Everything else to remain the same

cc: @takeutak @jonathan-m-hamilton @sfuji822

petermetz:
It is very easy to confuse PascalCase variables in JS/TS with classes.
We would much rather keep
lowerCamelCase for variables and PascalCase for classes
to make it easier to read the code.

fix #197

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -4,7 +4,15 @@
         "tslint:recommended"
     ],
     "jsRules": {},
-    "rules": {},
+    "rules": {
+        "variable-name": {
+            "options": [
+                "allow-leading-underscore",
+                "ban-keywords",
+                "check-format"
+            ]
+        }
+    },
     "rulesDirectory": [],
     "linterOptions": {
         "exclude": [


petermetz:
> I understand your intention, but I prefer to suspending this strict rule for a while since we are still under refactoring.

Works for me. Happy to keep this open until we are all ready to pull the plug.

takeutak:
LGTM

PR #204, opened 2020-06-25
docs(contributing-md): github Desktop on Windows know about the file …

### LINKED ISSUE #188: docs(contributing-md): github Desktop on Windows know about the file path length solution
**Describe the bug**

When checking out the project via Github Desktop, it crashes because of too long paths.

**To Reproduce**

Attempt to check out the project on Github Desktop

**Expected behavior**

Github Desktop does not crash.

**Screenshots**

image

image

**Additional context**

The CONTRIBUTING.md guide specifically says that one should check out the code via Github Desktop so this is an issue of high importance due to how newcomers to the project might get turned off if the project can't even be cloned properly...

cc: @hartm @jonathan-m-hamilton @sfuji822 @takeutak

petermetz:
…path length solution

Considered multiple other alternatives here but eventually
decided to go with just documenting the
workaround and
not refactor the directory tree under
contribs because at the end of the day

there will always be a possibility to experience this
issue since we cannot know in advance (or
account for)
the length of the path of the directory in which the user is cloning...

fix #188


petermetz:
> Since the issue may occur on any windows machine with the git client installed. I would prefer to explaining as a generic solution like:
> If you are using Git client on Windows, you will need to enable long paths option for the git client
> which you can do in PowerShell by executing this command:

Agreed 100%, thank you.
I updated the commit in place to have the proposed changes incorporated in it (did not want a second commit because it's a correction of the existing commit rather than it addressing some separate concern).

Issue #181, opened 2020-06-15
Support tools to build `test net` for demo apps


sfuji822:
**Is your feature request related to a problem? Please describe.**
Providing sample application is important for demonstration purpose.
But it is also needed to keep separated such demo app from main framework.

**Describe the solution you'd like**
We should provide support tool to build `test net` for specific Ledger platform to work with sample web application build on top of Cactus SDK.

**Describe alternatives you've considered**
Providing test case on each Cactus components, but it is merely work for unit tests.

**Additional context**
I am considering to provide solution as docker-compose environment which is  easy to build and run without deep knowledge.

petermetz:
@sfuji822 Is it required in the scope of this issue to have the test nets run multiple nodes or is it okay to have single node test nets?

Just asking because for the integration tests we have single node test nets that you can pull up via code, like this one brings up a Besu test ledger that is then used for ledger testing:
packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts

And the limitation of this is that it single node ledger only, so if we need multiple nodes in the test net then it is not a good fit in it's current form (could be improved of course).

RafaelAPB:
I think it is very relevant to us to be able to bootstrap **both** single node and multiple node test networks. Several features we are developing can be trivially tested with a single node network; but in production, many networks will have multiple nodes, and we want to test those interactions.

sfuji822:
@petermetz , I did not any preference on the number of node since my expectation on this feature is simply tasks on trying Cactus at docker-based playground. takeutak or I will submit related PR soon.

sfuji822:
@RafaelAPB , thanks for you comment. I like to implement this as minimal set at first, but I also agree with you on the potential needs on tricky configuration and we may extend it after first version.

RafaelAPB:
@sfuji822, you're welcome! Are you planning to extend the test ledger classes, then? It does require new images to implement multi-node networks, is that correct?

petermetz:
Closing as per what was done and described in #240

petermetz:
> @sfuji822, you're welcome! Are you planning to extend the test ledger classes, then? It does require new images to implement multi-node networks, is that correct?

@RafaelAPB Now I can answer this (after having seen the PR) This does not use/work with the `FabricTestLedger` class so we'll have to pick that up in a separate thread/issue.

PR #211, opened 2020-06-26
chore(tools/docker): consolidate Besu/Quorum AIO images

### LINKED ISSUE #207: chore(tools): Consolidate Docker All In One Images, Publishing
### Description
As a developer I want to have a consistent view of the images maintained by Cactus for testing so that it's not confusing to locate the besu or the quorum all in one images for example (both of which at the moment are just dumped under Peter's personal handle because it was the path of least resistance at the time).

Context: We have 2 all-in-one images that are designed for testing already (besu, quorum) and two more are in the works: Jacob is working on Corda and Sownak is doing Fabric 1.4.x

Links to existing images that need to be moved into Cactus (because our code references the images via DockerHub as it is right now):
- https://github.com/petermetz/besu-docker-all-in-one
- https://github.com/petermetz/quorum-docker-all-in-one

The tasks/issues for Corda and Fabric:
#132 
#143 

### Acceptance Criteria
1. Images are all located under `./tools/docker/$LEDGER_NAME-all-in-one/` directory trees
2. Publishing of said images is set up as part of the CI, images go to the Github image registry or whatever else is preferred by HL

cc: @ryjones @takeutak  @sfuji822 @jonathan-m-hamilton @jweate @sownak

petermetz:
Fixes #207 

Related to #181 but not yet sure how exactly

See currently available images on DockerHub:

https://hub.docker.com/repository/docker/hyperledger/cactus-besu-all-in-one/tags?page=1
https://hub.docker.com/repository/docker/hyperledger/cactus-quorum-all-in-one/tags?page=1

Note 1: We could create one for Vanilla Ethereum as well if needed, if the their default base image is not sufficient for whatever reason, but I think their base image is literally just the geth binary anyway so there isn't much complexity that we need to get rid of there that would justify us maintaining our own all-in-one image (except if we want multi node support in a single container for example).

Note 2: There are two commits in the PR that I'd prefer to merge in as they are instead of squashing because one is specifically for adding the quorum AIO image and the other one is for Besu so there is justification to have them in their own commits, but it also makes sense to have said two commits in the same PR given the same concern they are both addressing.

cc: @RafaelAPB @sownak @jweate


ryjones:
@petermetz nice; let's do this?

petermetz:
> @petermetz nice; let's do this?

@ryjones I'm all for it, but need to secure the approvals before it can be merged!

Issue #192, opened 2020-06-18
docs(whitepaper): terminology for assets and interaction types


DOBEN:
### Description
I want to update the white paper focusing on following changes (an explanatory draft of the proposed changes are given in the attached document):
https://chat.hyperledger.org/file-upload/YjNwc6wnBypHyMBvJ/Suggestions%20white%20paper.pdf?download

1. Defining and using consistent terminology with respect to the current names "data", "asset", and "data asset" in the example use case section. The following naming convention is proposed:
- FA = fungible asset (alternative naming: value token/coin T) – cannot be duplicated on different ledgers
- NFA = non-fungible asset (alternative naming: asset A) – cannot be duplicated on different ledgers
- D = data – can be duplicated on different ledgers
2. Defining three ledger interaction types (ledger transfer/ledger coordination/ledger entry-point coordination)
3. Using the definitions from 1. and 2. to create a table to classify and order all mentioned example use cases.
4. A section about "burning of tokens" and use cases in the appendix.

This table acts as a building block to describe the different flavors of how cactus can be used. Real-world use cases will likely be built on these simple foundation blocks leveraging a mix of them simultaneously.  

As a new person who is considering using cactus in a concrete business use case, this table will help to identify relevant example use cases that are aligned with a given concrete business use case.

petermetz:
@DOBEN Thank you for helping out, as mentioned in the call this week, I'm in general in support of all possible improvements! 
As for me, I will need some more time to read it thoroughly and pick on the details (as I'm sure others will too), but in the meantime I have a recommendation related to administrative side of things: 

The github issue/PR name should be something more specific that gets to the heart of what is the proposed change to reduce the cognitive load required when there are multiple issues circulating with similar titles. 
What I mean is that `Whitepaper improvements` could be replaced with `docs(whitepaper): terminology for assets and interaction types` so that we don't get confused if there are multiple issues opened by different people (all of them being some sort of improvement to the whitepaper). 
This is not the case yet, but as the community grows we expect (and hope) to have this as a problem. :-)

DOBEN:
Yes, good point. I changed the title :).

Mr1262:
BTC WALLET MMM

DOBEN:
Thank you for the helpful comments in this thread and the thread related to the pull request #216. I will close this issue since it was addressed and merged in #216.

PR #216, opened 2020-07-04
docs(whitepaper): terminology for assets and interaction types #192 

DOBEN:
This adds terminology for ledger object types, and blockchain interoperability types.

Signed-off-by: Doris Benda <doris_benda@yahoo.com>


whitepaper/whitepaper.md
```diff
@@ -33,6 +37,11 @@ Photo by Pontus Wellgraf on Unsplash
 <div style="page-break-after: always; visibility: hidden"><!-- \pagebreak --></div>
 
 - [1. Abstract](#1-abstract)
+  - [1.1 Terminology of Blockchain Interoperability](#11-terminology-of-blockchain-interoperability)
```
sfuji822, line 40: I am afraid proposed contents are not abstract. I might suggest you to add new chapter titled like: "Introduction".
The Introduction should cover what we will deliver with Cactus as integrated service builder.

whitepaper/whitepaper.md
```diff
@@ -124,12 +133,248 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
 
 We propose a protocol and it's implementation to connect as many of them as possible in an attempt to solve the fragmentation problem by creating a heterogeneous system architecture <sup>[1](#7-references)</sup>.
 
+
+
+There are two inherent problems<sup>[a](#12-footnotes)</sup> that have to be solved when connecting different blockchains: 
+* How to retrieve the
```
sfuji822, line 139: It could be possible to view the ledger information, but I have little concerns about expected function. Cactus will not provide any retrieval of ledger information. We need to clarification about the functions in this topic. You may know about the project Explorer for generic use  is proposed at TSC. If we can collaborate with them, it can be beneficial.

whitepaper/whitepaper.md
```diff
@@ -124,12 +133,248 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
...
+networkwide<sup>[b](#12-footnotes)</sup> ledger view aka the state of a connected blockchain from the outside?
+* How can other entities verify a given claim about the state of a connected blockchain
+from the outside? 
+
+`Cactus` operates several validator nodes above the
```
sfuji822, line 144: I thought this is also need to clarify your expectations. My assumed the relationship between Cactus SDK and Validator is following: 
- The Validator will provide the proof on a transaction is observed at the Ledger.
- The SDK (at business logic plugin) can complete set of transactions or issue another transaction to Ledger
Does it make sense to you ?

whitepaper/whitepaper.md
petermetz: > Cactus will not provide any retrieval of ledger information.

I would say that providing proofs about the ledger state is what one could call as "retrieval of ledger information"
Would also say that while Cactus itself may not be doing the ledger querying, the plugins certainly are. It's an interesting question how should this be phrased. If a Cactus plugin is doing something, can we say that Cactus is doing that thing?
Technically if someone wanted to (not me, but if someone had that motivation for whatever reason), they could implement the entire HL Explorer project as a Cactus plugin and then deploy that.

whitepaper/whitepaper.md
```diff
@@ -124,12 +133,248 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
...
+
+Difference between an asset (FA or NFA) and data (D)
+
+Unicity applies to FAs and NFAs meaning it guarantees that only one valid representation of a given asset exists in the system. It prevents double-spending of the same token/coin in different blockchains. The same data package can have several representations on different ledgers while an asset (FA or NFA) can have only one representation active at any time, i.e., an asset exists only on one blockchain while it is locked/burned on all other blockchains. If fundamental disagreement persists in the community about the purpose or operational upgrades of a blockchain, a hard fork can split a blockchain creating two representations of the same asset to coexist. For example, Bitcoin split into Bitcoin and Bitcoin Cash in 2017. Forks are not addressing blockchain interoperability so the definition of unicity applies in a blockchain interoperability context. A data package that was once created as a copy of another data package might divert from its original one over time because different blockchains might execute different state changes on their data packages.
```
petermetz, line 183: @DOBEN This just a question, not actual feedback. 
Would be curious to hear your thoughts about breaking the NFA category into two distinct ones one for digital NFA and one for physical (not saying we should, just wondering).
The reasoning is that the digital NFAs have their counterfeit-resistance backed by hard math/cryptography, but if I have a car tokenized on a ledger as an NFA then the only thing that cannot be counterfeited is the digital token not the car itself (I can create the same car and make sure even the serial numbers are a match).
While thinking and reading about supply chain and blockchain based solutions to it, I often noticed this pattern of misunderstanding by people where they thought that because a piece of information is signed and consensus was reached on it, then the claims contained within that piece of information must be true as well, but in reality there's only proof of someone making said claims, not the claims being true.

DOBEN: The last sentence can be summarized: Garbage in => Garbage out
Garbage (false claims) that are put into the blockchain and are then signed and included as a transaction in the blockchain through the consensus algorithm will still result in garbage (false claims) coming out of the blockchain.

The weakest point at any kind of blockchain NFA solution that tracks corresponding physical assets is always the link between the physical asset identifier (Serial number, QR code, barcodes, and RFID) and the physical asset. People will always be able to counterfeit an asset and copy-paste the serial number, QR code... of the genuine asset tracked on the blockchain on to the fake asset. Unfortunately, people don’t even have to make the effort most of the time to produce/counterfeit the same asset or a similar-looking asset, because as long as an asset identifier is copied-pasted on to a fake asset, the average user will just scan the QR code or look up the serial number on the blockchain and believe whatever claims are presented. There are ways around to use solutions by companies like Scantrust and AccessReal (i-Sprint) .... Nonetheless, there is no way in the physical world to secure 100% a physical asset to its identifier but it can be made as hard as possible to break the link. As long as it takes more effort (time or money) to break the link, it is good enough for most blockchain solutions. On the one hand, the cheaper an asset the easier it is to find a solution that is good enough to link the physical asset to its identifier. On the other hand, the more expensive an asset the more likely that there is a business use case to track the asset on a blockchain.

That said, digital NFAs are always counterfeit-resistance but physical NFAs might or might not be counterfeit-resistance. I hope that people will take this problem seriously in the future because there is no added value of tracking an NFA on the blockchain that can be easily counterfeited in the physical world.

whitepaper/whitepaper.md
```diff
@@ -124,12 +133,248 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
...
+<img src="https://render.githubusercontent.com/render/math?math=T_2(O_1||T_1||E_1) = \text{ Transaction 2 depends on }\quad\!\!\!\!O_1\quad\!\!\!\! \text{ or } \quad\!\!\!\!T_1\quad\!\!\!\! \text{ or }\quad\!\!\!\!E_1">
+
+	
+
+
+
+### 1.1.3 Burning or Locking of Assets
+
+
+To guarantee unicity, an asset (NFA or FA) has to be burned or locked before being released in another blockchain. Locked assets can be unlocked in case the asset is retransferred back to its original blockchain, whereas the burning of assets is an irreversible process. Burning of assets usually applies more to tokens/coins (FAs)<sup>[h](#12-footnotes)</sup> and can be seen as a donation to the community since the overall value of the currency increases.
```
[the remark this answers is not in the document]
DOBEN: I included the explanation from the footnote into the main text of the revised write-up and highlighted that burning/locking of assets can be avoided when wallets/accounts are present on both blockchains and even should be avoided in ordinary token exchanges, such as with BTC and ETH, because their minting process can not be leveraged on during a ledger transfer.

whitepaper/whitepaper.md
```diff
@@ -124,13 +134,235 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
...
+
+Burning of assets can be implemented as follows: 
+
+* Assets are sent to the address of the coinbase/generation transaction<sup>[h](#22-footnotes-introduction)</sup> in the genesis block. A coinbase/generation transaction is in every block of blockchains that rely on mining. It is the address where the reward for mining the block is sent to. Hence, this will burn the tokens/coins in the address of the miner that mined the genesis block. In many blockchain platforms, it is proven that nobody has the private key to this special address. 
+
+* Tokens/Coins are subtracted from the user account as well as optionally from the total token/coin supply value.
```
RafaelAPB, line 268: Additionally, this is a mechanism that many interoperability solutions use: they lock an asset on a source blockchain and release a representation of that asset on a target blockchain. E.g., Wanchain page 23

whitepaper/whitepaper.md
```diff
@@ -124,13 +134,235 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
...
+
```
RafaelAPB, line 276: I did not understand from "Private blockchains" up to "merkle trees". Why can't Cactus utilize a merkle tree hash as a proof of block validity? I think this could be an easy way to prove that a transaction we issued to Bitcoin or Ethereum, for example, was included into a valid block.

In fact, Cactus' validators could in theory build a merkle tree from a Fabric network, sign the hash of the merkle tree root, and deliver it as a summary/proof of the blockchain state.  

However, merkle trees are not used in Fabric, for example, because Fabric relies on endorsers' signatures to validate transactions - Fabric was not meant to be easily verifiable from the outside.

whitepaper/whitepaper.md
```diff
@@ -124,13 +134,235 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
...
+Difference between an asset (FA or NFA) and data (D)
+
+Unicity applies to FAs and NFAs meaning it guarantees that only one valid representation of a given asset exists in the system. It prevents double-spending of the same token/coin in different blockchains. The same data package can have several representations on different ledgers while an asset (FA or NFA) can have only one representation active at any time, i.e., an asset exists only on one blockchain while it is locked/burned on all other blockchains. If fundamental disagreement persists in the community about the purpose or operational upgrades of a blockchain, a hard fork can split a blockchain creating two representations of the same asset to coexist. For example, Bitcoin split into Bitcoin and Bitcoin Cash in 2017. Forks are not addressing blockchain interoperability so the definition of unicity applies in a blockchain interoperability context. A data package that was once created as a copy of another data package might divert from its original one over time because different blockchains might execute different state changes on their data packages. 
+
+
+### 2.1.2 Blockchain Interoperability Types
```
RafaelAPB, line 181: Please check page 18 from https://arxiv.org/pdf/2005.14282.pdf, and see if that classification makes sense for you.

This classification is not very clear, as the different types overlap. For instance, an atomic swap is a ledger transfer, as it locks and releases tokens, but it is also a ledger interaction, because there is a pair of actions issued in different ledgers.

whitepaper/whitepaper.md
```diff
@@ -124,13 +134,235 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
...
+
+Outside entities (verifier nodes) can request and register the public keys of the validator
+nodes of a blockchain network that they want to connect to. Therefore, they can verify the signed proofs of the state of the blockchain since they have the public keys of the validator nodes. This implies that the verifier nodes trust the validator nodes as such they trust the `Cactus` consortium operating the validator nodes.
+
+
+
+
+<img src="./cactus_overview.png" width="400"><br>Figure description: V1, V2, and V3 are validator nodes which provide proofs of the underlying blockchain network through ledger-specific plug-ins. V1, V2, and V3 run a consensus algorithm which is independent of the consensus algorithm run by the blockchain network nodes N1, N2, N3, N4, and N5.</img>
+
+## 2.1 Terminology of Blockchain Interoperability
```
RafaelAPB, line 156: If you think it is useful, we can add the following definitions, which are very useful in the context of Cactus pages 5-8 (https://arxiv.org/pdf/2005.14282.pdf):

source blockchain, and target blockchain
Cross-Chain Transaction (CC-Tx)
Cross-Blockchain Transaction (CB-Tx)
Cross-Chain Decentralized Application (CC-dApp)
Internet of Blockchains (IoB)

petermetz:
@DOBEN @RafaelAPB 
So, I was about to merge this just now, but I see @RafaelAPB came in with some extra suggestions post-approval (which is cool).
Is it okay for you to discuss these comments/suggestions and then if there's any changes agreed on do that in a follow-up PR instead? I want to close this one down with the merge now that it has enough approvals so we don't complicate the process anymore than it has to be.

DOBEN:
> @DOBEN @RafaelAPB
> So, I was about to merge this just now, but I see @RafaelAPB came in with some extra suggestions post-approval (which is cool).
> Is it okay for you to discuss these comments/suggestions and then if there's any changes agreed on do that in a follow-up PR instead? I want to close this one down with the merge now that it has enough approvals so we don't complicate the process anymore than it has to be.


Yes, since the pull request is already quite long. I feel, it might be better to complete it and follow up with more PR requests for any changes/additional paragraphs rather than expanding this pull request.
@Rafael: Thank you for the additional comments. Let’s keep the discussion going to find new pinpoints to work on. I will reply to the comments, so we can clarify and work on these ideas.

whitepaper/whitepaper.md
```diff
@@ -124,12 +133,248 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
...
+of their underlying blockchain and sign the result. Outside entities (verifier nodes)
+can verify these signed claims of the state of the blockchain since they have the public keys of the validator nodes. Since the claim of the state of the blockchain was produced by several validator nodes<sup>[c](#12-footnotes)</sup> with respect to the rules of the consensus algorithm, a networkwide ledger view is achieved.
+
+
+
+
+
+<img src="./cactus_overview.png" width="400"><br>Figure description: V1, V2, and V3 are validator nodes which connect through the SDK to the underlying blockchain network nodes N1, N2, and N3, respectively (red line). V1, V2, and V3 run a consensus algorithm which is independent of the consensus algorithm run by the nodes N1, N2, N3, N4, and N5.</img>
+
+## 1.1 Terminology of Blockchain Interoperability
```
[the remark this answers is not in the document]
DOBEN: @sfuji822: I was just wondering, if I do any changes now, would that mean that the whole approve process has to start again?

whitepaper/whitepaper.md
```diff
@@ -124,13 +134,235 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
 
 We propose a protocol and it's implementation to connect as many of them as possible in an attempt to solve the fragmentation problem by creating a heterogeneous system architecture <sup>[1](#7-references)</sup>.
 
-# 2. Example Use Cases
+# 2. Introduction to Blockchain Interoperability
+
+There are two inherent problems<sup>[a](#22-footnotes-introduction)</sup> that have to be solved when connecting different blockchains: 
+* How to provide a proof of the networkwide<sup>[b](#22-footnotes-introduction)</sup> ledger state of a connected blockchain from the outside?
+* How can other entities verify a given proof of the state of a connected blockchain
+from the outside?
```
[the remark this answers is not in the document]
DOBEN: If a blockchain uses Merkle trees to store its state than anyone (even un-trusted parties) can provide a proof which can be verified through the Merkle tree root and techniques like simple payment verification without trusted third parties. But if the blockchain does not have an support to verify its state from the outside than trusted third parties are often needed. 

Yes, Cactus is mainly about technical interoperability. There was once the discussion about using the Read_the_docs framework for documenting Cactus (https://readthedocs.org/). It might give us the opportunity to include chapters/paragraphs focusing on areas and use cases beyond technical blockchain interoperability. The white paper is probably focusing mainly on technical blockchain interoperability since this is the main deliverable of Cactus.

whitepaper/whitepaper.md
```diff
@@ -124,13 +134,235 @@ Blockchain technologies are growing in usage, but fragmentation is a big problem
...
+
+The validator nodes are ledger-specific plug-ins, hence a smart contract on the connected blockchain can enable the ledger-specific functionalities necessary for a validator node to observe the ledger state to finalize a proof. The validator nodes are easier discovered from the outside than the blockchain nodes. Hence, the benefit of operating the `Cactus` network to enable blockchain interoperability relies on the fact that for any cross-blockchain interaction the same type of validator node signatures can be used. That means, the cross-blockchain interaction can be done canonically with the validator node signatures in  `Cactus` rather than having to deal with many different ledger-specific types of blockchain node signatures.
```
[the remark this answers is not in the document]
DOBEN: Rather than having to discover for each blockchain their corresponding nodes, it is easier to discover all cactus validator nodes and then canonically achieve the blockchain interoperability through the cactus validator nodes proof producing capabilities. Therefore only the functionalities for discovering cactus validator nodes are needed in contrast to implementing for each blockchain the functionalities to discover their nodes and enabling these nodes to produce verifiable proofs about their state. I don’t know exactly if Cactus supports node discovery (e.g. through gossip protocol) or has a registry for all the validator nodes currently or if this is planned in the future.

The ledger-specific plug-ins have to provide the capabilities to produce proofs about the state of the connected blockchain. Smart contracts “can enable” this functionality but any other method is valid as well as you mentioned. I used “can enable” for that purpose because it is not mandatory to use smart contracts for the ledger-specific plug-ins.

whitepaper/whitepaper.md
DOBEN: The two bullet points are example implementations for “burning of assets”, but I assume, we can also include another list with the heading “example implementations for locking of assets” in the future. Since the concept of burning of assets is quite unique to blockchains, a new reader might wonder how to implement “burning of assets” whereas “locking of assets” seems to be more straight forward for a new reader to implement.

whitepaper/whitepaper.md
DOBEN: Cactus cannot rely natively on Merkle tree proofs since not all blockchains support it natively. Of course, it is possible to reproduce the complete state of a blockchain in a Merkle tree but it might cost you a lot of computational resources or storage if you only want to get a proof quite infrequently. It would kind of duplicate the world state of the blockchain but if that is not of your concern then yes a cactus validator node can be implemented to produce proofs of state of a blockchain by generating a Merkle tree.

whitepaper/whitepaper.md
DOBEN: The main difference between atomic swaps and ledger transfers are the number of assets in each blockchain after the exchange. If there are 100 tokens in Blockchain A and 100 tokens in Blockchain B initially then after an atomic swap the number is still the same in each blockchain, but after a ledger transfer of one token, there are 99 tokens in Blockchain A and 101 tokens in blockchain B.
I am aware that “ledger interaction” is probably a quite broad definition and ledger transfers can be classified as a very special case of ledger interactions. Nonetheless, because ledger transfers and atomic swaps have been already quite widely used in cryptocurrency blockchains, they should be described as separate types in my opinion. To not bore the reader with any further sub-classification, I kept the definition of ledger interaction quite broad so it can include all the blockchain interoperability cases that are not ledger transfers/atomic swaps.

whitepaper/whitepaper.md
DOBEN: Since this an addition, do you think we can move this to a new PR request?

RafaelAPB:
@DOBEN @petermetz @jonathan-m-hamilton agreed!

whitepaper/whitepaper.md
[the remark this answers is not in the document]
RafaelAPB: Yes!

whitepaper/whitepaper.md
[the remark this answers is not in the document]
RafaelAPB: Thanks for clarifying, agreed! We need to make sure the validators can produce such proofs

whitepaper/whitepaper.md
RafaelAPB: I understand now what you meant. How would it duplicate the world state?

whitepaper/whitepaper.md
RafaelAPB: Alright, understood. The mentioned classification refers to those as well, and at a later point we may discuss if we want to adopt it. In my point of view, it helps a lot classifying the existing solutions in a very clear way.

PR #223, opened 2020-07-08
feat(plugin-web-service-oidc): add Open ID Connect auth plugin

petermetz:
Cockpit auth implicit flow with PKCE

PKCE (RFC 7636) is an extension to the Authorization Code flow
to prevent certain attacks and to be able to securely perform
the OAuth exchange from public clients.
It is primarily used by mobile and JavaScript apps, but the
technique can be applied to any client as well.

https://oauth.net/2/pkce/


petermetz:
Checklist:

- [ ] Dependency: consortium management
- [ ] Dependency: per plugin instance distinction for the IDs (right now all plugin instances have the same ID and they are not individually routable)
- [ ] Dependency: routing mechanism that is aware of individual plugin instances not just plugin IDs
- [ ] Define how per consortium member OIDC providers would work (persistent storage is the biggest question)
- [ ] Implement per consortium member OIDC providers
- [ ] PoC in the cockpit package that uses multiple consortium members

sfuji822:
Can you explain how OID will be used in Cactus at the meeting ?
I believe Open ID Connect is known as a single-sign-on functionality. I doubt the acceptance of foreign identities by a blockchain.

petermetz:
> Can you explain how OID will be used in Cactus at the meeting ?
> I believe Open ID Connect is known as a single-sign-on functionality. I doubt the acceptance of foreign identities by a blockchain.

Fair questions.

It is aimed to be an application layer concern for those who want it (so optional to use like any other plugin).
Will produce much more details about this, but the truth is that I don't have all the answers yet either, just making a lot of it up as I go. I'm deducting most answers on the fly from this vision that it should be easy(er) for a dapp author to use Cactus to create an end user facing application that can run cross-chain transactions on behalf of the end user's identity(es).
Stay tuned for more in the near future, but it has other dependencies that I need to work on before this can even get to the stage where I'm comfortable switching it from a 'draft' to a real PR (of course help/feedback is welcome at any point in time including right now)

sfuji822 (dismissed):
LGTM.

RafaelAPB:
@petermetz this looks awesome. So the idea is to allow end users to connect over OIDC to Cactus, and then possibly run transactions on behalf of a certain user registered on the blockchain correct? 
So we assume some kind of mapping OIDC - blockchain identity, similarly to what happens today with full-stack dApps built on permissioned blockchains, I suppose?

petermetz:
> @petermetz this looks awesome. So the idea is to allow end users to connect over OIDC to Cactus, and then possibly run transactions on behalf of a certain user registered on the blockchain correct?
> So we assume some kind of mapping OIDC - blockchain identity, similarly to what happens today with full-stack dApps built on permissioned blockchains, I suppose?

@RafaelAPB 100% correct!

petermetz:
Closing this down temporarily as part of the PR spring cleaning. We'll re-open it later once I have more time to work on it. We bridged the biggest gap with an intermediary PR that added JWT authZ (this is about authN)
cc: @travis-payne

PR #227, opened 2020-07-20
feat(examples): add example01-car-trade and its validatorDriver

takeutak:
This PR is related with the issue #178 as an development of API systems and related with the issue #181 .


examples/example01-car-trade/validatorDriver/src/app.ts
```diff
@@ -0,0 +1,123 @@
...
+    var param = req.body.param;
+    console.log('req.body : ' + JSON.stringify(req.body));
+    console.log('validator : ' + validator);
+    console.log('func : ' + func);
+    console.log('param : ' + param);
+
+    xxx(validator, func, param, res, next);
+});
+
+function xxx(validator, func, param, res, next): void {
```
petermetz, line 50: I personally don't have anything against having a function called xxx, but you might want to avoid it anyway because of this alternative meaning explained here: https://en.wikipedia.org/wiki/.xxx
@takeutak What do you think?

examples/example01-car-trade/validatorDriver/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "version": "1.0.0",
+  "private": true,
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc --watch"
+  },
+  "devDependencies": {
+    "typescript": "^3.9.3"
+  },
+  "dependencies": {
```
petermetz, line 12: @takeutak Shouldn't this use the existing cactus npm packages as dependencies? 
My concern: If the example writes everything from scratch then it's an example of how to create a piece of software like Cactus, not an example of how to use Cactus the software that was already written by someone.

examples/example01-car-trade/validatorDriver/src/app.ts
sfuji822: Whoops. The cultural gap is here. 
We should be used more safe strings like foo or bar > @takeutak

petermetz:
> Function names are changed

@takeutak Commit lint is still failing, could you please fix that?

PR #250, opened 2020-08-07
test-incorrect-pr-title validate PR title and ALL commits

petermetz:
If this actually works as advertised, then we can
enable squash commits from the repo settings
once again because it will no longer be possible
to make the mistake of squash-merging a PR and
ending up with the PR title as the commit
message while the PR title is not a valid commit
message (yes, this has happened in the past).

### DIFF
--- a/.github/.semantic.yml
+++ b/.github/.semantic.yml
@@ -1,0 +1,10 @@
+# Always validate the PR title AND all the commits
+titleAndCommits: true
+
+# Prohibit use of Merge commits (eg on github: "Merge branch 'master' into feature/ride-unicorns")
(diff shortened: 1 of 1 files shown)


sfuji822 (approved):
LGTM.

petermetz:
After running this PR for a while to evaluate edge cases, it seems like there are a few (edge cases) that the bots do not cover (case in point, this PR is marked green right now by the checks despite having multiple intentional issues included in it.

Based on this documentation here, another direction we can take is Github Actions which is a little more involved (might require custom code) but since the bots do not work 100% out of the box, I don't see how can we avoid that.
https://github.com/zeke/semantic-pull-requests/commit/805ca94ea2890aab2013b041901698aeaccae7e8

Issue #247, opened 2020-08-05
feat(common): KeyConverter class to and from PEM/hex/buffer 


petermetz:
### Description
As a programmer I want to have a utility class in the common package (which is cross platform) that can convert back and forth between a raw byte stream, a hex string and PEM formatted public/private keys so that it's easier to work with these cryptographic primitives when developing new, higher level functionality.

### Acceptance Criteria
1. Cover all directions/permutations among the three mentioned formats
2. Test coverage both in NodeJS and browser (write the tests with tape)

suyukuoacn:
Peter, I will work on this feature.

petermetz:
Fixed by #255, closing

PR #246, opened 2020-08-04
feat(common): Stable Signature Generation from JS Objects

suyukuoacn:
This PR resolve issue #136 

This PR depend on PR #245 and should only be merged after PR #245 merged.

PR #255, opened 2020-08-17
feat(common): KeyConverter class to and from PEM/hex/buffe

suyukuoacn:
This PR resolve issue #247 

This PR depend on PR #245 and #246 as it uses functionality and test from other PRs.

### DIFF
--- a/packages/cactus-common/src/main/typescript/key-converter.ts
+++ b/packages/cactus-common/src/main/typescript/key-converter.ts
@@ -1,0 +1,128 @@
+import KeyEncoder from "key-encoder";
+
+export enum KeyFormat {
+  Raw = "raw",
+  Hex = "hex",
+  PEM = "pem",
+}
+
+export class KeyConverter {
+  private keyEncoder: KeyEncoder;
+
+  public readonly supportedKeyFormats: string[];
+
+  constructor() {
+    this.supportedKeyFormats = Object.values(KeyFormat);
+    this.keyEncoder = new KeyEncoder("secp256k1");
+  }
+
+  /**
+   * Convert public key from one format to another format
+   * @param key
+   * @param fromFormat
+   * @param toFormat
+   */
+  public publicKeyAs(
+    key: string | Uint8Array,
+    fromFormat: KeyFormat,
+    toFormat: KeyFormat
+  ): string | Uint8Array {
+    this.validateKeyFormatValue(fromFormat);
+    this.validateKeyFormatValue(toFormat);
+
+    let keyValue: string = "";
+    let convertToRaw = false;
+
+    if (key instanceof Uint8Array) {
+      keyValue = Buffer.from(key).toString("hex");
+    } else {
+      keyValue = key;
+    }
+
+    if (fromFormat === KeyFormat.Hex) {
+      fromFormat = KeyFormat.Raw;
+    }
+
+    if (toFormat === KeyFormat.Raw) {
+      convertToRaw = true;
+    } else if (toFormat === KeyFormat.Hex) {
+      toFormat = KeyFormat.Raw;
+    }
+
+    let resultKey: string | Uint8Array = this.keyEncoder.encodePublic(
+      keyValue,
+      fromFormat,
+      toFormat
+    );
+
+    if (convertToRaw) {
+      resultKey = Uint8Array.from(Buffer.from(resultKey, "hex"));
+    }
+
+    return resultKey;
+  }
+
+  /**
+   * Convert private key from one format to another format
+   * @param key
+   * @param fromFormat
+   * @param toFormat
+   */
+  public privateKeyAs(
+    key: string | Buffer,
+    fromFormat: KeyFormat,
+    toFormat: KeyFormat
+  ): string | Buffer {
+    this.validateKeyFormatValue(fromFormat);
+    this.validateKeyFormatValue(toFormat);
+
+    let keyValue = key;
+    let convertToRaw = false;
+
+    if (fromFormat === KeyFormat.Raw) {
+      if (key instanceof Buffer) {
+        keyValue = key.toString("hex");
+      }
+    } else if (fromFormat === KeyFormat.Hex) {
+      fromFormat = KeyFormat.Raw;
+    }
+
+    if (toFormat === KeyFormat.Raw) {
+      convertToRaw = true;
+    } else if (toFormat === KeyFormat.Hex) {
+      toFormat = KeyFormat.Raw;
+    }
+
+    let resultKey: string | Buffer = this.keyEncoder.encodePrivate(
+      keyValue,
+      fromFormat,
+      toFormat
+    );
+
+    if (convertToRaw) {
+      resultKey = Buffer.from(resultKey, "hex");
+    }
+
+    return resultKey;
+  }
+
+  /**
+   * This method will validate if the input key format match to one of the enum value.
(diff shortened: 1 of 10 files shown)


packages/cactus-common/src/test/typescript/unit/key-converter.test.ts
```diff
@@ -0,0 +1,395 @@
...
+  KeyConverter,
+  KeyFormat,
+} from "../../../main/typescript/key-converter";
+
+const keyConverter = new KeyConverter();
+
+let privKey: any;
+// generate secp256K1 private key
+do {
+  privKey = crypto.randomBytes(32);
```
sfuji822, line 19: use real key which can be generated on the fly.

packages/cactus-common/src/main/typescript/key-converter.ts
```diff
@@ -0,0 +1,128 @@
+import KeyEncoder from "key-encoder";
+
+export enum KeyFormat {
+  Raw = "raw",
+  Hex = "hex",
+  PEM = "pem",
+}
+
+export class KeyConverter {
```
sfuji822, line 9: I am little confused for the usage of this class. I 
If your intended usage was for browser, this class should be placed at tools or new one like "cactus-client-sdk".
Additionally, I might be wrong, but Typescript has standard API to handle cryptographic key data like, CryptoKey. https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.cryptokey.html
I thought we better to have a functionality to check whether key is valid or not.

packages/cactus-common/src/main/typescript/key-converter.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * Convert public key from one format to another format
+   * @param key
+   * @param fromFormat
+   * @param toFormat
+   */
+  public publicKeyAs(
+    key: string | Uint8Array,
```
sfuji822, line 26: I might be wrong, but Typescript has standard API to handle cryptographic key data like, CryptoKey. https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.cryptokey.html
For nodejs, they has another set of API for handling keys, can you use them instead ? https://nodejs.org/api/crypto.html#crypto_crypto_createpublickey_key
I thought we better to have a functionality to check whether key is valid or not.

packages/cactus-common/src/main/typescript/signer-key-pairs.ts
```diff
@@ -0,0 +1,26 @@
+import crypto from "crypto";
+import secp256k1 from "secp256k1";
+
+export interface ISignerKeyPairs {
+  privateKey: any;
+  publicKey: any;
+}
+
+export class Secp256k1Keys {
```
sfuji822, line 9: Will it be only option for signing algorithm ?
I hope to change API to allow using another algorithm as well.

packages/cactus-common/src/main/typescript/signer-key-pairs.ts
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * Generate random private and public secp256k1 key in Buffer format
+   * @return Generated key pair
+   */
+  static generateKeyPairsBuffer(): ISignerKeyPairs {
```
sfuji822, line 14: I am afraid this is duplicated with the functionality of standard lib.

packages/cactus-common/src/test/typescript/unit/key-converter.test.ts
```diff
@@ -0,0 +1,395 @@
...
+const hexPublic = Buffer.from(pubKey).toString("hex");
+const pemPublic = keyEncoder.encodePublic(
+  Buffer.from(pubKey).toString("hex"),
+  "raw",
+  "pem"
+);
+const hexPrivate = privKey.toString("hex");
+const pemPrivate = keyEncoder.encodePrivate(
+  privKey.toString("hex"),
+  "raw",
```
sfuji822, line 37: I thought we should have more variants here. 
Int8Buffer, CryptoKey, encrypted key, ...

sfuji822 (changes requested):
I am afraid some class files are misplaced. I would like to hear assumed usage of those classes.
I also suggested some changes to align standard library like Web Crypto API or nodeJS crypto lib.

packages/cactus-common/src/main/typescript/key-converter.ts
```diff
@@ -0,0 +1,128 @@
+import KeyEncoder from "key-encoder";
+
+export enum KeyFormat {
+  Raw = "raw",
+  Hex = "hex",
```
sfuji822, line 5: I think original code https://github.com/blockstack/key-encoder-js defined this as 'der'. Can you use same value to avoid  confusion?

sfuji822 (commented):
Can you check my comment reagarding 'hex' format ?
If using 'hex' is correct way format, then I will revoke my comment.

packages/cactus-common/src/main/typescript/key-converter.ts
suyukuoacn: Currently Cactus’s key converter does not support 'DER' format.  

Here's list of format and enum for the Cactus and the external key encoder representation.

PEM Format - Cactus' enum 'PEM', - Key-Encoder's enum 'PEM'
DER Format - Cactus not support, - Key-Encoder's enum 'DER'
HEX Format - Cactus' enum 'HEX', - Key-Encoder's enum 'RAW'
Buffer/Unit8Array Format - Cactus' enum 'RAW', - Key-Encoder not support.

sfuji822: My comment intended to use a meaningful name for the format.
Actually key-encoder obviously supports DER(Distinguished Encoding Rule), which is official name for the format.
Please refer related description at https://github.com/blockstack/key-encoder-js#encoding-public-keys

sfuji822: OK. Sorry for my misunderstanding.r
Is there any reason to eliminate the DER format which widely is used at may key management implementations.
If you are not willing to support it, I can approve this PR as is.

sfuji822:
> I am afraid some class files are misplaced. I would like to hear assumed usage of those classes.
> I also suggested some changes to align standard library like Web Crypto API or nodeJS crypto 

I will withdraw my comment.

sfuji822:
> I am afraid some class files are misplaced. I would like to hear assumed usage of those classes.
> I also suggested some changes to align standard library like Web Crypto API or nodeJS crypto lib.

I will withdraw my comment.

sfuji822:
> Can you check my comment reagarding 'hex' format ?
> If using 'hex' is correct way format, then I will revoke my comment.

I will withdraw my comment see my new comment below.

petermetz:
@sfuji822 Why did you close the pull request? Misclick?

sfuji822 (approved):
I will create a new item on the issue list, and I will pass this PR.

yochliu8 (commented):
[[](https://github.com/hyperledger/cactus/pull/278)](https://github.com/hyperledger/cactus/pull/278)

petermetz:
> [](#278)

@yochliu8 Sorry I don't get it. Could you please clarify?

Issue #236, opened 2020-07-21
docs(whitepaper): improve related work


RafaelAPB:
**Description**
As a contributor/maintainer/developer I want to have specific information about the related work on the blockchain interoperability space. At this moment, the whitepaper provides little insights on what is happening on other projects.

**Acceptance Criteria**
The whitepaper exposes different approaches to the blockchain interoperability problem, and how Cactus relates to them

cc @petermetz @jonathan-m-hamilton @takeutak @sfuji822

dilumb:
If relevant, we cover lots of blockchain migration patterns & scenarios in use at [https://arxiv.org/abs/1906.00239](url)

RafaelAPB:
@dilumb it looks appropriate. Feel free to add it to the related work!

PR #258, opened 2020-08-24
docs(whitepaper): Expanded Blockchain Migration

dilumb:
Expanded Blockchain Migration to include both application and entire blockchain migration, and patterns and challenges. `Consortium A` is replaced with 'User A' where user can be a single user or consortium of users.
Added Ref. 4 as discussed in #236.
Fixed wording for consistency with rest of the write-up, e.g., destination --> target, asset --> asset and data
Signed-off-by: Dilum Bandara <dilum.bandara@ieee.org>


RafaelAPB:
I’ve read the proposal, LGTM

whitepaper/whitepaper.md
```diff
@@ -518,51 +518,53 @@ gets replaced by a traditional bank account holding USD.
...
-
+| Use Case Attribute Name    | Use Case Attribute Value |
+| -------------------------- | -------------------------------------------- |
+| Use Case Title             | Blockchain Migration |
+| Use Case                   | 1. `User A` runs an application/service on a source blockchain that is either permissioned or permissionless.<br>2. `User A` decides to use another blockchain instance (permissioned or permissionless) to run his/her application. <br>3. `User A` migrates the existing assets and data to the target blockchain.<br>4. `User A` runs the application/service on the target blockchain.
```
sfuji822, line 524: I am afraid I do not see any "User A" in sequence diagram. 
If the "User A" is not mere a end-user, it should be name as "Administrator A" or something else.
I also want to clarify what is the application described here. If application is intermediates the transfer asset across blockchain ledgers, the statement should be "integrated service application can access to both source blockchain and targeted blockchain".

whitepaper/whitepaper.md
```diff
@@ -518,51 +518,53 @@ gets replaced by a traditional bank account holding USD.
...
+| Interworking patterns      | Value transfer, Data transfer |
+| Type of Social Interaction | Asset and Data Transfer |
+| Narrative                  | `User A` who's application/service is operating on a source blockchain (e.g., a Hyperledger Fabric instance), would want to migrate the assets, data, and functionality to a target blockchain (e.g., Hyperledger Besu) to expand their reach. However, such migration requires lots of technical effort and time. The `Blockchain Migration` feature from `Cactus` can provide support for doing so, by connecting to the source and target blockchains, and performing the migration task.
```
sfuji822, line 527: I saw that proposed texts eliminated the description about therole of "Consortium A" and "Consortium B". I believe if the migration is intended to transfer the ownership or origination, both consortium should ask participant to authorize the transfer during preparation phase.
I also want to see text about how legacy data is treated on "targeted blockchain". 
I think it should have some reference to the transactions on the source blockchain and targeted blockchain continue to add new records for the future.

whitepaper/whitepaper.md
```diff
@@ -518,51 +518,53 @@ gets replaced by a traditional bank account holding USD.
...
+| Actors                     | 1. `User A`: The entity operating on the source blockchain which aims at performing a migration to the target blockchain. |
+| Goals of Actors            | 1. `User A`: Wants to be able to operate his/her application on the target blockchain while gaining its benefits such as better performance, lower cost, or enhanced privacy. |
+| Success Scenario           | Asset and data are available on the target blockchain enabling `User A`'s application to operate on the target blockchain.|
+| Success Criteria           | Migration starts at a set time, and all desired assets and data, as well as their histroy have been migrated in a decentralized way. |
+| Failure Criteria           | All desired assets and data, as well as their histroy not on the target blockchain. `User A` is unable migrate to target blockchain for a number of reasons, such as inability to recreate the same smart contract logic, inability to arbitary recreate native assets on target blockchain, and lack of access to private keys of external accounts. It may also be cost and time prohibitive to recreate entire history on the target blockchain. |
```
sfuji822, line 532: I think proposed text is too much.
I think any side effects should not be occurred on the failure considering the nature of transition.
Because transition is one-way operation, retrying the operation should be considered instead.

sfuji822 (changes requested):
Over all, the role of User A is not clear to me. I am afraid proposed text should be revised.

whitepaper/whitepaper.md
dilumb: Happy to update the sequence diagram.

Depending on the scenario `User A` can be a mere end-user of a blockchain, an administrator, or even the governance body of a blockchain. So 'User A` is more generic. Any suggestion on preferred terminology?

Application is essentially a DApp. It doesn't do anything to support/perform the migration. For e.g., we could consider application using Cactus to initially talk to source blockchain and after migration still using Cactus to talk to target blockchain. We could also envision application using Cactus only for the migration, and directly talking to target blockchain after migration. Should I use DApp or something else?
Thanks for suggestions.

whitepaper/whitepaper.md
dilumb: Could you please explain a bit about 1st comment. I'm not clear about "... transfer the ownership or origination". Also, by "preparation phase" do you mean the "Initialization phase" in sequence diagram?

Legacy data can be transferred to target blockchain under several data fidelity levels ranging from no transfer to entire history depending on the use case. As per our work, we found 5 data fidelity levels, namely Fresh Start (nothing transfer), State Only, State and Transactions, Genesis and Transactions, Blockchain History (everything transfer). Also, we found 5 cases of migration (Relocate, Upgrade, Consolidate, Separation, and Archive). So whether there's some reference to transactions on source blockchain depends on the application. Should I add such details?

whitepaper/whitepaper.md
dilumb: Does too much text refer to "Failure Criteria"? While retrying will mostly solve a problem, blockchain migration likely require a single-shot process rather than flushing target database if something goes wrong. Anyway, would the concern be addressed if I remove "It may also be cost and time prohibitive ..." sentence?

dilumb:
> Over all, the role of User A is not clear to me. I am afraid proposed text should be revised.

As per off-line discussion with Rafael, `User` is reverted back to `Consortium`. Based on same discussion, sequence diagram enhanced. Hope other comments are addressed too.

dilumb:
Accidentally "Closed with unmerged commits"

whitepaper/whitepaper.md
sfuji822: I thought the assets recorded on source blockchain will be invalidated under agreement between participants on both source and targeted blockchain.
I was not convinced on the simple operation can do it.
There are two problem here,: 
1) how the participants of targeted ledger can trust the imported records from source ledger, 
2) how the transferred asset, which should not be exist, can be realized as "invalid" by the participants on sourced ledger without knowing/referencing targeted ledger.  
I suggested you to add some procedure for solving above issues.

whitepaper/whitepaper.md
sfuji822: No. I meant if CACTUS application will try to recover the failure it should be described, but this particular use case should not be fail unless it was rejected by participants.
More specially,  what can make it fail in your use case ? I believe that the transition should not be never rejected if system is correctly configured. And if it should be rejected there are no need to record the fact of the operation on both ledgers.

whitepaper/whitepaper.md
sfuji822: I just wanted to be consistent with texts and sequence diagrams. 
You wrote "Application is essentially a DApp. It doesn't do anything to support/perform the migration. ", but it was your assumption. We may implement pair of DApps(smart contracts) which one can delete transferring asset with creating a proof of burn, and another import the asset referring the proof. I thought that is the reasonable assumption for linking the trust over multiple blockchains.

whitepaper/whitepaper.md
dilumb: Retry is needed and likely to be essential in setting multiple assets/data and transactions like in any distributed system. Hence, not mentioned explicitly. However, recording of transactions replayed on the target blockchain can lead to rejection or result in a state that's inconsistent with the state on source blockchain. This may not be fixed by retrying. I added "... e.g., when transactions replayed on target blockchain are reordered". 
Other cases are inspired from Failure Criteria in use case 3.4 and 3.7 and focus on under what conditions migration can be done. Happy to remove these if they don't fit.

whitepaper/whitepaper.md
dilumb: 1. Consortium users have to trust some intermediatory (gateway, relay, inter-blockchain connector, or oracle) that shows the source blockchain data/assets to the target, e.g., proof of burn. This is where Cactus can help. This is similar to use case 3.6 where ledger 1 gets what it's missing from leader 2 (vice versa) through Cactus.
2. Assets can be burned and smart contracts can be destroyed (both typically refer as burning).
Expanded Narrative to include following:
"Assets on the source blockchain are burned and smart contracts are destructed during the migration to prevent double-spending. Proof-of-burn is verified on the target blockchain before creating the assets, smart contracts, or their state. ... 
As assets are burned and smart contracts are destructed, any future attempt to manipulate them on the source blockchain will fail. ... 
The `Blockchain Migration` feature from `Cactus` can provide support for doing so by connecting source and target blockchains; proving values and proof-of-burn of assets, smart contracts, and data imported from the source blockchain to the target; and then performing the migration task."

whitepaper/whitepaper.md
dilumb: Yes, this is certainly needed to ensure proper linking, and proof-of-burn is already being used in asset migration. However, this isn't essential in data migration unless there's is a need to have only one copy of state that data represents. I was trying not to go into such details in the current write up, especially this is the use case than implementation. Such discussion belongs to Sec. 4 and 5. Hope the changes made in response to next comment now cover this comment too.

dilumb:
Narative and Failure Criteria further updated based on @sfuji822 comments around burning assets on source blockchain and establishing trust.

whitepaper/whitepaper.md
sfuji822: IMHO, history of record cannot be transferred unless all participants of the sourced ledger are agreed to migrate the ledger hosting system. But think again, why the migration process is needed for this case.
That is an implementation issue , how old data format can be converted with new one, which is not interoperability.
I suggest you focusing on asset level migration only.

whitepaper/whitepaper.md
sfuji822: OK. I can accept them with assuming further works will be done in the future.

whitepaper/whitepaper.md
sfuji822: I believe 'some intermediatory' should be authorized by the consortium member with obvious consensus or agreement which might be outside of the system. Can you add some text to clarify this fact?

For your comment-2,  that can be happened only when all members agreed on burning/creating asset across source/destination leader with steps. 
1-user request smart-contract on destination ledger (via CACTUS) to transfer his/her asset, but it will be pending until confirmation or authorization is given from smart-contract on sourced ledger.
2-user request smart-contract on sourced ledger (via CACTUS) to burn his/her asset
3- smart-contract will be notified confirmation or authorization from sourced ledger smart-contract 
4- smart-contract on destination ledger create his/her asset, as the response to confirmation/authorization on step3)

I would like to ask you to rewrite the section to capture the above observation.

dilumb: 1. Add a sentence on use of intermediatory.
2. Added 4 steps suggested

whitepaper/whitepaper.md
dilumb: Remove all discussion related to history in multiple places.
Anyway, about this topic - I was thinking of recreating the same history by replaying the entire transaction trace (if someone really really need it). That way, any data format incompatibility issues can be addressed too. Even in our paper, we don't recommend this option due to cost, time, and even impracticality. If source blockchain is decommissioned after migration, having a copy of source blockchain history as a proof of exist entry is the easiest.

dilumb:
@sfuji822's comments are addressed

sfuji822 (approved):
LGTM

PR #275, opened 2020-09-05
feat(common): Checks and CodedError classes

### LINKED ISSUE #266: feat(common): preconditions utility with coded errors
### Description
As a developer I want to have an easier way to assert truthy-ness of values so that I'm more productive in general and less likely to omit/forget crucial validation of method input parameters for example.

There's a lot of "manual" validation in the code as of right now and it would be nice to have something instead that is similar to Guava's `Preconditions` class just for Typescript so that we can replace these

```typescript
    const fnTag = "DeployContractGoBinEndpointV1#constructor()";
    if (!options) {
      throw new Error(`${fnTag} options falsy.`);
    }
    if (!options.path) {
      throw new Error(`${fnTag} options.path falsy.`);
    }
```

with the much more concise version like this:
```typescript
    const fnTag = "DeployContractGoBinEndpointV1#constructor()";
    Checks.truthy(options, `${fnTag} options`);
    Checks.truthy(options.path, `${fnTag} options.path`);
```

### Acceptance Criteria
1. Create a class called `Checks` and another one called `CodedError`
2. Checks can do general assertions for truthy-ness and throw `CodedError` instances which are extended from the standard `Error` but add a `code` property which can make it easy to detect and equality compare two `CodedError` classes even if their `message` property varies because it has something in it that is a variable (like an ID of a transaction block or whatever else that's not static strings)
3. Both of these classes must work in the browser and have test coverage.

petermetz:
Utilities for easier one-liners that can verify arguments or
execution state in a much more concise way compared to
if conditions throwing manually. Makes the code less verbose
and requires less typing as well.

Also: added a bools utlity class for checking strict boolean
typing of a value and a new string utility method as well.

Fixes #266

### DIFF
--- a/packages/cactus-common/src/main/typescript/bools.ts
+++ b/packages/cactus-common/src/main/typescript/bools.ts
@@ -1,0 +1,15 @@
+export class Bools {
+  /**
+   * Determines if a value is strictly a boolean `true` or `false`. Anything else
+   * will result in the method returning `false`.
+   *
+   * Useful in cases where you have an optional boolean parameter that you need
+   * to assign a default value to by determining if it had been set or not.
+   *
+   * @param val The value to be decided on whether it's strictly boolean `true`
+   * or `false`.
+   */
+  public static isBooleanStrict(val: unknown): boolean {
+    return val === true || val === false;
+  }
+}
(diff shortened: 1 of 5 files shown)


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):
LGTM

PR #276, opened 2020-09-05
feat(common): Servers utility class

### LINKED ISSUE #260: refactor(cmd-api-server): move Servers class to common package
### Description
As a plugin author I want to have access to the `Servers` utility class so that I can use it to deal with HTTP/S servers easier.
An example is when wanting to shut down an HTTP/S server with a one liner method call that returns a promise (lacking that, you have to manually promisify the shutdown call).

The class currently is at:
`packages/cactus-cmd-api-server/src/main/typescript/common/servers.ts`

And it should be under:
`packages/cactus-common/src/main/typescript/servers.ts`

### Acceptance Criteria
1. Class is moved
2. Dependent code is updated in all packages to use the new import from the common package
3. Ensure that the common package only exports the Servers class under the node target (index.ts) and not for the web.

### LINKED ISSUE #267: feat(common): servers utility class for HTTP/S
### Description
As a developer who works a lot on different plugins that expose web services I want to have a utility class starting/shutting down HTTP/S servers in a promisifed way so that I can start/shutdown a server as needed with one-liners instead of 15 lines of callback hell.

There is already a `./common/servers.ts` file in the api server package but that should be moved to the common package instead and be extended with some additional functionality.

### Acceptance Criteria
1. Move `Servres` class to common package
2. Make sure to only export the Servers class for the node builds never the web ones.

petermetz:
Useful when working with HTTP servers especially for
test case authoring where there's a lot of server
creation for the purpose of allocating random ports
prior to launching an API server instance.

Fixes #260
Fixes #267
Depends on #275 


petermetz:
Depends on #275

jonathan-m-hamilton (approved):
LGTM

takeutak:
Could we discuss this component again at next week's Contributors' meeting?
I would like to discuss with you especially about the API processing of shutdown in components.

petermetz:
> Could we discuss this component again at next week's Contributors' meeting?
> I would like to discuss with you especially about the API processing of shutdown in components.

@takeutak Sure, happy to discuss. Please also note: We could also discuss here in the comments. It would be great if we could reduce the average turnaround time for the PRs by not having to wait for the weekly calls with Q/As (I understand sometimes it's a lengthier discussion that is not well suited for comments, but in other cases it ends up being just a few short sentences anyway)

takeutak (approved):
LGTM

PR #277, opened 2020-09-05
feat(test-tooling): Containers class ls() and exec() methods

petermetz:
Helpers to list files under an arbitrary path in a container's file
system and a more generic method that helps executing shell
commands (exec())

Depends on #275


jonathan-m-hamilton (approved):
LGTM

sfuji822 (approved):
LGTM.

PR #281, opened 2020-09-07
refactor(BLP,LP,RIF,examples): divide components into packages and ex…

takeutak:
…amples


petermetz:
@takeutak If you are still drafting this then I recommend considering setting the PR state to 'draft' that way we get a clearer view of what's ready for review and what isn't.
This will be more important once we start getting a larger number of PRs and have to scale up the review operations, but it is a good habit/practice to do in general IMO.

takeutak:
Thank you for Peter's advice!  Then I revised this PR.  But, I'm sorry that in this PR I mistook to push my PR from my master branch, not my temporary branch.
Then I just made another PR #326 , which includes the revision of this PR.  So I closed this PR at once.

PR #283, opened 2020-09-15
fix(test-tooling): getContainerInfo methods lookup criteria

### LINKED ISSUE #265: fix(test-tooling): getContainerInfo methods lookup criteria
**Describe the bug**

Currently all the test ledger implementation classes have a method like this (or similar) that look up the information about their own container by selecting it from the list of all running containers based on the image name.
This works when there's only ever a single container of that image is running but breaks when there are 2 or more containers with the same image (when tests are running in parallel for example).

**To Reproduce**

Read the code here and observe the issue:

As seen in the test ledger classes:

```typescript
  protected async getContainerInfo(): Promise<ContainerInfo> {
    const fnTag = "FabricTestLedgerV1#getContainerInfo()";
    const docker = new Docker();
    const image = this.getContainerImageName();
    const containerInfos = await docker.listContainers({});

    const aContainerInfo = containerInfos.find((ci) => ci.Image === image);

    if (aContainerInfo) {
      return aContainerInfo;
    } else {
      throw new Error(`${fnTag} no image "${image}"`);
    }
  }
```

**Expected behavior**

Container look ups are done in a way that makes it impossible to return the wrong container, e.g. based on the container ID for example.
When the container gets started/created, the test ledger class should save the resulting ID of it as a pointer to the actual container.

Also, it would be nice to pull this implementation to a more generic place such as a the `Containers` helper class and maybe rename it at the same time to something like `getContainerInfoById(id: string): Promise<ContainerInfo>`.

769c087af524451a1a7c6860ee5cb6d6de08461f

cc: @jonathan-m-hamilton @sfuji822 @takeutak @jweate @suyukuoacn @RafaelAPB

suyukuoacn:
Fixes #265

Signed-off-by: suyukuoacn <su-yu.kuo@accenture.com>

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts
@@ -11,14 +11,14 @@ export interface IBesuTestLedgerConstructorOptions {
   containerImageVersion?: string;
   containerImageName?: string;
   rpcApiHttpPort?: number;
-  envVars?: string[]
+  envVars?: string[];
 }
 
 export const BESU_TEST_LEDGER_DEFAULT_OPTIONS = Object.freeze({
   containerImageVersion: "latest",
   containerImageName: "hyperledger/cactus-besu-all-in-one",
   rpcApiHttpPort: 8545,
-  envVars: ["BESU_NETWORK=dev"]
+  envVars: ["BESU_NETWORK=dev"],
 });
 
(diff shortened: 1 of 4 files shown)


petermetz (approved):
LGTM

jonathan-m-hamilton (approved):
LGTM

sfuji822:
I felt that having control over the testing container is out of scope on CACTUS, since I considered it is violating governance of blockchain unless it was testing purposes only.
Can we discuss it during the CACTUS meeting?

takeutak (approved):
LGTM

sfuji822 (approved):
LGTM.

PR #288, opened 2020-09-24
build(typescript): upgrade to v4 compiler

### LINKED ISSUE #241: build: upgrade to Typescript 4.x
### Description

As a contributor I want to be able to use the variadic tuple types feature of the compiler so that application developers who use the project can get the best developer experience when using the SDK for example.

The above example is not just a generic one, there is a specific point in the source code at present where the variadic tuple types would be great: The ApiClient class of the SDK package (`./packages/cactus-sdk`) has a method dedicated to extending the API client object with other API clients (the `DefaultApi` classes generated by the `typescrit-axios` profile of the OpenAPI generator tool)

Feature announcement: https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#variadic-tuple-types

### Acceptance Criteria
1. Upgrade the compiler everywhere in the project
2. Tests are passing of course
3. There's no massive refactoring necessary for the change to happen (if it has to be done then open separate issues about those and tackle them prior to even attempting the upgrade)
4. Wait for the 4.0 *stable* version to come out first. It is possible that currently it is still in beta, but that is expected to change very soon if the past release cycles of Typescript are to be considered clues in that regard.

petermetz:
Fixes #241

The OpenAPI generator also had to be upgraded
because the code it generated with the older
version was not TS 4 compatible (that's why
there's new files in the generated folders)


jonathan-m-hamilton (approved):
LGTM

sfuji822 (approved):
LGTM.　
Can you explain the potential risk of migrating the V4 compiler as default?

petermetz:
> LGTM.　
> Can you explain the potential risk of migrating the V4 compiler as default?

Thank you @sfuji822 

The risk is that one might have to update the syntax in some places with the new version because technically major releases can have breaking changes, but in my experience, with Typescript this is rarely the case (been through all the major releases they had since the beta and only 1 and 2 were bigger changes, 3 and 4 were pretty much seamless).
For example in this case 99% of the patch is just about upgrading the auto-generated source, nothing that I had to do manually (the whole PR took very little time to prepare compared to what one would normally expect)

PR #298, opened 2020-09-29
docs(whitepaper): refactor sequence diagram

takeutak:
Related with Issue #178 .


whitepaper/communication-protocol-between-blp-and-lp.puml
```diff
@@ -16,82 +21,146 @@ end box
...
+rif --> admin: startService() OK
 deactivate rif
+...
 
 == Send a ledger operation request ==
 
-blp -> blp: event triggered
+euser -> rif: CREATE a trade
+activate rif
+rif -> rif: assign tradeID
```
petermetz, line 60: Would this mean that the routing interface has some sort of distributed storage associated with it so that it can manage the trade IDs?

petermetz (commented):
General question for the communication protocols: Can we make this integrate with the OpenAPI spec that powers the client library generation?

whitepaper/communication-protocol-between-blp-and-lp.puml
takeutak: Yes. On my assumption, the routing interface needs some storage and it can manage the trade IDs.
Since tradeID is the only way to identify a businessLogicID, I'm sure that there should be a mechanism to properly route LedgerEvents that are uniquely assigned and notified externally and internally.
I think that this feature is hard to implement outside of the routing interface, is there an alternative plan?

petermetz: > Yes. On my assumption, the routing interface needs some storage and it can manage the trade IDs.
> Since tradeID is the only way to identify a businessLogicID, I'm sure that there should be a mechanism to properly route LedgerEvents that are uniquely assigned and notified externally and internally.

@takeutak Thank you for clarifying!

> I think that this feature is hard to implement outside of the routing interface, is there an alternative plan?

I agree it's hard to implement and I'm not against having it in the routing interface either, but do have a suggestion to somehow make it more generic which might be done just by simply naming the `tradeID` something more generic like `subscriptionId` or `operationId` or `correlationId` or `routingKey`, etc. the idea being that there can be use cases that do not involve actual trades but still need the same kind of functionality where observing clients need to hold onto a unique identifier for the routing purposes.
Do you think that renaming could make sense as I outlined above? (names are just ideas)

Do you have anything specific (code/technology stack/dependencies, etc) about the storage implementation for the routing interface already or is it just the design vision for now that's shown on the diagram? I don't want to hijack this thread with the implementation details because this is about the design, but would love to continue the discussion about the actual implementation as well.

takeutak:
Thanks for Peter's comments.  I concretized the function names on the sequence as the above commits.

petermetz (approved):
> Thanks for Peter's comments. I concretized the function names on the sequence as the above commits.

@takeutak 
My pleasure. 

I'm not a 100% sure I fully get the trade ID persistence mechanism. Tried to locate some code, but in the [other PR (281)](https://github.com/hyperledger/cactus/pull/281) it says 'unknown repository' in the place of the branch reference for some reason.
Probably will be easier to interpret this once I can read the code behind it as well, so until then I'll just say LGTM to this so we don't hold up each other's progress.

jonathan-m-hamilton (approved):
LGTM!

PR #301, opened 2020-10-02
chore(contribs/accenture-bif): delete old code

### LINKED ISSUE #300: chore(vendored code): replace GPL code with Apache 2.0 code
**Describe the bug**

crypto/secp256k1 library and libsecp256k1 is vendored code from a GPL source.

petermetz:
Fixes #300

Deleted the whole folder with the old simple
asset transfer examples instead of applying a
targeted fix because this is the lowest effort
solution that gets the job done and we want
to focus on moving forward not keep fixing
up the old code that's really just there for
posterity's sake. This will also break some
links that are out there on the internet but
the sooner we fix those the better anyway.

Considered as an alternative just removing
the fabric specific parts of the old simple asset
transfer code but that also would've ended up
being a surgical move because of the dependencies
within the scripts that do the foriegn validator
key pre-sharing so instead of turning this into
a rabbit hole I'd just opted to take the whole
thing out and spend the saved up time on making
sure that we have the simple asset transfer
example done within the new codebase.


petermetz:
cc: @ryjones @jonathan-m-hamilton

ryjones:
@swinslow LGTM

jonathan-m-hamilton (approved):
LGTM

ryjones (approved):
LGTM

PR #311, opened 2020-10-13
docs: update README

RafaelAPB:
The current README presents potentially outdated information that holds details that are better suitable for the whitepaper. It lacks instructions on how to build and run the project, which can frustrate newcomers. 

I would propose us to simplify the README of the project, and as the example section matures, provide a practical example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Hyperledger Cactus is currently undergoing a major refactoring effort to enable 
 
 ## Scope of Project
 
-As blockchain technology proliferates, blockchain integration will become an increasingly important topic in the broader blockchain ecosystem.  For instance, people might want to trade between multiple different blockchains that are run on different platforms. The blockchain integration framework is a web application system designed to allow users to securely integrate different blockchains. It includes a set of libraries, data models, and SDK to accelerate development of an integrated services application. Our goal is to build a system that allows users of our code to securely conduct transactions between all of the most commonly used blockchains.  
+As blockchain technology proliferates, blockchain integration will become an increasingly important topic in the broader blockchain ecosystem.  For instance, people might want to trade between multiple different blockchains that are run on different platforms. The blockchain integration framework is a web application system designed to allow users to securely integrate different blockchains. It includes a set of libraries, data models, and SDK to accelerate development of an integrated services application. Our goal is to deliver a system that allows users of our code to securely conduct transactions between all of the most commonly used blockchains.
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


RafaelAPB:
We can add an updated version of the architecture, and forward readers to the whitepaper, as well

petermetz (approved):
LGTM

petermetz:
@RafaelAPB Would be nice to have a more specific commit message. `Update README` can mean a lot of different things. I'd say something like `docs(readme): reference build instructions` for example.

RafaelAPB:
I've chosen that specific message because the commit also removes parts of it. Would it still be reasonable to only refer the build instructions?

petermetz:
> I've chosen that specific message because the commit also removes parts of it. Would it still be reasonable to only refer the build instructions?

Tough question. It definitely makes sense IMO to not just say `updated readme`, but how much explanation should fit in the commit subject line (the top line) is difficult to say. 
My personal take: I'd use the commit body (e.g. lines after the first) to explain the rest of the story like that the theoretic explanations have been removed in favor of the rest of the documentation which contains the same information. 
Something like that. :-)

RafaelAPB:
Makes sense. Let's see first what the other reviewers think.

takeutak (changes requested):
This is about the change request mentioned in the commit below.
https://github.com/hyperledger/cactus/pull/311/commits/3bfab322e0180b3a3176151fe2d2e5a90d8a5092

takeutak (commented):
I have strengthened your proposal as the commit below.  The reason is written in the header of the commit below.  Do you have any comments?
https://github.com/hyperledger/cactus/commit/3bfab322e0180b3a3176151fe2d2e5a90d8a5092

RafaelAPB:
I agree, @takeutak ! Thanks!

petermetz:
CI is failing because of an unrelated problem: https://github.com/hyperledger/cactus/issues/319

takeutak (approved):
LGTM

jonathan-m-hamilton (approved):
LGTM, pending CI resolution

RafaelAPB:
Done @petermetz , thanks!

takeutak:
Excuse me, it look like that the current README ignores my suggestion.
Please wait for merging before solving the issue.

takeutak (changes requested):
As I said the above, the current status ignores my above commit.  Then I canceled my LGTM at once.

takeutak:
I pushed the following commit, which content is the same as the above my commit.
https://github.com/hyperledger/cactus/pull/311/commits/ab2542d4b0e2d53a95c19fa4c0270bca01319ff6

Now I approved this PR again.

takeutak (approved):
LGTM

PR #308, opened 2020-10-12
docs(examples): add example working in new architecture

petermetz:
Fixes #189 

**Depends on #337 => meaning that #337 must be merged first prior to this getting merged.**

Huge PR that incorporates all the remaining changes/features necessary to have an actual working cross ledger asset transfer example (not an atomic swap, just a simple one).

It contains changes in several different packages that got piled up together to shorten the development time instead of slowly iterating with smaller but more digestible pull requests (a trade off made to move things along faster in the initial phases of development where we don't yet have a stable release).

The interesting changes are in the last commit that actually adds the example, the older commits of the PR are not specific to the example code but had to be put there because the example code depends on them.

Recommended to start reviewing by testing it first via the instructions placed in the README.md file.

The example does not feature asset transfers among ledgers (export) just shows how different contracts on different ledgers can be used to query and join data from different ledgers. This is a shortcut I had to take to be able to finish with the PR in reasonable time (well, it's still already way later than it should've been, but it is what it is).

Once you launch the example it will show you a web UI that you can click around on to create data on both ledgers that are involved with the example (quorum, besu).

Will need to add more documentation in subsequent commits and also update the whitepaper.

This PR contains the backbone of all the changes we need to push through with the v0.2.0 release (there are a few minor issues still left after this but those are just quick administrative ones that are related to documentation links and such).

cc: @takeutak @jonathan-m-hamilton @sfuji822

PR #337, opened 2020-11-02
feat(example): plumbing, intra-node routing for #308

petermetz:
As per the change request: https://github.com/hyperledger/cactus/pull/308#issuecomment-720195413

> Moreover, I don't feel comfortable because the title of your PR is "docs" and this PR includes not only documents but also actual codes. Would you mind if you divide your PR into docs(example) and feat(example)?

cc: @takeutak


jonathan-m-hamilton (approved):
LGTM

takeutak:
What are the codes to change this pull-request?  Are they all 166 codes on "Files changed" or the part filtered with some commit name?

changelog.config.js
```diff
@@ -0,0 +1,92 @@
...
+    "breaking",
+    "issues",
+    "lerna",
+  ],
+  scopes: [],
+  types: {
+    build: {
+      description:
+        "Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)",
+      emoji: "🏗",
```
takeutak, line 33: These all emojis on this file are very confusing to me.  Would you mind if you delete these emojis?

takeutak (commented):
@petermetz I almost agree to these pull-request. But I'm wondering the config of emojis on changelog.config.js. Would you mind if you delete these config of emojis?

takeutak:
@petermetz Moreover, as above, I think there will be no problem except the changelog setting. However, from the next time, when issuing multiple Pull requests, please do not create Pull requests that depend heavily on other Pull requests, such as by separating branches. To be honest, it is very difficult to review.

changelog.config.js
petermetz: @takeutak Done: https://github.com/hyperledger/cactus/pull/337/commits/30ad1b42c5a749b62f7eb26f74c6796d9e39e8f8

petermetz:
> @petermetz I almost agree to these pull-request. But I'm wondering the config of emojis on changelog.config.js. Would you mind if you delete these config of emojis?

@takeutak Done: 30ad1b4

petermetz:
> However, from the next time, when issuing multiple Pull requests, please do not create Pull requests that depend heavily on other Pull requests, such as by separating branches.

@takeutak If we fix the root cause, then I can stop submitting dependent PRs, yes, but until then this is the only solution I could come up with so far. (always open to alternative solutions of course)
Root cause: The average PR review time is 2 to 3 weeks even for PRs where I only change a few lines of code. Without the dependent PRs, I would be forced to spend 80% of my time idling or just spend the same time rebasing+conflict resolving dozens of branches daily.

> To be honest, it is very difficult to review.

@takeutak I hear you. Could you please point out which step is the difficult one from the step by step guides I provided in the past? I'm happy to expend on said steps, go deeper into specific parts of it if you give me some pointers.

Also happy to do pair programming style reviews where we do screen sharing and I assist with performing the reviews via the either the GitHub UI (or other preferred tool such as VSCode)

takeutak:
LGTM

takeutak (approved):
LGTM.
@petermetz Please make the same correction for other dependent Pull requests. Otherwise, when merging dependent pull-requests, the modifications will be reversed by mistake.

petermetz:
> LGTM.

@takeutak Cheers!

> @petermetz Please make the same correction for other your dependent pull requests before they are merged. Otherwise, when merging dependent pull-requests, the modifications will be reversed by mistake.

@takeutak Yeah, thankfully, rebase takes care of that automatically (assuming no conflicts) It's a big part of the reason why I strongly prefer rebase to merges in general.

PR #356, opened 2020-11-04
chore: fix package scope  typo

petermetz:
Instead of Hyperledger Labs, the package scope should
be just Hyperledger since we've graduated from  being
a lab into a top level project.


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):
LGTM, because the PR is only fixing the typo.

PR #368, opened 2020-11-05
docs(whitepaper): updates on architecture diagram and terminology

sfuji822:
This PR resolves the issue regarding out-of-date documentations.

Signed-off-by: Shingo Fujimoto <shingo_fujimoto@fujitsu.com>

### DIFF
--- a/whitepaper/cacts_arch.svg
+++ b/whitepaper/cacts_arch.svg
@@ -1,0 +1,1777 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="210mm"
+   height="297mm"
+   viewBox="0 0 210 297"
+   version="1.1"
+   id="svg8"
+   sodipodi:docname="cacts_arch.svg"
+   inkscape:version="0.91 r13725"
+   inkscape:export-filename="C:\Users\shingo_fujimoto\Desktop\cactus_arch2020-10-26.png"
+   inkscape:export-xdpi="96"
+   inkscape:export-ydpi="96">
+  <defs
+     id="defs2">
+    <marker
+       inkscape:stockid="TriangleOutL"
+       orient="auto"
+       refY="0.0"
+       refX="0.0"
+       id="marker129997"
+       style="overflow:visible"
+       inkscape:isstock="true">
+      <path
+         id="path129999"
+         d="M 5.77,0.0 L -2.88,5.0 L -2.88,-5.0 L 5.77,0.0 z "
+         style="fill-rule:evenodd;stroke:#000000;stroke-width:1pt;stroke-opacity:1;fill:#000000;fill-opacity:1"
+         transform="scale(0.8)" />
+    </marker>
+    <marker
+       inkscape:isstock="true"
+       style="overflow:visible"
+       id="marker116049"
+       refX="0.0"
+       refY="0.0"
+       orient="auto"
+       inkscape:stockid="TriangleOutL">
+      <path
+         transform="scale(0.8)"
+         style="fill-rule:evenodd;stroke:#000000;stroke-width:1pt;stroke-opacity:1;fill:#000000;fill-opacity:1"
+         d="M 5.77,0.0 L -2.88,5.0 L -2.88,-5.0 L 5.77,0.0 z "
+         id="path116051" />
+    </marker>
+    <marker
+       inkscape:isstock="true"
+       style="overflow:visible"
+       id="marker109647"
+       refX="0.0"
+       refY="0.0"
+       orient="auto"
+       inkscape:stockid="TriangleOutL"
+       inkscape:collect="always">
+      <path
+         transform="scale(0.8)"
+         style="fill-rule:evenodd;stroke:#000000;stroke-width:1pt;stroke-opacity:1;fill:#000000;fill-opacity:1"
+         d="M 5.77,0.0 L -2.88,5.0 L -2.88,-5.0 L 5.77,0.0 z "
+         id="path109649" />
+    </marker>
+    <marker
+       inkscape:isstock="true"
+       style="overflow:visible"
+       id="marker105869"
+       refX="0.0"
+       refY="0.0"
+       orient="auto"
+       inkscape:stockid="TriangleOutL">
+      <path
+         transform="scale(0.8)"
+         style="fill-rule:evenodd;stroke:#000000;stroke-width:1pt;stroke-opacity:1;fill:#000000;fill-opacity:1"
+         d="M 5.77,0.0 L -2.88,5.0 L -2.88,-5.0 L 5.77,0.0 z "
+         id="path105871" />
+    </marker>
+    <marker
+       inkscape:isstock="true"
+       style="overflow:visible"
+       id="marker105203"
+       refX="0.0"
+       refY="0.0"
+       orient="auto"
+       inkscape:stockid="TriangleOutL">
+      <path
+         transform="scale(0.8)"
+         style="fill-rule:evenodd;stroke:#000000;stroke-width:1pt;stroke-opacity:1;fill:#000000;fill-opacity:1"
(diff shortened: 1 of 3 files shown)


whitepaper/cacts_arch.svg
```diff
@@ -0,0 +1,1777 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
petermetz, line 1: @sfuji822 This looks like it has the same content as `./whitepaper/cactus_arch.svg`, do we really need this or could it be deleted to avoid dupliation?

whitepaper/whitepaper.md
```diff
@@ -706,19 +706,25 @@ Once an API call to Hyperledger Cactus framework is requested by a User, Busines
...
-<img src="./architecture-with-plugin-and-routing.png" width="700">
+<img src="./cactus_arch.svg" width="700">
 
 ### 5.2.1 Definition of key components in system architecture
 
 Key components are defined as follows:
-- **Application user**: The entity submits API calls to "Cactus Routing Interface". Note: this component is exist outside of Cactus service system.
+- ~~**Application user**: The entity submits API calls to "Cactus Routing Interface". Note: this component is exist outside of Cactus service system.~~
 - **Business Logic Plugin**: The entity executes business logic and provide integration services that are connected with multiple blockchains. The entity is composed by web application or smart contract on a blockchain. The entity is a single plugin and required for executing Hyperledger Cactus applications.
+- **Command API Server (cmd-api-server)**: The server accepts a request from an End-user Application, and return a response depending on the status of the targeted trade. Trade ID will be assigned when a new trade is accepted.
```
petermetz, line 716: > The server accepts a request from an End-user Application, and return a response depending on the status of the targeted trade. Trade ID will be assigned when a new trade is accepted.

@sfuji822 Could it be that this description (quoted above) is better suited for a specific plugin than the entirety of the API server? And for the latter, potentially we could phrase it more generically with a focus on it being the host/shell/container/etc. for all the plugins?

whitepaper/whitepaper.md
```diff
@@ -706,19 +706,25 @@ Once an API call to Hyperledger Cactus framework is requested by a User, Busines
...
+- **End-user Application**: The entity submits API calls to request a trade, which invokes a set of transactions on Ledger by the Business Logic Plugin.
```
petermetz, line 717: > The entity submits API calls to request a trade, which invokes a set of transactions on Ledger by the Business Logic Plugin.

@sfuji822 Would it make sense to mention something more generic than requesting a `trade`? For example the end user application could submit an API call to request information (e.g. querying ledger state without modifying it).

whitepaper/whitepaper.md
```diff
@@ -706,19 +706,25 @@ Once an API call to Hyperledger Cactus framework is requested by a User, Busines
...
 - **Validator**: The entity monitors transaction records of Ledger operation, and it determines the result(success, failed, timeouted) from the transaction records.
 Validator ensure the determined result with attaching digital signature with "Validator key" which can be verified by "Verifier".
-- **Verifier**: The entity accepts only sucussfully verified operation results by verifying the digital signature of the validator. Note that "Validator" is apart from "Verifier" over a bi-directional channel.
-- **Cactus Routing Interface**: The entity is a routing service between "Business Logic Plugin" and  "Ledger Plugin(s)". The entity is also a routing service between Business Logic Plugin and API calls from "Application user(s)".
-- **Ledger-n**: DLT platforms(e.g. Ethereum, Quorum, Hyperledger Fabric, ...)
+- **Validator Server**: The server accepts a connection from Verifier, and it provides Validator API, which can be used for controlling and monitoring Ledger behind it. The LedgerConnector will be implemented for interacting with the Ledger nodes.
```
petermetz, line 723: > Validator API, which can be used for **controlling** and monitoring Ledger behind it

@sfuji822 Just want to converge on the mental models: What does the `controlling` here refer to? When I read validator, I imagined it would mostly just read ledger state and make statements about that rather than actually being in control of the ledger itself.

whitepaper/whitepaper.md
```diff
@@ -706,19 +706,25 @@ Once an API call to Hyperledger Cactus framework is requested by a User, Busines
...
+- **Verifier**: The entity accepts only sucussfully verified operation results by verifying the digital signature of the validator. Verifier will be instantiated by calling the VerifierFactory#create method with associated with the Validator to connect. Each Verifier may be temporarily enabled or disabled. Note that "Validator" is apart from "Verifier" over a bi-directional channel.
+- **Verifier Registroy**: The information about active Verifier. The VerifierFactory uses this information to instantiate Verifier for the Business Logic Plugin.
```
petermetz, line 725: @sfuji822 

> Verifier Registroy

Typo? https://en.wikipedia.org/wiki/Registry

> The information about active Verifier.

Is it implied there that there's only one verifier active or could it be multiple active verifiers simultaneously ?

whitepaper/cactus_arch.svg
```diff
@@ -0,0 +1,1777 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
petermetz, line 1: @sfuji822 

I added a boundary around the validators wrapping them with the API server blue box as well and added a legend entry for the same thing. This is is roughly what I was talking about during our last call.

C: Users shingo_fujimoto Desktop cactus_arch2020-10-26

whitepaper/whitepaper.md
sfuji822: I will fix the typo 'Registry'. 

> Is it implied there that there's only one verifier active or could it be multiple active verifiers simultaneously ?

I thought that, but we may clarify if Verifier(s) to be multiple instances or not. 
Our prototyped code have instanced interface to Validator once it was created by ValidatorFactory.
I would like to hear your opinion about this. > @petermetz

whitepaper/cactus_arch.svg
sfuji822: Humm. That is something different from my understanding.
The icon in the Validator was Ledger, which may consist of multiple nodes that participated in the blockchain network. 
However, Validator (API) Server will be the point of contact for the Ledger, and it will be an independent server from other Validators.
This is because Validator is a client or node of Ledger which is operated by the BLP of CACTUS.

whitepaper/whitepaper.md
sfuji822: My intention was Trade is changing Ledger behind LedgerPlugin, but queries will not change the state of Ledger.
Do you want to have the capability to make a query through BLP? > @petermetz

takeutak (approved):
LGTM

whitepaper/whitepaper.md
petermetz: Discussed at maintainer's call 2020-12-08, controlling as in submitting transactions, not as in owning the entire ledger.
@petermetz

sfuji822: I meant controlling is issuing transaction regarding the user's account or assets.
I will change the wording with "issuing signed transactions".
Could you accept this change? @petermetz

petermetz: > I meant controlling is issuing transaction regarding the user's account or assets.
> I will change the wording with "issuing signed transactions".
> Could you accept this change? @petermetz

@sfuji822 Yes, sounds good, thank you!

whitepaper/whitepaper.md
petermetz: > My intention was Trade is changing Ledger behind LedgerPlugin, but queries will not change the state of Ledger.

@sfuji822 Makes sense. Do you feel that the word `trade` can really encompasses all of our non-financial use cases? As-in, would you describe an end user application uploading cat pictures to a ledger as a `trade`?

> Do you want to have the capability to make a query through BLP? > @petermetz

@sfuji822 I have no strong opinions on that. Should be up to the author of the BLP IMO.

whitepaper/whitepaper.md
petermetz: > I thought that, 

@sfuji822 Sorry, wasn't sure which one of the two you meant by `that`. Do you mean a) `there's only one verifier active` or b) `could it be multiple active verifiers simultaneously`

whitepaper/cactus_arch.svg
petermetz: > Humm. That is something different from my understanding.
> The icon in the Validator was Ledger, which may consist of multiple nodes that participated in the blockchain network.
> However, Validator (API) Server will be the point of contact for the Ledger, and it will be an independent server from other Validators.
> This is because Validator is a client or node of Ledger which is operated by the BLP of CACTUS.

@sfuji822 Gotcha, I'll take a step back/zoom out a bit in that case: The validator. => Is the logic of the validator implemented as a plugin or is the logic prescribed/hardcoded?

whitepaper/cacts_arch.svg
sfuji822: I will delete it.

whitepaper/whitepaper.md
sfuji822: I did not intend to define generic API servers. I thought this should be named as "CACTUS node server" which provides API interfaces to both end-users and administrators.

PR #379, opened 2020-11-10
build(tools): pkg info and bundle validation scripts

petermetz:
Adds two utility scripts that come handy when
1. Authoring build tools for the project where
the package dependency graph has to be accessed
and processed programmatically for example.

2. Wanting to validate that there are no typos in
the bundle names of packages as declared in
their package.json files (which has been a
recurring theme hence this script to automate
the checks for it).


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):
LGTM

PR #386, opened 2020-11-16
docs(whitepaper): fix pdf syntax in section 3.8

petermetz:
1. The horizontal line (---) syntax was breaking the
PDF generation process which lead to 22 pages
being excluded from the final PDF file.
Adding the new line here fixes this syntax error
and therefore the PDF build no longer has this issue.

2. Automated tooling has removed the leading space
characters from some sentences so this is
also included in this change which I considered
small and unimportant enough to just include it
in the same changeset regardless of the fact that
it is actually a different change that is not related
to the PDF generation syntax error explained in 1)


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):
LGTM, as this pull-request includes only bug-fixes.

PR #384, opened 2020-11-10
fix(common): Servers#listen() port number validation

petermetz:
**Depends on #381**

# How to review

Because this PR is a dependent of PR #381, by default the diff will show the full set of changes including everything from #381 and its parents changes as well. Therefore, to review this pull request please only look at the last commit (hash may change after rebasing later on)

---

Commit: 1ce14fa3fee1a723f96e5ae284bbf91a0ecb7959
Parents: be429528c14c9230bceff31db070424cea702fe2
Author: Peter Somogyvari <peter.somogyvari@accenture.com>
Committer: Peter Somogyvari <peter.somogyvari@accenture.com>
Date: Tue Nov 10 2020 11:20:52 GMT-0800 (Pacific Standard Time) 

fix(common): servers#listen() port number validation

Fixes #383

Ensures that the when specifying port 0 the liste()
method does not throw an exception since it is
perfectly valid to bind to port zero in general.

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

---

PR #389, opened 2020-11-18
fix(test-tooling): fabric AIO image docker in docker support

### LINKED ISSUE #279: fix(test-tooling): fabric AIO image docker in docker support
**Describe the bug**

Right now we are forced to bind to host ports on the fabric AIO image to make it work as a temporary measure (was short on time and decided to defer tackling the DinD support)

**Expected behavior**

Fabric AIO image should be able to run multiple containers in parallel (right now it's not possible because of the host ports).
Having DinD would allow us to not have to bind to the docker.sock of the host either and so this way the chaincode containers
that the peer launches upon contract instantiation would be hosted within the AIO image and make the host port binding problem go away.

petermetz:
## How to Review

**<ins>This PR depends on another [PR](https://github.com/hyperledger/cactus/pull/384)</ins>**

**<ins>Before this one can be merged, #384 has to be approved and merged first.</ins>**

**<ins>Reviewing this one can be done as of right now by looking at the changes made specifically by the single commit that this branch commits on top of the branch of the #384</ins>**

## Commit to be reviewed:

```
fix(test-tooling): fabric AIO image docker in docker support

Fixes #279

Note: Although this image now uses Docker in Docker (DinD)
it does not actually solve the problem of randomized
ports completely because as it turns out there's a second
issue with the randomized ports namely that Fabric's
service discovery doesn't have a port mapping feature in it
so we are not able to specify the associations between
the public and private ports where the public (host) ports
are randomized and the private ones are the ones
returned by the service discovery algorithms.
For example, if you start the new AIO image with randomized
ports, then it will run one of the peers on port 7051 of the
AIO container and the host will map that to something random
typically somewhere in the 30000 to 40000 port range.
When the Fabric connector (that's running on the host machine)
instantiates a Fabric Gateway object of the Fabric Node SDK
it performs service discovery where the peer is described
by the service discovery as listening on port 7051, but from
the network of the host (where the Fabric connector is)
that port is not correct because the real one is the random
one as explained in the above paragraph.
What we need to solve this is a way to inject into the service
discovery mechanism our own port mappings so that the
Fabric connector related tests can run in parallel and the
CI can stop being flaky.

Because of what is explained above, in this commit we also
skip the Fabric AIO image dependent test cases until they
can be fixed via refactoring the test cases and the test
ledger class in a way so that it works with the new AIO
image. These changes will need to have some method deletions
and renames which will make the change quite big on its own
and therefore will go in a separate commit which will also
re-activate the test cases that are currently being skipped
as per this current change.

```

Fixes #279 

Depends on #384


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

takeutak:
LGTM

PR #394, opened 2020-11-20
feat(fabric): add run transaction endpoint

petermetz:
> Important: The CI will not pass on this PR until the parent PR is merged due to a dependency on the docker image that will only get updated once #389 has been merged in. Because of this, please disregard the fact that the CI is failing and review the pull request with the assumption (requesting benefit of the doubt) that the CI is (will be) passing once #389 has been merged.

## How to Review

**<ins>This PR depends on another [PR](https://github.com/hyperledger/cactus/pull/389) to ensure that the authors are not blocked in performing their work while waiting for the pull request to get reviewed. Please read the each point of the description below carefully!</ins>**

- It is strongly recommended that you review the parent PR first before reviewing this one because once that one (parent PR) is merged, this can be rebased onto the main branch which will literally disappear the additional commits that are shown in this PR but actually belong to the parent PR
- Before this one can be merged, #389 has to be approved and merged first.
- Reviewing this one can be done as of right now by looking at the changes made specifically by the single commit that this branch commits on top of the branch of the parent branch (parent PR)
- Feel free to ask @petermetz via any of your preferred communication channels for help if you get stuck using the GitHub UI or any git features during the review.
- You can use the 'commits' tab of the PR page of GitHub to filter down the diff to a specific commit which allows you to not be bothered by the other commits at all (which belong to the parent PR). To see which commit to filter the diffs down onto, see the section below titled as "commit to be reviewed".
- Important: The CI will not pass on this PR until the parent PR is merged due to a dependency on the docker image that will only get updated once the parent PR has been merged in. Because of this, please disregard the fact that the CI is failing and review the pull request with the assumption (requesting benefit of the doubt) that the CI is (will be) passing once the parent PR has been merged.

## Commit to be reviewed:

```
feat(fabric): add run transaction endpoint
    
    Primary change:
    ---------------
    
    1. The Fabric connector can now execute transactions via the Gateway
    object of the Fabric Node SDK. For this, the OpenAPI specs have been
    added so that the transaction execution endpoint is callable via the
    generated API client.
    
    Minor refactorings that were also necessary to accomodate 1)
    ------------------------------------------------------------
    
    2. The go binary endpoint has been refactored to accept go
    source code instead and a go mod file optionally. On the back
    of this change we also retired the code that was doing the
    contract deployments altogether as it would not have worked
    with the new Fabric AIO image which has an extra level of
    indirection in the sense that once you SSH into the running
    container of the image, you are not able to directly access a
[...]

PR #403, opened 2020-12-02
feat(fabric): user defined fabric samples version

### LINKED ISSUE #391: tools(fabric-aio): user defined Fabric Samples version

Right now (as of #389 ) the Fabric AIO image uses DinD (Docker in Docker) and pulls up a complete "first-network" ledger, verbatim as provided by the [Fabric Samples repo](https://github.com/hyperledger/fabric-samples/tree/release-1.4/fabcar)
and the version the Fabric bootstrap script gets fed through the parameters is hardcoded as `v1.4.8` which is exactly what we want for the Fabric 1.x connector's development, but once we start with writing the connector (or updating the existing one for) Fabric 2.x we'll need the [Cactus Fabric AIO image](https://github.com/hyperledger/cactus/tree/master/tools/docker/fabric-all-in-one) to be able to pull up a Fabric 2.x ledger (most likely something closer to 2.2.x to be exact).

1. The Dockerfile to have/specify an `ENV` variable for customizing the Fabric version that gets pulled at startup.
2. The `FabricTestLedger` class to have a constructor parameter that gets fed into the `ENV` variable mentioned in 1) 
3. Image can handle the differences between 1.x and 2.x Fabric Samples repo (I think they changed the directory structure and renamed some of them?)

Considered CLI and docker build arguments instead and also considered keeping the Fabric version hardcoded and publishing different versions of the AIO image that map their versions to the Fabric version they come equipped with.
These don't seem like bad ideas either, but I thought it best to start with the fully dynamic option, see how it works and then take it from there. 

**Additional context**

There's a PR open for upgrading the AIO image to support DinD: #389 

jagpreetsinghsasan:
# Commit to be reviewed
----------------------------

feat(fabric): user defined fabric samples version

    Primary Change
    --------------

    1. The fabric-all-in-one can now accept fabric versions as
    argument at runtime and corresponding code changes are incorporated.

    Minor refactorings that were also necessary to accomodate 1)

    2. The interface IFabricTestLedgerV1ConstructorOptions in the
    class fabric-test-ledger-v1 is modified to incorporate new
    variable fabricVersion. Corresponding changes are made to the
    functions to include the same.

    3. The client user name has been renamed from user1 to
    user2 as Hyperledger Fabric 2.xversions has user1 already
    registered.

    4. Go has been manually setup in the Dockerfile located
    at tools/docker/fabric-all-in-one/Dockerfile to support
    Hyperledger Fabric 2.x versions, as Go isn't setup by default
    in v2.x

    5. Check for existence has been put over the
    peer1.org{org_number}.example.com as peer1 is not setup
    for test-network.

    6. fabricVersion has been removed from FabricTestLedgerV1,
     and instead envVars map<string,string> has been included to
    support more dynamic features for the fabric-all-in-one
    image like fabricVersion, caVersion

    7. The healthcheck logic for the above mentioned
    Dockerfile has moved to the healthcheck.sh script to support
    Hyperledger Fabric 2.x versions. Changes are made
    for the run-fabric-network.sh script for the same.

    8. The Readme for the Dockerfile mentioned in 4) has been updated

    9. Created 2 seperate Dockerfile for 2.x and 1.4.x versions
    of Hyperledger Fabric

    10. 2 sepearte integration tests corresponding to Fabric v1.4.x
    and v2.2.x

    Fixes #391

Depends on #394 

Fixes #391

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
@@ -1,6 +1,8 @@
 import path from "path";
 import { EventEmitter } from "events";
 
+import compareVersions, { compare } from "compare-versions";
+
 import Docker, {
   Container,
   ContainerCreateOptions,
@@ -14,7 +16,7 @@ import {
   Gateway,
 } from "fabric-network";
 import FabricCAServices from "fabric-ca-client";
-import Joi from "joi";
+import Joi, { array } from "joi";
 import { ITestLedger } from "../i-test-ledger";
 import { Containers } from "../common/containers";
 import {
@@ -31,6 +33,7 @@ export interface IFabricTestLedgerV1ConstructorOptions {
   publishAllPorts: boolean;
   imageVersion?: string;
   imageName?: string;
+  envVars?: Map<string, string>;
   logLevel?: LogLevelDesc;
 }
 
@@ -40,6 +43,7 @@ export interface IFabricTestLedgerV1ConstructorOptions {
 const DEFAULT_OPTS = Object.freeze({
   imageVersion: "latest",
   imageName: "hyperledger/cactus-fabric-all-in-one",
+  envVars: new Map([["FABRIC_VERSION", "1.4.8"]]),
 });
 export const FABRIC_TEST_LEDGER_DEFAULT_OPTIONS = DEFAULT_OPTS;
 
@@ -53,6 +57,10 @@ const OPTS_JOI_SCHEMA: Joi.Schema = Joi.object().keys({
     .regex(/[a-z0-9]+(?:[._-]{1,2}[a-z0-9]+)*/)
     .min(1)
     .required(),
+  envVars: Joi.object().pattern(/.*/, [
+    Joi.string().required(),
+    Joi.string().min(1).required(),
+  ]),
 });
 
 export const FABRIC_TEST_LEDGER_OPTIONS_JOI_SCHEMA = OPTS_JOI_SCHEMA;
@@ -63,6 +71,7 @@ export class FabricTestLedgerV1 implements ITestLedger {
   public readonly imageVersion: string;
   public readonly imageName: string;
   public readonly publishAllPorts: boolean;
+  public readonly envVars: Map<string, string>;
 
   private readonly log: Logger;
 
@@ -84,6 +93,12 @@ export class FabricTestLedgerV1 implements ITestLedger {
     this.imageVersion = options.imageVersion || DEFAULT_OPTS.imageVersion;
     this.imageName = options.imageName || DEFAULT_OPTS.imageName;
     this.publishAllPorts = options.publishAllPorts;
+    this.envVars = options.envVars || DEFAULT_OPTS.envVars;
+
+    if (compareVersions.compare(this.getFabricVersion(), "1.4", "<"))
+      this.log.warn(
+        `This version of Fabric ${this.getFabricVersion()} is unsupported`
+      );
 
     this.validateConstructorOptions();
   }
@@ -99,6 +114,10 @@ export class FabricTestLedgerV1 implements ITestLedger {
 
   public getContainerImageName(): string {
     return `${this.imageName}:${this.imageVersion}`;
+  }
+
+  public getFabricVersion(): string {
+    return `${this.envVars.get("FABRIC_VERSION")}`;
   }
 
   public getDefaultMspId(): string {
@@ -126,7 +145,7 @@ export class FabricTestLedgerV1 implements ITestLedger {
     const fnTag = `${this.className}#enrollUser()`;
     try {
       const mspId = this.getDefaultMspId();
-      const enrollmentID = "user1";
+      const enrollmentID = "user2";
       const connectionProfile = await this.getConnectionProfileOrg1();
       // Create a new gateway for connecting to our peer node.
       const gateway = new Gateway();
@@ -199,7 +218,17 @@ export class FabricTestLedgerV1 implements ITestLedger {
     const fnTag = `${this.className}#getConnectionProfileOrg1()`;
     const cInfo = await this.getContainerInfo();
     const container = this.getContainer();
-    const ccpJsonPath = "/fabric-samples/first-network/connection-org1.json";
+    const CCP_JSON_PATH_FABRIC_V1: string =
+      "/fabric-samples/first-network/connection-org1.json";
+    const CCP_JSON_PATH_FABRIC_V2: string =
+      "/fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/connection-org1.json";
+    const ccpJsonPath = compareVersions.compare(
+      this.getFabricVersion(),
+      "2.0",
+      "<"
+    )
+      ? CCP_JSON_PATH_FABRIC_V1
+      : CCP_JSON_PATH_FABRIC_V2;
     const ccpJson = await Containers.pullFile(container, ccpJsonPath);
     const ccp = JSON.parse(ccpJson);
 
@@ -209,7 +238,7 @@ export class FabricTestLedgerV1 implements ITestLedger {
       const hostPort = await Containers.getPublicPort(privatePort, cInfo);
       ccp.peers["peer0.org1.example.com"].url = `grpcs://localhost:${hostPort}`;
     }
-    {
+    if (ccp.peers["peer1.org1.example.com"]) {
       // peer1.org1.example.com
       const privatePort = 8051;
       const hostPort = await Containers.getPublicPort(privatePort, cInfo);
@@ -236,8 +265,17 @@ export class FabricTestLedgerV1 implements ITestLedger {
       const privatePort = 7050;
       const hostPort = await Containers.getPublicPort(privatePort, cInfo);
       const url = `grpcs://localhost:${hostPort}`;
-      const ordererPemPath =
+      const ORDERER_PEM_PATH_FABRIC_V1: string =
         "/fabric-samples/first-network/crypto-config/ordererOrganizations/example.com/tlsca/tlsca.example.com-cert.pem";
+      const ORDERER_PEM_PATH_FABRIC_V2: string =
+        "/fabric-samples/test-network/organizations/ordererOrganizations/example.com/orderers/orderer.example.com/msp/tlscacerts/tlsca.example.com-cert.pem";
+      const ordererPemPath = compareVersions.compare(
+        this.getFabricVersion(),
+        "2.0",
+        "<"
+      )
+        ? ORDERER_PEM_PATH_FABRIC_V1
+        : ORDERER_PEM_PATH_FABRIC_V2;
       const pem = await Containers.pullFile(container, ordererPemPath);
 
       ccp.orderers = {
@@ -301,6 +339,9 @@ export class FabricTestLedgerV1 implements ITestLedger {
 
   public async start(omitPull: boolean = false): Promise<Container> {
     const containerNameAndTag = this.getContainerImageName();
+    const dockerEnvVars: string[] = new Array(...this.envVars).map(
+      (pairs) => `${pairs[0]}=${pairs[1]}`
+    );
 
     if (this.container) {
       await this.container.stop();
@@ -328,6 +369,8 @@ export class FabricTestLedgerV1 implements ITestLedger {
         "9051/tcp": {}, // peer0.org2.example.com
         "10051/tcp": {}, // peer1.org2.example.com
       },
+
+      Env: dockerEnvVars,
 
       // This is a workaround needed for macOS which has issues with routing
       // to docker container's IP addresses directly...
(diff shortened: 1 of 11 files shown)


packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -31,6 +31,7 @@ export interface IFabricTestLedgerV1ConstructorOptions {
   publishAllPorts: boolean;
   imageVersion?: string;
   imageName?: string;
+  fabricVersion?: string;
```
petermetz, line 34: Idea to extend this and make it more generic for the future uses: 
1. Rename the `fabricVersion` parameter to envVars
2. Change the parameter type to a `Map<string, string>` where the key is the variable name and the value is the ENV variable's value
3. The default value (`DEFAULT_OPTS`) can be a new map populated with the same information as before but this way: 

```typescript
envVars: new Map([ 
    [
        "fabricVersion", 
         "1.4.8"
    ] 
]),
```
4. In the start method just iterate over all map entries and convert them into an array of strings that are built in the expected format e.g. `${key}=${value}`, then just pass this in directly to dockerode as the `Env` property of the container.

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -199,7 +208,9 @@ export class FabricTestLedgerV1 implements ITestLedger {
     const fnTag = `${this.className}#getConnectionProfileOrg1()`;
     const cInfo = await this.getContainerInfo();
     const container = this.getContainer();
-    const ccpJsonPath = "/fabric-samples/first-network/connection-org1.json";
+    const ccpJsonPath = this.getFabricVersion().valueOf().includes("1.4.")
```
petermetz, line 211: Use [this](https://www.npmjs.com/package/compare-versions#human-readable-compare) (or something like it) to cover all possible integers for the supported versions.
It is okay for us to only support Fabric 1.4.x, we can throw for 1.3, 1.2, ... 0.1, etc...
Throw exceptions if the provided fabric version is not in the supported range.
Cover cases of major version being above 2.x (when Fabric 3.x inevitably hits, we want at least our code to fail gracefully instead of having undefined side effects/silent bugs). So if the version is greater or equal (GTE) than 3.0.0 then you can either print a warning saying that we are falling back to using the 2.x paths and this may not work (warning log level) OR you can just throw saying 3.x is not supported (we don't know yet which one is more appropriate because we have no idea if they'll change the paths again or not).

petermetz, line 211: There shouldn't be a need to call the `valueOf()` method, the `string` type already has the includes method on it IIRC (happy to be wrong on this one I did not actually test it just typing from memory)

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -199,7 +208,9 @@ export class FabricTestLedgerV1 implements ITestLedger {
...
+      ? "/fabric-samples/first-network/connection-org1.json"
```
petermetz, line 212: Pull these magic strings out to separate constant variables where the variables of the names show what they hold and which fabric version they are used for (good variable/method names make the code basically self-documenting).

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -209,7 +220,7 @@ export class FabricTestLedgerV1 implements ITestLedger {
       const hostPort = await Containers.getPublicPort(privatePort, cInfo);
       ccp.peers["peer0.org1.example.com"].url = `grpcs://localhost:${hostPort}`;
     }
-    {
+    if (JSON.stringify(ccp).indexOf("peer1.org1.example.com") > -1) {
```
petermetz, line 223: Prefer to have this check implemented in a more deterministic way. This can fail if a new, unrelated property gets added to the ccp object (unlikely, but can happen and therefore we don't want that edge case to be possible at all).

Example:

```typescript
const ccp = {
    description: "New ccp edition that omits peer1.org1.example.com",
    peers: {
        "peer0.org1.example.com": { ... }
    }
};

// now if you check here for the presence of `peer1.org1.example.com` 
// it will return true, but there is no such peer and it will start crashing
```

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -236,8 +247,9 @@ export class FabricTestLedgerV1 implements ITestLedger {
       const privatePort = 7050;
       const hostPort = await Containers.getPublicPort(privatePort, cInfo);
       const url = `grpcs://localhost:${hostPort}`;
-      const ordererPemPath =
-        "/fabric-samples/first-network/crypto-config/ordererOrganizations/example.com/tlsca/tlsca.example.com-cert.pem";
+      const ordererPemPath = this.getFabricVersion().valueOf().includes("1.4.")
```
petermetz, line 250: All 3 of my comments apply here as well regarding the version checks, the `valueOf()` and the magic strings.

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -329,6 +342,8 @@ export class FabricTestLedgerV1 implements ITestLedger {
         "10051/tcp": {}, // peer1.org2.example.com
       },
 
+      Env: [`FABRIC_VERSION=${this.fabricVersion}`],
```
petermetz, line 345: If everything is done as I suggested in my above comments then here you can just plug in the array of strings you generate from the envVars Map<string, string>

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -301,6 +313,7 @@ export class FabricTestLedgerV1 implements ITestLedger {
 
   public async start(omitPull: boolean = false): Promise<Container> {
     const containerNameAndTag = this.getContainerImageName();
+    const containerFabricVersion = this.getFabricVersion();
```
petermetz, line 316: Here you can do the `Map<string,string>` to `string[]` conversion instead.

tools/docker/fabric-all-in-one/Dockerfile
```diff
@@ -34,6 +34,14 @@ RUN apk add --no-cache curl
 # The file binary is used to inspect exectubles when debugging container image issues
 RUN apk add --no-cache file
 
+# Download and setup path variables for Go
```
petermetz, line 37: Recommend to also answer the `why` (IIRC it was because you need to compile the peer binary for Fabric 2.x due to lack of CLI container or something related to that right?)

tools/docker/fabric-all-in-one/Dockerfile
```diff
@@ -34,6 +34,14 @@ RUN apk add --no-cache curl
...
+RUN wget https://golang.org/dl/go1.13.4.linux-amd64.tar.gz
```
petermetz, line 38: Any reason why not using the latest stable go version? I think it was 1.15.5. If Fabric needs an older one for whatever reason that's fine then.

tools/docker/fabric-all-in-one/Dockerfile
```diff
@@ -71,14 +79,15 @@ RUN curl -sSL https://raw.githubusercontent.com/hyperledger/fabric/release-2.2/s
 RUN chmod +x bootstrap.sh
 # Run the bootstrap here so that at least we can pre-fetch the git clone and the binary downloads resulting in
 # faster container startup speed since these steps will not have to be done, only the docker image pulls.
-RUN /bootstrap.sh $FABRIC_VERSION $FABRIC_VERSION -d
+RUN /bootstrap.sh ${FABRIC_VERSION} 1.4.8 -d
```
petermetz, line 82: Typo? Default CA version in the Fabric bootstrap scripts is 1.4.9
https://github.com/hyperledger/fabric/blob/master/scripts/bootstrap.sh#L11

petermetz, line 82: We can have a separate variable for the CA version the same way we have one for the Fabric version. No need to hardcode it.

tools/docker/fabric-all-in-one/run-fabric-network.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/sh
 set -e
 
-# TODO: make the verion numbers here injectable to the script
-/bootstrap.sh 1.4.8 1.4.8 -b -s
+/bootstrap.sh ${FABRIC_VERSION} 1.4.8 -b -s
```
petermetz, line 4: Same two asks as above: 1) typo of default CA version, 2) we can have an ENV variable for the CA version as well (and it can be optional with the default value taken from the bootstrap script of Fabric itself, 1.4.9)

petermetz (changes requested):
@jagpreetsinghsasan For the most part see comments inline and there's one additional thing:
The image publishing has to be automated so we need to figure out how to update the `./hooks/post_push` script to ensure that two images get built/published at the same time (may not even be possible with a single Dockerfile we may need to create a second one that inherits from the first one because DockerHub auto-build will need it)

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

tools/docker/fabric-all-in-one/Dockerfile
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

tools/docker/fabric-all-in-one/Dockerfile
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

tools/docker/fabric-all-in-one/Dockerfile
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

tools/docker/fabric-all-in-one/run-fabric-network.sh
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
jagpreetsinghsasan: Incorporated the changes. Please have a re-look.

petermetz (approved):
LGTM

takeutak (approved):
LGTM

PR #414, opened 2020-12-04
feat(core-api): add consensus algorithms OpenAPI enum

petermetz:
## How to Review

**<ins>This PR depends on another [PR](https://github.com/hyperledger/cactus/pull/409) to ensure that the authors are not blocked in performing their work while waiting for the pull request to get reviewed. Please read the each point of the description below carefully!</ins>**

- It is strongly recommended that you review the parent PR first before reviewing this one because once that one (parent PR) is merged, this can be rebased onto the main branch which will literally disappear the additional commits that are shown in this PR but actually belong to the parent PR
- Before this one can be merged, #409 has to be approved and merged first.
- Reviewing this one can be done as of right now by looking at the changes made specifically by the single commit that this branch commits on top of the branch of the parent branch (parent PR)
- Feel free to ask @petermetz via any of your preferred communication channels for help if you get stuck using the GitHub UI or any git features during the review.
- You can use the 'commits' tab of the PR page of GitHub to filter down the diff to a specific commit which allows you to not be bothered by the other commits at all (which belong to the parent PR). To see which commit to filter the diffs down onto, see the section below titled as "commit to be reviewed".
- Important: The CI will not pass on this PR until the parent PR is merged due to a dependency on the docker image that will only get updated once the parent PR has been merged in. Because of this, please disregard the fact that the CI is failing and review the pull request with the assumption (requesting benefit of the doubt) that the CI is (will be) passing once the parent PR has been merged.

## Commit to be reviewed:

```
feat(core-api): add consensus algorithms OpenAPI enum

Enumerates a list of consensus algorithm families in
existence. Does not intend to be an exhaustive list, just a
practical one, meaning that we only include items here
that are relevant to Hyperledger Cactus in fulfilling its
own duties. This can be extended later as more
sophisticated features of Cactus get implemented. This
enum is meant to be first and foremest a useful
abstraction for achieving practical tasks, not an
encyclopedia and therefore we ask of everyone that this
to be extended only in ways that serve a practical
purpose for the runtime behavior of Cactus or Cactus
plugins in general. The bottom line is that we can accept
this enum being not 100% accurate as long as it 100%
satisfies what it was designed to do.

Fixes hyperledger#359

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>
```

Fixes #359

PR #402, opened 2020-12-02
refactor(core-api): move PluginRegistry to core package

petermetz:
## Commit to be reviewed:

```
commit 6c88f647baf6f5f70e47d2634697d47498e8bed4 (refactor/core-api/move-pluginregistry-to-core-package-380)
Author: Peter Somogyvari <peter.somogyvari@accenture.com>
Date:   Tue Dec 1 17:04:16 2020 -0800

    refactor(core-api): move PluginRegistry to core package
    
    Fixes #380
    
    Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

```

PR #331, opened 2020-10-28
docs(whitepaper): clear up node vs API server terminology

petermetz:
Fixes #249

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

PR #415, opened 2020-12-04
feat(core-api): getConsensusAlgorithmFamily() on connector API

### LINKED ISSUE #355: feat(core-api): getConsensusAlgorithm() on connector API

We need a way for the ledger connector plugins to be queried at runtime for the consensus algorithm they use.
This is not very meaningful for ledgers that are fixed on a single algorithm, but can be a must have for the ones that support multiple types and have it configurable. 

~~For one-trick pony ledgers we can just do it like this:~~

```typescript
export enum ConsensusAlgorithm {
  IBFT2_0,
  POW,
  POS,
  etc...
}
```

```typescript
public async getConsensusAlgorithm(): Promise<ConsensusAlgorithm> {
    rerturn ConsensusAlgorithm.PROOF_OF_WORK;
}
```

If the ledger itself provides APIs to query this, then we can just call that, otherwise we may need to have the operator of Cactus specify this in the consortium definition or at least the connector plugin's own configuration somehow.

```typescript
public async getConsensusAlgorithm(): Promise<ConsensusAlgorithm> {
    const consensusAlgorithm = ... // query ledger at runtime for the algorithm being used
    // maybe have caching introduced here as well? 
    // Not sure because then we would have to deal with cache invalidation...    
    return consensusAlgorithm;
}
```

Nothing, I'm pretty confident that we need this feature for the higher level, cross ledger transaction coordination protocol that doesn't exist yet, but is definitely coming.

**Additional context**

This should power the higher level API method proposed here:
#354 

petermetz:
## How to Review

**<ins>This PR depends on another [PR](https://github.com/hyperledger/cactus/pull/414) to ensure that the authors are not blocked in performing their work while waiting for the pull request to get reviewed. Please read the each point of the description below carefully!</ins>**

- It is strongly recommended that you review the parent PR first before reviewing this one because once that one (parent PR) is merged, this can be rebased onto the main branch which will literally disappear the additional commits that are shown in this PR but actually belong to the parent PR
- Before this one can be merged, #414 has to be approved and merged first.
- Reviewing this one can be done as of right now by looking at the changes made specifically by the single commit that this branch commits on top of the branch of the parent branch (parent PR)
- Feel free to ask @petermetz via any of your preferred communication channels for help if you get stuck using the GitHub UI or any git features during the review.
- You can use the 'commits' tab of the PR page of GitHub to filter down the diff to a specific commit which allows you to not be bothered by the other commits at all (which belong to the parent PR). To see which commit to filter the diffs down onto, see the section below titled as "commit to be reviewed".
- Important: The CI will not pass on this PR until the parent PR is merged due to a dependency on the docker image that will only get updated once the parent PR has been merged in. Because of this, please disregard the fact that the CI is failing and review the pull request with the assumption (requesting benefit of the doubt) that the CI is (will be) passing once the parent PR has been merged.

## Commit to be reviewed:

```
feat(core-api): getConsensusAlgorithmFamily() on connector API

The method is called getConsensusAlgorithmFamily()
because we don't need (for now) the level of granularity
that is provided by having a method that points to the exact
algorithm used by the ledger (can be added later).
Why? Because for now we are mostly interested in whether an
algorithm family guarantees transaction finality or not and this
can be determined just from the family since for example it
does not make much of a difference if you are talking about
Bitcon's or Ethereum 1's proof of work, they both just do not
guarantee transaction finality the same way for all our intents
and purposes at present.

The added benefit of only dealing in families instead of specific
algorithms is that we can hardcode the answers instead of
having to query the ledger or rely on operators to provide
this information via configuration (again, we'll probably end
up needing this in the future anyway, but for the upcoming
work the families should be enough).

Fixes #355

```

Fixes #355

### DIFF
--- a/packages/cactus-core-api/src/main/json/generated/openapi-spec.json
+++ b/packages/cactus-core-api/src/main/json/generated/openapi-spec.json
@@ -27,6 +27,15 @@
     ],
     "components": {
         "schemas": {
+            "ConsensusAlgorithmFamily": {
+                "type": "string",
+                "description": "Enumerates a list of consensus algorithm families in existence. Does not intend to be an exhaustive list, just a practical one, meaning that we only include items here that are relevant to Hyperledger Cactus in fulfilling its own duties. This can be extended later as more sophisticated features of Cactus get implemented. This enum is meant to be first and foremest a useful abstraction for achieving practical tasks, not an encyclopedia and therefore we ask of everyone that this to be extended only in ways that serve a practical purpose for the runtime behavior of Cactus or Cactus plugins in general. The bottom line is that we can accept this enum being not 100% accurate as long as it 100% satisfies what it was designed to do.",
+                "enum": [
+                    "org.hyperledger.cactus.consensusalgorithm.PROOF_OF_AUTHORITY",
+                    "org.hyperledger.cactus.consensusalgorithm.PROOF_OF_STAKE",
+                    "org.hyperledger.cactus.consensusalgorithm.PROOF_OF_WORK"
+                ]
+            },
             "PrimaryKey": {
                 "type": "string",
                 "minLength": 1,
--- a/packages/cactus-core-api/src/main/typescript/generated/openapi/typescript-axios/api.ts
+++ b/packages/cactus-core-api/src/main/typescript/generated/openapi/typescript-axios/api.ts
@@ -125,6 +125,17 @@ export interface CactusNodeMeta {
     publicKeyPem: string;
 }
 /**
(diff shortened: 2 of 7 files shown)


takeutak (commented):
In my understanding, the consensus algorithm of connector API seems to be related to the Consortium plugin and depends on the discussion in PR #331, #368. Please allow me to review this PR after the discussion of PR #331, #368.

petermetz:
> In my understanding, the consensus algorithm of connector API seems to be related to the Consortium plugin and depends on the discussion in PR #331, #368. Please allow me to review this PR after the discussion of PR #331, #368.

@takeutak Please take a look at my question here: https://github.com/hyperledger/cactus/pull/402#issuecomment-743407715

takeutak (approved):
LGTM

jonathan-m-hamilton (approved):
LGTM

PR #428, opened 2020-12-16
fix(ci): windows operating system

### LINKED ISSUE #140: fix(ci): windows operating system
### Description
As a developer I want to have CI that runs the tests on both windows, linux and OS X so that I don't get annoying bug reports about some script/feature/tool/etc. not working on a particular platform that I don't have a development machine for at home.

For this bug, we only focus on Windows in order to break down a large problem into smaller, incremental pieces.
macOS has it's own set of problems that are more severe that of Windows because GitHub Actions do not support docker on mac OS at all as of right now.

### Acceptance Criteria
1. CI executes on Ubuntu 18 (bonus point for other mainstream Linux distros that are common)
2. CI executes on Windows

petermetz:
1. Migrated the OpenAPI spec export script of
the cmd-api-server package to the new
template where the script is self executing if
invoked directly from a terminal.
2. The migration above also means that the
generated JSON file is now placed on a different path on the file system and is
also under version control.

The API server package was still using the legacy way of
invoking the OpenAPI spec exporter which required that
quotes to be used in the npm scripts that works fine on Linux
but not on Windows operating systems.

Fixes #140

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,10 +13,10 @@ jobs:
       fail-fast: true
       max-parallel: 1 # Fabric AIO image needs fixed host ports allocated
       matrix:
-        os: [ubuntu-20.04, ubuntu-18.04]
-        node-version: [v12.13.0, v14.15.1]
-        experimental: [false]
-        # include:
+        # os: [ubuntu-20.04, ubuntu-18.04]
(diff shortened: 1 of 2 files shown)


petermetz:
Going to close this for now because the timeline for working in this again is unclear at present and I don't want it to clutter the PR list.

PR #435, opened 2020-12-17
ci(github-workflows): launch CI on PR & push for all branches

### LINKED ISSUE #434: ci(github-workflows): launch CI on PR & push for all branches
### Description

Currently the `dev` branch does not have CI running on it, but we could use the verification it provides when the tests are passing (or the notification when they are not).

petermetz:
Before this change the only time the CI would get executed was when
a pull request was submitted against the $DEFAULT branch. After this
the CI will run for every commit and pull request that is issued against
the any branches.
The specific reason for this change is that we introduced a new branch
called dev which can be used to provide early access to pending changes
that may or may not be stuck in the review queue for longer periods of
time. Of course there's an additional overhead in maintaining this
dev branch by way of cherry picking and rebasing, but this is minor
compared to the positive effects it can bring to the community.

Fixes #434

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: Cactus CI Github Workflow
 
-# Triggers the workflow on pull request events
-on: [pull_request]
+# Triggers the workflow on push or pull request events
+on: [push, pull_request]
 
 jobs:
   build:


takeutak (commented):
Your changes appear to check all branches. However, for example, it is possible to push the code in the draft state to the branch, and it would be troublesome in such a case.
As for your request, it is enough to apply this change to the two branches of $DEFAULT and dev, so can you change it to apply to only the two branches of $DEFAULT and dev?

petermetz:
> it is possible to push the code in the draft state to the branch, and it would be troublesome in such a case.

@takeutak Could you please elaborate on how does it become troublesome? Would the CI running a draft branch and failing block you somehow from doing something?

jonathan-m-hamilton (approved):
LGTM

takeutak:
Please let me confirm one. In this suggestion, not only pull-request but also pushing codes will be targets for CI checking system, is it right?  I think I have a lot of trouble when we make a draft code which does not work at this time.
As a question, does this change increase the time required for git push? If the code check increases the time significantly, I think it is better to limit the target branch to main and dev.
If the time required for pushing does not change from the current situation, I think there is no problem.

petermetz:
> not only pull-request but also pushing codes will be targets for CI checking system, is it right?

@takeutak That's right. I think it's great to receive feedback about failing CI as early on as possible versus only getting it once the PR has been opened (my personal take).

> As a question, does this change increase the time required for git push? 

@takeutak It should not. The time your git push takes should not be affected by this change.

> If the code check increases the time significantly, I think it is better to limit the target branch to main and dev.
> If the time required for pushing does not change from the current situation, I think there is no problem.

@takeutak Agreed: If that were true, I would agree, myself as well regarding it being a problem. To the best of my knowledge this is not the case though. Definitely has never been the case in my workflows, ever, but I also do not want to presume that other people do the exact same things as I do so if somehow this change ends up slowing you down in *any* way, then please do let me know the details because that's definitely something we would need fix were it to happen.

takeutak (approved):
LGTM

Issue #441, opened 2021-01-01
feat(ledger-plugin): add signature/verification features into Validator/Verifier


takeutak:
**Is your feature request related to a problem? Please describe.**
Verifiers and Validators should have a feature set for signature and verification between their communication in order to be part of the LedgerPlugin. Previous pull-request on their codes did not meet all functionality required for this issue. Still, the CACTUS core developer agreed on passing through it instead of rejecting the PR by adding a new issue item to remind fixing the identified issue.

**Describe the solution you'd like**
Fulfill missing functionalities such as signature and verification between their communication according to the architecture on CACTUS whitepaper.

PR #442, opened 2021-01-01
feat(ledger-plugin): add signature/verification features into Validator/Verifier

takeutak:
This PR is related with Issue #441 .


takeutak:
Contents:
- Add a new feature into Verifiers and Validators for signature and verification between their communication

The following README describes their boot method and how to use them
- https://github.com/takeutak/cactus/blob/feat_lp_add_signature_verification_features_into_Validator_Verifier/examples/cartrade/README.md

packages/routing-interface/util/RIFUtil.ts
```diff
@@ -43,7 +43,7 @@ export class RIFUtil {
             return JSON.parse(jsonStr);
         }
         catch (error) {
-            logger.error(error)
+            // logger.error(error)
```
petermetz, line 46: @takeutak 
I recommend disambiguation between legit JSON parsing errors and the case of the try block just running 

```typescript
JSON.parse("null")
```

which will return null the same way the catch block will. Currently there's no way for the caller to find out which one has happened.
With the logging commented out even if they are looking at the logs they won't be able to tell the difference.

packages/jest.config.js
```diff
@@ -0,0 +1,201 @@
+/*
```
petermetz, line 1: @takeutak Is this file scoped for the entire project (e.g. all packages) or just a sub-set of packages?

cc: @sfuji822 @hartm @jonathan-m-hamilton

packages/routing-interface/util/RIFUtil.ts
takeutak: Thank you for your comment.  I corrected this file as your comment.
https://github.com/hyperledger/cactus/pull/442/commits/7c88afda62b140e9573f920a313e953215408058

packages/jest.config.js
takeutak: Thank you for your suggestion.
As I said on the maintainer call, this file should be included by PR #444, not by this PR.  Then I deleted this file.
Moreover, I answered your question on the comment on PR #444.

takeutak:
@petermetz I answered to your questions as the above. Would you mind if you confirm this pull-request again?

petermetz (approved):
LGTM

jonathan-m-hamilton (approved):
LGTM

PR #454, opened 2021-01-06
fix: reverting accidental manual change in generated code

### LINKED ISSUE #453: fix(connector-besu): revert manual change in generated code
**Describe the bug**

The `api.ts` file of the besu connector package has a manual change in it that was not coming from a change in the openapi spec ts file.

**To Reproduce**

Open the mentioned file.
Perform a clean build.
Observe that your git index is no longer empty because the generator restored the state of the api.ts file (e.g. erased the manual change)

**Expected behavior**

Whenever the build runs, there should NOT be any files changing under source control (only exception is when issuing releases)

cc: @sfuji822 @hartm @takeutak @jonathan-m-hamilton

petermetz:
This must've slipped in during a recent refactor.

Fixes #453


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #464, opened 2021-01-10
refactor: convert OpenAPI Typescript specs to JSON

### LINKED ISSUE #462: refactor: convert Open API Typescript specs to JSON
### Description

As a developer I want to have the most convenient way of working with the Open API specs of Cactus possible so that I'm more productive and more likely to contribute to the success of the project.

So far we believed that the most convenient option available was to have the spec files in Typescript and then run those scripts via NodeJS processes to export into JSON which then can be referenced.
The need for this was due us not wanting to have to specify hard-coded HTTP paths in multiple places which is always ripe for the multiple code locations to get out of sync and cause hard to detect bugs powered by never ending human errors. 

However:
With the consistent usage of operation IDs, the above mentioned mechanism will not be necessary anymore as we can define the OpenAPI specs in pure JSON, then import said JSON to Typescript files which will get embedded at compile time and the HTTP paths will be possible to be looked up at runtime via the operation IDs. 

This structure will also allow us to validate the matching of operation IDs in the Typescript code and the raw JSON files via automated testing and therefore eliminating the possibility of human error. 

As an additional positive side effect this change will also speed up our build processes and unlock the possibility for us to have more files

### Acceptance Criteria
1. The exporting step of Open API spec Typescript files to JSON files is no longer necessary (because the Typescript files are replaced by the raw JSON files) 
2. It is possible to reference the JSON spec files from within each other thereby allowing us to break up the growing files (especially the one in the core-api package) into smaller, more maintainable pieces without having to copy paste the exporting logic to even more places)
3. Automated tests are in place to ensure that every and all operation IDs referenced in the Typescript code are matching up to the ones in the JSON files.

cc: @takeutak @jonathan-m-hamilton @takeutak @hartm

petermetz:
Primary change
=============

Why?
Because we wanted to be able to easily use the
file imports functionality of the OpenAPI generator
when one OpenAPI spec can refernces another through
the $ref attribte (by providing the local file path
or a remote URL)
https://swagger.io/docs/specification/using-ref/

1. All the openapi-spec.ts files have been removed.
2. Instead of them (^^) we now use raw JSON files
that are placed under src/main/json/openapi.json
3. The npm scripts we needed to invoke the Typescript
files to export the specs to raw JSON were removed
as well since they are no longer required.
4. The main reason behind originally going with
Typescript files for the OpenAPI specs was so that
we could leverage the compiler and the VSCode
auto-complete features provided through it when
typing out the specification files. It has been
decided that the downsides (mainly additional
complexity) are too great for not enough gain
given the fact that it has been discovered recently
that there are IDE addons/extensions that can provide
the same auto-complete/linting functionality for
OpenAPI spec raw JSON files such as this one for
VSCode: 42crunch.vscode-openapi

Additional minor changes
========================

1. The `cactus-api-client` package no longer exports
the generated API client code (typescript-axios)
of the cmd-api-server package (which was something
that did not make any sense in hindsight)

2. The consortium manual plugin plugin's API client methods
now use a shorthand operation ID to drive their names in the
generated code which makes it so that
the names are much much shorter compared to how they were
before when it was (implicitly) derived from the full HTTP path
of the endpoints instead.

3. GetConsortiumEndpointV1 and GetNodeJwsEndpoint no longer
use the deprecated "path" member on their constructor
objects ("options") and instead leverage their own methods
that are designed to provide the same information and more.
(These mechanisms have been fine grained over time to minimze
the room for human error when defining/updating HTTP paths).

Fixes #462 

cc: @hartm


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #470, opened 2021-01-13
fix(npm): clean script was missing folders

### LINKED ISSUE #469: fix(npm): clean script del-cli syntax
**Describe the bug**

1. The clean script does not clean all dist directories properly with the current way of it being invoked through lerna
2. Also, it only cleans the generated typescript code for the api-client package, but not for all the other packages that have generated typescript code as well.

**To Reproduce**

Run the `clean` script and observe that build artifacts remain in place.

**Expected behavior**

When I run `clean` I want all the build artifacts to be gone.

Dev machine.

`v0.3.0`

petermetz:
1. No longer uses lerna and instead specifies glob patterns
2. Expands the deletion of generated Typescript code to all
packages where it's present not just the api-client package.

Fixes #469


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #482, opened 2021-01-13
refactor(core-api): remove IPluginKeychain#rotateEncryptionKeys()

### LINKED ISSUE #477: refactor(core-api): remove IPluginKeychain#rotateEncryptionKeys()
### Description

Remove the `rotateEncryptionKeys()` method from the `IPluginKeychain` interface in the `core-api` package.

Also remove the related `getEncryptionAlgorithm(): string;` method which also serves no purpose at the moment. We can re-add it later if needed, but I don't want to have baggage when we hit `GA` so better to be aggressive with removing things while we still can without consequences to backwards compatibility.

Why?
Key rotation should be managed by ops people running the keychain backend not people running the Cactus deployment (these are likely to be separate services in a production grade deployment).

petermetz:
Also remove the getEncryptionAlgorithm method.

Fixes #477


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #483, opened 2021-01-14
build(npm): watch script

### LINKED ISSUE #149: build(npm): watch script
### Description
As a developer I want to be able to change the code and have it rebuilt automatically so that its convenient for me to work on the code which makes me more likely to contribute and increases my productivity in general when I do.

### Acceptance Criteria
1. A top level `watch` npm script that launches per package watching of the source code
2. `watch` does not rebuild the whole project when I change code in one package, it just rebuilds that one package so that turnaround is quick enough.
3. `watch` doesn't involve any sort of non-cross platform dark magic like observing file system changes in a way that breaks on any of the supported development operating systems like Windows/Linux/OS X/ etc.

petermetz:
Adds a top level watch script that invokes the package
level scripts for watching all Typescript sources and
automatically recompiling them whenever their contents change.

It also watches the openapi.json files so if you edit one
of those the API client code will automatically get
regenerated as well.

The big deal here with this feature is that it reduces
the time it takes for a change to compile from minutes
to seconds on average which is great, especially when
you are working from a machine with weaker hardware not
the top of the line stuff.

Future possible improvement is to also run the unit
tests upon every successful compilation as well.

To use it, just run this in the project root directory:

```sh
npm run watch
```

Fixes #149


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #520, opened 2021-01-29
build(deps): migrate to ESLint+Prettier

### LINKED ISSUE #518: fix(deps): migrate to ESLint
**Describe the bug**

TSLint has been deprecated for a while and now we are having issues making it work, specifically the VSCode extension doesn't seem to work at all.

**To Reproduce**

Try to configure the project from scratch, go make some mistakes in the code style and then watch how the `lint` script fails but your IDE doesn't show the errors (which is way too mundane and annoying and time consuming to fix manually)

**Expected behavior**

Whenever I type code and if it has code style/formatting issues, it should at least highlight every issue and for most cases it should also provide an automatic way of fixing it or all auto fixable issues.

**Additional context**

This came up as part of the on-boarding for a few new contributors recently and therefore is being fast tracked to resolution.

petermetz:
TSLint has been deprecated for quite a while. We kept it going
as long as we could but now with the onboarding of a few new
people issues are coming up with it not working as well as it
could so it was time to make the move to ESLint finally.

There's hundrds of warnings emitted by ESLint (mostly not errors
because we nerfed the errors down to warnings for now just to
keep it easier to accomplish the task without having to rewrite
a thousand lines of code all at once).

Follow-up tasks can be created for the existing warnings so that
eventually we can turn off the --quiet flag on ESLint which for
the moment makes it so that when you run the lint task you are
only seeing the errors in the output and not the warnings.

The biggest gain from ESLint here is that the IDE support will
hopefully work on everyone's dev machines not just the person
who originally configured it.

Fixes #518


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #523, opened 2021-02-01
chore(linter): enable no-unused-vars error

petermetz:
Depends on #520

```sh
This makes it so that the linter will error
out when there are unused variables.
This is more important than it looks because unused imports can contribute
a lot of overhead to the build artifact sizes.

```


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #528, opened 2021-02-02
feat(common): servers.startOnPort() host arg #527

### LINKED ISSUE #527: feat(common): servers.startOnPort() host arg

Right now if I call `Servers.startOnPort(...)` there's only one argument, for the port and the host defaults to the loopback interface which makes the whole thing useless if you need to bind to `0.0.0.0` or any other network interface.

Add an optional `host` argument that defaults to loopback for both security and backwards compatibility reasons.

Considered not having the host arg as optional but decided against it since we have a default value for it that can be considered both secure and backwards compatible (a rare, but most welcome sight in software engineering)

**Additional context**

This came up as part of the ongoing work to containerize the project, more specifically the supply chain app example (the containerization of which is mostly equivalent to having containerized the entire project anyway since it uses most of the existing packages of Cactus - aka it's a top level package).

petermetz:
Extends the existing functionality of the utility method so that it can
accept a 'host' parameter where callers can
specify if they want to bind to the wildcard
network interface of `0.0.0.0` for example.

The default option of the host argument is
localhost so that we are *secure by default*
even if this method is only used for testing
code, it cannot hurt to be on the safe side.

Fixes #527

### DIFF
--- a/packages/cactus-common/src/main/typescript/servers.ts
+++ b/packages/cactus-common/src/main/typescript/servers.ts
@@ -20,17 +20,14 @@ export class Servers {
    * @param server The server object that will be shut down.
    */
   public static async shutdown(server: Server | SecureServer): Promise<void> {
+    const fnTag = `Servers#shutdown()`;
     if (!server) {
-      throw new TypeError(`Servers#shutdown() server was falsy. Need object.`);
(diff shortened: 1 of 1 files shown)


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #530, opened 2021-02-03
fix(deps): missing uuid from plugins #529

### LINKED ISSUE #529: fix(deps): missing uuid from plugins
**Describe the bug**

The `uuid` package is being used by two plugin packages while it not being declared as a dependency. (it works for tests because the uuid package is also a dependency in the parent package most likely so node's module resolution algorithm ends up finding it when there's a `./node_modules` directory present in the project root with the dependencies installed.

**To Reproduce**

Open these files and observe the lack of `uuid` dependency:

`./packages/cactus-plugin-consortium-manual/package.json`
`./packages/cactus-plugin-keychain-memory/package.json`

**Expected behavior**

`uuid` should be a dependency in these packages since they use it as an import.

Inside a docker container running on a dev machine.

`main`

`./packages/cactus-plugin-consortium-manual/package.json`
`./packages/cactus-plugin-keychain-memory/package.json`

**Additional context**

This came up while containerizing the supply-chain-app example where we had to delete the root folder's node_modules directory while trying to reduce the image size from the initial 3 GB (yikes).
After having deleted said directory the `uuid` module failed to resolve in the mentioned packages which is how this bug was discovered.

petermetz:
Adds the missing dependency "uuid" to the packages where applicable.

Fixes #529

### DIFF
--- a/packages/cactus-plugin-consortium-manual/package-lock.json
+++ b/packages/cactus-plugin-consortium-manual/package-lock.json
@@ -136,6 +136,12 @@
 				"@types/mime": "*"
 			}
 		},
+		"@types/uuid": {
+			"version": "8.3.0",
+			"resolved": "https://registry.npmjs.org/@types/uuid/-/uuid-8.3.0.tgz",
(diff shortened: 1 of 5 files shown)


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

Issue #512, opened 2021-01-25
feat(besu): Eth atomic swaps using HTLCs


petermetz:
**Is your feature request related to a problem? Please describe.**

A common approach to move assets across different DLTs in a trustless manner is the use of atomic swaps.
The proposed implementation is using HashTimeLockedContracts (HTLCs)
We need a Cactus plugin for executing atomic swaps via HTLCs for ETH (e.g. the default currency on the Ethereum/based ledger(s)). There's a separate task for ERC-20 tokens that is otherwise the same task as this.

**Describe the solution you'd like**

Some inspiration could be taken from [this](https://github.com/jelly-swap/jelly-eth-htlc), but the solidity code itself is just the tip of the ice-berg since we have to work within the plugin architecture and work out other bells and whistles such as contract deployment so that the overall experience for a developer using Cactus as a framework is pleasant rather than overwhelming.

There's a lot more information in the parent issue where the common theme is set for all these sub-tasks related to HTLC development for Cactus.

On a high level the solution should consist of 
1. the solidity code and the boilerplate to make that solidity code callable via the Cactus API clients:
2. The plugin package under `./packages/cactus-plugin-htlc-eth-besu` (naming things is a difficult problem, feel free to suggest better names)
3. OpenAPI spec json file (in the plugin package)
4. Generated API client in Typescrit (based on the OpenAPI spec)
5. The plugin has to be able to deploy the solidity code to the underlying ledger (solidity code to be embedded as a JSON resource, see parent issue for explanation)

**Describe alternatives you've considered**

`N/A`

**Acceptance Criteria**

- [x] Interactions with the services must be defined via openapi
- [x] Plugin must check the required smart contracts are deployed or try to initiate a deployment if possible
- [x] Plugin must handle HTLC creation to a specific receiver, with a specific ether amount, a hashed secret and a limit timestamp
- [x] Plugin must handle HTLC resolution for both secret disclosure (provided manually as parameter, the core HTLC plugin would be in charge of secret recovery from different chains), and asset recovery due to timeouts (Executed manually).
- [x] Plugin must handle HTLC state queries
- [x] Test coverage of 75%+
**Additional context**

Parent: feat(core-api): atomic swaps with HTLCs #510
Sibling: feat(besu): erc-20 HTLC #513

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton

kikoncuo:
[...]

PR #532, opened 2021-02-03
Feature/htlc eth besu

kikoncuo:
Added functionality to used HTLCs for Eth.

Interface defined in JSON, created for #512 

Status WIP: 
- Using the core besu (instead of using web3)
- Finishing test coverage using TAP


kikoncuo:
Rebase required @AzaharaC

PR #526, opened 2021-02-02
fix: webpack 5 NodeJS polyfills for web GUI

petermetz:
Fixes #486

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

Issue #344, opened 2020-11-04
docs(examples): containerize supply chain app example


petermetz:
### Description

Depends on https://github.com/hyperledger/cactus/issues/345

As a person evaluating Cactus for my business use-case I want to have an ever more streamlined way of running the currently available supply chain example so that I spend most of my available time actually discovering Cactus not fighting with the build system/compiler/my own OS level dependencies/package managers and all the other usual suspects that make computing difficult.

### Acceptance Criteria
1. There is a `Dockerfile` which gets auto-built and published to DockerHub similar how we are already doing it with the AIO test images

cc: @jonathan-m-hamilton @takeutak @sfuji822

petermetz:
Hmm, this was done already, maybe we forgot to close the issue.

PR #555, opened 2021-02-06
docs(examples): containerize supply-chain-app #344

### LINKED ISSUE #486: fix(examples): supply chain app regression - polyfills
**Describe the bug**

Multiple issues:
1. Buffer polyfill not present
2. process polyfill not present
3. ~~www root directory mapping doesn't work~~ (will fix as part of another issue)
4. ~~API client methods not updated to match operation names (some are, some aren't)~~ (will fix as part of another issue)
5. ~~npm i --no-package-lock fails~~ (will fix as part of another issue - via containerization)

**To Reproduce**

Follow the steps presented in the README and watch it fail.

**Expected behavior**

Supply chain app example works, GUI can be used to demonstrate the whole example end to end.

Affects the latest revision but also `v0.3.0` it seems.

Quorum, Besu

petermetz:
Depends on #526 
Depends on #528 
Depends on #530 

### Commit to be reviewed:

```sh
docs(examples): containerize supply-chain-app

Uses a Docker-in-Docker container to run the supply
chain app example which itself has functionality that
spawns docker containers to host the AIO ledger images.
This makes it a 100% turn-key way of providing examples
to anyone who have at least the ability to run docker
containers built for the linux amd64 platform.

Fixes #486

```

### DIFF
--- a/examples/supply-chain-app/Dockerfile
+++ b/examples/supply-chain-app/Dockerfile
@@ -1,0 +1,139 @@
+FROM node:12.20.1-alpine3.12 as builder
+
+RUN apk update
+
+# Need JDK for the openapi-generator that is part of the build process
+RUN apk add --no-cache openjdk8=8.252.09-r0
+
+# Need git because some of our npm depedencies might be coming
+# straight from github instead of being an npm package on npmjs.com.
+RUN apk add --no-cache git
+
+# Some install scripts of the npm package fabric-network need python
+RUN apk add --no-cache python3 py3-pip
+
+RUN npm install modclean -g
+
+WORKDIR /
+RUN mkdir /app/
+WORKDIR /app/
+COPY ./ ./
+RUN npm ci
+RUN ./node_modules/.bin/lerna clean --yes
+RUN ./node_modules/.bin/lerna bootstrap
+RUN npm run build:dev:backend
+RUN npm run webpack:dev:web
+RUN npm run build:dev:frontend -- --scope='@hyperledger/cactus-example-supply-chain-frontend'
+RUN ./node_modules/.bin/lerna clean --yes
+RUN ./node_modules/.bin/lerna bootstrap -- --production --no-optional
+
+# RUN modclean --run --patterns="default:*" --path ./examples/cactus-example-supply-chain-backend/
+# RUN modclean --run --patterns="default:*" --path ./examples/cactus-example-supply-chain-business-logic-plugin/
+# RUN modclean --run --patterns="default:*" --path ./examples/cactus-example-supply-chain-frontend/
+# RUN modclean --run --patterns="default:*" --path ./packages/cactus-api-client/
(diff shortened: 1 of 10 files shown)


jonathan-m-hamilton (approved):
LGTM

sfuji822 (commented):
I commented minior issue.
If those are intened, I will withdraw my comments.

examples/supply-chain-app/Dockerfile
```diff
@@ -0,0 +1,139 @@
...
+RUN apk add --no-cache python3 py3-pip
+
+RUN npm install modclean -g
+
+WORKDIR /
+RUN mkdir /app/
+WORKDIR /app/
+COPY ./ ./
+RUN npm ci
+RUN ./node_modules/.bin/lerna clean --yes
```
sfuji822, line 22: Why you did not use npx ?

examples/supply-chain-app/Dockerfile
```diff
@@ -0,0 +1,139 @@
...
+# Add bash for convenience, sh has less features
+RUN apk add --no-cache bash=5.0.17-r0
+
+# Need curl for healthchecks
+RUN apk add --no-cache curl=7.69.1-r3
+
+# The file binary is used to inspect exectubles when debugging container image issues
+RUN apk add --no-cache file=5.38-r0
+
+RUN apk add --no-cache nodejs=12.20.1-r0
```
sfuji822, line 67: It looks wierd to update nodejs and npm again.

examples/supply-chain-app/Dockerfile
petermetz: @sfuji822 Usually I leverage npx to stay operating system neutral with the commands I'm executing, but in the container there's no need for it because the container image is fixed for a specific OS that I know is guaranteed to work with the syntax of `./node_moules/.bin/...`.
The other part of it is that npx is slower than directly invoking the executable file so that's why I only use npx when I want to make sure that the command works on different operating systems as well (for example I had bugs some time ago where windows was failing with the forward slashes).

examples/supply-chain-app/Dockerfile
petermetz: > It looks wierd to update nodejs and npm again.

@sfuji822 I agree. :-) Unfortunately the Alpine Linux packagers decided to package them separately for some reason so it's not actually installing npm twice because on Alpine Linux if you install the nodejs package then that will not give you npm out of the box like it does on other operating systems.

examples/supply-chain-app/Dockerfile
sfuji822: I got it. It will be OK for now.

petermetz: @sfuji822 Cool, thanks. Any chance that you could approve the pull request? The governance says that it cannot be merged without approval.

sfuji822 (approved):
LGTM.

Issue #563, opened 2021-02-09
fix(connector-besu): Calls to contracts with parameters are failing


kikoncuo:
**Describe the bug**

When performing a call to a smart contract in order to read a value stored in the state, if the call contains parameters the operation will fail

**To Reproduce**
In this example calling getSingleStatus with the param id, the node will revert the transaction
```
const result = await this.connector.invokeContract({
  contractAbi: this.abi[environment.CONTRACT_NAME],
  contractAddress: environment.CONTRACT_ADDRESS,
  invocationType: EthContractInvocationType.CALL,
  methodName: "getSingleStatus",
  params: [id],
  web3SigningCredential: {
    ethAccount: environment.ACCOUNT_ADDRESS,
    type: Web3SigningCredentialType.PRIVATEKEYHEX,
    secret: environment.PRIVATE_KEY,
  },
});
```

**Expected behavior**

The value is returned

kikoncuo:
The parameters are passed in call() execution. That's for extra params (IE: sender, gas...) not the function params.
https://github.com/hyperledger/cactus/blob/add9cd94ed7fe4a7facda89749895f7b04de89a8/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts#L201

The params are already specified in https://github.com/hyperledger/cactus/blob/add9cd94ed7fe4a7facda89749895f7b04de89a8/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts#L198

kikoncuo:
Working on #564

PR #565, opened 2021-02-09
fix(connector-besu): avoid req.params repetition

kikoncuo:
Fix for #563


petermetz (approved):

jordigiam:
Added test for calls with params

takeutak (approved):

Issue #458, opened 2021-01-06
feat(monitoring): add prometheus metrics support to the OpenAPI specs of plugins & API server


petermetz:
**Is your feature request related to a problem? Please describe.**

Right now, if we were to run Cactus in production, there'd have to be custom logging/metrics implemented in order to be able to respond to (seemingly) simple questions such as "what is the current, effective TPS on my Cactus node that I operate?"
Or how many transactions have succeeded/failed in the last 24 hours? etc..

**Describe the solution you'd like**

1. A prometheus `.../metrics` endpoint exposed by all plugins and the API server as well.
2. Would also be nice to have a pre-built grafana Dashboard so that people who deploy Cactus can get a detailed and visual overview of what their node(s) are doing.
3. It's important that we have each plugin the ability to host it's own metrics because it will be difficult to define a common set of metrics that can work for all of them (I think?). Also, we do  want to have the ability to monitor the plugins individually, e.g. I don't want to only able to see the summed up transaction per second values for a collection of plugins that talk to different ledgers, I want to see each plugin by themselves as well. Granularity is important

**Describe alternatives you've considered**

The other big player in the monitoring/APM game is the ELK stack, I'm not opposed to baking that in as well, but what I definitely do not want is making them mutually exclusive. People should be able to choose and without much hassle or confusion.

**Additional context**

Kubernetes is (probably) the most popular container orchestration platform out there with a steady growth in adoption. [1] [2]
Because of this, longer term we should have ops capabilities that make life easier for people who run Cactus nodes on Kubernetes. Prometheus (anecdotally, seems) to be the monitoring solution of choice for people using Kubernetes in production and that's why I figured we should make some effort here, eventually: At the time of this writing, I'm not putting this down as a 1.0 milestone requirement, but will put it in 1.2 because I believe it will become very important as soon as we have production deployments of Cactus out there in the wild.

[1] https://enterprisersproject.com/article/2020/6/kubernetes-statistics-2020
[2] https://www.datadoghq.com/container-report/

cc: @takeutak @jonathan-m-hamilton @hartm @sfuji822

jagpreetsinghsasan:
I would like to work on this (currently working).

jagpreetsinghsasan:
Created issue #531 for the fabric plugin

jagpreetsinghsasan:
Acceptance Criteria:
1. A prometheus `../metrics` endpoint exposed by all the plugins.
[...]

PR #576, opened 2021-02-10
feat(fabric): prometheus exporter support

### LINKED ISSUE #531: feat(fabric): add prometheus metrics support to the OpenAPI specs of the plugin
### Description
As a developer, I want to integrate Prometheus Exporter for the Hyperledger Fabric Cactus plugin, Issue #458 details the description in depth and covers up for all the underlying plugins.

cc: @petermetz @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo

jagpreetsinghsasan:
## Commit to be reviewed

feat(fabric): add prometheus metrics support to the openapi specs of the plugin

	Primary Change
	--------------

	1. The fabric ledger connector plugin now includes the prometheus metrics exporter integration
	2. OpenAPI spec now has api endpoint for the getting the prometheus metrics

	Refactorings that were also necessary to accomodate 1) and 2)

	3. The generate-sdk command now has DateTime mapping to Date (to read Date as Date object instead of string)
	4. GetPrometheusMetricsV1 class is created to handle the corresponding api endpoint
	5. IPluginLedgerConnectorFabricOptions interface in PluginLedgerConnectorFabric class now has a prometheusExporter optional field
	6. The PluginLedgerConnectorFabric class has relevant functions and codes to incorporate prometheus exporter
	7. run-transaction-endpoint-v1.test is changed to incorporate the prometheus exporter for both fabric 1.4.x and 2.x versions
	8. Added Readme.md on the prometheus exporter usage

	Fixes #531

Sub issue of #458 
Fixes #531


packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v1-4-x/run-transaction-endpoint-v1.test.ts
```diff
@@ -173,6 +174,23 @@ test("runs tx on a Fabric v1.4.8 ledger", async (t: Test) => {
     t.ok(car277.Record.owner, `Car object has "Record"."owner" property OK`);
     t.equal(car277.Record.owner, carOwner, `Car has expected owner OK`);
   }
-
+  await new Promise((resolve) => setTimeout(resolve, 200000));
+  {
+    const promExporter = plugin.getPrometheusExporter();
```
petermetz, line 179: @jagpreetsinghsasan How will this work in a browser environment where you don't have access to the `PluginLedgerConnectorFabric` instance?

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/get-prometheus-exporter-metrics/get-prometheus-exporter-metrics-v1.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import { PluginLedgerConnectorFabric } from "../plugin-ledger-connector-fabric";
+import { PrometheusExporter } from "../prometheus-exporter/prometheus-exporter";
+
+export interface IGetPrometheusExporterMetricsV1Options {
+  logLevel?: LogLevelDesc;
+  connector: PluginLedgerConnectorFabric;
+}
+
+export class GetPrometheusExporterMetricsV1 implements IWebServiceEndpoint {
```
petermetz, line 27: @jagpreetsinghsasan Could you add the `Endpoint` suffix to the class and the options interface names? I know it makes it super long but hopefully keeping with the convention will pay dividends on the long run.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/get-prometheus-exporter-metrics/get-prometheus-exporter-metrics-v1.ts
```diff
@@ -0,0 +1,91 @@
...
+    registerWebServiceEndpoint(app, this);
+    return this;
+  }
+
+  async handleRequest(req: Request, res: Response): Promise<void> {
+    const fnTag = "GetPrometheusExporterMetrics#handleRequest()";
+    this.log.debug(`POST ${this.getPath()}`);
+
+    try {
+      let prometheusExporterObject: PrometheusExporter = new PrometheusExporter(
```
petermetz, line 68: @jagpreetsinghsasan Question: Are we wiping the state on each request by creating a new instance every time? As in, would this test pass?
1. transact() method of plugin runs
2. I get the metrics, total tx count === 1
3. transact() method of plugin runs
4. I get the metrics, total tx count === 2 (if we wipe the state then it looks like it would be 1 still).

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/prometheus-exporter/prometheus-exporter.ts
```diff
@@ -0,0 +1,51 @@
+import promClient from "prom-client";
+import { Transaction, Transactions } from "./response.type";
+import { collectMetrics } from "./data-fetcher";
+import { PrometheusExporterMetricsResponse } from "../generated/openapi/typescript-axios";
+
+export interface PrometheusExporterOptions {
+  pollingIntervalInMin?: number;
+}
+
+export interface PrometheusExporterResponse {
```
petermetz, line 10: @jagpreetsinghsasan Is this redundant with the response definition coming from the openapi.json?

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/prometheus-exporter/prometheus-exporter.ts
```diff
@@ -0,0 +1,51 @@
+import promClient from "prom-client";
+import { Transaction, Transactions } from "./response.type";
+import { collectMetrics } from "./data-fetcher";
+import { PrometheusExporterMetricsResponse } from "../generated/openapi/typescript-axios";
+
+export interface PrometheusExporterOptions {
```
petermetz, line 6: @jagpreetsinghsasan Any chance to rename this to have an `I` prefix? (convention)

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v1-4-x/run-transaction-endpoint-v1.test.ts
```diff
@@ -173,6 +174,23 @@ test("runs tx on a Fabric v1.4.8 ledger", async (t: Test) => {
     t.ok(car277.Record.owner, `Car object has "Record"."owner" property OK`);
     t.equal(car277.Record.owner, carOwner, `Car has expected owner OK`);
   }
-
+  await new Promise((resolve) => setTimeout(resolve, 200000));
```
petermetz, line 177: @jagpreetsinghsasan Can we make it work without having to do an explicit wait?

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v1-4-x/run-transaction-endpoint-v1.test.ts
jagpreetsinghsasan: I have fixed in the new commit.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/get-prometheus-exporter-metrics/get-prometheus-exporter-metrics-v1.ts
jagpreetsinghsasan: Fixed

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/get-prometheus-exporter-metrics/get-prometheus-exporter-metrics-v1.ts
jagpreetsinghsasan: Fixed.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/prometheus-exporter/prometheus-exporter.ts
jagpreetsinghsasan: Fixed.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/prometheus-exporter/prometheus-exporter.ts
jagpreetsinghsasan: Fixed.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v1-4-x/run-transaction-endpoint-v1.test.ts
jagpreetsinghsasan: Fixed

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/get-prometheus-exporter-metrics/get-prometheus-exporter-metrics-endpoint-v1.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  async handleRequest(req: Request, res: Response): Promise<void> {
+    const fnTag = "GetPrometheusExporterMetrics#handleRequest()";
+    this.log.debug(`POST ${this.getPath()}`);
+
+    try {
+      const resBody = await this.opts.connector.getPrometheusExporterMetrics();
+      res.status(200);
+      res.send(resBody.result.replace(/\\/g, ""));
```
petermetz, line 72: @jagpreetsinghsasan If you are returning plain text (which I'm guessing Prometheus forces you to do) then the response definition in openapi.json should match that as well. Right now it looks like there's a disconnect between what the openapi.json specifies and what the endpoint implementation actually returns

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
```diff
@@ -286,6 +286,42 @@
                         "type": "string"
                     }
                 }
+            },
+            "PrometheusExporterOptions": {
```
petermetz, line 290: @jagpreetsinghsasan Do we still need this in the request? IIRC HTTP GET requests cannot have a body.

packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
```diff
@@ -286,6 +286,42 @@
...
+                    }
+                }
+            },
+            "PrometheusExporterMetricsRequest": {
+                "type": "object",
+                "required": [
+                    "promExporter"
+                ],
+                "properties": {
+                    "promExporter":{
```
petermetz, line 309: @jagpreetsinghsasan Do we still need this in the request? IIRC HTTP GET requests cannot have a body.

packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
```diff
@@ -368,6 +404,40 @@
...
+                "x-hyperledger-cactus": {
+                    "http": {
+                        "verbLowerCase": "get",
+                        "path": "/api/v1/plugins/@hyperledger/cactus-plugin-ledger-connector-fabric/get-prometheus-exporter-metrics"
+                    }
+                },
+                "operationId": "getPrometheusExporterMetricsEndpointV1",
+                "summary": "Get the Prometheus Metrics",
+                "parameters": [],
+                "requestBody": {
```
petermetz, line 419: @jagpreetsinghsasan Do we still need this in the request? IIRC HTTP GET requests cannot have a body.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
```diff
@@ -243,6 +284,12 @@ export class PluginLedgerConnectorFabric
         functionOutput: outUtf8,
       };
       this.log.debug(`transact() response: %o`, res);
+      const endTransactionTime = new Date();
+      this.prometheusExporter.addCurrentTransaction(
```
petermetz, line 288: @jagpreetsinghsasan If initially we only support the count of total transactions (which I agree with) then there's no need for us to store the start and end times. We could keep them anyway, but then the code is a bit harder to read for those who are trying to see how it works because the start and end transaction times are not actually being used by the exporter. The other thing is that if we leave them in, then we also have to come up with some way to clear them out of memory over time to avoid memory leaks => with the current format if I read it right it just keeps pumping the array with new objects into infinity, so it's easier (since we only return totalTxCount anyway) to just not create nor store any new objects and instead just increment a single variable that stores the tx count.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/prometheus-exporter/Readme.md
```diff
@@ -0,0 +1,39 @@
+# Prometheus Exporter
```
petermetz, line 1: @jagpreetsinghsasan Please move the content from this file to the `README.md` file in the package's root directory and increase the header levels by one in each section (e.g. `#` becomes `##`, `##` becomes `###` etc..)
This file itself then can be deleted so that in the src/main/typescript directory with don't end up having documentation that's not typescript code.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1.test.ts
```diff
@@ -174,5 +176,22 @@ test("runs tx on a Fabric v2.2.0 ledger", async (t: Test) => {
     t.equal(car277.Record.owner, carOwner, `Car has expected owner OK`);
   }
 
+  // await new Promise((resolve) => setTimeout(resolve, 200000));
+  {
+    const promExporter = plugin.getPrometheusExporter();
+    const res = await apiClient.getPrometheusExporterMetricsEndpointV1({
+      promExporter: promExporter,
```
petermetz, line 183: @jagpreetsinghsasan Same as above, if the endpoint is now `HTTP GET` then the request body is not actually being used IIRC and therefore can be removed altogether.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/prometheus-exporter/metrics.ts
```diff
@@ -0,0 +1,7 @@
+import { Gauge } from "prom-client";
+
+export const totalTxCount = new Gauge({
+  name: "totalTxCount",
```
petermetz, line 4: @jagpreetsinghsasan While I was testing this, typing it to the Prometheus expression console, I realized that once we have multiple metrics, the naming here would become ambiguous. Any chance that you could update the metrics name to uniquely identify that 1) it's a Cactus metric and 2) within Cactus it's specific to fabric?
Something like this `cactus_fabric_total_tx_count` could work (and this would have the added benefit of following the (seeming) convention in Prometheus to have underscores in the names, not camelCase

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
jagpreetsinghsasan: Fixed

packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
jagpreetsinghsasan: Fixed

packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
jagpreetsinghsasan: Fixed

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
jagpreetsinghsasan: Fixed

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/prometheus-exporter/Readme.md
jagpreetsinghsasan: Fixed

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1.test.ts
jagpreetsinghsasan: Fixed

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/prometheus-exporter/metrics.ts
jagpreetsinghsasan: fixed

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/get-prometheus-exporter-metrics/get-prometheus-exporter-metrics-endpoint-v1.ts
jagpreetsinghsasan: The issue is, the response definition looks like
image

But the response I am getting from the server is something like
'# HELP cactus_fabric_total_tx_count Total transactions executed\n# TYPE cactus_fabric_total_tx_count gauge\n     'cactus_fabric_total_tx_count{type=\"cactus_fabric_total_tx_count\"} 3';

It is escaping the double quotes using '\'
Which is not acceptable by prometheus
So I used the above code to remove the same. (using replace fx)
Is there any other way to fix it?

petermetz: @jagpreetsinghsasan I've removed the replacement logic and it was working fine in both the automated tests and the manual tests (where I actually pull up prometheus and see if the graph gets charted properly with the total tx count).
Here are my changes on top of your branch that I'd recommend applying as part of this change request: https://github.com/petermetz/cactus/commit/82aa01cf431ef1e2e947f6d6470151508e6f861b

jagpreetsinghsasan: Fixed.

petermetz (approved):

petermetz (approved):

takeutak (approved):

PR #583, opened 2021-02-16
fix(test-tooling): removed latest tag

### LINKED ISSUE #421: fix: test ledger image tag references to be exact
**Describe the bug**

Currently the test ledger classes all default to `latest` as the default tag if unspecified via constructor options.
This is bad because whenever we have a change to any one of the images, there's this limbo and false negative (failing) CI runs if the new version of the image has some incompatibility (which they most of the time do). 

**To Reproduce**

Make a breaking change to an image, try to run the integration tests or have a PR that passes the CI and observe your own demise.

**Expected behavior**

Test ledger classes declare explicit/exact tag references associated with a specific SHA so that a given version of the code has stable test outcomes as long as the docker registry does not start disappearing image versions or changing them.

Every environment suffers from this.

Every release suffers from this.
Currently on `v0.2.0`

Fabric is the notorious culprit because our image has issues there (and the fix is stuck in the review queue for now).

**Additional context**

This is a great example of why we need reproducible builds, which the docker containers are a part of.

cc: @takeutak @sfuji822 @jonathan-m-hamilton @jagpreetsinghsasan

jordigiam:
Fixes #421

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/besu/besu-test-ledger.ts
@@ -18,7 +18,7 @@ export interface IBesuTestLedgerConstructorOptions {
 }
 
 export const BESU_TEST_LEDGER_DEFAULT_OPTIONS = Object.freeze({
-  containerImageVersion: "latest",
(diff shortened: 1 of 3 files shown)


petermetz (approved):
LGTM

takeutak (approved):

Issue #590, opened 2021-02-20
fix(example): fix README on car-trade


takeutak:
**Describe the bug**
The README on example/car-trade lacks some information

**To Reproduce**
Fix README

petermetz:
@takeutak Can this issue be closed now that the linked PR has been merged?

PR #591, opened 2021-02-20
fix(example): fix README on examples/car-trade

takeutak:
related with Issue #590

### DIFF
--- a/examples/cartrade/README.md
+++ b/examples/cartrade/README.md
@@ -6,7 +6,7 @@ Before you begin, you need to check that you have all the prerequisites installe
 - OS: Linux (recommended: Ubuntu18.04 or CentOS7)
 - Docker (recommend: v17.06.2-ce or greater)
 - Docker-compose (recommend: v1.14.0 or greater)
-- node.js (recommend: v12)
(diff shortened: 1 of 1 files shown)


kikoncuo (commented):
NodeJS 12 makes sense, no need to specify the minor version, our specific version we are using is 12.13.0.
Ubuntu 18.04 makes sense due to our environments and tests.
We don't do our tests against CentOS, but if that's the version you are directly using we can keep it, otherwise we can just leave Ubuntu

takeutak:
Thank you for your comment.
As for Node 12, I think "NodeJS v 12" is enough, but I added the subversion of Node.js that I tested as recommended version just in case.
I have tested this code on both Ubuntu 18.04 and CentOS and both work fine. Therefore, I think there is no need to modify the OS notation.
I think there is no problem here. Can you give me a comment if you have anything?

kikoncuo (approved):
No other comments looks good!

petermetz (approved):

petermetz:
FYI: re-triggered the CI just now, looked like a fluke.

PR #596, opened 2021-02-23
docs(examples): supply chain app with modifiable settings

### LINKED ISSUE #347: docs(examples): supply chain app WWW root config
### Description

The supply chain app example configures its API server at the moment with a WWW path for the front-end that is hardcoded in the back-end package. This makes it inflexible and in certain situations will just break the example because the back-end will not serve the UI if the path is incorrect.

We need to have a way of passing in a CLI/ENV arg/variable with the WWW root path so that this can be remedied by the users regardless of how they choose to install/try out the example.

### Acceptance Criteria
1. The example back-end can serve the front-end assets regardless of what is the `current working directory` where it was started from.

cc: @sfuji822 @jonathan-m-hamilton @takeutak

jordigiam:
Resolve #347


examples/supply-chain-app/package.json
```diff
@@ -48,7 +48,8 @@
   "homepage": "https://github.com/hyperledger/cactus#readme",
   "dependencies": {
     "@hyperledger/cactus-example-supply-chain-backend": "file:../cactus-example-supply-chain-backend",
-    "@hyperledger/cactus-example-supply-chain-frontend": "file:../cactus-example-supply-chain-frontend"
+    "@hyperledger/cactus-example-supply-chain-frontend": "file:../cactus-example-supply-chain-frontend",
+    "dotenv": "^8.2.0"
```
petermetz, line 52: @jordigiam Please lock the version down to an exact one (so no caret, wildcard, etc).

petermetz (changes requested):

examples/supply-chain-app/package.json
jordigiam: Done! Thanks for the advice!

petermetz (approved):

takeutak (approved):

PR #607, opened 2021-03-01
fix(api-client): flaky DefaultConsortiumProvider test #605

### LINKED ISSUE #605: fix(api-client): flaky DefaultConsortiumProvider test
**Describe the bug**

There's this one test that I've seen the second time failing now (and it's a fluke with a timeout).

**To Reproduce**

It's random so no repro steps.

**Expected behavior**

Tests should be consistent.

**Logs/Stack traces**

https://github.com/hyperledger/cactus/pull/602/checks?check_run_id=1985681878#step:10:27

**Screenshots**

`N/A`

GitHub Action Execution environment. Never seen it on my dev machines.

See the ci.yml file for the OS details.

`main`

The manual consortium management plugin.
 
**Additional context**

There is a test that uses a timeout to assert whether something was successful or not. Currently this timeout is a 100ms which appears to be too low nowadays (could indicate that our GH actions are getting less CPU power than before?).
Likely just a matter of bumping up the timeout to 10x or 100x the current value, but would be nice if the need for the timeout could be completely eliminated (while preserving the test case and the assertions themselves of course).

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo

petermetz:
Fixes #605

### DIFF
--- a/packages/cactus-api-client/src/test/typescript/integration/default-consortium-provider.test.ts
+++ b/packages/cactus-api-client/src/test/typescript/integration/default-consortium-provider.test.ts
@@ -2,7 +2,6 @@ import { AddressInfo } from "net";
 
 import test, { Test } from "tape";
 
-// import { ApiClient } from "../../../main/typescript/public-api";
 import { DefaultApi as ConsortiumManualApi } from "@hyperledger/cactus-plugin-consortium-manual";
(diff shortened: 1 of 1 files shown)


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #618, opened 2021-03-02
feat(core-api): common consortium interface

### LINKED ISSUE #427: feat(core-api): define common consortium plugin interface

Currently we have the manual consortium plugin [1] as the only implementation in existence that is out there to provide other plugins with a list of nodes and other consortium related metadata. 

What we need is the common interface already implied by the functionality of [1] to be actually be present in a Typescript interface that lives in the `core-api` package so that the implementation of other consortium management plugins can use it as well.

[1] `./packages/cactus-plugin-consortium-manual/src/main/typescript/plugin-consortium-manual.ts`

1. Define the request/response definitions (just pull the ones currently in the consortium manual  plugin package)
2. Define something like this (again, nothing new here just pulling the existing functionality into a separate interface for re-use by others)
    ```typescript
    import {
      ICactusPlugin,
      GetNodeJwsRequest,
      GetNodeJwsResponse,
      GetConsortiumJwsRequest,
      GetConsortiumJwsResponse,
    } from "@hyperledger/cactus-core-api";
    export interface IPluginConsortium extends ICactusPlugin {

      /**
       * Returns the JSON Web Signature of the consortium metadata, as issued by the
       * **current node** that the request has been issued to.
       * @param req The parameters of the request, if any.
       */
      getNodeJws(req: GetNodeJwsRequest): Promise<GetNodeJwsResponse>;

      /**
       * Returns the JSON Web Signature of the consortium metadata, as issued by
       * **all the nodes in the consortium** that are known to be in the consortium
       * by this current node that the request has been issued to.
       *
       * @param req The parameters of the request, if any.
       */
      getConsortiumJws(req: GetConsortiumJwsRequest): Promise<GetConsortiumJwsResponse>;
    }
    ```

Nothing, I'm pretty sure we need to do this because without this other's are blocked from authoring their own consortium plugins which is not what we want.

**Additional context**

N/A

jordigiam:
Resolve #427

### DIFF
--- a/packages/cactus-core-api/src/main/typescript/plugin/consortium/i-plugin-consortium.ts
+++ b/packages/cactus-core-api/src/main/typescript/plugin/consortium/i-plugin-consortium.ts
@@ -1,0 +1,27 @@
+import { ICactusPlugin } from "../i-cactus-plugin";
+
+/**
+ * Common interface to be implemented by plugins which are implementing the connection to ledgers.
+ */
+export interface IPluginConsortium<
+  GetNodeJwsIn,
+  GetNodeJwsOut,
+  GetConsortiumJwsIn,
+  GetConsortiumJwsOut
+> extends ICactusPlugin {
+  /**
+   * Returns the JSON Web Signature of the consortium metadata, as issued by the
+   * **current node** that the request has been issued to.
+   * @param req The parameters of the request, if any.
+   */
+  getNodeJws(req?: GetNodeJwsIn): Promise<GetNodeJwsOut>;
+
+  /**
+   * Returns the JSON Web Signature of the consortium metadata, as issued by
+   * **all the nodes in the consortium** that are known to be in the consortium
+   * by this current node that the request has been issued to.
+   *
+   * @param req The parameters of the request, if any.
+   */
+  getConsortiumJws(req?: GetConsortiumJwsIn): Promise<GetConsortiumJwsOut>;
+}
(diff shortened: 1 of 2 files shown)


packages/cactus-core-api/src/main/typescript/plugin/consortium/i-plugin-consortium.ts
```diff
@@ -0,0 +1,29 @@
+import { ICactusPlugin } from "../i-cactus-plugin";
+
+/**
+ * Common interface to be implemented by plugins which are implementing the connection to ledgers.
+ */
+export interface IPluginConsortium<
+  GetNodeJwsRequest,
```
petermetz, line 7: @jordigiam Generic type parameters having the exact same name as the classes can be a bit confusing in other places of the code where the actual classes will be imported alongside the type parameters being mentioned. Any chance that you could give them some other name in the interest of disambiguation? Could go with just the old school single letter kind of things from java like `T1, T2, T3, T4` but I also think that those are not very expressive so maybe something in-between that still conveys the meaning and yet different from the actual class names like `GetNodeJwsIn, GetNodeJwsOut, GetConsortiumJwsIn, GetConsortiumJwsOut`. These also have the added benefit that the in/out are more generic that can apply to function calls whereas the words `response` and `request` always make (me) think of HTTP only.

petermetz (changes requested):

packages/cactus-core-api/src/main/typescript/plugin/consortium/i-plugin-consortium.ts
jordigiam: Sure! Makes sense.

petermetz (approved):

takeutak (approved):

PR #659, opened 2021-03-12
feat(keychain-vault): add prometheus exporter

### LINKED ISSUE #536: feat(keychain-vault): add prometheus metrics support to the OpenAPI specs of the plugin
### Description
As a developer, I want to integrate Prometheus Exporter for the Keychain Vault Cactus plugin, Issue #458 details the description in depth and covers up for all the underlying plugins.

cc: @petermetz @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo

### LINKED ISSUE #537: feat(keychain-memory): add prometheus metrics support to the OpenAPI specs of the plugin
### Description
As a developer, I want to integrate Prometheus Exporter for the Keychain Memory Cactus plugin, Issue #458 details the description in depth and covers up for all the underlying plugins.

cc: @petermetz @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo

jagpreetsinghsasan:
Commit to be reviewed
feat(keychain-vault): add prometheus metrics	
	
	Primary Change
	--------------

	1. The keychain-vault plugin now includes the prometheus metrics exporter integration
	2. OpenAPI spec now has api endpoint for the getting the prometheus metrics

	Refactorings that were also necessary to accomodate 1) and 2)

	3. GetPrometheusMetricsV1 class is created to handle the corresponding api endpoint
	4. IPluginKeychainVaultOptions interface in PluginKeychainVault class now has a prometheusExporter optional field
	5. The PluginKeychainMemory class has relevant functions and codes to incorporate prometheus exporter
	6. Added Readme.md on the prometheus exporter usage

Resolve #536

### DIFF
--- a/packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
+++ b/packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
@@ -4,6 +4,7 @@ import { Server as SecureServer } from "https";
 import { Express } from "express";
 import { Optional } from "typescript-optional";
 import Vault from "node-vault";
+import HttpStatus from "http-status-codes";
 
 import {
   Logger,
@@ -18,8 +19,18 @@ import {
   IWebServiceEndpoint,
   PluginAspect,
 } from "@hyperledger/cactus-core-api";
-import { GetKeychainEntryEndpointV1 } from "./web-services/get-keychain-entry-endpoint-v1";
-import { SetKeychainEntryEndpointV1 } from "./web-services/set-keychain-entry-endpoint-v1";
+
+// TODO: Writing the getExpressRequestHandler() method for
+// GetKeychainEntryEndpointV1 and SetKeychainEntryEndpointV1
+// import { GetKeychainEntryEndpointV1 } from "./web-services/get-keychain-entry-endpoint-v1";
+// import { SetKeychainEntryEndpointV1 } from "./web-services/set-keychain-entry-endpoint-v1";
+
+import { PrometheusExporter } from "./prometheus-exporter/prometheus-exporter";
+
+import {
+  IGetPrometheusExporterMetricsEndpointV1Options,
+  GetPrometheusExporterMetricsEndpointV1,
+} from "./web-services/get-prometheus-exporter-metrics-endpoint-v1";
 
 export interface IPluginKeychainVaultOptions extends ICactusPluginOptions {
   logLevel?: LogLevelDesc;
@@ -38,7 +49,19 @@ export interface IPluginKeychainVaultOptions extends ICactusPluginOptions {
    * The `VAULT_TOKEN` which the backing Vault instance will accept as valid.
    */
   token: string;
+  /**
+   * Prometheus Exporter object for metrics monitoring
+   */
+
+  prometheusExporter?: PrometheusExporter;
+
+  /**
+   * The HTTP path prefix where the KV Secrets Engine is mounted.
+   */
+  kvSecretsMountPath?: string;
 }
+
+export const K_DEFAULT_KV_SECRETS_MOUNT_PATH = "secret/";
 
 export class PluginKeychainVault implements ICactusPlugin, IPluginWebService {
   public static readonly CLASS_NAME = "PluginKeychainVault";
@@ -48,7 +71,9 @@ export class PluginKeychainVault implements ICactusPlugin, IPluginWebService {
   private readonly endpoint: string;
   private readonly log: Logger;
   private readonly instanceId: string;
+  private readonly kvSecretsMountPath: string;
   private readonly backend: Vault.client;
+  public prometheusExporter: PrometheusExporter;
 
   public get className() {
     return PluginKeychainVault.CLASS_NAME;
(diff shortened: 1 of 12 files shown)


takeutak (approved):

packages/cactus-plugin-keychain-vault/src/main/typescript/prometheus-exporter/metrics.ts
```diff
@@ -0,0 +1,10 @@
+import { Gauge } from "prom-client";
+
+export const K_CACTUS_KEYCHAIN_VAULT_TOTAL_KEY_COUNT =
+  "cactus_keychain_vault_total_key_count";
+
+export const totalKeyCount = new Gauge({
```
petermetz, line 6: @jagpreetsinghsasan We might need a different name/description for this that properly conveys the fact that the vault itself could have millions of keys in it that are pre-existing and were not set by the Cactus keychain plugin (and therefore the counter doesn't know about their existence either).
Example: Big enterprise deploys Cactus and they connect it to the existing Vault deployment that they had been running for years and dozens of other (unrelated) applications of theirs store thousands of secrets in that Vault instance. In this case if someone looked at our metrics and saw that the "total keys present in vault" metrics is, let's say `5` they could become very, very concerned that either Cactus wiped out the rest of the keys or something else has gone very wrong (if our metrics to be taken at face value).

I'm thinking maybe something like this could prevent the people in the above example from having a stressful time:
`The number of keys that were set in the backing Vault deployment via this specific keychain plugin instance.`

For the metric name, not sure, it's tough to name things... maybe something like this (but feel free to suggest something else)
`K_CACTUS_KEYCHAIN_VAULT_MANAGED_KEY_COUNT`
I replaced the `TOTAL` with `MANAGED` trying to convey the information that this is the key count that's under the management of this plugin, or something like that. Just an idea, open to going with something else.

packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts
```diff
@@ -1,14 +1,40 @@
...
+  PluginKeychainVault,
+} from "../../../main/typescript/public-api";
+
+import { K_CACTUS_KEYCHAIN_VAULT_TOTAL_KEY_COUNT } from "../../../main/typescript/prometheus-exporter/metrics";
+
+import { DefaultApi as KeychainVaultApi } from "../../../main/typescript/public-api";
+
+const logLevel: LogLevelDesc = "TRACE";
 
 // test("PluginKeychainVault", (t1: Test) => {
```
petermetz, line 35: @jagpreetsinghsasan If we don't need this code then it's best to remove it. Or if you plan to add it back in the future and left it here commented so that it doesn't have to be written from scratch again (which is a valid reason IMO) then just make sure to leave a TODO item in the code explaining the latter and also open a GH issue for it.
I'm guilty of this too, so apologies for not setting the right example here (I was the one originally commenting out this test without any explanation so shame on me for that... :-) )

packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts
```diff
@@ -35,79 +61,182 @@
...
+    t.true(
+      res.data.includes(promMetricsOutput),
+      "Total Key Count 1 recorded as expected. RESULT OK",
+    );
+  }
+
+  t.end();
+});
+
+test.skip("rotateEncryptionKeys() fails fast", async (t: Test) => {
```
petermetz, line 211: @jagpreetsinghsasan Same idea for skipping tests as for the commented code: please leave a TODO or FIXME comment and create a GH issue for those as well so that in the future when we have more time someone can come around and fix these. And once again sorry for not setting the right example here, I've been setting tests to skip myself as well without explanation but it should always be explained right there in the code in a comment why the test must be skipped.

packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts
```diff
@@ -35,79 +61,182 @@
...
+  const plugin = new PluginKeychainVault(options);
+
+  const promise = plugin.rotateEncryptionKeys();
+  const expected = /not implemented/;
+  await t.rejects(promise, expected, "rotateEncryptionKeys() rejects OK");
+
+  t.end();
+});
+
+test.skip("getEncryptionAlgorithm() returns null", (t: Test) => {
```
petermetz, line 227: @jagpreetsinghsasan Same idea for skipping tests as for the commented code: please leave a TODO or FIXME comment and create a GH issue for those as well so that in the future when we have more time someone can come around and fix these. And once again sorry for not setting the right example here, I've been setting tests to skip myself as well without explanation but it should always be explained right there in the code in a comment why the test must be skipped.

packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts
```diff
@@ -35,79 +61,182 @@
...
-//     t.notok(plugin.getEncryptionAlgorithm(), "encryption algorithm falsy OK");
-
-//     t.end();
-//   });
+test("get,set,has,delete alters state as expected", async (t: Test) => {
+  const vaultTestContainer = new VaultTestServer({});
+  await vaultTestContainer.start();
+
+  const ci = await Containers.getById(vaultTestContainer.containerId);
+  const vaultIpAddr = await Containers.getContainerInternalIp(ci);
```
petermetz, line 81: @jagpreetsinghsasan Unfortunately this does not work on mac OS, the internal container IPs don't get routed by Docker on mac OS and based on discussions in the reported bugs on Docker for Mac, this is not likely to change in the future either because of some OS level limitation/security feature. This is why you need to use the internal-ip package like we discussed in some other test cases.

petermetz (changes requested):

packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
```diff
@@ -139,21 +192,74 @@ export class PluginKeychainVault implements ICactusPlugin, IPluginWebService {
...
+   * not and deciding whether the keychain has the entry ot not
+   * based on this.
+   */
   async has(key: string): Promise<boolean> {
-    const list = await this.backend.list(key);
-    return list.length > 0;
+    const path = this.pathFor(key);
+    try {
+      const res = await this.backend.read(path);
+      this.log.info(`=========================================`);
```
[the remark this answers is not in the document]
jagpreetsinghsasan: Oh, I wasn't aware of such security vulnerability. I will fix it asap.

packages/cactus-plugin-keychain-vault/src/main/typescript/prometheus-exporter/metrics.ts
jagpreetsinghsasan: This makes perfect sense. The enterprise can decide to use existing vault for cactus purpose and this name will then be troublesome. I shall fix it

packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts
jagpreetsinghsasan: https://github.com/hyperledger/cactus/issues/676

petermetz (approved):

takeutak (approved):

takeutak (approved):

PR #695, opened 2021-03-20
feat(test-tooling): pruneDockerResources() observability #694

### LINKED ISSUE #694: feat(test-tooling): containers#pruneDockerResources() observability
### Description
As a contributor I want to have a way to see what is the outcome of the docker resources pruning so that it is easier for me to debug infrastructure/hardware related CI failures that I encounter.

A good example of such issue is when the GitHub CI Action started failing because the disk was full with docker images that we created in the test cases.
It is particularly hard to debug these failures because the test runner will show a particular test failing and that failure of that test case will not be reproducible on any other machine unless that machine also has the disk full issue.

### Acceptance Criteria
1. The pruning method can log it's own results for itself via a log level parameter
2. The pruning method also returns the outcome of the pruning process so that callers can either log that for themselves or pass it further up to their callers when applicable.

petermetz:
Primary change
-----------------------

Adds request and response definitions to
the pruneDockerResources() method to
allow callers to control the log levels
and also to get a complete report of the
pruning processes' outcome from the
returned object instead of just returning
with `void`.

Miscellaneous changes
----------------------------------

Significantly improved the documentation
for the utility method in question.

Fixes #694

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/common/containers.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/common/containers.ts
@@ -6,7 +6,30 @@ import Dockerode from "dockerode";
 import tar from "tar-stream";
 import fs from "fs-extra";
 import { Streams } from "../common/streams";
-import { Checks, Strings } from "@hyperledger/cactus-common";
+import { Checks, LoggerProvider, Strings } from "@hyperledger/cactus-common";
+import { LogLevelDesc } from "loglevel";
+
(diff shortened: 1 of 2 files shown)


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

petermetz:
Will merge this without the CI passing due to this being a dependency of the hot-fix that will (hopefully) make the false negatives go away.

PR #697, opened 2021-03-22
feat(test-tooling): utility function docker prune in GH action #696

### LINKED ISSUE #696: feat(test-tooling): utility function prune-docker-all-if-github-action
### Description

As a contributor I want to have a convenience function to call when I need to clean up before and after my test cases so that I'm more productive when working on Cactus.

Depends on #694

### Acceptance Criteria
1. Abstract away the detection algorithm's details regarding how we detect if it's a GitHub Action or not that is running the code
2. Provide adequate code documentation that also explains that this is a development only utility function.

petermetz:
## Dependencies

Depends on #695 

## Commit to be reviewed

feat(test-tooling): utility function docker prune in GH action #696

Primary change
--------------

Added a convenience function that tests can call in order to easily
get rid of dangling docker resources that could be eating up the
file system's available space on disk (which is an issue that we've
started hitting more and more as we kept increasing the number
of test cases that use containers for end to end testing).
A good problem to have but nevertheless it is a problem that we
need to be vigilant about cleaning up after our test cases to avoid
overusing the resources of the CI runner.

Miscellaneous change(s)
-----------------------

Also added another utility function to determine if the current code
is being executed in a GitHub Action Workflow runner or not.
The presence and value of an environment variable is used to
achieve this feat so it definitely is not bullet proof but the trade-
off seemed fair given that these utility functions are only meant to
be used by the test code anyway.

Fixes #696 

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/common/containers.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/common/containers.ts
@@ -6,7 +6,30 @@ import Dockerode from "dockerode";
 import tar from "tar-stream";
 import fs from "fs-extra";
 import { Streams } from "../common/streams";
-import { Checks, Strings } from "@hyperledger/cactus-common";
+import { Checks, LoggerProvider, Strings } from "@hyperledger/cactus-common";
+import { LogLevelDesc } from "loglevel";
+
+export interface IPruneDockerResourcesRequest {
+  logLevel?: LogLevelDesc;
+}
+
+/**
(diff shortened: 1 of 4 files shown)


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

petermetz:
Will merge this without the CI passing due to this being a dependency of the hot-fix that will (hopefully) make the false negatives go away.

PR #706, opened 2021-03-23
fix(whitepaper): build fails on Ubuntu 18 due to glibc 2.29 #703

### LINKED ISSUE #703: fix(whitepaper): build fails on Ubuntu 18 due to glibc 2.29
**Describe the bug**

Back when the whitepaper build tooling was created it was working fine on Ubuntu 18 but nowadays it appears that the required glibc is not included in the repos for Ubuntu 18.
Rather than going for hacks it would be best to just containerize the build process that comes with the benefits of:
1. Also making it work on Windows (and Ubuntu 18 of course)
2. Will be easier to make the whitepaper build part of the CI as well

**To Reproduce**

Try to do this on an Ubuntu 18 machine and observe how it fails due to glibc being older than the required (and not being possible to be upgraded from the standard apt repos either)

```sh
cd whitepaper
make configure
make pdf
```

**Expected behavior**

make task building the pdf just works, no issues.

**Logs/Stack traces**

```sh
wkhtmltopdf --dpi 150 --disable-smart-shrinking hyperledger-cactus-whitepaper.html --allow  - hyperledger-cactus-whitepaper.pdf
wkhtmltopdf: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by wkhtmltopdf)
wkhtmltopdf: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by wkhtmltopdf)
Makefile:13: recipe for target 'pdf' failed
make: *** [pdf] Error 1
```

**Screenshots**

`N/A`

`Dev machine`

Ubuntu 18.04

`main` @ a28edcf2a02a8d8e8fcc876e3c4eb40931f0fd9a

`N/A`

**Additional context**

https://askubuntu.com/questions/1143268/how-to-install-a-libc6-version-2-29

petermetz:
Refactored the whitepaper build toolchain to be based on
entirely just containers so that it is easier to set up the build
which now only requires a working docker installation and
make (for the Makefile).

Also refactored the generated files to be placed under the
./whitepaper/build/ directory instead of being dumped
straight into the ./whitepaper/ directory so that it is
easier to differentiate between what are build files and
what are not.

There are 3 different pdf generation tasks in order to provide
contributors with a way of comparing the differnet pdfs rendring
issues (since we have a few bugs active in that sense still).
The supposedly "best" pdf render is produced by the make
target called "pdf-wk-with-flags" which is exported to the
build directory with a similar suffix in the filename as well.

The new way to do a full build is by issuing the `make` command
from the ./whitepaper/ directory of the project.

A Dockerfile was added accordingly and is tagged for the current
revision on DockerHub as:
petermetz/cactus-whitepaper-builder:2021-03-22-fix-703

Fixes #703


petermetz:
cc: @AzaharaC

AzaharaC:
I have been testing this change but not working for me. I have the following error because I don't have access to the docker image:

Unable to find image 'petermetz/cactus-whitepaper-builder:2021-03-22-fix-703' locally
docker: Error response from daemon: pull access denied for petermetz/cactus-whitepaper-builder, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
Makefile:34: recipe for target 'html' failed
make: *** [html] Error 125

whitepaper/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM pandoc/latex:2.12
```
takeutak, line 1: Why do you use Docker in the whitepaper? It seems that latex container is not needed because whitepaper is markdown style.

petermetz: It's incidental: I need pandoc for the Markdown => HTML conversion. I picked their latex image instead of the one they called 'core' because according to their docs the 'core' is lacking most features (some may not be related to latex) compared to the image called 'latex'.

takeutak: @petermetz Thanks for your replying. I understand that you need pandoc.
However, this is just a document, so I think it is enough for users to install pandoc by themselves. I don't know why you want to offer it as a Docker container, but could you tell me why you need it?

petermetz: > so I think it is enough for users to install pandoc by themselves.

It's for the convenience of the users/contributors. My guiding principle here is that anything that takes more than 3 clicks to achieve is something I want to make simpler to reduce cognitive overhead especially for newcomers.

> I don't know why you want to offer it as a Docker container

I don't mean to use the docker container as a way of distributing the document itself.
This container is for the tool-chain used to build/compile different formats of the document such as HTML and PDF. 
There are other issues open with tasks to have automated publishing of the most up to date version of the whitepaper in PDF format as well and having a container will play nicely into that.

@takeutak

petermetz: @takeutak Kindly requesting that you review or follow-up on the review discussion within this pull request (whichever is applicable). Thank you very much in advance.

takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #730, opened 2021-03-26
feat(corda-connector): dsl to support collections, enums #622

### LINKED ISSUE #622: feat(corda-connector): dsl support jvm collection types and enums

Collection types are hard coded right now and only Lists are supported (arrays in JSON).
We need to have support for lists in a way that's future proof (e.g. avoids hard coding factory invocations) or at the very least we should make sure that sufficiently robust of a solution is in place for detecting when the list factory method should be called (e.g. do we only check for the List interface implementations or go deeper with specific classes like ArrayList?)

Need a simple way to express collections in the DSL. Could potentially have an additional property on the JvmType definition indicating that something is a collection, but I'm trying not to extend the DSL at all because ultimately we want the DSL to completely go away anyway and just be able to send pure JSON that gets fed into Jackson without issues or us having to implement/maintain any sort of additional logic.

1. The `isCollection` flag on the JvmType
2. Support static/global factory functions such as the kotlin `listOf(...)` function. (more research needed regarding what that exactly gets mapped to in Java for example.
3. Implement constructable classes for each major collection type (set,map,list,array,etc). Constructable means that it extends some collection class and the only thing it does is giving a public, callable constructor with a list of elements.

**Additional context**

This issue is a follow-up after having completed the initial implementation of the corda connector.

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo

petermetz:
## Dependencies

Depends on #725

## Commit to review

Author Date: 2021-03-25 17:11:08 -0700
Committer Date: 2021-03-25 17:19:53 -0700 

feat(corda-connector): dsl to support collections, enums #622

Primary change
============

You can now express via the JSON DSL otherwise non-constructable
types such as ByteArray, DoubleArrays, enums, lists, maps, sets,
etc.

The collections are being powered by a pre-populated map of so called
"exotic" types which are named as such because they do not follow
the Java OOP ideals of having a constructor responsible for creating
the class instances.

Added test coverage for all the above as well.

Miscellaneous changes:
===================

Deleted the Kotlin test that was dependent on a corda network being
pulled up and cordapps deployed because we are covering all
of this via the existing tests written in Typescript anyway.
With the above done, now the dockerfile can be switched back
to execute the tests of the corda plugin server (which are only
unit tests at this point that is more than enough since we
have coverage elsewhere).

Tagged the container image built from this revision as:
hyperledger/cactus-connector-corda-server:2021-03-25-feat-622

Fixes #622 


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #752, opened 2021-03-30
fix(docs): supply chain example container image exact versions #751

### LINKED ISSUE #751: fix(docs): supply chain example container image exact versions
**Describe the bug**

Right now the dockerfile of the supply chain app example specifies exact versions to apk when installing dependencies and over time this has broken the build.

**To Reproduce**

Try to build the image. See that it fails.

**Expected behavior**

Image build succeeds.

**Logs/Stack traces**

`N/A`

**Screenshots**

`N/A`

`N/A/`

`N/A`

Fabric, Besu, Quorum

**Additional context**

The docker hub automated builds are also failing due to the same bug.

petermetz:
Remove the specific versions from the Dockerfile so that instead what
gets installed is "whatever's latest that day". This is not preferred, but
since Alpine Package Manager disappears versions of packages after
the fact, we do not have a choice but to abandon our quest for
reproducible builds in this particular instance.

Fixes #751

### DIFF
--- a/examples/supply-chain-app/Dockerfile
+++ b/examples/supply-chain-app/Dockerfile
@@ -3,7 +3,7 @@ FROM node:12.20.1-alpine3.12 as builder
 RUN apk update
 
 # Need JDK for the openapi-generator that is part of the build process
-RUN apk add --no-cache openjdk8=8.275.01-r0
+RUN apk add --no-cache openjdk8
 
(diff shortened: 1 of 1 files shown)


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #793, opened 2021-04-14
feat(core-api): jwt authorization #770

### LINKED ISSUE #770: feat(core-api): role based access control through oauth2 scopes

I want to be able to configure the API server with an arbitrary OAuth2 identity provider so that it's easy to plug Cactus plugins into existing infrastructure and at the same time Cactus itself does not have to be burdened (for now at least) with providing a full implementation of an identity provider itself.

1. API server can be configured with OAuth2 provider parameters such as this example [1] as seen at the official docs of the passport js oauth 2 strategy documentation: https://github.com/jaredhanson/passport-oauth2
2. OAuth2 scopes work as well (and are checked for so that effective RBAC is implemented)
3. The web service plugin interface is updated with functions to allow for finer control, for example the plugin instances should be able to specify themselves what scopes they need and also if they wish to be secure endpoints or not (e.g. can anonymous users hit it or not)

[1]
```javascript
{
    authorizationURL: 'https://www.example.com/oauth2/authorize',
    tokenURL: 'https://www.example.com/oauth2/token',
    clientID: EXAMPLE_CLIENT_ID,
    clientSecret: EXAMPLE_CLIENT_SECRET,
    callbackURL: "http://localhost:3000/auth/example/callback"
  },
  function(accessToken, refreshToken, profile, cb) {
    User.findOrCreate({ exampleId: profile.id }, function (err, user) {
      return cb(err, user);
    });
  }
```

**Additional context**

This came up during the work on #540 and is also heavily related to all the other identity related tasks that we have defined so far. The idea here is to do the absolute bare minimum that we need for a release candidate that we can honestly stand behind and say it is ready for production.

#546
#144

petermetz:
## Dependencies

Depends on #810
Depends on #824

## Commit to review

feat(core-api): jwt authorization #770

Primary change(s):
------------------

1. Verification of JSON web tokens for all endpoints by default.
2. The ability for endpoints to control on top of JWT valdity
the requirement of the JWT payload containing a certain scope

Miscellaneous change(s):
------------------------

1. Refactor of all the endpoints and plugins to make this change
technically possible.

Fixes #770


kikoncuo (approved):
This is critical for production apps

takeutak:
I think there should be no problem, but I would like to check the code a little, so could you give me some time until the approval? If there is no problem, I will approve it by next Monday.

petermetz:
> @petermetz I think there should be no problem, but I would like to check the code a little, so could you give me some time until the approval? If there is no problem, I will approve it by next Monday.

@takeutak Sure thing, thank you very much!

takeutak (approved):

takeutak:
LGTM to this PR. 
The content of this PR is not directly related, but I think that we may need to sort out the authentication functions. There is a possibility that it will be put on the agenda at a later date. If I do so at a late date, I would like to ask for your cooperation.

PR #802, opened 2021-04-16
feat(connector-besu): containerize plugin

### LINKED ISSUE #788: feat(connector-besu): containerize plugin

Cloud has became the de-facto way of running applications on computers.
Containers are rapidly becoming (have become?) the de-facto way of running applications in the cloud.
We therefore need to make sure that Cactus is not just built and distributed as the language native artifacts (e.g. npm packages/webpack bundles) but that it's also distributed as containers.
This doesn't just mean containerizing the API server that can host the plugins, but also means containerizing each plugin individually as well OR providing a generic image of the API server that can be configured to run any and all of the available Cactus plugins on-demand.

I want to be able to whisper things like this to my terminal and expect to have a fully functional connector for Ledger X up and running in a few seconds and with the config.json file mounted at the specified path/volume that gets picked up by the API server running inside the container as well.

```sh
echo "{ \"some-configuration\": 42 } " > config.json

docker run \
  --rm \
  --volume ./config.json:/opt/hyperledger/cactus/cactus-plugin-ledger-connector-besu/config.json \ 
  hyperledger/cactus-plugin-ledger-connector-besu
```

**Acceptance criteria**

1. Dockerfile for generate de docker image with the plugin connector and the api server running.

jordigiam:
Resolve #788


packages/cactus-cmd-api-server/Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM node:12.20.1-alpine3.12
```
petermetz: @jordigiam My PR adding this just got merged, should be safe to rebase onto main now and then this doesn't need to be in the diff here (and I'm guessing the other containerize PRs are going to be the same)

packages/cactus-cmd-api-server/README.md
```diff
@@ -8,10 +8,15 @@
 - [Deployment Scenarios](#deployment-scenarios)
```
petermetz: @jordigiam Same as for the Dockerfile above

petermetz (changes requested):

packages/cactus-cmd-api-server/Dockerfile
jordigiam: Done

packages/cactus-cmd-api-server/README.md
jordigiam: Done

petermetz (approved):

packages/cactus-plugin-ledger-connector-besu/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM cactus-api-server:latest
+
+ARG NPM_PKG_VERSION=latest
+
+RUN npm i @hyperledger/cactus-plugin-keychain-memory@${NPM_PKG_VERSION} --production
```
petermetz, line 5: @jordigiam Sorry, missed this one earlier: The keychain memory plugin => only good for development but this image is meant for production so I was wondering if it's possible/doable/easy enough to remove the keychain plugin from the image itself and then have it install at runtime (by adding an entry for it in the `PLUGINS` array in the config of the examples so that the examples still work, but the keychain memory plugin is not hardcoded in the image)
Apologies once again for missing this earlier!

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-besu/Dockerfile
jordigiam: No problem! Done

packages/cactus-plugin-ledger-connector-besu/README.md
```diff
@@ -99,6 +99,125 @@ enum Web3SigningCredentialType {
...
+```
+
+**Terminal Window 2 (Cactus API Server)**
+```sh
+docker run \
+  --network host \
+  --rm \
+  --publish 3000:3000 \
+  --publish 4000:4000 \
+  --env PLUGINS='[{"packageName": "@hyperledger/cactus-plugin-ledger-connector-besu", "type": "org.hyperledger.cactus.plugin_import_type.LOCAL",  "options": {"rpcApiHttpHost": "http://localhost:8545", "instanceId": "some-unique-besu-connector-instance-id"}}]' \
```
petermetz, line 170: @jordigiam Is it not needed to add here the keychain memory plugin? I was just guessing that it is based on you having included the npm install of it in the image. What I'm trying to make sure is that these example commands (all of them) still work with the updates to the image you just made, if you can confirm that then I'm good.

jordigiam: Without keychain memory is working, but only with the main version of the plugins:

Captura de pantalla 2021-04-22 a las 11 30 37

petermetz: Understood! That's good enough, we should publish v0.5.0 soon then so this problem fixes itself when the fresh package versions hit the @latest dist-tag on npm. Thanks for double checking this @jordigiam

petermetz (approved):

takeutak (approved):

PR #803, opened 2021-04-16
feat(connector-quorum): containerize plugin

### LINKED ISSUE #787: feat(connector-quorum): containerize plugin

jordigiam:
Resolve #787


petermetz (approved):

packages/cactus-plugin-ledger-connector-quorum/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM cactus-api-server:latest
+
+ARG NPM_PKG_VERSION=latest
+
+RUN npm i @hyperledger/cactus-plugin-keychain-memory@${NPM_PKG_VERSION} --production
```
petermetz, line 5: @jordigiam Sorry, missed this one earlier: The keychain memory plugin => only good for development but this image is meant for production so I was wondering if it's possible/doable/easy enough to remove the keychain plugin from the image itself and then have it install at runtime (by adding an entry for it in the `PLUGINS` array in the config of the examples so that the examples still work, but the keychain memory plugin is not hardcoded in the image)
Apologies once again for missing this earlier!

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-quorum/Dockerfile
jordigiam: Done!

packages/cactus-plugin-ledger-connector-quorum/README.md
```diff
@@ -86,6 +86,125 @@ To check that all has been installed correctly and that the pugin has no errors,
...
+
+#### Running the container
+
+Launch container with plugin configuration as an **environment variable**:
+```sh
+docker run \
+  --rm \
+  --publish 3000:3000 \
+  --publish 4000:4000 \
+  --env PLUGINS='[{"packageName": "@hyperledger/cactus-plugin-ledger-connector-quorum", "type": "org.hyperledger.cactus.plugin_import_type.LOCAL",  "options": {"rpcApiHttpHost": "http://localhost:8545", "instanceId": "some-unique-quorum-connector-instance-id"}}]' \
```
petermetz, line 111: @jordigiam Is it not needed to add here the keychain memory plugin? I was just guessing that it is based on you having included the npm install of it in the image. What I'm trying to make sure is that these example commands (all of them) still work with the updates to the image you just made, if you can confirm that then I'm good.

jordigiam: If I delete the keychain-memory from the dockerfile I can't install the keychain-memory with the docker run command, it failed with a error: "Cannot find module "uuid""

petermetz: Added dependency in the description. Approving now because once that issue is resolved the examples in the README here should start working automatically (once npm has the new code published on the default dist-tag)

petermetz (approved):

takeutak (approved):

petermetz:
@elenaizaguirre Please push them to a separate branch and send a link and then let's continue the discussion there (with the diff being visible)

PR #812, opened 2021-04-18
feat(core-api): hasTransactionFinality() on connector API

### LINKED ISSUE #354: feat(core-api): hasTransactionFinality(): boolean on connector API

Need a way to ask the connector at runtime whether its ledger has transaction finality guaranteed or not.

**Depends on #355**

A method on the connector that returns a boolean, something like:

For ledgers that only support one consensus algorithm (public permissionless ledgers mostly I guess?)

```typescript
public hasTransactionFinality(): Promise<boolean> {
    return false;
}
```

In cases where the ledger has configurable consensus the function's return value would depend on which one is configured for the current ledger instance.

```typescript
public async hasTransactionFinality(): Promise<boolean> {
    const algorithmsWithTxFinality = ["IBFT2", "RAFT", "Other_Algo"];
    const algorithmsWithOutTxFinality = ["Proof_of_Work", "Some_Other_Algo"];
    const currentAlgo = await this.getConsensusAlgorithm();
    const unrecognizedAlgorithm = !algorithmsWithTxFinality.includes(currentAlgo) &&  !algorithmsWithOutTxFinality.includes(currentAlgo);
    if (unrecognizedAlgorithm) { // fail fast when algorithm is unrecognized
        throw new Error(`Unrecognized consensus algorithm: ${currentAlgo}`);
    } else {
        return algorithmsWithTxFinality.includes(currentAlgo);
    }
}
```

To avoid the problem of a ledger introducing a new consensus algorithm that does support TX finality and then us returning incorrect results based on that, we must check the `currentAlgo` for presence in both lists and throw if it isn't present in either one of them. This guarantees that we fail fast instead of non-deterministic behavior.

Another way of going about this would be to have the consensus algorithms defined in the consortium definition which would provide us with the means to have the consensus algorithm locked down. Not sure if we should do this because the problem is regarding enforcement: Big mistakes can still happen if a consortium member accidentally changes the consensus algorithm of their own ledger without seeking an update in the consensus definition first.

jscode017:
Adding function hasTransactionFinality() on connector API, to let ledger query whether the configured consensus has transaction finality at run time.
Fixes #354

### DIFF
--- a/packages/cactus-cmd-api-server/src/test/typescript/fixtures/plugin-ledger-connector-stub/plugin-ledger-connector-stub.ts
+++ b/packages/cactus-cmd-api-server/src/test/typescript/fixtures/plugin-ledger-connector-stub/plugin-ledger-connector-stub.ts
@@ -14,7 +14,10 @@ import {
   ICactusPluginOptions,
 } from "@hyperledger/cactus-core-api";
 
-import { PluginRegistry } from "@hyperledger/cactus-core";
+import {
+  PluginRegistry,
+  consensusHasTransactionFinality,
+} from "@hyperledger/cactus-core";
 
 import {
   Checks,
@@ -128,7 +131,11 @@ export class PluginLedgerConnectorStub
   > {
     return ConsensusAlgorithmFamily.AUTHORITY;
   }
+  public async hasTransactionFinality(): Promise<boolean> {
+    const currentConsensusAlgorithmFamily = await this.getConsensusAlgorithmFamily();
 
+    return consensusHasTransactionFinality(currentConsensusAlgorithmFamily);
+  }
   public async transact(req: unknown): Promise<unknown> {
     const fnTag = `${this.className}#transact()`;
     Checks.truthy(req, `${fnTag} req`);
--- a/packages/cactus-core/src/main/typescript/consensus-has-transaction-finality.ts
+++ b/packages/cactus-core/src/main/typescript/consensus-has-transaction-finality.ts
@@ -1,0 +1,28 @@
+import {
+  ConsensusAlgorithmFamily,
+  ConsensusAlgorithmFamiliesWithTxFinality,
+  ConsensusAlgorithmFamiliesWithOutTxFinality,
+} from "@hyperledger/cactus-core-api";
+
+export function consensusHasTransactionFinality(
+  consensusAlgorithmFamily: ConsensusAlgorithmFamily,
+): boolean {
+  const isInConsensusAlgorithmFamiliesWithTxFinality = (Object.values(
+    ConsensusAlgorithmFamiliesWithTxFinality,
+  ) as string[]).includes(consensusAlgorithmFamily.toString());
+
+  const isInConsensusAlgorithmFamiliesWithOutTxFinality = (Object.values(
+    ConsensusAlgorithmFamiliesWithOutTxFinality,
+  ) as string[]).includes(consensusAlgorithmFamily.toString());
+
+  const unrecognizedConsensusAlgorithmFamily =
+    !isInConsensusAlgorithmFamiliesWithTxFinality &&
+    !isInConsensusAlgorithmFamiliesWithOutTxFinality;
+
+  if (unrecognizedConsensusAlgorithmFamily) {
+    throw new Error(
+      `Unrecognized consensus algorithm family: ${consensusAlgorithmFamily}`,
+    );
+  }
+  return isInConsensusAlgorithmFamiliesWithTxFinality;
+}
--- a/packages/cactus-core/src/main/typescript/public-api.ts
+++ b/packages/cactus-core/src/main/typescript/public-api.ts
@@ -9,3 +9,5 @@ export {
   AuthorizationOptionsProvider,
   IEndpointAuthzOptionsProviderOptions,
 } from "./web-services/authorization-options-provider";
+
+export { consensusHasTransactionFinality } from "./consensus-has-transaction-finality";
--- a/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
+++ b/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
@@ -20,7 +20,10 @@ import {
   IPluginKeychain,
 } from "@hyperledger/cactus-core-api";
 
-import { PluginRegistry } from "@hyperledger/cactus-core";
+import {
+  PluginRegistry,
+  consensusHasTransactionFinality,
+} from "@hyperledger/cactus-core";
 
 import {
   Checks,
@@ -220,7 +223,11 @@ export class PluginLedgerConnectorBesu
   > {
     return ConsensusAlgorithmFamily.AUTHORITY;
   }
+  public async hasTransactionFinality(): Promise<boolean> {
+    const currentConsensusAlgorithmFamily = await this.getConsensusAlgorithmFamily();
 
+    return consensusHasTransactionFinality(currentConsensusAlgorithmFamily);
+  }
   public async invokeContract(
     req: InvokeContractV1Request,
   ): Promise<InvokeContractV1Response> {
--- a/packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
+++ b/packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
@@ -13,7 +13,7 @@ import {
   ICactusPluginOptions,
   ConsensusAlgorithmFamily,
 } from "@hyperledger/cactus-core-api";
-
+import { consensusHasTransactionFinality } from "@hyperledger/cactus-core";
 import {
   Checks,
   Logger,
@@ -64,6 +64,11 @@ export class PluginLedgerConnectorCorda
     ConsensusAlgorithmFamily
   > {
     return ConsensusAlgorithmFamily.AUTHORITY;
+  }
+  public async hasTransactionFinality(): Promise<boolean> {
+    const currentConsensusAlgorithmFamily = await this.getConsensusAlgorithmFamily();
+
+    return consensusHasTransactionFinality(currentConsensusAlgorithmFamily);
   }
 
   public getAspect(): PluginAspect {
--- a/packages/cactus-core-api/src/main/json/openapi.json
+++ b/packages/cactus-core-api/src/main/json/openapi.json
@@ -68,6 +68,24 @@
                     "org.hyperledger.cactus.consensusalgorithm.PROOF_OF_WORK"
                 ]
             },
+            "ConsensusAlgorithmFamiliesWithTxFinality":{
+                "type": "string",
+                "description": "Enumerates a list of consensus algorithm families that provide immediate finality",
+                "enum": [
+                    "org.hyperledger.cactus.consensusalgorithm.PROOF_OF_AUTHORITY",
+                    "org.hyperledger.cactus.consensusalgorithm.PROOF_OF_STAKE"
+                ]
+
+            },
+            "ConsensusAlgorithmFamiliesWithOutTxFinality":{
+                "description":"Enumerates a list of consensus algorithm families that do not provide immediate finality",
+                "type": "string",
+                "enum": [
+                    "org.hyperledger.cactus.consensusalgorithm.PROOF_OF_WORK"
+                ],
+                "x-enum-varnames": ["WORK"]
+
+            },
             "PrimaryKey": {
                 "type": "string",
                 "minLength": 1,
--- a/packages/cactus-core-api/src/main/typescript/generated/openapi/typescript-axios/api.ts
+++ b/packages/cactus-core-api/src/main/typescript/generated/openapi/typescript-axios/api.ts
@@ -125,6 +125,25 @@ export interface CactusNodeMeta {
     publicKeyPem: string;
 }
 /**
+ * Enumerates a list of consensus algorithm families that do not provide immediate finality
+ * @export
+ * @enum {string}
+ */
+export enum ConsensusAlgorithmFamiliesWithOutTxFinality {
+    WORK = 'org.hyperledger.cactus.consensusalgorithm.PROOF_OF_WORK'
+}
+
+/**
+ * Enumerates a list of consensus algorithm families that provide immediate finality
+ * @export
+ * @enum {string}
+ */
+export enum ConsensusAlgorithmFamiliesWithTxFinality {
+    AUTHORITY = 'org.hyperledger.cactus.consensusalgorithm.PROOF_OF_AUTHORITY',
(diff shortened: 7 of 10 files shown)


petermetz:
@jscode017 That test failure is shown by the CI logs is not related to the changes you made, if you rebase your branch onto #810 (and mark it as a dependency of this PR) then the tests should start passing. Alternatively just wait a little until #810 gets merged and then rebase onto `main` to achieve the same outcome.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -230,7 +232,26 @@ export class PluginLedgerConnectorBesu
   > {
     return ConsensusAlgorithmFamily.AUTHORITY;
   }
-
+  public async hasTransactionFinality(): Promise<boolean> {
+    const currentConsensusAlgorithmFamily = await this.getConsensusAlgorithmFamily();
+
+    const unrecognizedConsensusAlgorithmFamily =
+      !(Object.values(
```
petermetz, line 239: @jscode017 I would cache the result of 
```typescript
(Object.values(
      ConsensusAlgorithmFamiliesWithTxFinality,
    ) as string[]).includes(currentConsensusAlgorithmFamily.toString())
```

So that at the end it can be returned as-is without further computation needed.

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
```diff
@@ -65,6 +67,26 @@ export class PluginLedgerConnectorCorda
   > {
     return ConsensusAlgorithmFamily.AUTHORITY;
   }
+  public async hasTransactionFinality(): Promise<boolean> {
+    const currentConsensusAlgorithmFamily = await this.getConsensusAlgorithmFamily();
+
+    const unrecognizedConsensusAlgorithmFamily =
```
petermetz, line 73: @jscode017 The part below could be pulled out into a utility function that could be placed in the `cactus-core` package (not `cactus-core-api`).

```typescript
  const unrecognizedConsensusAlgorithmFamily =
      !(Object.values(
        ConsensusAlgorithmFamiliesWithTxFinality,
      ) as string[]).includes(currentConsensusAlgorithmFamily.toString()) &&
      !(Object.values(
        ConsensusAlgorithmFamiliesWithOutTxFinality,
      ) as string[]).includes(currentConsensusAlgorithmFamily.toString());

    if (unrecognizedConsensusAlgorithmFamily) {
      throw new Error(
        `Unrecognized consensus algorithm family: ${currentConsensusAlgorithmFamily}`,
      );
    }
    return (Object.values(
      ConsensusAlgorithmFamiliesWithTxFinality,
    ) as string[]).includes(currentConsensusAlgorithmFamily.toString());
```

Then the `hasTransactionFinality()` methods could all just reuse that method instead of duplicating the same code in every connector.

petermetz (changes requested):

takeutak (approved):

takeutak:
LGTM

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
jscode017: Yeah, I agree that pulled it out would be a much better way. 
But which class should I put the function to? Both ConsortiumRepository and PluginRegistry in cactus-core do not seems to be a good place?

jscode017: BTW, my wrap up function may look like this:
```
public consensusHasTransactionFinality(consensusAlgorithmFamily:ConsensusAlgorithmFamily) boolean{
    const isInConsensusAlgorithmFamiliesWithTxFinality=(Object.values(     ConsensusAlgorithmFamiliesWithTxFinality ) as string[]).includes(consensusAlgorithmFamily.toString());
    
    const isInConsensusAlgorithmFamiliesWithOutTxFinality=(Object.values(  ConsensusAlgorithmFamiliesWithOutTxFinality) as string[]).includes(consensusAlgorithmFamily.toString());
    
    const unrecognizedConsensusAlgorithmFamily =! isInConsensusAlgorithmFamiliesWithTxFinality && !isInConsensusAlgorithmFamiliesWithOutTxFinality;

    if (unrecognizedConsensusAlgorithmFamily) {
      throw new Error(
        `Unrecognized consensus algorithm family: ${consensusAlgorithmFamily}`,
      );
    }
    return isConsensusAlgorithmFamiliesWithTxFinality;
    }
```

petermetz:
@jscode017 Rebasing onto `main` should fix that CI test failure (it is not caused by the changes you made)

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
jscode017: Sorry I get a stupid question here.
I add 
`export { consensusHasTransactionFinality } from "./consensus-has-transaction-finality";`
or
`export * from "./consensus-has-transaction-finality";`
But keep getting
`src/main/typescript/public-api.ts(7,49): error TS2307: Cannot find module './consensus-has-transaction-finality' or its corresponding type declarations.`

petermetz: Hmm, that's strange indeed. Can you push to the remote so I can check it out as well?

jscode017: Sure , I've pushed it to the remote.

petermetz: Ran the build on it, to me it was saying that you are missing `ConsensusAlgorithmFamily` from your import statement (which then makes that file not compile and that leads to the compilation of public-api.ts bugging out.

This is how I got it to build:

```typescript
import {
  ConsensusAlgorithmFamiliesWithTxFinality,
  ConsensusAlgorithmFamiliesWithOutTxFinality,
  ConsensusAlgorithmFamily,
} from "@hyperledger/cactus-core-api";

export function consensusHasTransactionFinality(
  consensusAlgorithmFamily: ConsensusAlgorithmFamily,
): boolean {
  const isInConsensusAlgorithmFamiliesWithTxFinality = (Object.values(
    ConsensusAlgorithmFamiliesWithTxFinality,
  ) as string[]).includes(consensusAlgorithmFamily.toString());

  const isInConsensusAlgorithmFamiliesWithOutTxFinality = (Object.values(
    ConsensusAlgorithmFamiliesWithOutTxFinality,
  ) as string[]).includes(consensusAlgorithmFamily.toString());

  const unrecognizedConsensusAlgorithmFamily =
    !isInConsensusAlgorithmFamiliesWithTxFinality &&
    !isInConsensusAlgorithmFamiliesWithOutTxFinality;

  if (unrecognizedConsensusAlgorithmFamily) {
    throw new Error(
      `Unrecognized consensus algorithm family: ${consensusAlgorithmFamily}`,
    );
  }
  return isInConsensusAlgorithmFamiliesWithTxFinality;
}
//export defualt consensusHasTransactionFinality;

```

jscode017: Almost finish.
But there is a small issue remains.
Would want to know how to add dependencies for this project.
I have uploaded a new commit, but somehow cound not refactor "cactus-plugin-ledger-connector-corda".
I would want like to add "@hyperledger/cactus-core-api": "0.4.1" in cactus-plugin-ledger-connector-corda/package.json. 
But instead break dependencies and got bunch of things like below:
```
src/main/typescript/plugin-factory-ledger-connector.ts(4,8): error TS2307: Cannot find module '@hyperledger/cactus-core-api' or its corresponding type declarations.
src/main/typescript/plugin-ledger-connector-corda.ts(15,8): error TS2307: Cannot find module '@hyperledger/cactus-core-api' or its corresponding type declarations.
src/main/typescript/plugin-ledger-connector-corda.ts(16,10): error TS2305: Module '"../../../node_modules/@hyperledger/cactus-core/dist/types/main/typescript"' has no exported member 'consensusHasTransactionFinality'.
src/main/typescript/plugin-ledger-connector-corda.ts(22,8): error TS2307: Cannot find module '@hyperledger/cactus-common' or its corresponding type declarations.
src/main/typescript/plugin-ledger-connector-corda.ts(55,27): error TS2339: Property 'instanceId' does not exist on type 'IPluginLedgerConnectorCordaOptions'.
src/main/typescript/plugin-ledger-connector-corda.ts(60,36): error TS2339: Property 'instanceId' does not exist on type 'IPluginLedgerConnectorCordaOptions'.
src/main/typescript/public-api.ts(15,39): error TS2307: Cannot find module '@hyperledger/cactus-core-api' or its corresponding type declarations.
src/main/typescript/public-api.ts(21,43): error TS2554: Expected 0 arguments, but got 1.
src/main/typescript/web-services/deploy-contract-jars-endpoint.ts(11,8): error TS2307: Cannot find module '@hyperledger/cactus-core-api' or its corresponding type declarations.
src/main/typescript/web-services/deploy-contract-jars-endpoint.ts(18,8): error TS2307: Cannot find module '@hyperledger/cactus-common' or its corresponding type declarations.
src/test/typescript/integration/deploy-cordapp-jars-to-nodes.test.ts(4,33): error TS2307: Cannot find module '@hyperledger/cactus-test-tooling' or its corresponding type declarations.
src/test/typescript/integration/deploy-cordapp-jars-to-nodes.test.ts(5,30): error TS2307: Cannot find module '@hyperledger/cactus-common' or its corresponding type declarations.
src/test/typescript/integration/deploy-cordapp-jars-to-nodes.test.ts(9,8): error TS2307: Cannot find module '@hyperledger/cactus-test-tooling' or its corresponding type declarations.
src/test/typescript/integration/jvm-kotlin-spring-server.test.ts(7,8): error TS2307: Cannot find module '@hyperledger/cactus-test-tooling' or its corresponding type declarations.
src/test/typescript/integration/jvm-kotlin-spring-server.test.ts(8,30): error TS2307: Cannot find module '@hyperledger/cactus-common' or its corresponding type declarations.
src/test/typescript/integration/jvm-kotlin-spring-server.test.ts(12,8): error TS2307: Cannot find module '@hyperledger/cactus-test-tooling' or its corresponding type declarations.
```

packages/cactus-core/src/main/typescript/consensus-has-transaction-finality.ts
```diff
@@ -0,0 +1,29 @@
...
+    !isInConsensusAlgorithmFamiliesWithOutTxFinality;
+
+  if (unrecognizedConsensusAlgorithmFamily) {
+    throw new Error(
+      `Unrecognized consensus algorithm family: ${consensusAlgorithmFamily}`,
+    );
+  }
+  return isInConsensusAlgorithmFamiliesWithTxFinality;
+}
+//export defualt consensusHasTransactionFinality;
```
petermetz, line 29: @jscode017 Recommend to remove commented lines unless you really need them for some specific reason (assuming here this was just forgotten and no longer needed at all)

packages/cactus-core/src/main/typescript/public-api.ts
```diff
@@ -4,3 +4,4 @@ export {
   ConsortiumRepository,
   IConsortiumRepositoryOptions,
 } from "./consortium-repository";
+export * from "./consensus-has-transaction-finality";
```
petermetz, line 7: @jscode017 Recommendation here is to explicitly export not with the `*` character (easier to read what exactly goes in the public API that way because then there's no need to open the file up itself and skim through it's exports).
There's exceptions to this practice when there's a lot of exports for example in a generated file, but here it's just the one so I'd say it's easier on the reader if you export it explicitly

```typescript
export { theFunction } from "./consensus-has-transaction-finality";
```

petermetz (changes requested):

packages/cactus-core/src/main/typescript/consensus-has-transaction-finality.ts
jscode017: My bad, forget to do a little clean up

packages/cactus-cmd-api-server/src/test/typescript/fixtures/plugin-ledger-connector-stub/plugin-ledger-connector-stub.ts
```diff
@@ -128,7 +131,11 @@ export class PluginLedgerConnectorStub
   > {
     return ConsensusAlgorithmFamily.AUTHORITY;
   }
+  public async hasTransactionFinality(): Promise<boolean> {
```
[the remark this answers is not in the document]
petermetz: It's perfectly fine as it is IMO. The stub is just there to help with testing so the expectation is that if someone wants to use it to simulate different consensus algorithm families then they can either sub-class it or override the default behavior some other way.

The only thing I need now is the commits to be squashed into a single one so that there are no in-between states (every commit on the main branch should be valid, passing, tests, etc.
So if you squash and then force push (make sure to have the commit signed off) then I'll smash that approve button right after that. Thank you for the contribution and working with me on hammering out the details!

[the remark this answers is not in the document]
jscode017: Thanks for the help for my first pull request!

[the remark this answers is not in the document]
petermetz: @jscode017 About to approve and merge this, but I wanted to give one more piece of advice regarding the mechanics of how PR reviews are done:
If someone reviews the PR (e.g. when I literally click button submit review) from that point on it won't notify me until you hit the 're-request review' button (that little refresh icon next to my name in the top right corner among the reviewers).

This is not an actual issue, but I wanted to explain this anyway because I just saw by chance that you had fixed everything that I was asking for more than a week ago but I only noticed it now (otherwise I would've approved+merged the same day)

petermetz (approved):

packages/cactus-cmd-api-server/src/test/typescript/fixtures/plugin-ledger-connector-stub/plugin-ledger-connector-stub.ts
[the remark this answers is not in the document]
jscode017: Sorry for that, I would hit the 're-request review' next time.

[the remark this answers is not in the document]
petermetz: No worries at all, thank you again for the PR!

PR #842, opened 2021-04-23
docs(examples): fix supply chain app process.env authorization protocol

petermetz:
The AUTHORIZATION_PROTOCOL=NONE
configuration parameter had to be specified because without it the config-
service bails out.


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #843, opened 2021-04-23
fix(connector-besu): network update only if present in keychain

petermetz:
Only attempt to set/update the networks object on the contract JSON if
it was present in the keychain, leave it alone
otherwise.


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

Issue #295, opened 2020-09-25
ci(ledger-connector): pilot light performance benchmarks


petermetz:
### Description
As a maintainer I want to get an automated warning if a PR passes the tests, but has an adverse impact on performance as measured by a set of baseline benchmarks so that if a PR looks good to me, but in fact reduces performance to a 10th of what it used to be because of some mistake hidden in the code, then I am made aware of this as soon as possible (which hopefully makes it much easier to keep the code performant)

### Acceptance Criteria
1. Performance checks are part of the CI tests and the reports are easily accessible/readable (from the CI logs?)
2. The performance checks can be executed on the average developer machine and does not require/depend on external infrastructure to be set up (ideally docker should be the only dependency)

petermetz:
FYI: https://www.npmjs.com/package/loadtest

travis-payne:
I'll have a look at this

kikoncuo:
Talk with Hyperledger Caliper 
Mark Wagner
Performance and Scale working group

travis-payne:
https://github.com/hyperledger/cactus/pull/845

PR #845, opened 2021-04-23
feat: incorporating load testing into our CI pipeline

travis-payne:
This changes utilizes Artillery to load test our API endpoint. This is a first cut implementation.

BREAKING CHANGE: 🧨 Behaviors in a cloud environment is currently untested and could impact
CI pipeline.

Artillery works by sending HTTP requests at different intervals as defined in [benchmark-script.yaml](https://github.com/hyperledger/cactus/pull/845/files#diff-db633280d17272defd50724836cb82a91e5a6e23f040d8e261c5eb75a5d53382).

I've defined 3 phases, a warm up phase, a ramp up phase, and finally a phase imitating a sustained high load.

We can further define different scenarios, target endpoints, and user actions.

After running the test, Artillery produces a report in json format, which is read back in to pull out the metrics we need. We can then define a threshold, and write an assert to ensure we're meeting that requirement.

✅ Closes: #295

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,8 +43,9 @@
     "webpack:prod": "lerna run webpack:prod",
     "webpack:prod:web": "lerna run webpack:prod:web",
     "webpack:prod:node": "lerna run webpack:prod:node",
-    "test:all": "tap --node-arg=--max-old-space-size=4096 --timeout=3600 \"packages/cactus-*/src/test/typescript/{unit,integration}/\"",
+    "test:all": "tap --node-arg=--max-old-space-size=4096 --timeout=3600 \"packages/cactus-*/src/test/typescript/{unit,integration,benchmark}/\"",
     "test:unit": "tap --node-arg=--max-old-space-size=4096 --timeout=600 \"packages/cactus-*/src/test/typescript/unit/\"",
+    "test:benchmark": "tap --jobs=1 --no-timeout  \"packages/cactus-*/src/test/typescript/benchmark/\"",
     "test:browser": "karma start karma.conf.js",
     "test:integration": "tap --node-arg=--max-old-space-size=4096 --jobs=1 --timeout=3600 \"packages/cactus-*/src/test/typescript/integration/\"",
     "changelog": "conventional-changelog --infile CHANGELOG.md --outfile CHANGELOG.md && git add CHANGELOG.md",
@@ -122,6 +123,7 @@
     "*/*/src/**/*.{js,ts}": "eslint --fix"
   },
   "dependencies": {
-    "@types/ssh2": "0.5.46"
+    "@types/ssh2": "0.5.46",
+    "artillery": "1.7.2"
   }
 }
--- a/packages/cactus-cmd-api-server/src/test/typescript/benchmark/artillery-api-benchmark.test.ts
+++ b/packages/cactus-cmd-api-server/src/test/typescript/benchmark/artillery-api-benchmark.test.ts
@@ -1,0 +1,151 @@
+import test, { Test } from "tape-promise/tape";
+import { v4 as uuidv4 } from "uuid";
+import { JWK } from "jose";
+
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -122,6 +123,7 @@
     "*/*/src/**/*.{js,ts}": "eslint --fix"
   },
   "dependencies": {
-    "@types/ssh2": "0.5.46"
+    "@types/ssh2": "0.5.46",
+    "artillery": "^1.7.2"
```
petermetz, line 127: @travis-payne Please lock it down to an exact version (e.g. remove the caret).
https://github.com/hyperledger/cactus/blob/main/CONTRIBUTING.md#on-reproducible-builds

packages/cactus-cmd-api-server/src/test/typescript/benchmark/artillery-api-benchmark.test.ts
```diff
@@ -0,0 +1,143 @@
...
+} from "../../../main/typescript/public-api";
+
+import { DefaultApi as ApiServerApi } from "../../../main/typescript/public-api";
+
+const log = LoggerProvider.getOrCreate({
+  level: "INFO",
+  label: "logger-test",
+});
+
+const util = require("util");
```
petermetz, line 25: @travis-payne the `import { ... } from "..."` syntax is preferred unless there's some specific reason to not use it. Sorry for the nitpicking!

petermetz (changes requested):

packages/cactus-cmd-api-server/src/test/typescript/benchmark/artillery-api-benchmark.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  t.equal(res.status, 200);
+
+  // Sets apihost in process env variables to artillery script can read it.
+  process.env.apiHost = apiHost;
+  await fireArtilleryCommand(t);
+
+  // Reading in Artillery JSON output
+  let report = null;
+  try {
+    const jsonString = fs.readFileSync("./report.json");
```
petermetz, line 110: @travis-payne Please also dump the whole JSON report to the logs just in case we want to take a deeper look into the results/debug something/etc.

packages/cactus-cmd-api-server/src/test/typescript/benchmark/artillery-api-benchmark.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    report = JSON.parse(jsonString);
+  } catch (err) {
+    log.error(err);
```
petermetz, line 113: @travis-payne More nits, sorry: Logging the exception is cool, but it should also say something about the context of the error in case the stack trace gets mangled/disappeared for whatever reason and the only info we have is the exception message.
For example here if the JSON syntax was invalid for whatever reason then a generic exception message will not help the person debugging it later to determine what was being parsed (and if they have no stack trace with line numbers then they'll definitely have no clue where it was even thrown).

Based on the above I recommend something like this:

```typescript
log.error(`Could not parse the artillery report JSON from the file system: ${jsonString}`,err);
```

petermetz (changes requested):

petermetz:
@travis-payne As a next step, you could look into publishing the report JSON file generated by artillery to some purpose-made git branch that we can hook up to GitHub pages for automatic publishing: https://github.com/marketplace/actions/add-commit?version=v7.2.0

(It's okay to address this in a separate PR/task)

petermetz (commented):
@travis-payne Is this the latest code? I couldn't find the logging of the exception + logging of the JSON.

travis-payne:
> @travis-payne Is this the latest code? I couldn't find the logging of the exception + logging of the JSON.

I might've overwritten some code, have a look now

petermetz (approved):

takeutak (approved):

PR #827, opened 2021-04-20
docs(readthedocs): readthedocs documentation

AzaharaC:
Resolve #221

PR #846, opened 2021-04-23
docs(readthedocsio): add packages readme files

### LINKED ISSUE #340: docs(readthedocsio): include package readmes on website
### Description
As a potential user of Cactus I want to have a single, unified experience for reading documentation with the individual packages' README.md files included so that I don't spend most of my time that I have dedicated to Cactus at looking up documentation but rather I spend it reading the documentation which brings me closer to my end goals faster and therefore increases the chance of me actually liking Cactus and joining the community around it.

Context:
We have a ReadTheDocs.io page bootstrapped, but it's not ready yet.

### Acceptance Criteria
1. When I go on the Cactus ReadTheDocs page and enter a search term for something that is let's say in the README.md file of the `@hyperledger/cactus-api-client` package, then I actually get a hit instead of a "No results" message.
2. The solution is automated in the sense that any chance made to the README.md files of the packages will automatically get pushed to ReadTheDocs as well upon re-build (which should be triggered when a PR is merged). In short: copy pasting around the README.md files is not acceptable as a solution.

AzaharaC:
Resolve #340 
Depends on #827

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -44,4 +44,5 @@ A pluggable, enterprise-grade framework to transact on multiple distributed ledg
    maintainers
    Whitepaper <whitepaper-init.rst>
    regulatory-and-industry-initiatives-reading-list
+   Components <packages.rst>
 
--- a/docs/source/packages.rst
+++ b/docs/source/packages.rst
@@ -1,0 +1,31 @@
+Cactus Components
+-----------------
+
+This section contains the components to form Hyperledger Cactus.
+
+.. toctree::
+   :maxdepth: 1
+
+   Busines Logic <packages/business-logic-plugin.md>
+   Api Client <packages/cactus-api-client.md>
+   CMD Api Server <packages/cactus-cmd-api-server.md>
+   Cactus Cockpit <packages/cactus-cockpit.md>
+   Cactus Common <packages/cactus-common.md>
+   Catus Core <packages/cactus-core.md>
+   Core Api<packages/cactus-core-api.md>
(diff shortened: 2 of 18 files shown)


petermetz (approved):

takeutak:
As I questioned on PR827, What will it look like when you actually launch this site? (Will it be like the site below?)
https://hyperledger-cactus.readthedocs.io/en/latest/

My concern is that this document will simply reflect the contents of markdown on the site?
If a function to pick up and document comments from sources is being considered, it may be inconvenient to develop the V0.X version.

petermetz:
@takeutak 

> What will it look like when you actually launch this site? (Will it be like the site below?)
> https://hyperledger-cactus.readthedocs.io/en/latest/

It will be far better aesthetically and functionally than that. What you see on https://hyperledger-cactus.readthedocs.io/en/latest/ at present was my my feeble attempt at getting us started back in the day. This will be vastly different.

There are build instructions documented as part of the other readthedocs PR of Azahara I believe so you could check it out locally to see how it looks.
 
> My concern is that this document will simply reflect the contents of markdown on the site?

For the most part yes. To me that gives me cozy feelings rather than concern because re-using the markdown files is great IMO instead of having to duplicate them.

> If a function to pick up and document comments from sources is being considered, it may be inconvenient to develop the V0.X version.

It is not being considered in the scope of this PR. 
Would be nice to have https://typedoc.org/ generated docs as well at some point and that could be an additional menu item on the ReadTheDocs page labeled something like "API Reference", but that would have to be done as an improvement later.

Issue #848, opened 2021-04-23
fix(ci): dependent issues bot workflow has no job id


petermetz:
Hi Peter,

Thanks for contacting GitHub Support.

There's a mismatch in what you have set as the name of the required status check for the main branch (Dependabot Issues) and what the check is actually called when the workflow runs (check)

https://github.com/hyperledger/cactus/actions/runs/772009240/workflow

```yaml
jobs:
  check:
    runs-on: ubuntu-latest
---
```
You can rename the job in that workflow to Dependabot Issues to match the required status check name:
```yaml
jobs:
  check:
    runs-on: ubuntu-latest
    name: Dependabot Issues
----
```
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idname

The check will then be displayed as Dependabot Issues - if this was the intended name - and will satisfy the protected branches rule.

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo @jagpreetsinghsasan @arnab-roy

PR #849, opened 2021-04-23
fix(ci): dependent issues bot workflow has no job id #848

petermetz:
There's a mismatch in what you have set as the name of the required status check for the main branch (Dependabot Issues) and what the check is actually called when the workflow runs (check)

https://github.com/hyperledger/cactus/actions/runs/772009240/workflow

jobs:
  check:
    runs-on: ubuntu-latest
---
You can rename the job in that workflow to Dependabot Issues to match the required status check name:

jobs:
  check:
    runs-on: ubuntu-latest
    name: Dependabot Issues
----
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idname

The check will then be displayed as Dependabot Issues - if this was the intended name - and will satisfy the protected branches rule.


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #854, opened 2021-04-23
feat(test-tooling): add OpenEthereumTestLedger #851

### LINKED ISSUE #851: feat(test-tooling): add OpenEthereumTestLedger

Need an xdai test ledger class so we can test the new xdai connector that's dropping soon.

Same as with the rest of the connector's test ledgers, a class that manages a container under the hood so that we can write end to end tests for the connector..

None, I'm pretty sure we need this.

**Additional context**

This is needed for the carbon accounting app #540 because I realized after the fact that they are not on the main net but xdai instead (makes a lot of sense, tx fees are crazy on the main net).

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo @jagpreetsinghsasan @arnab-roy

petermetz:
We did not need to roll our own contanier image for this
so the only thing new here is the class in the test tooling
package that makes it convenient to write test cases against
OpenEthereum. The container image used by the class behind
the scenes is just the stock OpenEthereum image from DockerHub.

Fixes #851

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/openethereum/openethereum-test-ledger.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/openethereum/openethereum-test-ledger.ts
@@ -1,0 +1,245 @@
+import { EventEmitter } from "events";
+import Docker, { Container } from "dockerode";
+import { v4 as internalIpV4 } from "internal-ip";
+import Web3 from "web3";
+import { Account } from "web3-core";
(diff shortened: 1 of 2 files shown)


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #884, opened 2021-04-29
refactor(consortium-manual): move logic from endpoint to plugin class

### LINKED ISSUE #429: refactor(plugin-consortium-manual): move plugin logic from endpoint classes to plugin class
### Description

Just move the code out of the endpoint classes into the plugin class to have the separation of concerns best practice/programming pattern enforced: The endpoint class is the joint between the raw HTTP requests and the plugin classes and is only responsible for validating that the requests are compliant with what is being described in the OpenAPI specs.
So because of the above they should not contain the actual logic of the plugin because that belongs in the plugin.

### Acceptance Criteria
1. `packages/cactus-plugin-consortium-manual/src/main/typescript/consortium/get-node-jws-endpoint-v1.ts` has the logic moved to the plugin class at `./packages/cactus-plugin-consortium-manual/src/main/typescript/plugin-consortium-manual.ts`. The constructor options argument of the endpoint takes the plugin as an input (if not already) so that it can invoke the relevant method of the plugin in the place of the code that is being moved there
    ```typescript
    public async createJws(): Promise<JWSGeneral> {
      const fnTag = "GetNodeJwsEndpoint#createJws()";
      const { keyPairPem, consortium } = this.options;
      try {
        const keyPair = JWK.asKey(keyPairPem);
        const payload = jsonStableStringify({ consortium });
        const _protected = {
          iat: Date.now(),
          jti: uuid.v4(),
          iss: "Hyperledger Cactus",
        };
        // TODO: double check if this casting is safe (it is supposed to be)
        return JWS.sign.general(payload, keyPair, _protected) as JWSGeneral;
      } catch (ex) {
        throw new Error(`${fnTag} ${ex.stack}`);
      }
  }
    ```
 
2. Same as 1) but for `packages/cactus-plugin-consortium-manual/src/main/typescript/consortium/get-consortium-jws-endpoint-v1.ts`
    ```typescript
    const nodes2d = this.options.consortium.members?.map((m) => m.nodes);
    const nodes = flatten(nodes2d);
    const requests: Promise<AxiosResponse<GetNodeJwsResponse>>[] = nodes
      .map((cnm) => cnm.nodeApiHost)
      .map((host) => new Configuration({ basePath: host }))
      .map((configuration) => new DefaultApi(configuration))
      .map((apiClient) =>
        apiClient.apiV1PluginsHyperledgerCactusPluginConsortiumManualNodeJwsGet()
      );
    const responses = await Promise.all(requests);
  
    const signatures: JWSRecipient[] = [];
  
    responses
      .map((apiResponse) => apiResponse.data)
      .map((getNodeJwsResponse) => getNodeJwsResponse.jws)
      .forEach((aJws: JWSGeneral) =>
        aJws.signatures.forEach((signature) => signatures.push(signature))
      );
  
    const [response] = responses;
    const jws = response.data.jws;
    jws.signatures = signatures;
  
    const body = { jws };
    ```
3. The new methods on the plugin class at `./plugin at `./packages/cactus-plugin-consortium-manual/src/main/typescript/plugin-consortium-manual.ts` to have a single input parameter in the form of the request body and the output needs to be the response body object (both of these types should be defined by the open API spec of the `core-api` package, there's another issue open specifically for this: #427 

Depends on #427 

AzaharaC:
Resolve #429

### DIFF
--- a/packages/cactus-plugin-consortium-manual/src/main/typescript/plugin-consortium-manual.ts
+++ b/packages/cactus-plugin-consortium-manual/src/main/typescript/plugin-consortium-manual.ts
@@ -4,6 +4,9 @@ import { Optional } from "typescript-optional";
 import { promisify } from "util";
 import express, { Express } from "express";
 import bodyParser from "body-parser";
+import { JWS, JWK } from "jose";
+import jsonStableStringify from "json-stable-stringify";
+import { v4 as uuidv4 } from "uuid";
 
 import {
   ConsortiumDatabase,
@@ -12,6 +15,8 @@ import {
   IWebServiceEndpoint,
   ICactusPlugin,
   ICactusPluginOptions,
+  JWSGeneral,
+  JWSRecipient,
 } from "@hyperledger/cactus-core-api";
 
 import { PluginRegistry, ConsortiumRepository } from "@hyperledger/cactus-core";
@@ -31,6 +36,11 @@ import {
   IGetPrometheusExporterMetricsEndpointV1Options,
   GetPrometheusExporterMetricsEndpointV1,
 } from "./consortium/get-prometheus-exporter-metrics-endpoint-v1";
+
+import {
+  Configuration,
+  DefaultApi,
+} from "./generated/openapi/typescript-axios";
 export interface IWebAppOptions {
   port: number;
   hostname: string;
@@ -39,6 +49,7 @@ export interface IWebAppOptions {
 export interface IPluginConsortiumManualOptions extends ICactusPluginOptions {
   keyPairPem: string;
   consortiumDatabase: ConsortiumDatabase;
+  consortiumRepo: ConsortiumRepository;
   prometheusExporter?: PrometheusExporter;
   pluginRegistry?: PluginRegistry;
   logLevel?: LogLevelDesc;
@@ -176,7 +187,7 @@ export class PluginConsortiumManual
 
     const endpoints: IWebServiceEndpoint[] = [];
     {
-      const options = { keyPairPem, consortiumRepo };
+      const options = { keyPairPem, consortiumRepo, plugin: this };
       const endpoint = new GetConsortiumEndpointV1(options);
       endpoints.push(endpoint);
       const path = endpoint.getPath();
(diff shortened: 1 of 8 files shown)


packages/cactus-plugin-consortium-manual/src/main/typescript/plugin-consortium-manual.ts
```diff
@@ -216,4 +227,46 @@ export class PluginConsortiumManual
   public getAspect(): PluginAspect {
     return PluginAspect.CONSORTIUM;
   }
+
+  public async createJws(): Promise<JWSGeneral> {
```
petermetz, line 231: @AzaharaC Please name this method to match the endpoint that it is the backing implementation of, e.g. something like `getNodeJws()`

packages/cactus-plugin-consortium-manual/src/main/typescript/plugin-consortium-manual.ts
```diff
@@ -216,4 +227,46 @@ export class PluginConsortiumManual
...
+    const _protected = {
+      iat: Date.now(),
+      jti: uuidv4(),
+      iss: "Hyperledger Cactus",
+    };
+    // TODO: double check if this casting is safe (it is supposed to be)
+    return JWS.sign.general(payloadJson, keyPair, _protected) as JWSGeneral;
+  }
+
+  public async getJws(): Promise<JWSGeneral> {
```
petermetz, line 247: @AzaharaC Please name this method to match the endpoint that it is the backing implementation of, e.g. something like `getConsortiumJws()`

packages/cactus-plugin-consortium-manual/src/test/typescript/unit/consortium/get-node-jws-endpoint-v1.test.ts
```diff
@@ -75,16 +76,16 @@ test("Can provide JWS", async (t: Test) => {
   await pluginConsortiumManual.getOrCreateWebServices();
   await pluginConsortiumManual.registerWebServices(expressApp);
 
-  const epOpts: IGetNodeJwsEndpointOptions = {
+  /*  const epOpts: IGetNodeJwsEndpointOptions = {
```
petermetz, line 79: @AzaharaC It's cool to delete this if no longer needed, if we later figure that it is needed we can always take a peek at the git history to restore it so no need to keep a backup of it in the comments.

packages/cactus-plugin-consortium-manual/src/test/typescript/unit/consortium/get-node-jws-endpoint-v1.test.ts
```diff
@@ -75,16 +76,16 @@ test("Can provide JWS", async (t: Test) => {
...
     plugin: pluginConsortiumManual,
     consortiumRepo,
     keyPairPem,
-  };
+  };*/
   const pubKeyPem = keyPair.toPEM(false);
 
-  const ep = new GetNodeJwsEndpoint(epOpts);
+  // const ep = new GetNodeJwsEndpoint(epOpts);
```
petermetz, line 86: @AzaharaC It's cool to delete this if no longer needed, if we later figure that it is needed we can always take a peek at the git history to restore it so no need to keep a backup of it in the comments.

petermetz (changes requested):

petermetz (approved):

PR #913, opened 2021-05-04
refactor(core-api): get is replaced with post

### LINKED ISSUE #430: refactor(core-api): change consortium and node JWS endpoints to have a POST HTTP verb from GET
### Description

For the sake of being a little more future proof it would be better to have these endpoints accept request bodies (`POST` HTTP verb) so that we can for example add parameters later on pertaining to the behavior of the endpoints. An idea would be to specify the majority threshold we'd want for a consortium JWS to be considered valid and this could be specified in the request roughly translating to behavior where the developer interfacing with the Cactus API could say `Give me the JWS for your consortium (e.g. all nodes sign it) and I'm okay with X% of those being off/invalid)`

Files where the change can be made:
1. `./packages/cactus-plugin-consortium-manual/src/main/typescript/consortium/get-node-jws-endpoint-constants.ts`
2. `./packages/cactus-plugin-consortium-manual/src/main/typescript/consortium/get-consortium-jws-endpoint-constants.ts`

In theory the rest should be done because we have the endpoint wiring code set up in a way that this automatically makes the endpoints change to `POST` from `GET` without any further action.

> Once you make the changes, the build has to run so that the api-client gets regenerated that will also need to be included in the commit when resolving this issue.

### Acceptance Criteria
1. Both instances of 
    ```typescript 
    public static readonly HTTP_VERB_LOWER_CASE: string = "get";
    ``` 
    is replaced with

    ```typescript
    public static readonly HTTP_VERB_LOWER_CASE: string = "post";
    ```
    (notice the verb change)
2. Make sure the tests where the API client's method for these requests gets invoked are updated if necessary (they may start insisting on a request body object being passed in, even if it's just an empty object literal. For this reason, this issue depends on #427 which is the one where all the request/response objects will be defined.

Depends on #427 

AzaharaC:
Resolve #430

### DIFF
--- a/packages/cactus-plugin-consortium-manual/src/main/json/openapi.json
+++ b/packages/cactus-plugin-consortium-manual/src/main/json/openapi.json
@@ -61,15 +61,25 @@
             "PrometheusExporterMetricsResponse": {
                 "type": "string",
                 "nullable": false
+            },
+            "GetNodeJwsRequest": {
+                "type": "object",
+                "properties": {
+                }
+            },
+            "GetConsortiumJwsRequest": {
+                "type": "object",
+                "properties": {
+                }
             }
         }
     },
     "paths": {
         "/api/v1/plugins/@hyperledger/cactus-plugin-consortium-manual/consortium/jws": {
-            "get": {
+            "post": {
                 "x-hyperledger-cactus": {
                     "http": {
-                        "verbLowerCase": "get",
+                        "verbLowerCase": "post",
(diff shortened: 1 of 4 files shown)


petermetz:
@AzaharaC If you rebase onto https://github.com/hyperledger/cactus/pull/915 that should fix that test failure which crept in due to auto-upgrading of go dependencies in the hello world chain code we have. (then you need to mark this PR as dependent on https://github.com/hyperledger/cactus/pull/915)

petermetz (approved):

takeutak (approved):

PR #921, opened 2021-05-05
fix(cmd-api-server): config-service example - authorization JSON

petermetz:
Depends on #993 

The custom formatter that was introduced the parse the JSON
config for the authorizer was causing problems.
This was overlooked at the time of the implementation of the authz
feature because Peter (yours truly) is an idiot but also to a smaller
extent because there was no automated test coverage for this specific
issue which this commit is now rectifying by adding a new test case.

Longer term we should look into using a different configuration
parsing library that has more flexibility on how to handle JSON
and validations around it.

There was a second bug masked by the first which is that the
"algorithms" array property of the express-jwt middleware
options was also not being used correctly, but to get to that
first the initial bug with the parsing had to be fixed.


petermetz:
cc: @TonyRowntree

kikoncuo (approved):

takeutak (approved):

petermetz:
@takeutak Did you mean to dismiss your approval on this PR or was it by accident? I see in the logs that you approved it and then re-requested review from yourself a little later.

PR #923, opened 2021-05-06
build(lint): start using cspell for grammar in code

### LINKED ISSUE #321: build: start using cspell for grammar in code
### Description
As a developer whose native tongue is *not English* I want to have at least a bare minimum spell checking in my code so that I don't end up making silly grammatical mistakes and I also don't have to paste my code manually into a spell checker for example.

There's a great (but not perfect) tool for making English grammar spell checks part of the linting process that we could use on the source code that matches this glob pattern for example `./packages/**/*.ts`.
Can be expanded to markdown later as we go, but wanted to start small and simple so for now the glob pattern only applies to Typescript source files and only under `./packages/` which is where the code based on the new architecture is residing (don't want to waste too much time fixing up the mistakes in the old code. 

### Acceptance Criteria
1. cspell get executed as part of the lint npm script and also in a pre-commit hook just to be sure (maybe we already run the lint script in the pre-commit hook, can't remember TBH)
2. cspell is passing (meaning that if there were grammatical errors these should also be fixed by the PR so that the CI does not fail
3. cspell is part of the CI (this should happen implicitly if you made it part of the `lint` npm script, but wanted to explicitly state it here as a requirement anyway)

travis-payne:
Utilsing cSpell to prevent common spelling mistakes in code.

This is implemented as part of the npm lint script, which gets executed both
on a commit and during the CI pipeline. Parameters for this are defined in .cspell.json

Resolve #321


petermetz (approved):
Full disclosure: I didn't sort through each and every change one by one, just sort of skimmed it :-)

petermetz:
@travis-payne Looks like this CI failure is legit (at a quick glance) it claims there are 4 more cspell errors.

takeutak (approved):

petermetz:
@travis-payne Squash and rebase please

PR #925, opened 2021-05-06
chore(deps): upgrade to npm v7

petermetz:
This doesn't actually do anything apart from changing all the
package-lock.json files to the new format mandated by versin 7 of npm.

The real change is that everyone now has to upgrade to npm v7 with a
command like this (or exactly this)
npm i -g npm@7

Personally, what I did is went all in and upgraded my NodeJS version to
v16.1.0 (current LTS) because we have to keep up with the times anyway.
If you use the excellent node version manager (nvm) tool then it's as
simple as saying: "nvm install 16" and boom, you are upgraded.
It is highly recommended that everyone on the team uses nvm because
it makes life so much easier when having to switch between different
NodeJS versions.

Depends on #926


petermetz:
Note to self: **do not merge this without also sending an email to the mailing list and the chat explaining that this change has happened and that we recommend everyone upgrades to NodeJS v16.x or at least to upgrade their npm to v7 otherwise version control will keep flickering the package-lock.json files)**

takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #926, opened 2021-05-06
build(webpack): upgrade to v5.36.2

petermetz:
Previously we were using a fork of Peter's because there was a missing
feature that we needed and the PR approval in webpack took a long time
but now it has been approved and released onto the upstream so
we can just go back to using the official release channel of webpack
and that's what this commit does.

The other reason why this suddenly became so important is because
npm v7 is failing to install webpack directly from GitHub forks which
may be easy to fix with some flag to make npm ignore the SSL issues
but that's not how we roll since one of our core design principles is
secure by default.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3379,9 +3379,9 @@
       }
     },
     "@types/eslint": {
-      "version": "7.2.6",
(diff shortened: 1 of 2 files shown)


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #927, opened 2021-05-06
chore(release): publish 0.5.0

petermetz:


sfuji822:
@petermetz  @kikoncuo 
Can you explain why you added "alpha" for this version only?
# I thought our codes are all in the alpha or beta stage

I also think we need to discuss release planning toward V1 in the next contributor meeting, like what feature must be included for specific milestone version.

petermetz:
> @petermetz @kikoncuo
> Can you explain why you added "alpha" for this version only?
> 
> # I thought our codes are all in the alpha or beta stage

@sfuji822 It's because the release management scripts themselves are in alpha/beta stage as well so they are being refined as we speak and I just added support for the mentioned scripts to be able to issue these alpha/beta tags, we just didn't have this functionality before in our release management capabilities.

Everything before the 1.0.0 release is implied to be in alpha/beta stage according to semantic versioning conventions indeed.

> I also think we need to discuss release planning toward V1 in the next contributor meeting, like what feature must be included for specific milestone version.

Sure thing, please feel free to add that as a discussion item to the agenda.

takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #949, opened 2021-05-08
feat(test-tooling): improvements to fabric aio

RafaelAPB:
1. feat(test-tooling): obtain connection profile from multiple orgs 

Allows obtaining a connection profile from any org that is created.
Adds a state database attribute to the test ledger (will be necessary for operations regarding private data)

2. fix(test-tooling): deletes volume created by faio

When a faio instance is deployed, it creates a volume. That volume is not destroyed when faio is shutdown. This commit destroys the created volume.

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
@@ -25,6 +25,7 @@ import {
   LoggerProvider,
   Bools,
 } from "@hyperledger/cactus-common";
+import Dockerode from "dockerode";
 
 /*
  * Contains options for Fabric container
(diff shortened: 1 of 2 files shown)


takeutak (approved):

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -36,6 +36,7 @@ export interface IFabricTestLedgerV1ConstructorOptions {
   envVars?: Map<string, string>;
   logLevel?: LogLevelDesc;
   emitContainerLogs?: boolean;
+  stateDatabase?: string;
```
petermetz, line 39: @RafaelAPB Is the `stateDatabase` argument somehow needed for the `getConnectionProfileOrgX` method? (I did a quick search but couldn't find a reference for it so I wasn't sure what it is used for)

RafaelAPB:
@petermetz is this good to go, or you'd rather prefer me to remove the stateDatabase string (although it is related to fabric-aio-improvements)?

petermetz:
@RafaelAPB If it's not being used for anything at the moment then I'd drop it, or at least document it in the code that this is not being used right now but it is going to be used/will be good for something.

RafaelAPB:
Closing, as this is covered by #1084

PR #964, opened 2021-05-13
feat(aws-sm): added keychain plugin for aws secret manager

### LINKED ISSUE #912: feat(keychain-aws-secretmanager): create aws secret manager plugin
### Description
As a developer, I want to create a plugin for aws secret manager and perform tasks related to secret management like createSecret and retrieveSecret sort of functionalities.

For reference,
https://docs.aws.amazon.com/secretsmanager/latest/userguide/intro.html

### Acceptance Criteria
1. New package to be created as `@hyperledger/cactus-plugin-keychain-aws-sm`
2. Plugin must implement the central keychain plugin Typescript interface we have in the core-api package
3. End to end tests to be written that verify that the plugin does indeed work by leveraging localstack
4. In a separate PR to this add a helper class in the test-tooling package called `LocalstackTestContainer` which launches a localstack container
5. Make sure to have the healthcheck of the local stack container passing as part of the start() method's return value (the promise should only resolve once the container is healthy and the localstack services are up and running,  ready to serve requests)
6. The helper class to have methods to expose the credentials, port numbers, etc. that are required to configure/instantiate the `PluginKeychainAwsSm` class which is to be implemented as part of the package from point 1)

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo @petermetz  @arnab-roy

jagpreetsinghsasan:
# Commit to be reviewed
feat(aws-sm): added keychain playing for aws secret manager

        Primary Change
        ---
        1. Added new package cactus-plugin-keychain-aws-sm under packages/
        2. Added LocalStack under cactus-test-tooling/src/main/typescript for local aws secret manager deployment and testing

        Refactorings that were also necessary to incorporate 1) and 2)
        ---
        3. Updated public-api.ts under packages/cactus-test-tooling/src/main/typescript for exporting the Localstack class, its interface and constants

Resolves #912


packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
```diff
@@ -0,0 +1,124 @@
...
+  // Using awsCredentialType: AwsCredentialType.FromCodeVariable
+  {
+    const options2: IPluginKeychainAwsSmOptions = {
+      instanceId: uuidv4(),
+      keychainId: uuidv4(),
+      awsEndpoint: localstackHost,
+      awsRegion: "us-east-1",
+      awsProfile: "default",
+      awsCredentialType: AwsCredentialType.InMemory,
+      awsAccessKeyId: "fake",
```
petermetz, line 89: @jagpreetsinghsasan Could you also look into adding a negative case for the credentials? E.g. a `plugin3` instance which has credentials that we know are invalid and then a single call to each of the keychain methods with test assertions checking that an exception is thrown with something that is indicative to the user of what is actually going wrong. Flexible on what that error message is (I assume the AWS SDK has something good for this) as long as it does let the user know that it was some sort of authorization/authentication error instead something that doesn't (think of these random errors like "cannot ready property undefined of null blah blah").

If LocalStack has a way of actually checking the AWS access key ID and secret access key to match a pre-defined value then you can use that, but if the only thing LocalStack is capable of is accepting literally anything as a valid set of credentials then you may just need to fall back on pointing the plugin to the real AWS services for `plugin3` so that we can test the negative case where all it does is fail.

packages/cactus-test-tooling/src/main/typescript/localstack/localstack.ts
```diff
@@ -0,0 +1,152 @@
...
+export const K_DEFAULT_LOCALSTACK_IMAGE_VERSION = "0.12.10";
+export const K_DEFAULT_LOCALSTACK_HTTP_PORT = 4566;
+
+/**
+ * > **Do not use this image in production.**
+ *
+ * Class responsible for programmatically managing a container that is running
+ * the official LocalStack image in development mode.
+ */
+export class LocalStack {
```
petermetz, line 30: @jagpreetsinghsasan I recommend renaming this to `LocalStackContainer` to make it clear that the class itself is not the entire thing that LocalStack is but rather just a small utility/helper for managing a container that is LocalStack.
If you do the renaming please also make sure to rename the .ts file itself in the same manner.

packages/cactus-test-tooling/src/main/typescript/localstack/localstack.ts
```diff
@@ -0,0 +1,152 @@
...
+          ExposedPorts: {
+            [`${K_DEFAULT_LOCALSTACK_HTTP_PORT}/tcp`]: {},
+          },
+          // This is a workaround needed for macOS which has issues with routing
+          // to docker container's IP addresses directly...
+          // https://stackoverflow.com/a/39217691
+          PublishAllPorts: true,
+          HostConfig: {
+            // NetworkMode: "host",
+            CapAdd: ["IPC_LOCK"],
```
petermetz, line 97: @jagpreetsinghsasan Didn't find anything in the [localstack readme](https://hub.docker.com/r/localstack/localstack) about the IPC_LOCK capability being necessary so I think we could just remove this to adhere to the [principle of least privileges](https://en.wikipedia.org/wiki/Principle_of_least_privilege)

packages/cactus-test-tooling/src/main/typescript/public-api.ts
```diff
@@ -62,6 +62,14 @@ export {
   K_DEFAULT_VAULT_DEV_ROOT_TOKEN,
 } from "./vault-test-server/vault-test-server";
 
+export {
+  ILocalStackOptions,
+  LocalStack,
```
petermetz, line 67: @jagpreetsinghsasan Also take care to rename here if you use the VSCode renaming feature

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
```diff
@@ -0,0 +1,124 @@
...
+import {
+  IPluginKeychainAwsSmOptions,
+  PluginKeychainAwsSm,
+  AwsCredentialType,
+} from "../../../main/typescript/public-api";
+
+const logLevel: LogLevelDesc = "TRACE";
+
+test("get,set,has,delete alters state as expected", async (t: Test) => {
+  const localStackContainer = new LocalStack({});
```
petermetz, line 23: @jagpreetsinghsasan Recommending to pass in the `logLevel` to the LocalStackContainer options instead of just the empty object.

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/webservices/set-keychain-entry-endpoint-v1.ts
```diff
@@ -0,0 +1,76 @@
...
+  public getVerbLowerCase(): string {
+    return this.getOperation()["x-hyperledger-cactus"].http.verbLowerCase;
+  }
+
+  public getPath(): string {
+    return this.getOperation()["x-hyperledger-cactus"].http.path;
+  }
+
+  public getExpressRequestHandler(): IExpressRequestHandler {
+    throw new Error("Method not implemented.");
```
petermetz, line 74: @jagpreetsinghsasan Please create a follow-up issue to finish the endpoint's request handler as well + write tests that use the API client instead of the plugin instance directly.

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/public-api.ts
```diff
@@ -0,0 +1,18 @@
...
+import { IPluginFactoryOptions } from "@hyperledger/cactus-core-api";
+
+import { PluginFactoryKeychain } from "./plugin-factory-keychain";
+export { PluginFactoryKeychain } from "./plugin-factory-keychain";
+
+export {
+  IPluginKeychainAwsSmOptions,
+  PluginKeychainAwsSm,
+  AwsCredentialType,
+} from "./plugin-keychain-vault-aws-sm";
```
petermetz, line 12: @jagpreetsinghsasan Typo? `vault`

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-vault-aws-sm.ts
```diff
@@ -0,0 +1,267 @@
...
+  ICactusPlugin,
+  ICactusPluginOptions,
+  IPluginWebService,
+  IWebServiceEndpoint,
+  PluginAspect,
+} from "@hyperledger/cactus-core-api";
+
+export enum AwsCredentialType {
+  LocalFile = "FROM_AWS_CREDENTIAL_FILE",
+  InMemory = "FROM_CODE_VARIABLE",
```
petermetz, line 30: @jagpreetsinghsasan I'd update the enum values to match the left side of their declaration

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-vault-aws-sm.ts
```diff
@@ -0,0 +1,267 @@
+import { Server } from "http";
+import { Server as SecureServer } from "https";
```
petermetz, line 2: @jagpreetsinghsasan Same request here with the `import type { ... }` syntax

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-vault-aws-sm.ts
```diff
@@ -0,0 +1,267 @@
+import { Server } from "http";
```
petermetz, line 1: @jagpreetsinghsasan Same request here with the `import type { ... }` syntax

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-aws-sm-remote-adapter.ts
```diff
@@ -0,0 +1,154 @@
...
+    return data.value as any;
+  }
+
+  public async set<T>(key: string, value: T): Promise<void> {
+    // FIXME Does it make any sense to have the set() method be generic?
+    await this.backend.setKeychainEntry({ key, value: value as any });
+  }
+
+  public async delete(key: string): Promise<void> {
+    // FIXME Pretty sure aws secret manager can do delete so we don't have to hack it like this
```
petermetz, line 137: @jagpreetsinghsasan Can you confirm if this is leftover from the copy or is the AWS SM also not have an API for deleting entries like Vault?

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-aws-sm-remote-adapter.ts
```diff
@@ -0,0 +1,154 @@
+import { Server } from "http";
```
petermetz, line 1: @jagpreetsinghsasan Same request here with the `import type { ... }` syntax

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-aws-sm-remote-adapter.ts
```diff
@@ -0,0 +1,154 @@
...
+import { Server as SecureServer } from "https";
```
petermetz, line 2: @jagpreetsinghsasan Same request here with the `import type { ... }` syntax

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,103 @@
...
+  "mainMinified": "dist/cactus-plugin-keychain-aws-sm.node.umd.min.js",
+  "browser": "dist/cactus-plugin-keychain-aws-sm.web.umd.js",
+  "browserMinified": "dist/cactus-plugin-keychain-aws-sm.web.umd.min.js",
+  "module": "dist/lib/main/typescript/index.js",
+  "types": "dist/types/main/typescript/index.d.ts",
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "generate-sdk": "openapi-generator generate --input-spec src/main/json/openapi.json -g typescript-axios -o ./src/main/typescript/generated/openapi/typescript-axios/",
```
petermetz, line 15: @jagpreetsinghsasan Depending on which PR gets merged first please watch out for having to update this and also to regenerate the API client sources (the other PR being the one where Travis upgrades the generator to the latest and greatest for the whole project)

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,103 @@
+{
+  "name": "@hyperledger/cactus-plugin-keychain-aws-sm",
+  "version": "0.4.1",
```
petermetz, line 3: @jagpreetsinghsasan Version will have to be updated as well because we have a release in the queue so keep an eye out for this as well (I know for a fact that the release will be issued before the merging of this PR)

packages/cactus-plugin-keychain-aws-sm/README.md
```diff
@@ -0,0 +1,3 @@
+# `@hyperledger/cactus-plugin-keychain-aws-sm`
+
+## TO-DO
```
petermetz, line 3: @jagpreetsinghsasan Please create a follow-up issue to write a README.md file with examples for the package.

petermetz (changes requested):

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  // Using awsCredentialType: AwsCredentialType.FromAwsCredentialFile
+  {
+    const options1: IPluginKeychainAwsSmOptions = {
+      instanceId: uuidv4(),
+      keychainId: uuidv4(),
+      awsEndpoint: localstackHost,
+      awsRegion: "us-east-1",
+      awsProfile: "default",
+      awsCredentialType: AwsCredentialType.LocalFile,
```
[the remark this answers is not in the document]
jagpreetsinghsasan: Nice. I didn't think of this test case of parallel execution.

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-aws-sm-remote-adapter.ts
jagpreetsinghsasan: I copied it from the vault one because even in vault remote adapter, the other endpoints aren't implemented yet

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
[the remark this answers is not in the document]
jagpreetsinghsasan: I had to put a 1 second sleep, otherwise, the file is not getting created.. Used synchronous methods but still no luck.

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
jagpreetsinghsasan: Yes, the LocalStack accepts anything as the access key ID and secret access keys.
So how should I point it to real AWS services?

packages/cactus-plugin-keychain-aws-sm/package.json
jagpreetsinghsasan: Do we know what's the next release, 0.4.2 or 0.5.0? (Like is it planned yet)

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-aws-sm-remote-adapter.ts
jagpreetsinghsasan: I have updated the newly created issue for adding this endpoint too or maybe update the set-keychain-entry endpoint with some tag to represent deletion.

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/webservices/set-keychain-entry-endpoint-v1.ts
jagpreetsinghsasan: Created #967

packages/cactus-plugin-keychain-aws-sm/README.md
jagpreetsinghsasan: Created #968

packages/cactus-plugin-keychain-aws-sm/package.json
petermetz: @jagpreetsinghsasan It's going to be `0.5.0-alpha.0` (see https://github.com/hyperledger/cactus/pull/927 )

You can't update it just yet though because then there will be other problems, but once #927 is merged, then it will be possible (and necessary at that point).

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-aws-sm-remote-adapter.ts
petermetz: I didn't mean the endpoint implementation, but specifically the delete method's implementation being the way it is where instead of calling a delete method it sets it with an empty string (which is different from calling a delete method on the AWS SDK - if it exists).
So my question is about the underlying libraries that we use, not the plugins themselves.

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/a.ts
```diff
@@ -0,0 +1,15 @@
+// import fs from "fs"
```
petermetz, line 1: @jagpreetsinghsasan If this is just a leftover scratchpad then it could be deleted.

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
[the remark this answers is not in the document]
petermetz: Use the `fs` module instead of streams (sent more details via other messaging channel)

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
petermetz: > Yes, the LocalStack accepts anything as the access key ID and secret access keys.

I get that that's the default behavior, but do they have maybe a configuration parameter in their documentation that would make it so that it actually checks the credentials against some pre-defined value?
I can imagine them maybe having a parameter that you can start the container with like `--access-key-id=myfakekey` and then it would actually only accept `myfakekey` as a valid access key id and nothing else.
I just want to make sure that you checked their documentation for something like this and if it isn't there either then we could open a feature request in the local stack repo and in the meantime fall back to using the real AWS services as you mentioned.

petermetz (changes requested):

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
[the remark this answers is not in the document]
jagpreetsinghsasan: I did this

`    let tmpDirPath = "tmpDirPath";
    await (async() => {
      tmpDirPath = await fs.promises.mkdtemp(path.join(os.tmpdir(), "cactus-"))
      await fs.promises.writeFile(`${tmpDirPath}/credentials`, "[default]\naws_secret_access_key = test\naws_access_key_id = test", "utf-8")
    })();`

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
jagpreetsinghsasan: I see. Let me check that out.

jagpreetsinghsasan: The paid version (pro) of localstack supports IAM based authentication.
( I got a workaround of this in my head.. if it works, we can still do IAM access control testing)
image

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-aws-sm-remote-adapter.ts
jagpreetsinghsasan: Yes, this will be replaced with deleteKeychainEntry( ) method just like we have to setKeychainEntry( )
image

jagpreetsinghsasan: But I didnt define the deleteKeychainEntry because its not defined in the openapi of cactus-core-api

jagpreetsinghsasan: Hi @petermetz 
I have mentioned this to be fixed in the issue #967, to be fixed with the overall web services endpoints like get and set keychainEntry

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/index.web.ts
```diff
@@ -0,0 +1 @@
+export {};
```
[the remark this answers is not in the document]
petermetz: @jagpreetsinghsasan ?

petermetz (changes requested):

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
```diff
@@ -0,0 +1,149 @@
...
+  test.onFinish(async () => {
+    await localStackContainer.stop();
+    await localStackContainer.destroy();
+  });
+
+  // Using awsCredentialType: AwsCredentialType.FromAwsCredentialFile
+  {
+    // Create aws credential file in a local directory
+    let tmpDirPath = "tmpDirPath";
+    await (async () => {
```
petermetz, line 49: @jagpreetsinghsasan Is this anonymous wrapper function needed? Does it not compile without it?

petermetz (changes requested):

packages/cactus-plugin-keychain-aws-sm/src/main/typescript/plugin-keychain-aws-sm-remote-adapter.ts
petermetz: @jagpreetsinghsasan Could you also open a separate issue to add the delete method to the core-api keychain definitions?

jagpreetsinghsasan: I have added that to the issue #967 (third point of the description of the issue) while creating it.

packages/cactus-plugin-keychain-aws-sm/src/test/typescript/integration/plugin-keychain-aws-sm.test.ts
jagpreetsinghsasan: Yes, its no longer needed. Fixed.

petermetz:
@jagpreetsinghsasan Just double checking, is this ready for review?

jagpreetsinghsasan:
Yes, this is ready for review

petermetz:
> Yes, this is ready for review

@jagpreetsinghsasan Got it, thanks. To speed things up next time, please use this: https://github.blog/changelog/2019-02-21-re-request-review-on-a-pull-request/

petermetz (changes requested):
@jagpreetsinghsasan I see some things resolved in the comments but the code looks the same as before. Please take a second pass on all the review comments and then hit the re-request review.

petermetz (commented):
@jagpreetsinghsasan @elenaizaguirre Same request here regarding the openapi.json cross-references. (see my other comment on the other keychain plugin PRs for details)

petermetz (approved):

takeutak:
Would you wait my review since I'm asking about this PR on #1017?

takeutak (approved):

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,102 @@
+{
+  "name": "@hyperledger/cactus-plugin-keychain-aws-sm",
+  "version": "0.5.0",
```
petermetz, line 3: @jagpreetsinghsasan Please update to `0.6.0` (just issued the release) and then we are ready to go.

packages/cactus-plugin-keychain-aws-sm/package-lock.json
```diff
@@ -0,0 +1,1065 @@
+{
+	"name": "@hyperledger/cactus-plugin-keychain-aws-sm",
+	"version": "0.5.0",
```
petermetz, line 3: @jagpreetsinghsasan Please update to `0.6.0` (just issued the release) and then we are ready to go.

packages/cactus-plugin-keychain-aws-sm/package-lock.json
```diff
@@ -0,0 +1,1065 @@
...
+	"lockfileVersion": 2,
+	"requires": true,
+	"packages": {
+		"": {
+			"name": "@hyperledger/cactus-plugin-keychain-aws-sm",
+			"version": "0.5.0",
```
petermetz, line 9: @jagpreetsinghsasan Please update to `0.6.0` (just issued the release) and then we are ready to go.

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,102 @@
...
+      "url": "https://accenture.com"
+    }
+  ],
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/hyperledger/cactus/issues"
+  },
+  "homepage": "https://github.com/hyperledger/cactus#readme",
+  "dependencies": {
+    "@hyperledger/cactus-common": "0.5.0",
```
petermetz, line 86: @jagpreetsinghsasan Please update to `0.6.0` (just issued the release) and then we are ready to go.

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,102 @@
...
+    "@hyperledger/cactus-core": "0.5.0",
```
petermetz, line 87: @jagpreetsinghsasan Please update to `0.6.0` (just issued the release) and then we are ready to go.

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,102 @@
...
+    "@hyperledger/cactus-core-api": "0.5.0",
```
petermetz, line 88: @jagpreetsinghsasan Please update to `0.6.0` (just issued the release) and then we are ready to go.

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,102 @@
...
+    "aws-sdk": "2.903.0",
+    "axios": "0.21.1",
+    "http-status-codes": "2.1.4",
+    "typescript-optional": "2.0.1"
+  },
+  "devDependencies": {
+    "@hyperledger/cactus-test-tooling": "0.5.0",
```
petermetz, line 95: @jagpreetsinghsasan Please update to `0.6.0` (just issued the release) and then we are ready to go.

petermetz (changes requested):
@jagpreetsinghsasan Please update to all Cacts package version occurrences to `0.6.0` from `0.5.0` (just issued the release) and then run the configure script to also update the lock files as well. Commit all of those changes and then we are ready to merge.

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,102 @@
...
+  "browser": "dist/cactus-plugin-keychain-aws-sm.web.umd.js",
+  "browserMinified": "dist/cactus-plugin-keychain-aws-sm.web.umd.min.js",
+  "module": "dist/lib/main/typescript/index.js",
+  "types": "dist/types/main/typescript/index.d.ts",
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "generate-sdk": "openapi-generator-cli generate -i ./src/main/json/openapi.json -g typescript-axios -o ./src/main/typescript/generated/openapi/typescript-axios/",
+    "tsc": "tsc --project ./tsconfig.json",
```
petermetz, line 16: @jagpreetsinghsasan The `tsc` script can be safely deleted now with the new build system in place.

petermetz (changes requested):

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,102 @@
...
+    "watch": "npm-watch",
+    "pretsc": "npm run generate-sdk",
```
petermetz, line 18: @jagpreetsinghsasan The `pretsc` script can be safely deleted now with the new build system in place.

packages/cactus-plugin-keychain-aws-sm/package.json
```diff
@@ -0,0 +1,102 @@
...
+    "webpack": "npm-run-all webpack:dev webpack:prod",
+    "webpack:dev": "npm-run-all webpack:dev:node webpack:dev:web",
+    "webpack:dev:web": "webpack --env=dev --target=web --config ../../webpack.config.js",
+    "webpack:dev:node": "webpack --env=dev --target=node --config ../../webpack.config.js",
+    "webpack:prod": "npm-run-all webpack:prod:node webpack:prod:web",
+    "webpack:prod:web": "webpack --env=prod --target=web --config ../../webpack.config.js",
+    "webpack:prod:node": "webpack --env=prod --target=node --config ../../webpack.config.js"
+  },
+  "watch": {
```
petermetz, line 27: @jagpreetsinghsasan Please migrate the watch script like the rest of the packages.

petermetz (changes requested):

petermetz (changes requested):
@jagpreetsinghsasan Almost forgot, please also 
1. migrate the tsconfig.json inside your package to match the new boilerplate (see any of the other packages for an example).
2. Add an entry to the root tsconfig.json file under the `references` array that points to the new package that is being added in the PR
3. Do this to all the pending PRs that are adding new packages

petermetz (approved):

PR #1015, opened 2021-06-08
feat(besu): private transaction support

travis-payne:
This is a first cut of implementing private transaction support into Cactus for besu.

This main issue with this was trying to figure out a meaningful way of testing this functionality, the initial attempt was to use the Besu private network example in a docker container as part of our CI pipeline, as this is how I tested this in development. However this fires up quite a lot of nodes and uses a silly amount of computing resources. Mocking the responses of these transactions would also equate to meaningless tests and we're then not testing the "private" part of the private transactions, and we already have tests that check regular transactions.

A potential solution to this could be to fire up multiple of our "besu-all-in-one" containers in and perhaps try to get them to talk to each other, or even fire up multiple nodes within that container itself. Either way, this ends up making our CI suite quite the choke point. 

Any further ideas from @kikoncuo @petermetz would be welcome.

This was manually tested by doing the following:

- Firing up at local private network by using ``npx quorum-dev-quickstart`` and then running ``./run.sh``
- Locally running ``private-deploy-contract-from-json.test.ts`` which is currently commented out (as it will fail as we don't have a private network running in our CI pipeline.


petermetz:
@travis-payne I don't have any better ideas apart from the ones already outlined by you above so the best thing I can do is to offer to help out with the investigation.
What we can do to move things along is this:
1. You confirm that manual tests were executed and that the code works (and document the test script in a bullet pointed fashion in the git commit message)
2. You document in the package README that the tx privacy feature is developer preview for now due to lack of test coverage
3. Maintainers review the code with the assumption that it works (e.g. if there were tests they would actually pass)
4. This PR gets merged, fingers crossed
5. In parallel: You open a follow-up issue to write the test and the necessary components it would use (a test ledger class that supports tx privacy for the purpose of tests)
6. Then at some point we figure all that out and add the tests in another PR, happy days.

takeutak (dismissed):

PR #1016, opened 2021-06-09
feat(besu): private transaction support

travis-payne:
This change enables private transaction support by allowing extra parameter in the regular transaction endpoint.

This was incredibly difficult to test as it requires firing up at minimum 6 nodes (3 Orion/Tessera, 3 Besu) + validators which our CI pipeline cannot handle as it was using up to 20GB of RAM.  Even running this on a local machine was quite intensive. The second solution was to write mock responses from the nodes, however this defeats the point of the testing the "private" part of the transaction, and we already have tests that do this.

This was manually tested by following these steps:

- Firing up a Besu private network by running ``npx quorum-dev-quickstart`` followed by ``./run.sh``
- Running ``private-deploy-contract-from-json.test.ts`` as a standalone file to see a contract get deployed as a private transaction.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,8 @@
   },
   "dependencies": {
     "@types/ssh2": "0.5.46",
-    "artillery": "^1.5.6",
-    "cspell": "5.4.0"
+    "cspell": "5.4.0",
+    "artillery": "1.7.2",
+    "web3": "1.3.6"
   }
 }
--- a/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
+++ b/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
@@ -6,6 +6,7 @@ import type { Socket as SocketIoSocket } from "socket.io";
 import type { Express } from "express";
 import { promisify } from "util";
 import { Optional } from "typescript-optional";
+
 import Web3 from "web3";
 import { AbiItem } from "web3-utils";
 
@@ -71,6 +72,9 @@ import { WatchBlocksV1Endpoint } from "./web-services/watch-blocks-v1-endpoint";
 
 export const E_KEYCHAIN_NOT_FOUND = "cactus.connector.besu.keychain_not_found";
 
+const EEAClient = require("web3-eea");
+// const PrivateTransaction = require("web3-eea/src/privateTransaction");
+
 export interface IPluginLedgerConnectorBesuOptions
   extends ICactusPluginOptions {
   rpcApiHttpHost: string;
@@ -94,6 +98,7 @@ export class PluginLedgerConnectorBesu
   public prometheusExporter: PrometheusExporter;
   private readonly log: Logger;
   private readonly web3: Web3;
+  private web3EEA: any;
   private readonly pluginRegistry: PluginRegistry;
   private contracts: {
     [name: string]: Contract;
@@ -134,7 +139,19 @@ export class PluginLedgerConnectorBesu
       `${fnTag} options.prometheusExporter`,
     );
 
+    const web3Provider = new Web3.providers.HttpProvider(
+      this.options.rpcApiHttpHost,
+    );
+
+    this.web3 = new Web3(web3Provider);
+    this.initializeEEAClient();
+
     this.prometheusExporter.startMetricsCollection();
+  }
+
+  public async initializeEEAClient() {
+    const chainId = await this.web3.eth.getChainId();
+    this.web3EEA = new EEAClient(this.web3, chainId);
   }
 
   public getPrometheusExporter(): PrometheusExporter {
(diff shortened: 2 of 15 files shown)


takeutak (approved):

package.json
```diff
@@ -126,7 +126,8 @@
   },
   "dependencies": {
     "@types/ssh2": "0.5.46",
-    "artillery": "^1.5.6",
-    "cspell": "5.4.0"
+    "cspell": "5.4.0",
+    "artillery": "1.7.2",
+    "web3": "1.3.6"
```
petermetz, line 131: @travis-payne Is web3 here by mistake?

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -71,6 +72,9 @@ import { WatchBlocksV1Endpoint } from "./web-services/watch-blocks-v1-endpoint";
 
 export const E_KEYCHAIN_NOT_FOUND = "cactus.connector.besu.keychain_not_found";
 
+const EEAClient = require("web3-eea");
+// const PrivateTransaction = require("web3-eea/src/privateTransaction");
```
petermetz, line 76: @travis-payne Please remove commented code if no longer needed.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -94,6 +98,7 @@ export class PluginLedgerConnectorBesu
   public prometheusExporter: PrometheusExporter;
   private readonly log: Logger;
   private readonly web3: Web3;
+  private web3EEA: any;
```
petermetz, line 101: @travis-payne Could the web3-eea typings be used here by any chance?

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -134,9 +139,21 @@ export class PluginLedgerConnectorBesu
       `${fnTag} options.prometheusExporter`,
     );
 
+    const web3Provider = new Web3.providers.HttpProvider(
+      this.options.rpcApiHttpHost,
+    );
+
+    this.web3 = new Web3(web3Provider);
+    this.initializeEEAClient();
```
petermetz, line 147: @travis-payne Calling this here could (as-in, it definitely will) lead to race conditions, to resolve the conundrum I opened this issue: https://github.com/hyperledger/cactus/issues/1035

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -473,6 +526,25 @@ export class PluginLedgerConnectorBesu
       secret,
     } = web3SigningCredential as Web3SigningCredentialPrivateKeyHex;
 
+    // Run transaction to EEA client here if private transaction
+
+    if (req.privateTransactionConfig) {
+      const options = {
```
petermetz, line 532: @travis-payne Could this be made more concise with the spread operator?

e.g.
```typescript
      const options = {
        ...transactionConfig,
        privateKey: secret,
        ...req.privateTransactionConfig,
        restriction: "restricted",
      };

```

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -459,11 +490,33 @@ export class PluginLedgerConnectorBesu
         return { transactionReceipt: ledgerReceipt };
       default:
         throw new Error(
-          `${fnTag} Unrecognized ReceiptType: ${req.consistencyStrategy.receiptType}`,
+          `${fnTag} Unrecognized ReceiptType: ${request.consistencyStrategy.receiptType}`,
         );
     }
   }
 
+  public async transactPrivate(options: any): Promise<RunTransactionResponse> {
```
petermetz, line 498: @travis-payne How much effort would it be to narrow the type here from `any` to something more specific? Assuming the worst case scenario is we could add an interface in our local web3-eea typings that inherits from the web3 tx config and then adds the privateFor/From parameters (and a couple of other ones I saw in your code being used)

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/private-deploy-contract-from-json.test.ts
```diff
@@ -0,0 +1,126 @@
+// import test, { Test } from "tape-promise/tape";
```
petermetz, line 1: @travis-payne I recommend uncommenting this and then skipping it e.g. 
`test.skip(testCase, async (t: Test) => {`

Preferred this way so that at least it still gets compiled and so the shock will be smaller for someone who eventually lands on implementing the test case (because at least there won't be compilation errors when they uncomment the code)

petermetz (changes requested):

PR #991, opened 2021-05-27
feat(azure-kv): added keychain plugin for azure keyvault

jagpreetsinghsasan:
# Commit to be reviewed
------------------------------
feat(azure-kv): added keychain plugin for azure keyvault

        Primary Change
        ---
        1. Added new package cactus-plugin-keychain-azure-kv under packages/
        2. Added PluginKeychainAzureKvMock class to mock the functions of SecretClient under packages/cactus-plugin-keychain-azure-kv/src/test/typescript/mock/plugin-keychain-azure-kv-mock.ts

Resolves #971

Signed-off-by: Jagpreet Singh Sasan <jagpreet.singh.sasan@accenture.com>

PR #1017, opened 2021-06-10
feat(google-sm): added keychain plugin for google secret manager

### LINKED ISSUE #983: feat(keychain-google-sm): create google secret manager plugin
### Description
As a developer, I want to create a plugin for azure secret manager and perform tasks related to secret management like createSecret and retrieveSecret sort of functionalities.

### Acceptance Criteria
1. New package to be created as `@hyperledger/cactus-plugin-keychain-google-sm`
2. Plugin must implement the central keychain plugin Typescript interface we have in the core-api package
3. The helper class to have methods to expose the credentials, port numbers, etc. that are required to configure/instantiate the `PluginKeychainGoogleSm` class which is to be implemented as part of the package from point 1)

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo @petermetz  @arnab-roy

jagpreetsinghsasan:
Commit to be reviewed

feat(google-sm): added keychain plugin for google secret manager   

        Primary Change
        ---
        1. Added new package cactus-plugin-keychain-google-sm under packages/
        2. Added PluginKeychainGoogleSmMock class to mock the functions of SecretManagerServiceClient under packages/cactus-plugin-keychain-google-sm/src/test/typescript/mock/plugin-keychain-google-sm-mock.ts

Resolves #983


petermetz (changes requested):
@jagpreetsinghsasan Please apply the same changes to this one as for the Azure and the AWS plugins (see other 2 PRs for more info)

packages/cactus-plugin-keychain-google-sm/src/main/typescript/plugin-keychain-google-sm-remote-adapter.ts
```diff
@@ -0,0 +1,151 @@
...
+  public getKeychainId(): string {
+    return this.keychainId;
+  }
+
+  public async has(key: string): Promise<boolean> {
+    try {
+      await this.backend.getKeychainEntryV1({ key });
+      return true;
+    } catch (ex) {
+      // FIXME check for errors being thrown due to something other than
```
petermetz, line 118: @jagpreetsinghsasan Please narrow it down to only the exception that gets thrown when the key is not present

packages/cactus-plugin-keychain-google-sm/src/main/typescript/plugin-keychain-google-sm.ts
```diff
@@ -0,0 +1,182 @@
...
+  public getPackageName(): string {
+    return `@hyperledger/cactus-plugin-keychain-vault`;
+  }
+
+  async rotateEncryptionKeys(): Promise<void> {
+    throw new Error("Method not implemented.");
+  }
+
+  public getEncryptionAlgorithm(): string {
+    return null as any;
```
petermetz, line 121: @jagpreetsinghsasan Is there any documentation about the encryption algorithm used by GCP SM?

packages/cactus-plugin-keychain-google-sm/src/main/typescript/plugin-keychain-google-sm.ts
```diff
@@ -0,0 +1,182 @@
...
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  async set<T>(key: string, value: T): Promise<void> {
+    const parent = key.split("?")[0];
+    const secretId = key.split("?")[1];
+    const secret = await this.googleSmClient.createSecret({
```
petermetz, line 160: @jagpreetsinghsasan Is there some logic here that we need to take care of so that the set method is re-entrant?

petermetz (changes requested):

packages/cactus-plugin-keychain-google-sm/src/main/typescript/plugin-keychain-google-sm.ts
jagpreetsinghsasan: https://cloud.google.com/secret-manager/docs/encryption

I think it is **AES-256**

packages/cactus-plugin-keychain-google-sm/src/main/typescript/plugin-keychain-google-sm-remote-adapter.ts
jagpreetsinghsasan: I didnt write this method, because the getKeychainEntryV1( ) and similar methods aren't implemented and tested yet, in this PR. (Forgot to created the issue for this and the readme)

Created now
https://github.com/hyperledger/cactus/issues/1097 (get,set,delete keychainEntry endpoints) (point 3 of acceptance criteria specifically for this review comment)
https://github.com/hyperledger/cactus/issues/1096 (For Readme.md)

packages/cactus-plugin-keychain-google-sm/src/main/typescript/plugin-keychain-google-sm.ts
jagpreetsinghsasan: I added the comment in test case, but forgot to mention here.
Basically the key should consist of parent and secretId seperated by ?
For example, key = "projects/my-project?my-secret"

I have added that now.

petermetz: @jagpreetsinghsasan What I mean: would this code work or would the second invocation crash? (If it crashes it's not re-entrant, if it doesn't crash, then it is)

```typescript
await keychain.set("x?y", "something");
await keychain.set("x?y", "something");
```

jagpreetsinghsasan: It works for the second time too, so I believe that this method is re-entrant

petermetz (approved):
@elenaizaguirre @jagpreetsinghsasan Same request here as I mentioned in the other PR, please coordinate the merging of your relevant PRs regarding the cross-spec references.

takeutak:
Would you tell me how to position the codes included in this PR for review? (as the same question to #964, #991)
- Is this package required features or optional ones on Cactus?
- How do you intend to use the package on Cactus use-case?

jagpreetsinghsasan:
Hi @takeutak 
> Would you tell me how to position the codes included in this PR for review?  

I didnt understand this statement, can you explain it?


> Is this package required features or optional ones on Cactus?  

This plugin is another keychain plugin, but for cloud vault (like Google secret manager in this case). As keychain plugins are optional to choose from depending upon client requirements, this plugin is also optional.

> How do you intend to use the package on Cactus use-case?

The usage of this plugin is demonstrated in the test case (packages/cactus-plugin-keychain-google-sm/src/test/typescript/integration/plugin-keychain-google-sm.test.ts) (exactly like the vault keychain plugin, just with its own set of input requirements).

takeutak (approved):

takeutak:
@jagpreetsinghsasan Thanks for your kind description. I understood.

petermetz (changes requested):
@jagpreetsinghsasan Please update to all Cactus package version occurrences to 0.6.0 from 0.5.0 (just issued the release) and then run the configure script to also update the lock files as well. Commit all of those changes and then we are ready to merge. Don't forget the squash the commits into a single one.

.cspell.json
```diff
@@ -90,7 +90,13 @@
         "uuidv",
         "vscc",
         "wasm",
-        "Xdai"
+        "Xdai",
```
petermetz, line 93: @jagpreetsinghsasan These look like they are already in the file once (duplicates). Could you double check and if they really are then remove them from the diff.

packages/cactus-plugin-keychain-google-sm/package-lock.json
```diff
@@ -0,0 +1,3040 @@
+{
```
petermetz, line 1: @jagpreetsinghsasan package-lock.json files are now obselete and need to be deleted since the yarn migration (happened today). Please delete this file from the commit (e..g delete the file, then commit that deletion).

packages/cactus-plugin-keychain-google-sm/src/main/typescript/plugin-keychain-google-sm-remote-adapter.ts
```diff
@@ -0,0 +1,155 @@
+import type { Server } from "http";
```
petermetz, line 1: @jagpreetsinghsasan Sorry, /facepalm moment here. We do not need the remote adapter for this package because the implementation runs natively in NodeJS anyway, so it's pointless. The reason why we had the remote adapter for the vault keychain because that one originally was written in Rust and so there it is necessary, but otherwise no need, so this file can be deleted altogether, apologies for not realizing this earlier.

petermetz (changes requested):
@jagpreetsinghsasan Found a few more things last minute, please also apply these changes to the other two keychain plugin PRs in the queue. Sorry for not noticing these little tweaks earlier.

.cspell.json
jagpreetsinghsasan: Xdai, parameterizable were required, rest weren't.

packages/cactus-plugin-keychain-google-sm/package.json
```diff
@@ -0,0 +1,92 @@
...
+  "types": "dist/types/main/typescript/index.d.ts",
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "generate-sdk": "openapi-generator-cli generate -i ./src/main/json/openapi.json -g typescript-axios -o ./src/main/typescript/generated/openapi/typescript-axios/",
+    "codegen:openapi": "npm run generate-sdk",
+    "codegen": "run-p 'codegen:*'",
+    "watch": "npm-watch",
+    "pretsc": "npm run generate-sdk",
```
petermetz, line 19: @jagpreetsinghsasan The `pretsc` script can be safely deleted now with the new build system in place.

petermetz (changes requested):

petermetz (changes requested):
@jagpreetsinghsasan Almost forgot, please also 
1. migrate the tsconfig.json inside your package to match the new boilerplate (see any of the other packages for an example).
2. Add an entry to the root tsconfig.json file under the `references` array that points to the new package that is being added in the PR
3. Do this to all the pending PRs that are adding new packages

petermetz (approved):
@jagpreetsinghsasan LGTM, please resolve conflict(s) by rebasing and then we are good to go.

jagpreetsinghsasan:
Sure, if this is merged after azure-kv, then it will not have merge conflicts, or else, there might be conflict again (I have added azure-kv to the tsconfig.json here as well)

PR #1030, opened 2021-06-14
style(corda): linter fix deploy contract

Leeyoungone:

pinged @petermetz because I don't have review section 😢


petermetz:
@Leeyoungone The semantic PR check is failing because you are missing a space character from the commit message after the `:` (we need to adhere to these strict formats so that we can auto-generate the release notes from the commit messages)

Leeyoungone:
@petermetz re-pinging!

takeutak:
Just to confirm, does this PR includes file changing? There seems to be nothing on the tab of "File changed".

petermetz:
> Just to confirm, does this PR includes file changing? There seems to be nothing on the tab of "File changed".

@takeutak Strange indeed. My best guess is that the same PR was accidentally submitted twice, first one got merged, second one became an empty diff after the rebase.

@Leeyoungone If the above theory is correct then this can be closed as a no-op given that it was already merged. Please let us know and then we can proceed accordingly.

Leeyoungone:
@petermetz I think your theory is correct, I apologize for the confusion. I'll make sure to be more thorough in my other PRs next time. 

@takeutak Thank you for pointing this out, I appreciate it ⭐ !!

petermetz:
@Leeyoungone Absolutely no worries, stuff like this happens all the time. I once accidentally bought a plane ticket twice to the same day and destination, but with different airlines... :-)

PR #1007, opened 2021-06-03
feat(plugin-object-store-ipfs): add IPFS plugin implementation

petermetz:
## Dependencies

Depends on #1032
Depends on #1033 
 
## Commit to review

Author: Peter Somogyvari <peter.somogyvari@accenture.com>
Author Date: Mon Jun 14 2021 12:57:43 GMT-0700 (Pacific Daylight Time)
Committer: Peter Somogyvari <peter.somogyvari@accenture.com>
Committer Date: Mon Jun 14 2021 13:12:33 GMT-0700 (Pacific Daylight Time) 

feat(plugin-object-store-ipfs): add IPFS plugin implementation

Has the ability to get/set or determine presence of key.

Tests verify that it is possible to overwrite the same file
(key) with different content e.g. data is mutable (or at
least appears to be mutable on the file system abstraction
level vs the hashes)

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

PR #716, opened 2021-03-24
docs(roadmap.md): add roadmap with epics for next release #550

petermetz:
## Dependencies

Depends on #656 (only so that the CI actually passes, not a code dependency since this is only a documentation change)

## Commit to review

Author: Peter Somogyvari <peter.somogyvari@accenture.com>
Committer: Peter Somogyvari <peter.somogyvari@accenture.com>
Date: Tue Mar 23 2021 20:41:04 GMT-0700 (Pacific Daylight Time) 

docs(roadmap.md): add roadmap with epics for next release #550

WORK IN PROGRESS
- Epics need to be created in ZenHub still

Fixes #550

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

PR #1032, opened 2021-06-14
feat(core-api): add plugin object store interface definition

petermetz:
This will be implemented by object store plugins so as to
have guarantee that at runtime the plugins can be used the
way the caller expects them to.


takeutak:
@petermetz Just to confirm, what is the intent behind the functionality around this IPFS plugin? (What data are you going to put on distributed storage?)  **related with #1007, #1033

petermetz:
> @petermetz Just to confirm, what is the intent behind the functionality around this IPFS plugin? (What types of data are you going to put on distributed storage?) **related with #1007, #1033

@takeutak Nothing planned at the moment. It will be up to the BLP authors to use it as they see fit, store the kind of data they want as part of their business logic.

takeutak:
@petermetz Thanks to answer! Is this IPFS function required by releasing V1-RC as the roadmap #716 ? If this is not so, I would like to arrange the core functions on the roadmap on #716 and I'm sure that we should concentrate on the development of core functions. (related with #1007, #1033)

petermetz:
@takeutak Cheers! It's not needed for 1.0-RC, but it is needed for something else internally.

takeutak:
@petermetz Thanks for your comment. Would you change the IPFS plugin to IPFS extension feature (an optional feature) since this is not a required one?

takeutak:
@petermetz Thanks for your commit.
As I discussed on the previous, IPFS plugin seems to be suitable to be moved to an extension feature (= optional feature).
The features about ObjectRequest (GetObjectRequestV1, HasObjectRequestV1, etc.) seem to be ones for IPFS plugins. Could you mind if you move these features to some extension?

petermetz:
> @petermetz Thanks for your commit.
> As I discussed on the previous, IPFS plugin seems to be suitable to be moved to an extension feature (= optional feature).
> The features about ObjectRequest (GetObjectRequestV1, HasObjectRequestV1, etc.) seem to be ones for IPFS plugins. Could you mind if you move these features to some extension?

@takeutak I would need to create a brand new package for it. Is that acceptable?

takeutak (approved):

kikoncuo (approved):

PR #1038, opened 2021-06-15
docs(examples): fix typos and backend build

Zzocker:
- Fixed Typos with logLevel
- Fixed fabric chaincode deployment error

### DIFF
--- a/examples/cactus-example-carbon-accounting-backend/package.json
+++ b/examples/cactus-example-carbon-accounting-backend/package.json
@@ -14,6 +14,7 @@
   "bin": "./dist/lib/main/typescript/carbon-accounting-app-cli.js",
   "scripts": {
     "tsc": "tsc --project ./tsconfig.json",
+    "start" : "CONFIG_FILE=./example-config.json node dist/lib/main/typescript/carbon-accounting-app-cli.js",
     "solidity": "hardhat compile",
     "pretsc": "npm run solidity",
     "watch": "npm-watch",
--- a/examples/cactus-example-carbon-accounting-backend/src/utility-emissions-channel/typescript/src/index.ts
+++ b/examples/cactus-example-carbon-accounting-backend/src/utility-emissions-channel/typescript/src/index.ts
@@ -1,9 +1,4 @@
-import {
-  ChaincodeResponse,
-  ChaincodeServerOpts,
-  ChaincodeStub,
-  Shim,
-} from "fabric-shim";
+import { ChaincodeResponse, ChaincodeStub, Shim } from "fabric-shim";
 import { EmissionsRecordInterface } from "./lib/emissions";
 
 import { EmissionsRecordContract } from "./lib/emissionsRecordContract";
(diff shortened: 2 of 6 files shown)


examples/cactus-example-carbon-accounting-backend/src/utility-emissions-channel/typescript/src/index.ts
```diff
@@ -580,16 +580,18 @@ class EmissionsChaincode {
   }
 }
 
-const ccServerOpt: ChaincodeServerOpts = {
-  ccid: process.env.CHAINCODE_CCID,
-  address: process.env.CHAINCODE_ADDRESS,
-  tlsProps: null,
-};
+// const ccServerOpt: ChaincodeServerOpts = {
```
petermetz, line 583: @Zzocker It's okay to just delete code that you'd want to comment out like this (it's in the git history anyway it can be retrieved at will anytime in the future if needed)

examples/cactus-example-carbon-accounting-backend/package.json
```diff
@@ -14,6 +14,8 @@
   "bin": "./dist/lib/main/typescript/carbon-accounting-app-cli.js",
   "scripts": {
     "tsc": "tsc --project ./tsconfig.json",
+    "build:dev" : "npm run tsc || true && cp -r src/utility-emissions-channel dist/lib/",
```
petermetz, line 17: @Zzocker Is this meant to be executed inside the package directory or from the project root through lerna?

petermetz (changes requested):
@Zzocker Looks legit, just has a minor nit and a question that I added inline.

If you amend the commit message the semantic PR check should stop complaining. Recommending this:

`docs(examples): fix typos and backend build`

examples/cactus-example-carbon-accounting-backend/package.json
Zzocker: it should be executed inside package directory.

petermetz: @Zzocker And this works fine? E.g. `cd ./examples/cactus-example-carbon-accounting-backend/ && npm run build:dev` doesn't crash? If yes, that's cool, I'm just wondering, because I would've expected this to not work due to how lerna symlinks the common dependencies.

Zzocker: sorry , I forgot cactus uses lerna for building multiple packages present in cactus. 
For running the carbon accounting backend :
```
npm run build:dev:example-carbon-accounting-backend
npm run start:example-carbon-accounting
```
from project root.

takeutak:
Thanks for contributing. After Peter's change requests, I will review it again.

petermetz (approved):

takeutak (approved):

examples/cactus-example-carbon-accounting-backend/package.json
petermetz: @Zzocker No worries, looks good now with the changes. 
FYI: We have a show-stopper bug on the main branch that is making all the tests fail for all the PRs so it will be a minute before this one is possible to be merged, hang in there and keep an eye out for this one https://github.com/hyperledger/cactus/pull/1073 being merged (after which you can rebase onto upstream/main and your tests should start passing for this PR at the same time)

PR #1048, opened 2021-06-16
feat(besu): block validator endpoint

### LINKED ISSUE #955: feat(besu): sign transaction endpoint with new parameters
We only have endpoints that can sign the data of a transaction they read from the ledger.

**Describe the solution you'd like.**

We need to add a new parameter in the request, that indicates the percentage of valid signatures that are necessary to accept the transaction. Then in the response we need to add new return parameters: the number of valid/invalid signatures, the percentage based ratio of valid signatures and a boolean whether the valid signatures percentage ratio is above or below the value of the new parameter specified by the request.

### Iterate over other Cactus nodes to obtain their signatures

For this there should be two different possibilities.
1. The caller specifies in the request an array of node API hosts (e.g. what you put in the `basePath` parameter of the API client constructor options object)
2. The caller specifies that they want the consortium management plugin to be used for obtaining the list of Cactus nodes in the consortium (all of this is optional because we do not want a hard dependency on the consortium plugin).

For a code example on how to iterate over the a list of nodes and call an API endpoint on each, you can check out the manual consortium management plugin's get consortium JWS method implementation which asks itself for the node list and then iterates over it.

Recommended request parameters to be added in addition to the ones already explained in the second paragraph are:
1. `nodeHosts` string array (optional)
2. `nodeHostsProvider` (enum, mandatory) possible values are `ConsortiumPlugin` and `Request`.

If the `nodeHostsProvider` is set to `Request` then the `nodeHosts` request parameter becomes mandatory meaning that we return with a HTTP 400 BAD_REQUEST code if it's missing or empty.
If the `nodeHostsProvider` is set to `ConsortiumPlugin` then we retrieve the consortium plugin instance from the plugin registry and get the list of node hosts from there.

If the plugin registry has no consortium plugin in it at all, then we return with a HTTP 400 BAD_REQUEST explaining in the response that there is no consortium management plugin installed in the Cactus node so this is not a viable option for node host providers and the user (caller) must provide the node hosts manually.

### Signature validation

Use the `jose` npm dependency to validate the signatures.

An example of how that works is in the config-service.ts file

packages/cactus-cmd-api-server/src/main/typescript/config/config-service.ts

```typescript
 /**
   * Validation that prevents operators from mistakenly deploying a key pair
   * that they may not be operational for whatever reason.
   *
   * @throws If a dummy sign+verification operation fails for any reason.
   */
  validateKeyPairMatch(): void {
    const fnTag = "ConfigService#validateKeyPairMatch()";
    // FIXME most of this lowever level crypto code should be in a commons package that's universal
    const keyPairPem = ConfigService.config.get("keyPairPem");
    const keyPair = JWK.asKey(keyPairPem);

    const jws = JWS.sign({ hello: "world" }, keyPair);
    try {
      JWS.verify(jws, keyPair);
    } catch (ex) {
      throw new Error(`${fnTag} Invalid key pair PEM: ${ex && ex.stack}`);
    }
  }
```

AzaharaC:
Resolve #955

### DIFF
--- a/packages/cactus-plugin-ledger-connector-besu/package.json
+++ b/packages/cactus-plugin-ledger-connector-besu/package.json
@@ -69,9 +69,11 @@
   },
   "homepage": "https://github.com/hyperledger/cactus#readme",
   "dependencies": {
+    "@hyperledger/cactus-cmd-api-server": "1.0.0-rc.3",
     "@hyperledger/cactus-common": "1.0.0-rc.3",
     "@hyperledger/cactus-core": "1.0.0-rc.3",
     "@hyperledger/cactus-core-api": "1.0.0-rc.3",
+    "@hyperledger/cactus-plugin-consortium-manual": "1.0.0-rc.3",
     "axios": "0.21.4",
     "express": "4.17.1",
     "joi": "17.4.2",
@@ -85,7 +87,9 @@
     "web3-core": "1.5.2",
     "web3-eea": "0.11.0",
     "web3-eth": "1.5.2",
-    "web3-utils": "1.5.2"
+    "web3-utils": "1.5.2",
+    "jose": "4.1.0",
+    "key-encoder": "2.0.3"
   },
   "devDependencies": {
     "@hyperledger/cactus-plugin-keychain-memory": "1.0.0-rc.3",
--- a/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
+++ b/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
@@ -1,5 +1,6 @@
 import { Server } from "http";
 import { Server as SecureServer } from "https";
+import { JWK, JWS } from "jose";
 
 import type { Server as SocketIoServer } from "socket.io";
 import type { Socket as SocketIoSocket } from "socket.io";
@@ -31,6 +32,7 @@ import {
   IPluginWebService,
   ICactusPlugin,
   ICactusPluginOptions,
+  //  IPluginConsortium,
 } from "@hyperledger/cactus-core-api";
 
 import {
@@ -41,10 +43,6 @@ import {
 import {
   Checks,
   CodedError,
-  IJsObjectSignerOptions,
-  JsObjectSigner,
-  KeyConverter,
-  KeyFormat,
   Logger,
   LoggerProvider,
   LogLevelDesc,
@@ -64,6 +62,7 @@ import {
   RunTransactionRequest,
   RunTransactionResponse,
   SignTransactionRequest,
+  GetNodeSignatureRequest,
   SignTransactionResponse,
   Web3SigningCredentialCactusKeychainRef,
   Web3SigningCredentialPrivateKeyHex,
@@ -74,11 +73,16 @@ import {
   GetBlockV1Response,
   GetBesuRecordV1Request,
   GetBesuRecordV1Response,
+  NodeHostsProviderType,
+  Configuration,
+  DefaultApi,
 } from "./generated/openapi/typescript-axios";
 
 import { InvokeContractEndpoint } from "./web-services/invoke-contract-endpoint";
 import { isWeb3SigningCredentialNone } from "./model-type-guards";
 import { BesuSignTransactionEndpointV1 } from "./web-services/sign-transaction-endpoint-v1";
+import { BesuGetNodeSignatureV1 } from "./web-services/get-node-signature-v1-endpoint";
+
 import { PrometheusExporter } from "./prometheus-exporter/prometheus-exporter";
 import {
   GetPrometheusExporterMetricsEndpointV1,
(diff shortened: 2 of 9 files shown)


packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -29,10 +30,10 @@ import {
 import {
   Checks,
   CodedError,
-  IJsObjectSignerOptions,
-  JsObjectSigner,
-  KeyConverter,
-  KeyFormat,
+  //  IJsObjectSignerOptions,
```
petermetz, line 33: @AzaharaC If not used, please delete the commented code

packages/cactus-plugin-ledger-connector-besu/package.json
```diff
@@ -86,6 +86,7 @@
     "@hyperledger/cactus-common": "0.5.0",
     "@hyperledger/cactus-core": "0.5.0",
     "@hyperledger/cactus-core-api": "0.5.0",
+    "@hyperledger/cactus-plugin-consortium-manual": "0.5.0",
```
petermetz, line 89: @AzaharaC We cannot depend on a specific implementation of the consortium plugin. Please use the `IPluginConsortium` interface instead when looking up the consortium plugin instance in the plugin registry and remove the manual consortium plugin from the list of dependencies.

packages/cactus-test-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-validator-besu/obtain-signatures.test.ts
```diff
@@ -0,0 +1,417 @@
...
+    t.comment(`AddressInfo: ${JSON.stringify(addressInfoC1)}`);
+
+    const configuration = new ConfigurationConsortium({ basePath: nodeC1Host });
+    const api = new DefaultApiConsortium(configuration);
+    const res = await api.getNodeJws();
+    t.ok(res, "API response object is truthy");
+  }
+  t.comment(`Set up first node OK`);
+
+  // 2. Instantiate plugin registry which will provide the web service plugin with the key value storage plugin
```
petermetz, line 288: @AzaharaC It's cool to remove these if you didn't need them for the test.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -672,22 +696,103 @@ export class PluginLedgerConnectorBesu
...
-    const pkRaw = converter.privateKeyAs(pem, KeyFormat.PEM, KeyFormat.Raw);
+      if (consortium === undefined) {
+        throw new Error(`${fnTag} Consortium plugin not found`);
+      }
+
+      const keyPairPem = consortium.options.keyPairPem;
+      const totalNodes = consortium.getNodeCount();
+
+      //Get consortium nodes
+      const nodesAll = consortium.options.consortiumRepo.allNodes;
```
petermetz, line 712: @AzaharaC Please use the get consortium jws endpoint of the consortium plugin to obtain the list of nodes.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -672,22 +696,103 @@ export class PluginLedgerConnectorBesu
       throw new CodedError(msg, E_KEYCHAIN_NOT_FOUND);
     }
 
-    const pem: string = await keychain.get(keychainRef);
+    if (nodeHostsProvider === NodeHostsProviderType.ConsortiumPlugin) {
```
petermetz, line 699: @AzaharaC Please also include the other NodeHostsProviderType implementation

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-besu/package.json
AzaharaC: To do this, I have changed the code for this:
```
if (nodeHostsProvider === NodeHostsProviderType.ConsortiumPlugin) {
      const consortium = this.pluginRegistry.plugins.find(
        (plugin) => plugin.getInstanceId() == consortiumPluginId,
      ) as IPluginConsortium<unknown, unknown, unknown, unknown>;
```
Then I have this line: ```const nodesAll = await consortium.getConsortiumJws();``` When I want to use the constant ```nodesAll ``` it gives me an error saying that it is unknown type and I can't access its properties.

To use this variable I think I would have to do something like this: ```const nodesAll: GetConsortiumJwsOut = await consortium.getConsortiumJws();```, but I would need to import ```GetConsortiumJwsOut``` from the consortium plugin and we would return to the initial dependency problem

petermetz: @AzaharaC You can cast it to `any` initially until a better solution is available (you'll lose auto-completion/type info but it will compile)

petermetz:
@AzaharaC Are you still working on this?

petermetz:
@AzaharaC No worries, the Corda 5.x work that you are doing is higher priority than this so we can leave this until after that too.

petermetz:
@AzaharaC Closing this down for now just to tidy up the PR queue a bit. We can re-open it anytime later.

PR #1056, opened 2021-06-17
feat(odap) odap implement endpoints, plugin, node

jscode017:
@petermetz , @RafaelAPB, @ruychi
In the packages/cactus-odap directory
I defined enpoints and message format at src/json/openapi.json,
created  endpoints in src/typescript/web-services
created gateway methods in src/typescript/gateway

And try to understand the supply-chain-example source code, and modify it
But I would want to know whether I built the right thing in followings:
 src/typescript/business-logic-plugin: develop odap plugin
src/typescript/backend: added odap plugin in a plugin registry and create an API server(I am not sure I build this right)

And I would try to read the source code, understand how to call plugins in API server
@RafaelAPB , I wondered where I could use the adapter pattern, is there any point I miss?


jscode017:
Update:
Is this a way to call endpoints by plugin?
`import { OdapGateWayNode } from "../backend/odap-backend";
import { Checks } from "@hyperledger/cactus-common";
interface IOdapGateWayExampleOptions{
  port: number;
}

export class OdapGateWayExample {
  public static readonly CLASS_NAME = "OdapExample";
  public get className(): string {
    return OdapGateWayExample.CLASS_NAME;
  }
  private readonly port: number;
  constructor(public readonly options: IOdapGateWayExampleOptions) {
    const fnTag = `${this.className}#constructor()`;

    Checks.truthy(options, `${fnTag} arg options`);
    Checks.nonBlankString(options.port, `${fnTag} options.port`);
    this.port = options.port;
  }
  public async start() {
    const odapGateWayNode = new OdapGateWayNode({});
    
    const startResultA = await odapGateWayNode.start(4000);
    const exampleApiServerA = startResultA.apiServerA;
    const exampleOdapGateWayApiClientA = startResultA.OdapGateWayApiClientA;

    const pluginRegistry1 = await exampleApiServer1.getOrInitPluginRegistry();
    const odapGatewayPlugin = pluginRegistry1.findOneByPackageName(
      "@hyperledger/cactus-odap-odap-gateway-business-logic-plugin",
    );

    const odapGateWayWebservices = odapGatewayPlugin.getOrCreateWebServices();

    Checks.truthy(odapGateWayWebservices);
    /*
    And call methods/endpoints by this webservices
    */
}
}
`

petermetz:
@jscode017 

> how could I use the plugin? 

You can cast it the actual ODAP plugin type that has all the methods declared on top of what `ICactusPlugin` has.

>  do I call the endpoints by exampleOdapGateWayApiClientA?

Yep, that's correct!

> I wondered where I could use the adapter pattern, is there any point I miss?

This an example of how we use it when certain plugins are implemented in programming languages that are not transpiled to JS (e.g. NodeJS cannot load them as modules)

packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-factory-keychain.ts
packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault-remote-adapter.ts

PR #1073, opened 2021-06-20
test(connector-fabric): fix v1.4.8 golang chaincode test

petermetz:
Pinned all the hello world chaincode's dependencies to exact versions
that are known to have worked in the past for builds. We needed this
because the 1.13 go version of the Fabric CLI container could no
longer build (tidy+vendor) the dependencies that were built with the
newer versions of go such as 1.16 and 1.17.

More info about the root cause can be read here:
https://github.com/golang/go/issues/40067

By locking the dependencies versions to older ones we we ensured that
the older go installation of the Fabric CLI container does not fail.

The known-to-be-working dependency versions are coming from
packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/fixtures/go/hello-world-contract-fabric-v14/hello-world-contract-go-mod.ts


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1077, opened 2021-06-21
fix(xdai-connector) invokeContract for eth-method with sender's role …

Zzocker:
- web3.eth.estimateGas : works considering called solidity method do not throw an exception. So, for method having `modifier` with access control on `msg.sender` calling estimateGas without `from` field throws error.to make it work ,`transactionConfig.from = web3SigningCredential.ethAccount ` before calling `estimateGas`


petermetz (approved):

petermetz:
@Zzocker Please also include the information in the PR description in the commit message itself

Zzocker:
> @Zzocker Please also include the information in the PR description in the commit message itself

@petermetz Do you want me to enrich the commit message with more information about the PR?

petermetz:
> > @Zzocker Please also include the information in the PR description in the commit message itself
> 
> @petermetz Do you want me to enrich the commit message with more information about the PR?

@Zzocker Yup, what you put in the PR description is good IMO, I'd just copy paste that into the commit message to make sure it's not lost if we migrate from github to some other platform (which is unlikely, but not impossible)

takeutak (approved):

Issue #1078, opened 2021-06-21
feat(connector-xdai): add interval to pollForTxReceipt


Zzocker:
**Is your feature request related to a problem? Please describe.**
While waiting for transaction receipt for a given txHash (contract running ropsten test network, node connected via infura ) I saw more then  500 eth_getTransactionReceipt calls in my infura dashboard. 

**Describe the solution you'd like**
Rather then doing a while loop ([here](https://github.com/hyperledger/cactus/blob/526c4fb122e519a60d742b58e48003b06b9522b4/packages/cactus-plugin-ledger-connector-xdai/src/main/typescript/plugin-ledger-connector-xdai.ts#L488)) continuously,it should loop at given interval. this interval can be passed as a field to `ConsistencyStrategy`.

**Describe alternatives you've considered**
Implemented above described using a new instance of Web3

petermetz:
@Zzocker FYI you can link together (for future reference) a PR of yours with a corresponding issue so that the issue automatically gets closed as soon as the PR gets merged.
The way to do it is to use the `Fixes #ISSUE_NUMBER` syntax. More info here:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #1081, opened 2021-06-22
feat(connector-xdai): add interval to pollForTxReceipt

Zzocker:
#1078 


packages/cactus-plugin-ledger-connector-xdai/src/main/json/openapi.json
```diff
@@ -55,6 +55,11 @@
                         "type": "integer",
                         "minimum": 0,
                         "description": "The number of blocks to wait to be confirmed in addition to the block containing the transaction in question. Note that if the receipt type is set to only wait for node transaction pool ACK and this parameter is set to anything, but zero then the API will not accept the request due to conflicting parameters."
+                    },
+                    "callInterval" : {
```
petermetz, line 59: @Zzocker Please rename this to `pollIntervalMs` (ms for millis)

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-xdai/src/main/typescript/plugin-ledger-connector-xdai.ts
```diff
@@ -478,34 +478,38 @@ export class PluginLedgerConnectorXdai
...
-  }
+    return new Promise<TransactionReceipt>(async( resolve: (value: TransactionReceipt | PromiseLike<TransactionReceipt>) => void, reject: (reason?: any) => void)=>{
+      if (consistencyStrategy.receiptType === ReceiptType.NodeTxPoolAck){
+        consistencyStrategy.blockConfirmations = 0
+      }  
+      const intervalMs = consistencyStrategy.callInterval
+        const timeoutMs = consistencyStrategy.timeoutMs || Number.MAX_SAFE_INTEGER
+        const timeoutAt = (new Date()).getTime() + timeoutMs
+        let tries = 0
+        const interval = setInterval(async ()=>{
```
petermetz, line 489: @jagpreetsinghsasan I recommend using a one-liner to implement the wait instead of an explicit setInterval call wrapping everything.
It ends up being the same under the hood but this way the change becomes much smaller, easier to read.
The one liner I usually do to "sleep" goes like this
```typescript
await new Promise((resolve) => setInterval(resolve, pollIntervalMs));
```

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-xdai/src/main/typescript/plugin-ledger-connector-xdai.ts
Zzocker: something like this ?

```typescript
  public async pollForTxReceipt(
    txHash: string,
    consistencyStrategy: ConsistencyStrategy,
  ): Promise<TransactionReceipt> {
    const fnTag = `${this.className}#pollForTxReceipt()`;
    const intervalMs = consistencyStrategy.callInterval
    let receipt:TransactionReceipt
    try {
      receipt = await new Promise<TransactionReceipt>((resolve: (value: TransactionReceipt ) => void, reject: (reason?: any) => void)=>{
          setInterval(async()=>{
            // fetch tx Receipt
            // check block confirmations
            // check timeout, if timed out call reject
          },intervalMs)
      })
    } catch (error) {
      throw error
    }
    return receipt
}
```

petermetz: @Zzocker No, you'd just put in the one-liner to sleep in the body of while loop

petermetz (approved):

takeutak (approved):

PR #1084, opened 2021-06-23
feat(test-tooling): faio features and improvements

RafaelAPB:
Adds several improvements:
1) destroys faio volume when tests end
2) allows to recover a test ledger (setContainer)
3) allows to add organizations

Replaces PR #949 cc @petermetz 


RafaelAPB:
It looks like the CI fails because it cannot find an image: hyperledger/cactus-fabric-all-in-one:2021-04-21-2016750 ; I believe this happens within the Corda connector tests. @petermetz do you have any input on this?

petermetz:
> It looks like the CI fails because it cannot find an image: hyperledger/cactus-fabric-all-in-one:2021-04-21-2016750 ; I believe this happens within the Corda connector tests. @petermetz do you have any input on this?

@RafaelAPB Checking into it, not yet sure if this was a random flake or a legit failure. Re-trying now, will know more once it concludes.

packages/cactus-test-tooling/package.json
```diff
@@ -82,6 +82,7 @@
   "homepage": "https://github.com/hyperledger/cactus#readme",
   "dependencies": {
     "@hyperledger/cactus-common": "0.5.0",
+    "@types/js-yaml": "^4.0.1",
```
petermetz, line 85: @RafaelAPB Please lock down the dependency versions to exact ones.

packages/cactus-test-tooling/package.json
```diff
@@ -91,11 +92,13 @@
     "internal-ip": "6.2.0",
     "is-port-reachable": "3.0.0",
     "joi": "14.3.1",
+    "js-yaml": "4.1.0",
     "keycloak-admin": "1.14.11",
     "node-ssh": "11.1.1",
     "p-retry": "4.4.0",
     "run-time-error": "1.4.0",
     "tar-stream": "2.1.2",
+    "temp": "^0.9.4",
```
petermetz, line 101: @RafaelAPB Please lock down the dependency versions to exact ones.

packages/cactus-test-tooling/package.json
```diff
@@ -107,6 +110,7 @@
     "@types/node-ssh": "7.0.0",
     "@types/ssh2": "0.5.46",
     "@types/ssh2-streams": "0.1.8",
-    "@types/tar-stream": "2.1.0"
+    "@types/tar-stream": "2.1.0",
+    "@types/temp": "^0.9.0"
```
petermetz, line 114: @RafaelAPB Please lock down the dependency versions to exact ones.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/add-orgs.ts
```diff
@@ -0,0 +1,106 @@
...
+  FabricTestLedgerV1,
+  pruneDockerAllIfGithubAction,
+} from "@hyperledger/cactus-test-tooling";
+
+import { LogLevelDesc } from "@hyperledger/cactus-common";
+import path from "path";
+import { STATE_DATABASE } from "@hyperledger/cactus-test-tooling";
+import { LedgerStartOptions } from "../../../../../../cactus-test-tooling/dist/types/main/typescript/fabric/fabric-test-ledger-v1";
+
+const testCase = "obtains configuration profiles from Fabrix 2.x ledger";
```
petermetz, line 13: @RafaelAPB Could this be merged with the other test case?

petermetz (changes requested):
@RafaelAPB Could you move the contents of the
`packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/fixtures/typescript/addOrgX/`
directory to
`packages/cactus-plugin-ledger-connector-fabric/src/test/resources/fixtures/addOrgX/`

Having shell scripts/yaml/json/etc in the typescript directory might cause problems later on with the compiler.

RafaelAPB:
@petermetz done, thanks for the comments.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/add-orgs.ts
```diff
@@ -0,0 +1,106 @@
+import test, { Test } from "tape-promise/tape";
```
petermetz, line 1: @RafaelAPB Apologies I just spotted this, please add a `.test` suffix to the filename e.g. `add-org.test.ts`

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/obtain-profiles.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  const connectionProfileOrg1 = await ledger.getConnectionProfileOrgX("1");
+  t.isEquivalent(connectionProfile, connectionProfileOrg1);
+
+  const connectionProfileOrg2 = await ledger.getConnectionProfileOrgX("2");
+  t.ok(connectionProfileOrg2, "getConnectionProfileOrg2() out truthy OK");
+
+  t.ok(
+    connectionProfileOrg1 !== connectionProfileOrg2,
+    "getConnectionProfileOrg2() out truthy OK",
+  );
```
petermetz, line 51: @RafaelAPB The assertion message is a bit misleading here because we are asserting strict non-equality between two variables, but the message claims that it's about the truthiness of the second variable.

```suggestion
  t.doesNotEqual(
    connectionProfileOrg1,
    connectionProfileOrg2,
    "connectionProfileOrg1 !== connectionProfileOrg2 OK",
  );
```

packages/cactus-test-tooling/package.json
```diff
@@ -82,6 +82,7 @@
   "homepage": "https://github.com/hyperledger/cactus#readme",
   "dependencies": {
     "@hyperledger/cactus-common": "0.5.0",
+    "@types/js-yaml": "4.0.1",
```
petermetz, line 85: @RafaelAPB Recommend moving this to the `devDependencies` section because the type information package is not needed at runtime (because Typescript at runtime is just Javascript)

```suggestion
```

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -468,12 +1375,29 @@ export class FabricTestLedgerV1 implements ITestLedger {
...
+          if (volume.Name) {
+            const volumeToDelete = docker.getVolume(volume.Name);
+            await volumeToDelete.remove();
+          } else {
+            this.log.debug("Volume", volume, "could not be removed");
+          }
+        });
+      } else {
+        throw new Error(`${fnTag} Container not found, nothing to destroy.`);
+      }
```
petermetz, line 1397: @RafaelAPB The indentations can be reduced here with equivalent logic being in place (which is good because it reduces the chance of arrow programming [1] occurring. Recommended change below is to throw if `this.container` is falsy and then there's no need to have the rest of the logic on an explicit `else` branch.

[1]: https://blog.codinghorror.com/flattening-arrow-code/

```suggestion
  if (!this.container) {
    throw new Error(`${fnTag} Container not found, nothing to destroy.`);
  }
  const docker = new Dockerode();
  const containerInfo = await this.container.inspect();
  const volumes = containerInfo.Mounts;
  await this.container.remove();
  volumes.forEach(async (volume) => {
    this.log.debug("Removing volume: ", volume);
    if (volume.Name) {
      const volumeToDelete = docker.getVolume(volume.Name);
      await volumeToDelete.remove();
    } else {
      this.log.debug("Volume", volume, "could not be removed");
    }
  });
```

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -323,6 +370,815 @@ export class FabricTestLedgerV1 implements ITestLedger {
...
+        this.extraOrgs[i].orgName,
+        this.extraOrgs[i].orgChannel,
+        this.extraOrgs[i].certificateAuthority,
+        this.extraOrgs[i].stateDatabase,
+        this.extraOrgs[i].port,
+      );
+    }
+  }
+  // req: AddOrganizationFabricV2Request
+  // returns promise <AddOrganizationFabricV2Response>
```
petermetz, line 961: @RafaelAPB The comments do not appear to match the implementation (mismatch in parameters and return type as well). 
I recommend removing the comment to avoid confusing people who would only read the comments through the IDEs documentation viewer feature. Alternatively the documentation could be made up to date but I do not want to put undue extra burden on you with too much nitpicking (I'm doing enough of that already)

```suggestion
```

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -323,6 +370,815 @@ export class FabricTestLedgerV1 implements ITestLedger {
...
+      /*
+      const tmpDirPrefix = `hyperledger-cactus-${this.className}`;
+      const tmpDirPath = temp.mkdirSync(tmpDirPrefix);
+      const tmpDirPathDocker = temp.mkdirSync(
+        path.join(tmpDirPrefix, "docker"),
+      );
+    */
+
+      //
+
```
petermetz, line 1017: @RafaelAPB I recommend dropping the commented code if no longer needed. (assumption is that this commented code was a dead end during the implementation phase that was reasonable to be kept handy while it was uncertain whether it would be needed or not, but if in the end it turned out that it wasn't, then I'd prefer if it just went away to make the rest of the code around it easier to read). 
It is up to your discretion to decide if this is indeed the case or if there's real value in keeping the comment around for future reference (and if you say that that is the case then I will happily stand down of course)

```suggestion
```

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -323,6 +370,815 @@ export class FabricTestLedgerV1 implements ITestLedger {
...
+      /*
+      for (const sourceFile of sourceFiles) {
+        const { filename, filepath, body } = sourceFile;
+        const relativePath = filepath || "./";
+        const subDirPath = path.join(tmpDirPath, relativePath);
+        fs.mkdirSync(subDirPath, { recursive: true });
+        const localFilePath = path.join(subDirPath, filename);
+        fs.writeFileSync(localFilePath, body, "base64");
+      }
+      */
```
petermetz, line 1085: @RafaelAPB Recommending to remove the comments if they are just dead-ends (same considerations apply as above)

```suggestion
```

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -323,6 +370,815 @@ export class FabricTestLedgerV1 implements ITestLedger {
...
+        CA_PORT: caPort,
+        CONFIG_TX_GEN_PATH: path.join(remoteDirPath),
+        COMPOSE_FILE_COUCH: path.join(
+          remoteDirPath,
+          "docker",
+          couchdbFile.filename,
+        ),
+        COMPOSE_FILE: path.join(remoteDirPath, "docker", composeFile.filename),
+        COMPOSE_FILEe_CA: path.join(remoteDirPath, "docker", caFile.filename),
+        //COMPOSE_PROJECT_NAME: "cactusfabrictestnetwork_test",
```
petermetz, line 1118: @RafaelAPB Recommending to remove the comments if they are just dead-ends (same considerations apply as above)

```suggestion
```

petermetz (changes requested):

RafaelAPB:
@petermetz  done, thanks once again.

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-test-ledger-v1.ts
```diff
@@ -323,6 +370,795 @@ export class FabricTestLedgerV1 implements ITestLedger {
...
+
+  public async addExtraOrgs(): Promise<void> {
+    const fnTag = `FabricTestLedger#addOrgX()`;
+    const { log } = this;
+    if (!this.extraOrgs) {
+      throw new Error(`${fnTag}: there are no extra orgs`);
+    }
+
+    log.debug(`
+    Adding ${this.extraOrgs.length} orgs`);
```
petermetz, line 948: @RafaelAPB I recommend not putting a new line character inside the logging statement, makes it easier to read later on if all log messages are one line each unless it's an object dump for example.

petermetz (changes requested):

petermetz:
> @petermetz done, thanks once again.

@RafaelAPB No worries. I did see a few change requests that you resolved but the code remained the same, could you please double check if you pushed everything to the remote?

RafaelAPB:
@petermetz double checked, everything was pushed.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/add-orgs.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+import {
+  FabricTestLedgerV1,
+  pruneDockerAllIfGithubAction,
+} from "@hyperledger/cactus-test-tooling";
+
+import { LogLevelDesc } from "@hyperledger/cactus-common";
+import path from "path";
+import { STATE_DATABASE } from "@hyperledger/cactus-test-tooling";
+import { LedgerStartOptions } from "../../../../../../cactus-test-tooling/dist/types/main/typescript/fabric/fabric-test-ledger-v1";
```
petermetz, line 11: @RafaelAPB Need to import via the official package name otherwise this will be broken in production deployments, so instead of the relative path you need to import ` from "@hyperledger/cactus-test-tooling";`

And if it doesn't compile that way it must be because you forgot to `export` in the `public-api.ts` file of the test tooling package.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/add-orgs.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    "connection profile from org1 is different from the one from org2 OK",
+  );
+
+  //Should return error, as there is no org101 in the default deployment of Fabric AIO image nor it was added
+  const error = "/Error.*/";
+
+  await t.rejects(
+    async () => await ledger.getConnectionProfileOrgX("org101"),
+    error,
+  );
```
petermetz, line 87: @RafaelAPB  Can be made simpler (I think) by applying this change below:

```suggestion
  await t.rejects(ledger.getConnectionProfileOrgX("org101"), error);
```

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/add-orgs.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  //Let us add org 3 and retrieve the connection profile
+
+  const tearDown = async () => {
+    await ledger.stop();
+    await ledger.destroy();
+  };
+
+  test.onFinish(tearDown);
```
petermetz, line 96: @RafaelAPB Sorry for just catching this: The teardown logic should be wired up right after the ledger variable has been created. This makes sure that if the ledger.start() method throws the onFinish will have hooked up the teardown handler by that point. Otherwise the container can hang around and make following test tries fail within the same CI execution.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/obtain-profiles.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    envVars: new Map([["FABRIC_VERSION", "2.2.0"]]),
+  });
+  await ledger.start();
+
+  const tearDown = async () => {
+    await ledger.stop();
+    await ledger.destroy();
+  };
+
+  test.onFinish(tearDown);
```
petermetz, line 36: @RafaelAPB Same here as above, nowadays I ask everyone to have the teardown handlers registered before calling the ledger start method for the reason explained in my previous comment.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/obtain-profiles.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    connectionProfileOrg2,
+    "connection profiles are different",
+  );
+
+  //Should return error, as there is no Org3 in the default deployment of Fabric AIO image
+  const error = "/Error.*/";
+  await t.rejects(
+    async () => await ledger.getConnectionProfileOrgX("3"),
+    error,
+  );
```
petermetz, line 59: @RafaelAPB Could be simplified as

```suggestion
await t.rejects(ledger.getConnectionProfileOrgX("3"), error);
```

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/add-orgs.test.ts
RafaelAPB: Thanks Peter, very insightful!

petermetz (approved):

takeutak (approved):

RafaelAPB:
@petermetz rebased, hopefully it will work now :)

petermetz:
@RafaelAPB Thank you for all the effort and especially for putting up with my nit-picking!
:+1: :+1: :+1: :+1: :+1:

Issue #959, opened 2021-05-10
feat(LedgerPlugin): add a proto code of LedgerPlugin for Hyperledger Indy


takeutak:
**Is your feature request related to a problem? Please describe.**
When we integrate with blockchain identity application, Cactus needs connectivity to Hyperledger Indy.

**Describe the solution you'd like**
To add LedgerPlugin for Hyperledger Indy

PR #1099, opened 2021-06-28
feat(validator): Indy validator and indy-testnet files

izuru0:
Added new code that talks to indy pool. It submits request to indy it received and sends response back.
The python code will be updated in the future so that users can simply pip install them and run (although it is incomplete at the moment). The directory structure reflects this intention.

Also included are updated version of indy-testnet files. The code described above needs this environment to properly run.


takeutak:
resolve the issue #959

takeutak (approved):

petermetz (changes requested):
@izuru0 Thank you for the contribution! Could you add a remote adapter for this in Typescript so that the API server can pull it in as a plugin as well?

takeutak:
@petermetz Thanks for your comment. As you commented, the current main language on Cactus is TypeScript so your suggestion seems natural to terms of the current codes. However, as we discussed, in order to start collaboration with other projects, it is needed to adapt not only TypeScript but also other programming languages such as python.

Then, I think that it should be dealt as another issue to adapt TypeScript package, but this issue is not directly related with this pull-request. Could you review this pull-request again?

izuru0:
@petermetz Thank you for your comment.

It is nice to have interoperability between API server and connector (possibly not just for Indy but other ledgers too).

I think we can design a wrapper to make indy validator (in python) to be callable, but I suspect that it could take some time to design a standard interface between API server in typescript and (arbitrary) connector in python.

For that reason, can I ask this code to be accepted for the time being, and consider the designing/implementing such interface as other task?

petermetz:
> @petermetz Thank you for your comment.
> 
> It is nice to have interoperability between API server and connector (possibly not just for Indy but other ledgers too).
> 
> I think we can design a wrapper to make indy validator (in python) to be callable, but I suspect that it could take some time to design a standard interface between API server in typescript and (arbitrary) connector in python.
> 
> For that reason, can I ask this code to be accepted for the time being, and consider the designing/implementing such interface as other task?

@izuru0 Sure thing.

petermetz (changes requested):
@izuru0 Please squash into a single commit, then it LGTM.
Let me know if you need help with the git mechanics, happy to help.

takeutak:
@izuru0 It seems that he created a new pull request #1117 having the same codes instead of squashing. So I think it is no problem to close this PR #1099. Is it OK?

@petermetz So would you mind if you review the new PR #1117 again?

petermetz:
> @izuru0 It seems that he created a new pull request #1117 having the same codes instead of squashing. So I think it is no problem to close this PR #1099. Is it OK?
> 
> @petermetz So would you mind if you review the new PR #1117 again?

@takeutak @izuru0 Any chance that you could just do the squash instead of me reviewing the entire PR again? It takes 30 seconds to squash, if you don't know how to do it I can help.

petermetz (approved):

PR #1102, opened 2021-06-29
feat(besu): record locator

### LINKED ISSUE #365: feat(besu): define data ledger record locator model type

Depends on: #364

There is no well defined scheme (OpenAPI spec types) in the besu connector plugin for referencing any piece of data on a besu ledger.

Define a type called `BesuRecordLocator` in the `openapi-spec.ts` file of the besu connector that is capable of describing/uniquely identifying any piece of data (aka record) on a besu ledger so that we can then re-use this type across different endpoint definitions for the besu connector where it is needed.

1. The type defined should be aware of the existing consortium definition meaning that it is safe to assume that a `ledgerId` property will uniquely identify the ledger and encode enough information for the ledger to be reached via Cactus connector plugins or the API client.
2. The `BesuRecordLocator` type must be capable of referencing a particular piece of state **within** a contract
3. The `BesuRecordLocator` type must be capable of referencing a particular contract's `CALL`'s return data.

Been looking into what others have concluded on this topic so far (not a new problem) and there are some good schemes, but it would be best for us to support the consortium definition we have in-house already meaning that we do not need to store in the record locator the network host of a ledger for example, nor the credentials required to access it because all of that is taken care of in different components of the system: see the paragraph above.

**Additional context**

Consortium definition is about to be refactored as well => #349 

AzaharaC:
Resolve #365

### DIFF
--- a/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
+++ b/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
@@ -68,6 +68,8 @@ import {
   GetTransactionV1Response,
   GetBlockV1Request,
   GetBlockV1Response,
+  GetBesuRecordV1Request,
+  GetBesuRecordV1Response,
 } from "./generated/openapi/typescript-axios";
 
 import { InvokeContractEndpoint } from "./web-services/invoke-contract-endpoint";
@@ -84,6 +86,7 @@ import { GetTransactionEndpoint } from "./web-services/get-transaction-endpoint"
 import { GetPastLogsEndpoint } from "./web-services/get-past-logs-endpoint";
 import { RunTransactionEndpoint } from "./web-services/run-transaction-endpoint";
 import { GetBlockEndpoint } from "./web-services/get-block-v1-endpoint-";
+import { GetBesuRecordEndpointV1 } from "./web-services/get-besu-record-endpoint-v1";
 
 export const E_KEYCHAIN_NOT_FOUND = "cactus.connector.besu.keychain_not_found";
 
@@ -259,6 +262,13 @@ export class PluginLedgerConnectorBesu
     }
     {
       const endpoint = new BesuSignTransactionEndpointV1({
+        connector: this,
+        logLevel: this.options.logLevel,
+      });
+      endpoints.push(endpoint);
+    }
+    {
+      const endpoint = new GetBesuRecordEndpointV1({
         connector: this,
         logLevel: this.options.logLevel,
       });
@@ -773,5 +783,57 @@ export class PluginLedgerConnectorBesu
   ): Promise<GetBlockV1Response> {
     const block = await this.web3.eth.getBlock(request.blockHashOrBlockNumber);
     return { block };
+  }
+  public async getBesuRecord(
+    request: GetBesuRecordV1Request,
+  ): Promise<GetBesuRecordV1Response> {
+    const fnTag = `${this.className}#getBesuRecord()`;
+    //////////////////////////////////////////////
+    let abi: any = undefined;
+    const resp: GetBesuRecordV1Response = {};
+    const txHash = request.transactionHash;
+
+    if (txHash) {
+      const transaction = await this.web3.eth.getTransaction(txHash);
+      if (transaction.input) {
+        resp.transactionInputData = transaction.input;
+        return resp;
+      }
+    }
+
+    if (request.invokeCall) {
+      if (request.invokeCall.contractAbi) {
+        if (typeof request.invokeCall.contractAbi === "string") {
+          abi = JSON.parse(request.invokeCall.contractAbi);
+        } else {
+          abi = request.invokeCall.contractAbi;
+        }
+      }
+      const { contractAddress } = request.invokeCall;
+      const contractInstance = new this.web3.eth.Contract(abi, contractAddress);
+      const methodRef = contractInstance.methods[request.invokeCall.methodName];
+      Checks.truthy(
+        methodRef,
+        `${fnTag} YourContract.${request.invokeCall.methodName}`,
+      );
+      const method: ContractSendMethod = methodRef(
+        ...request.invokeCall.params,
+      );
+
+      if (
+        request.invokeCall.invocationType === EthContractInvocationType.Call
+      ) {
+        const callOutput = await (method as any).call();
+        const res: GetBesuRecordV1Response = {
+          callOutput,
+        };
+        return res;
+      } else {
+        throw new Error(
+          `${fnTag} Unsupported invocation type ${request.invokeCall.invocationType}`,
+        );
+      }
+    }
+    return resp;
   }
 }
--- a/packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/get-record-locator.test.ts
+++ b/packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/get-record-locator.test.ts
@@ -1,0 +1,341 @@
+import test, { Test } from "tape-promise/tape";
+import { v4 as uuidv4 } from "uuid";
+import { PluginRegistry } from "@hyperledger/cactus-core";
+import {
+  EthContractInvocationType,
+  Web3SigningCredentialType,
+  PluginLedgerConnectorBesu,
+  PluginFactoryLedgerConnector,
+  ReceiptType,
+  InvokeContractV1Request,
+  BesuApiClientOptions,
+  BesuApiClient,
+  GetBesuRecordV1Request,
+} from "../../../../../main/typescript/public-api";
+import { PluginKeychainMemory } from "@hyperledger/cactus-plugin-keychain-memory";
+import {
+  BesuTestLedger,
+  pruneDockerAllIfGithubAction,
+} from "@hyperledger/cactus-test-tooling";
+import {
+  LogLevelDesc,
+  IListenOptions,
+  Servers,
+} from "@hyperledger/cactus-common";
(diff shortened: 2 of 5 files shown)


packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -730,4 +732,60 @@ export class PluginLedgerConnectorBesu
...
+    request: BesuRecordLocatorV1Request,
+  ): Promise<BesuRecordLocatorV1Response> {
+    const fnTag = `${this.className}#getBesuRecord()`;
+    this.log.info(request);
+    //////////////////////////////////////////////
+    let abi: any = undefined;
+    const resp: BesuRecordLocatorV1Response = {};
+    const trans = request.transactionHash;
+
+    if (trans != undefined) {
```
petermetz, line 745: @AzaharaC Truthiness[1] can be checked by simply putting the variable in the if condition parentheses (see below)

[1]: https://basarat.gitbook.io/typescript/recap/truthy

```suggestion
    if (txHash) {
```

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -730,4 +732,60 @@ export class PluginLedgerConnectorBesu
...
+      const transaction = await this.web3.eth.getTransaction(trans);
+      if (transaction.input) {
+        resp.ledgerId = this.instanceId;
```
petermetz, line 748: @AzaharaC The ledger ID is a consortium database primary key (so different from the instance ID of the plugin itself).

You want the ledgerId to be optional otherwise we'd be making the consortium plugin a hard-dependency by accident.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -730,4 +732,60 @@ export class PluginLedgerConnectorBesu
...
+        resp.transactionInputData = transaction.input;
+        return resp;
+      }
+    }
+
+    if (request.invokeCall != undefined) {
```
petermetz, line 754: @AzaharaC Same point about truthiness checks as above.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -730,4 +732,60 @@ export class PluginLedgerConnectorBesu
...
+      if (request.invokeCall.contractAbi != undefined) {
```
petermetz, line 755: @AzaharaC Same point about truthiness checks as above.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -730,4 +732,60 @@ export class PluginLedgerConnectorBesu
...
+      const method: ContractSendMethod = methodRef(
+        ...request.invokeCall.params,
+      );
+
+      if (
+        request.invokeCall.invocationType === EthContractInvocationType.Call
+      ) {
+        const callOutput = await (method as any).call();
+        const res: BesuRecordLocatorV1Response = {
+          ledgerId: this.instanceId,
```
petermetz, line 778: @AzaharaC Same as above regarding ledgerId !== instanceId

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -730,4 +732,60 @@ export class PluginLedgerConnectorBesu
...
+          callOutput,
+        };
+        return res;
+      } else {
+        throw new Error(
+          `${fnTag} Unsupported invocation type ${request.invokeCall.invocationType}`,
+        );
+      }
+    }
+    resp.ledgerId = this.instanceId;
```
petermetz, line 788: @AzaharaC Same as above regarding ledgerId !== instanceId

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/get-record-locator.test.ts
```diff
@@ -0,0 +1,303 @@
...
+  /**
+   * Constant defining the standard 'dev' Besu genesis.json contents.
+   *
+   * @see https://github.com/hyperledger/besu/blob/1.5.1/config/src/main/resources/dev.json
+   */
+  const firstHighNetWorthAccount = "627306090abaB3A6e1400e9345bC60c78a8BEf57";
+  const besuKeyPair = {
+    privateKey:
+      "c87509a1c067bbde78beb793e6fa76530b6382a4c0241e5e4a9ec0a0f44dc0d3",
+  };
```
petermetz, line 52: @AzaharaC We have a newer, less copy-pasty way of getting the high net worth genesis account (same Priv/Pub keys as here but obtained from the besu test ledger directly).

```suggestion
  const firstHighNetWorthAccount = besuTestLedger.getGenesisAccountPubKey();
  const besuKeyPair = {
    privateKey: besuTestLedger.getGenesisAccountPrivKey(),
  };
```

petermetz (changes requested):
@AzaharaC I commented about the ledgerID not being the same as the instance ID, then I thought about how we could go about actually obtaining the ledger ID for the response and came up with this: let's just leave the ledgerID out of the picture for now (and please leave explaining this in a comment on the GH issue as well, referencing this discussion here).

Reason: I'd rather get us going with something simpler first and see how it works out, we can bake in the consortium plugin optional parts later.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
petermetz: @AzaharaC Please remove the ledgerId, assigning the instanceId to it could mislead people.

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/get-record-locator.test.ts
```diff
@@ -0,0 +1,302 @@
...
+      params: [0],
+      gas: 1000000,
+      signingCredential: {
+        ethAccount: testEthAccount.address,
+        secret: testEthAccount.privateKey,
+        type: Web3SigningCredentialType.PrivateKeyHex,
+      },
+    };
+
+    const { callOutput } = await connector.getBesuRecord({ invokeCall: req2 });
```
petermetz, line 285: @AzaharaC Please use the API client instead of the connector for at least one of these tests to ensure that the endpoint class is also covered by the tests. Apologies for not noticing this earlier!

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/deploy-contract/get-record-locator.test.ts
```diff
@@ -0,0 +1,378 @@
...
+   * @see https://github.com/hyperledger/besu/blob/1.5.1/config/src/main/resources/dev.json
+   */
+  const firstHighNetWorthAccount = besuTestLedger.getGenesisAccountPubKey();
+  const besuKeyPair = {
+    privateKey: besuTestLedger.getGenesisAccountPrivKey(),
+  };
+  ////////////////////////////////////////////////////////////////////
+  const keyEncoder: KeyEncoder = new KeyEncoder("secp256k1");
+  const keychainId = uuidv4();
+  const keychainRef = uuidv4();
```
petermetz, line 81: @AzaharaC This keychain ref does not appear to be used (if I read the code right) and could be removed, along with the dependency in package.json itself.

packages/cactus-plugin-ledger-connector-besu/package.json
```diff
@@ -86,6 +86,8 @@
     "@hyperledger/cactus-common": "0.5.0",
     "@hyperledger/cactus-core": "0.5.0",
     "@hyperledger/cactus-core-api": "0.5.0",
+    "@hyperledger/cactus-cmd-api-server": "0.5.0",
```
petermetz, line 89: @AzaharaC You cannot depend on the API server in the plugin packages (would lead to circular dependencies).

packages/cactus-plugin-ledger-connector-besu/package.json
```diff
@@ -86,6 +86,8 @@
...
+    "key-encoder": "2.0.3",
```
petermetz, line 90: @AzaharaC See my other comment below, this can be removed AFAICT.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
petermetz: @AzaharaC Please remove the ledgerId, assigning the instanceId to it could mislead people.

petermetz (changes requested):

petermetz (approved):

takeutak (approved):

petermetz:
@AzaharaC Let me know if you need help with conflict resolutoin!

AzaharaC:
> @AzaharaC Let me know if you need help with conflict resolutoin!

Conflict resolved

PR #1106, opened 2021-07-02
fix(connector-corda): kotlin compilation error due to missing method

petermetz:
When we added the prometheus monitoring endpoint to the
OpenAPI spec, the kotlin implementation was not updated
and this has lead to the container image build process failing
which was not noticed because currently the GHA CI does not
build the container images and therefore you can get away
with introducing issues to the container image build.
(this is soon to be rectified)


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1110, opened 2021-07-06
remove magic strings

### LINKED ISSUE #1104: test: fix PKI magic strings in besu tests - genesis account
### Description

We have dozens of test cases where the the default besu genesis accounts are hardcoded in magic strings.
The `BesuTestLedger` class has methods retrieving the same values and therefore we should migrate the tests over to use those instead of the magic strings.

What it boils down to is applying this change in all the other test cases where the private/public key pair of the high net worth genesis account appears in magic strings.
The change (example): https://github.com/hyperledger/cactus/pull/1102#discussion_r661956904

### Acceptance Criteria
1. If I search for `627306090abaB3A6e1400e9345bC60c78a8BEf57` or `c87509a1c067bbde78beb793e6fa76530b6382a4c0241e5e4a9ec0a0f44dc0d3` in the entire project, there are no test cases where I see it. It only appears inside the besu test ledger class.

cc: @ty-lazar @Jeff-Ushry @twiggins120 @awadhana @Leeyoungone

awadhana:
Fixes #1104

Signed-off-by: Hana Awad <awadhana0825@gmail.com>


petermetz (changes requested):
LGTM once the linter changes are addressed.

petermetz (approved):

takeutak (approved):

PR #1121, opened 2021-07-11
ci(github): add commit lint action to PRs

petermetz:
This is supposed to check every single commit in a PR
for having the correct message format.

Previously the semantic PR action was only checking
the PR title or the first commit message, so you could
get in invalid commit messages by adding multiple
commits post-approval which the semantic PR action
would not check properly.


takeutak (approved):

takeutak:
@petermetz Thanks for your commit. LGTM, but could you mind if you tell me how to do when some person would add a commit with invalid commit messages (Does the person have to make the same commit again?)

petermetz:
> @petermetz Thanks for your commit. LGTM, but could you mind if you tell me how to do when some person would add a commit with invalid commit messages (Does the person have to make the same commit again?)

@takeutak Thank you! 

If a commit message is invalid you have multiple options to fix the issue. The simplest one is to use the `--amend` flag of the commit command the more involved scenarios with multiple commits might need you to use the `reword` action from the rebase editor.

Examples both can be seen here and if needed I'm also happy to show it to you via screen sharing: https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message

jonathan-m-hamilton (approved):
LGTM

PR #1124, opened 2021-07-12
refactor(connector-fabric): upgrade to Fabric 2.x Node SDK #1123

### LINKED ISSUE #1123: refactor(connector-fabric): upgrade to Fabric 2.x Node SDK
### Description

The Cactus packages are relying in 1.4.x Fabric SDK dependencies at the moment (even when communicating with 2.x Fabric ledgers) which is something that we need to fix by upgrading the dependency versions and then if needed create a separate Fabric connector package for 1.4.x ledgers (so that we can have the Fabric Node SDK 1.4.x dependency in there, separately).
Though the best would be if we just did not need to use the 1.4.x Fabric SDK at all.

### Acceptance Criteria
1. No trace left of the Fabric 1.4.x SDK among the dependencies (if possible)
2. All Fabric tests are still passing both 1.4.x. and 2.x ledger-wise.

petermetz:
The Cactus packages were relying on 1.4.x Fabric SDK dependencies
prior to this change (even when communicating with 2.x Fabric ledgers)
which is something that we needed to fix by upgrading the dependency
versions.

Fixes #1123


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

petermetz:
@jscode017 FYI This just got merged, should make things easier on your ODAP work as well.

PR #1132, opened 2021-07-14
feat(connector-fabric): identity json signing credentials #1130

### LINKED ISSUE #1130: feat(connector-fabric): identity json signing credentials
### Description
Add support to people providing their wallet identity as part of the run transaction request instead of the keychain being the only option.

### Acceptance Criteria
1. Test case verifies that this actually works as intended
2. Add a utility function that returns a working `Gateway` instance from the credentials.
3. Keychain becomes an optional component of the Fabric connector (right now you must use it because it's the only option to provide the identity)

petermetz:
Introduces a new, optional (for now) parameter on the run tx
endpoint's request object called gatewayOptions which is capable
of including everything that one needs to instantiate a gateway
object of the underlying Fabric Node SDK.

This change makes it possible to not need the keychain plugin
at all when someone does not need/want it.

Fixes #1130

Depends on #1124

### DIFF
--- a/packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
+++ b/packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
@@ -55,7 +55,9 @@
             "ChainCodeLifeCycleCommandResponses": {
                 "type": "object",
                 "required": [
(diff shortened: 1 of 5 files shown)


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1133, opened 2021-07-14
feat(connector-fabric): endorsing peers request arg #1122

### LINKED ISSUE #1122: feat(connector-fabric): endorsing peers request arg
### Description
As a user of the Fabric connector I want to have control over the list of endorsers of my transaction so that I can specify this explicitly (with the default being all peers).

### Acceptance Criteria
1. The run transaction endpoint is extended with a new `endorsingPeers` parameter that is a string array of MSP IDs.

petermetz:
Makes it possible to set the endorsing peers on a transaction
when submitted through the Fabric connector.

For example these values can be passed in to the new parameter
and it will set the endorsers up accordingly prior to submitting
the transaction to the ledger:
- org1.example.com
- Org1MSP
- org2.example.com
- Org2MSP

(You only need either the org domain or the msp ID, no need
to specify both of them for it to work).

For a working example, see this test case:
packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1.test.ts

Depends on #1123
Depends on #1130

Fixes #1122

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -29,6 +29,7 @@
         "escc",
         "execa",
         "faio",
+        "fidm",
         "fsouza",
         "GETHKEYCHAINPASSWORD",
         "ghcr",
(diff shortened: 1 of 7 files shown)


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1134, opened 2021-07-14
feat(core): add installOpenapiValidationMiddleware

### LINKED ISSUE #847: fix(cmd-api-server): OpenAPI spec validation missing from plugin REST endpoints
**Describe the bug**

Right now the constraints we define for request/response objects are not checked when a request comes in. (except for the api-server's built in endpoints).

**To Reproduce**

Send an invalid request to some plugin endpoint that is invalid in a certain way and watch how it crashes the plugin resulting in internal server error vs. bad request. 

**Expected behavior**

Fail with `bad request` even before the plugin's back-end code is reached if a request is sent that does not comply with the open api specs defined for that endpoint.

**Logs/Stack traces**

`N/A`

**Screenshots**

N/A

N/A

N/A

`main` @ today

 All of them.

**Additional context**

The mechanics of this are very simple and there's also a working example in the code of the api-server.ts file, but what needs to be given some more thought prior to jumping in is how to do this within the confines of the plugin architecture.
More specifically one of the questions that needs answering prior to actual work is who should be responsible for registering the validation middleware? Is it the api-server when hooking up the plugin's web services or should we delegate this entirely to the plugin instance itself?

My recommendation is below (but this could turn out to be a bad idea so be ready for the need to do some refactoring once we see how it plays out)

The interface of the web service plugins should be extended with a getOpenApiSpecs() method that returns an OpenAPI v3 spec document (see the `openapi-types` package)
Then the api-server calls up this method when it's installing the web services of the plugin, obtains the open api specs and creates a validator instance from it, hooks that middleware in to the api-server's express JS object.

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo @jagpreetsinghsasan @arnab-roy @TonyRowntree

elenaizaguirre:
This method allows us to re-use the API server's internal
mechanisms to configure the OpenAPI spec validation in
test cases where we cannot depend on the API server itself
due to circular dependencies.
So this method is designed to be used both by the API server
and the test cases at the same time.

relationed with #847

### DIFF
--- a/examples/cactus-example-carbon-accounting-business-logic-plugin/src/main/typescript/business-logic-plugin/carbon-accounting-plugin.ts
+++ b/examples/cactus-example-carbon-accounting-business-logic-plugin/src/main/typescript/business-logic-plugin/carbon-accounting-plugin.ts
@@ -5,6 +5,8 @@ import { Optional } from "typescript-optional";
 import { Express } from "express";
 import { v4 as uuidv4 } from "uuid";
 
+import OAS from "../../json/openapi.json";
+
 import {
   Logger,
   Checks,
@@ -99,6 +101,10 @@ export class CarbonAccountingPlugin
     const label = this.className;
     this.log = LoggerProvider.getOrCreate({ level, label });
     this.instanceId = options.instanceId;
+  }
+
+  public getOpenApiSpec(): unknown {
+    return OAS;
   }
 
   async registerWebServices(app: Express): Promise<IWebServiceEndpoint[]> {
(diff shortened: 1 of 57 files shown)


petermetz:
@elenaizaguirre Please check on the test failures when you get the chance.

petermetz:
@elenaizaguirre Is this ready for review?

elenaizaguirre:
> @elenaizaguirre Is this ready for review?

@petermetz Not yet

elenaizaguirre:
@petermetz As this issue is bigger than expected and new features in main branch cause continuous changes in this validation, It is possible to divide this issue into smaller ones to make it easier to validate?

petermetz:
@elenaizaguirre Yup, totally fine to break it up to smaller pieces.

elenaizaguirre:
> @elenaizaguirre Yup, totally fine to break it up to smaller pieces.

@petermetz Thank you very much for your help. Finally it works. 
Are you agree with merge this code and create validation tests in independent PRs?

petermetz (approved):

takeutak (approved):

PR #1145, opened 2021-07-16
style: fix openAPI spec linter warning endpoints

### LINKED ISSUE #1126: style: fix OpenAPI spec linter warnings in endpoints
### Description

We have linter warnings about the endpoint classes having methods that do not declare their return types.
The workaround can be seen below that will declare explicitly the return type without having to construct any additional new types to map to the JSON files (open API specs)

```typescript
  public get oasPath(): typeof OAS.paths["/api/v1/plugins/@hyperledger/cactus-plugin-ledger-connector-fabric/run-transaction"] {
    return OAS.paths[
      "/api/v1/plugins/@hyperledger/cactus-plugin-ledger-connector-fabric/run-transaction"
    ];
  }

  public getPath(): string {
    return this.oasPath.post["x-hyperledger-cactus"].http.path;
  }

  public getVerbLowerCase(): string {
    return this.oasPath.post["x-hyperledger-cactus"].http.verbLowerCase;
  }

  public getOperationId(): string {
    return this.oasPath.post.operationId;
  }
```

### Acceptance Criteria
1. All linter warnings in the project that are about the OAS path objects in the endpoint classes' methods are fixed (no need to fix all linter warnings just the relevant ones)

cc: @ty-lazar @Jeff-Ushry @twiggins120 @awadhana @Leeyoungone

Leeyoungone:
Fixes #1126


petermetz (changes requested):
@Leeyoungone Sorry to to just having come up with this mid-flight your PR: Could you open one more PR separately that will become a dependency of this PR.
The new PR should make the change setting up the stage for this one to be a little smaller (because right now it's huge with 100+ changed files).

What the new PR should contain is just the remapping of the TS typings from `dist/types/...` to `dist/lib/...` 

This way we'll have two smaller PRs (and two smaller commits respectively) which is always easier to review and also to interpret after the fact based on the commit messages.

The commit message in the new PR should explain that the change is necessary in order to support the use-case of importing JSON files within the d.ts files which must be in under the same directory path as the implementation files to ensure that this can happen.

petermetz:
@Leeyoungone Forgot to ask one more thing: Please mark this PR dependent on the other PR once that new PR has been opened. Thank you again!

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/api-client/besu-api-client.ts
```diff
@@ -23,6 +23,7 @@ export class BesuApiClient extends DefaultApi {
   private readonly log: Logger;
   private readonly wsApiHost: string;
   private readonly wsApiPath: string;
+  getPrometheusExporterMetricsV1: any;
```
petermetz, line 26: @Leeyoungone Is this some leftover change that is otherwise not necessary? Just asking because it doesn't seem to be used in the code of the class.

petermetz (changes requested):
@Leeyoungone Just one small nit, but otherwise this is looking ready for prime time! Bringing it out of draft state now.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/api-client/besu-api-client.ts
Leeyoungone: I think you're right, I don't remember having to alter the besu-api-client.ts file, thank you!

petermetz: @Leeyoungone No worries! If you remove the line instead of commenting it that the file will nicely disappear from the diff altogether (and people won't wonder what the purpose of that commented code might be!). :-)

Once that's done, please squash + rebase and then it's LGTM

petermetz (approved):

takeutak (approved):

PR #1151, opened 2021-07-20
fix(connector-corda): fix build broken by operationId rename

petermetz:
Also: Update the readmes of other packages where the old name of
the operation/endpoint was still being used as well.


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1152, opened 2021-07-20
docs(examples): increase supply chain image boot responsiveness #780

### LINKED ISSUE #780: docs(examples): supply chain image boot goes silent for 10 minutes
**Describe the bug**

The container appears as if it hung because it takes a long time to download the AIO images and it's not saying anything in the logs for that period of time.

**To Reproduce**

Launch the example container as instructed by the readme and watch how it appears to hang for about 10 minutes while it is actually hard at work setting everything up.

**Expected behavior**

Logs should show a little more liveness so people don't abandon the example assuming that it's broken.

The demo has an adequate amount of documentation around what conditions, exactly, it needs to work.

**Logs/Stack traces**

N/A

**Screenshots**

N/A

Dev machine

N/A

`main` @ 0.4.1

N/A

**Additional context**

N/A

petermetz:
Added logging to all 3 of the test ledger classes that are being used
which are taking up most of the boot time.
Now users can see the setup process as it chugs along step by step and
in general be much more confident while waiting that things are indeed
happening instead of the container just being in a zombie state.

Also added an additional point in the README.md file claryfying that
one needs to have a working installation of docker on their machine.
Decided to not go into any more detail than that because the (linked)
docker for desktop documentation is the best/most robust resrouces
available there is and it would be hard to compete with it if we were
to produce our own installation instructions.

Fixes #780


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1160, opened 2021-07-22
feat(besu): support for private transactions

### LINKED ISSUE #952: feat(besu): Add private transaction support

Feature request: I need to be able to issue private transactions to a Besu ledger via the Cactus API client.

1. Either a separate endpoint next to the current POST /run-transaction one or as an extension of it. Not yet sure which one is more feasible/makes more sense.
2. Test coverage for the new feature
3. Private transactions must not be logged by Cactus for obvious reasons.
4. Documentation to explain specifically that this is not a privacy preserving feature, it only allows access to the pre-existing Besu APIs as-is

Nothing. There is no doubt in my mind that we need this feature.

Additional context

Here's the Besu documentation regarding private transactions:
https://besu.hyperledger.org/en/stable/Concepts/Privacy/Private-Transactions/

cc: @takeutak @sfuji822 @jonathan-m-hamilton @petermetz @kikoncuo @travis-payne

travis-payne:
fixes #952

### DIFF
--- a/packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
+++ b/packages/cactus-plugin-ledger-connector-fabric/src/main/json/openapi.json
@@ -55,7 +55,9 @@
             "ChainCodeLifeCycleCommandResponses": {
                 "type": "object",
                 "required": [
-                    "queryInstalledList", "approveForMyOrgList", "installList"
+                    "queryInstalledList",
+                    "approveForMyOrgList",
+                    "installList"
                 ],
                 "properties": {
                     "packaging": {
@@ -184,6 +186,42 @@
                     }
                 }
             },
+            "GatewayOptions": {
+                "type": "object",
+                "required": ["identity", "wallet"],
+                "properties": {
+                    "connectionProfile": {
+                        "$ref": "#/components/schemas/ConnectionProfile"
+                    },
+                    "discovery": {
+                        "$ref": "#/components/schemas/GatewayDiscoveryOptions",
+                        "nullable": false
+                    },
+                    "eventHandlerOptions": {
+                        "$ref": "#/components/schemas/GatewayEventHandlerOptions",
+                        "nullable": false
+                    },
+                    "identity": {
+                        "type": "string"
+                    },
+                    "wallet": {
+                        "type": "object",
+                        "minProperties": 1,
+                        "maxProperties": 1,
+                        "properties": {
+                            "keychain": {
+                                "$ref": "#/components/schemas/FabricSigningCredential"
+                            },
+                            "json": {
+                                "type": "string",
+                                "nullable": false,
+                                "minLength": 1,
+                                "maxLength": 65535
+                            }
+                        }
+                    }
+                }
+            },
             "DefaultEventHandlerStrategy": {
                 "type": "string",
                 "enum": [
@@ -201,6 +239,10 @@
                 "additionalProperties": false,
                 "properties": {
                     "commitTimeout": {
+                        "type": "number",
+                        "nullable": false
+                    },
+                    "endorseTimeout": {
                         "type": "number",
                         "nullable": false
                     },
@@ -293,6 +335,10 @@
                         "type": "object",
                         "nullable": true
                     },
+                    "gatewayOptions": {
+                        "$ref": "#/components/schemas/GatewayOptions",
+                        "nullable": false
+                    },
                     "signingCredential": {
                         "$ref": "#/components/schemas/FabricSigningCredential",
                         "nullable": false
@@ -329,7 +375,6 @@
                             "nullable": true
                         }
                     },
-
                     "endorsingParties": {
                         "type": "array",
                         "nullable": false,
--- a/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/common/create-gateway.ts
+++ b/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/common/create-gateway.ts
@@ -1,0 +1,92 @@
+import { DefaultEventHandlerOptions } from "fabric-network";
+import { DefaultEventHandlerStrategies, Wallets } from "fabric-network";
+import { Gateway } from "fabric-network";
+import { GatewayOptions as FabricGatewayOptions } from "fabric-network";
+import { Checks, LoggerProvider } from "@hyperledger/cactus-common";
+import { LogLevelDesc } from "@hyperledger/cactus-common";
+import { PluginRegistry } from "@hyperledger/cactus-core";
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -136,7 +136,8 @@
   },
   "dependencies": {
     "@types/ssh2": "0.5.46",
+    "cspell": "5.4.0",
     "artillery": "1.7.2",
-    "cspell": "5.4.0"
+    "web3": "1.3.6"
```
petermetz, line 141: @travis-payne The project root package.json is meant to only hold the dependencies needed to perform the building of the child packages where the code is => please remove this dependency declaration from.

packages/cactus-plugin-ledger-connector-besu/README.md
```diff
@@ -97,9 +97,13 @@ enum Web3SigningCredentialType {
     NONE = 'NONE'
 }
 ```
+
 > Extensive documentation and examples in the [readthedocs](https://readthedocs.org/projects/hyperledger-cactus/) (WIP) 
 
 
+## Transaction Privacy Feature
+Private transactions using Besu are currently enabled, however this feature is developer preview for now due to a lack of test coverage.
```
petermetz, line 105: @travis-payne Is it though? With tests now being available I think we can take this out.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -150,9 +156,21 @@ export class PluginLedgerConnectorBesu
       `${fnTag} options.prometheusExporter`,
     );
 
+    // const web3Provider = new Web3.providers.HttpProvider(
```
petermetz, line 159: @travis-payne If the commented code is no longer useful then I recommend to just remove it completely.

petermetz: @travis-payne What was the reason behind removing the HTTP provider? (IIRC we had both WS and HTTP available - might be wrong though)

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -150,9 +156,21 @@ export class PluginLedgerConnectorBesu
...
+    //   this.options.rpcApiHttpHost,
+    // );
+
+    // this.web3 = new Web3(web3Provider);
+    this.initializeEEAClient();
```
petermetz, line 164: @travis-payne Now that we have the async plugin initializers from Tony have been merged we can leverage that for calling the EEAClient init.
Note: You'll have to call the onInitPlugin method manually on the plugin instance in the test cases (because normally that would be done by the API server)

See: https://github.com/hyperledger/cactus/blob/main/packages/cactus-core-api/src/main/typescript/plugin/i-cactus-plugin.ts#L72

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -379,8 +399,37 @@ export class PluginLedgerConnectorBesu
...
-      const success = true;
+      let callOutput;
+      let success = false;
+      if (req.privateTransactionConfig) {
+        const data = method.encodeABI();
+        const fnParams = {
+          to: contractInstance.options.address,
+          data,
+          privateFrom: req.privateTransactionConfig.privateFrom,
+          // FIXME implement support for the rest of the supported signing credential types where applicable
```
petermetz, line 410: @travis-payne Please add support for optionally specifying the signing credentials via a keychain reference (emphasis on optionally, we also need to be able to specify everything in the request directly)

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -507,11 +569,38 @@ export class PluginLedgerConnectorBesu
...
       default:
         throw new Error(
-          `${fnTag} Unrecognized ReceiptType: ${req.consistencyStrategy.receiptType}`,
+          `${fnTag} Unrecognized ReceiptType: ${request.consistencyStrategy.receiptType}`,
         );
     }
   }
 
+  public async transactPrivate(options: any): Promise<RunTransactionResponse> {
+    if (!this.web3EEA) {
```
petermetz, line 578: @travis-payne Once you migrate over to the async plugin initializer I'd refactor this check to throw instead of lazy loading.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -507,11 +569,38 @@ export class PluginLedgerConnectorBesu
...
+      await this.initializeEEAClient();
+    }
+
+    const txHash = await this.web3EEA?.eea.sendRawTransaction(options);
```
petermetz, line 582: @travis-payne Once you refactor the if statement above to throw instead, the `?` will no longer be necessary because the compiler will understand that the type has been narrowed down sufficiently.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -507,11 +569,38 @@ export class PluginLedgerConnectorBesu
...
+      throw new RuntimeError(`eea.sendRawTransaction provided no tx hash.`);
+    }
+    return this.getPrivateTxReceipt(options.privateFrom, txHash);
+  }
+
+  public async getPrivateTxReceipt(
+    privateFrom: string,
+    txHash: string,
+  ): Promise<RunTransactionResponse> {
+    const txPoolReceipt = await this.web3EEA?.priv.getTransactionReceipt(
```
petermetz, line 593: @travis-payne Once you refactor the if statement above to throw instead, the `?` will no longer be necessary because the compiler will understand that the type has been narrowed down sufficiently.

petermetz (changes requested):
@travis-payne Please squash the commits into a single one.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
travis-payne: This bit of code isn't even in main branch, we only use the HTTP provider to sign a transaction (line 717 on main). I think I was just testing something. Will remove.

PR #1165, opened 2021-07-23
build: use yarn 1.19.0 instead of 1.22.10

petermetz:
Needed to downgrade because of a bug in 1.22.10 where it was
unable to install dependencies into workspaces individually.

See details here: https://github.com/yarnpkg/yarn/issues/8405


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1166, opened 2021-07-23
build(deps): fix sync-npm-deps-to-tsc-projects tool

petermetz:
1. The ./tools/sync-npm-deps-to-tsc-projects.ts script
had a missing part in it which would fill the root
tsconfig.json file with the references to the packages
of the monorepo.

2. Point 1 could not be achieved without removing the
cactus-cockpit package from the dependencies of
cmd-api-server because the script mentioned above fails
if there are dependencies declared which do not exist
on the file-system anymore (which is exactly what
happened when we deprecated the cockpit package and then
forgot to remove it from the list of dependencies.

3. All the tsconfig.json files are updated by the sync
script (from point 1.)

4. Adds skip lib check true to the ipfs package's tsconfig
file so that the issue with uint8array dependency's own typings
don't break the build. For reference, the errors look like this:

- Error: node_modules/uint8arrays/dist/to-string.d.ts(20,24): error TS2307:
  Cannot find module './util/bases' or its corresponding type declarations.

- Error: node_modules/uint8arrays/dist/to-string.d.ts(21,34): error TS2307:
  Cannot find module './util/bases' or its corresponding type declarations.

### DIFF
--- a/examples/cactus-example-carbon-accounting-backend/tsconfig.json
+++ b/examples/cactus-example-carbon-accounting-backend/tsconfig.json
@@ -41,7 +41,7 @@
   ],
   "include": [
     "./src",
-    "src/**/*.json",
+    "src/**/*.json"
   ],
   "exclude": [
     "./src/utility-emissions-channel"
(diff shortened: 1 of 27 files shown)


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1170, opened 2021-07-26
Core api

twiggins120:
Fixes issues


takeutak (approved):

petermetz:
@takeutak Apologies for the double PR - it was not possible to force push fixes to the original (this) one because the branch name contained special characters in it (and the only way to rename a backing branch of a PR in GH is to close down and resubmit a new PR from a different branch altogether.

PR #1177, opened 2021-07-30
docs(contributing.md): do not duplicate pull requests

petermetz:
It's been a common problem with people erasing review history
by opening new pull requests so now it's in the CONTRIBUTING.md
specifically being called out as something that's undesired.


takeutak:
Thanks for updating contributing.md!
However, I think that there are times when I want to create a new pull-request with the same content because the pull-request I submitted first cannot work for some reason. In that case, I would like you to make it possible to create the next pull-request if you specify it in the comment column of the pull-request that I submitted first.

jonathan-m-hamilton (approved):
LGTM

petermetz:
> new pull-request with the same content because the pull-request I submitted first cannot work for some reason. 

@takeutak My assumption was that it is impossible to have a scenario where the content of the pull request is the same and yet the first doesn't work but the second one does. Could you provide an example by any chance of this happening? 
Flaky tests aside, the same code should always produce the same outcomes e.g. either work or not consistently.

takeutak:
> @takeutak My assumption was that it is impossible to have a scenario where the content of the pull request is the same and yet the first doesn't work but the second one does. Could you provide an example by any chance of this happening?

For example, I assume errors other than the codes itself, such as wrong operation of GitHub (e.g. mistake to name the commit message). Of course, if there is a certain amount of mistake, it can be corrected by Git operation, but I think that there is no guarantee that any mistake can be corrected by maintainers.

Of course, as you said, it is a big problem that the review history of old PR disappears when a new PR is opened. However, in order to prevent that, I think it is enough to put the link of the old PR in the new PR.

petermetz:
@takeutak Understood. Would it help if I added an exception clause to the description where I'd add an item specifically for the case when the branch name is containing special characters and need to be changed (which is impossible to do without re-opening a brand new PR)?

I would prefer not to add commit messages being incorrect to the list of exceptions because they can be fixed in about 30 seconds with an interactive rebase or commit amend operation and then a force push with lease.

takeutak:
@petermetz
>  Would it help if I added an exception clause to the description where I'd add an item specifically for the case when the branch name is containing special characters and need to be changed (which is impossible to do without re-opening a brand new PR)?
Thanks. I agree with your idea. Would you add such an exception clause? If you do so, I will approve it.

> I would prefer not to add commit messages being incorrect to the list of exceptions because they can be fixed in about 30 seconds with an interactive rebase or commit amend operation and then a force push with lease.
Understood. I agree with it.

petermetz:
@takeutak Understood, thank you. I've updated the description to explain the exception.

> I may ask your help

Sure thing!

PR #1179, opened 2021-07-30
ci(tools): validate bundle names packages

petermetz:
This script is now part of the ci.sh script which
ensures that there is no human error when defining
the webpack bundle names in the package.json files
for each package in the project.

The idea is to reduce the manual review burden on the
maintainers who can also miss these type of issues.


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1169, opened 2021-07-26
feat(connector-iroha): adds connector plugin

hxlaf:
1. Enables Cactus to manipulate Iroha ledger
2. Supports 19/20 commands and 14/15 queries of the Iroha ledger. (removePeer and fetchCommits have not been fully implemented)
3. Possibility of starting Iroha and Postgres with random ports. 
4. Possibility of passing arbitrary Iroha admin/node keypairs to the test ledger.
5. Offers a comprehensive run transaction test suite to validate the support for Iroha ledger commands and queries.
6. Provides an example of asset transfer between two separate Iroha nodes.

PR #1183, opened 2021-08-03
feat(iroha): add Iroha AIO image and Iroha test ledger

hxlaf:
1. Provides port randomization
2. Possibility to launch from within test cases, no need for docker-compose
3. Can be linked to postgres container via the default host network interface instead of being forced to deal with docker networks and all the statefulness they bring into the picture
4. Contains a workaround to the FILE -> ENV configuration precedence of Iroha (opened an issue about this - PR is dropping soon but in the meantime we needed it to work anyway so that's why the entrypoint script has the jq trickery in there to update the configuration file based on what's in the environment variables for postgres credentials)


petermetz (changes requested):
@hxlaf Are the examples necessary to be included for the AIO image? In general I'd prefer to trim the diff down to the smallest possible size that still makes sense and achieves what's needed (e.g. have the container functionality available that is required to make the test cases in the other PR work)

petermetz (changes requested):
@hxlaf I'd also include the test ledger class' code from the test tooling package so that a test can be added for it which then verifies that hte image itself operates as intended. 
Then in the next PR that same test can be morphed into the final test which also verifies the connector itself

hxlaf:
> @hxlaf Are the examples necessary to be included for the AIO image? In general I'd prefer to trim the diff down to the smallest possible size that still makes sense and achieves what's needed (e.g. have the container functionality available that is required to make the test cases in the other PR work)

hxlaf:
> @hxlaf I'd also include the test ledger class' code from the test tooling package so that a test can be added for it which then verifies that hte image itself operates as intended.
> Then in the next PR that same test can be morphed into the final test which also verifies the connector itself

Hi Peter,
I updated the files according to my interpretation of your critique.
image

petermetz (changes requested):
@hxlaf It's shaping up great, thank you! 

One more nit regarding the commit message: I would disappear the intermediary messages from the squashed message because if you look at it from the perspective of the main branch for example it doesn't make sense to say that the commit is fixing the cspell tab size because that was something was introduced on this branch to begin with so in the future if someone is reading the commit message they would ask themselves "What was wrong with the tab size that this commit changed - it is 2 before and after the commit as well so I don't get it.".

It's a minor thing that I don't like to annoy people with, but it is important in the longer term to have our commit messages make sense to someone who is reading it in the future 1, 5 or 10 years from now because that's when it will be very important to be able to trace what changes were made and why exactly. Of course, specifically the whole tab size thing is unlikely to be some hot topic in the future that would ever need someone going through all the old commits and reconstructing history, but I'm making the point anyway because it's a good habit to have in general to pay attention to this principle about commit messages so that it just happens automatically and then it really becomes important when you are making high impact changes to the APIs or breaking changes between major version upgrades, etc. 

Another great example to demonstrate the importance of these things is security related topics. There could be something in the public API surface that annoys/constraints what people can do in order to ensure security and then in the future people could propose to change that without knowing the negative impact on security (which could be devastating).

This got way too long so I'll stop talking and also cc some more people because this is definitely an important principle that I want well publicized. :-)

cc: @Leeyoungone  @Jeff-Ushry @twiggins120 @awadhana => This is Monday's pro tip about software engineering. ;-)

hxlaf:
> @hxlaf It's shaping up great, thank you!
> 
> One more nit regarding the commit message: I would disappear the intermediary messages from the squashed message because if you look at it from the perspective of the main branch for example it doesn't make sense to say that the commit is fixing the cspell tab size because that was something was introduced on this branch to begin with so in the future if someone is reading the commit message they would ask themselves "What was wrong with the tab size that this commit changed - it is 2 before and after the commit as well so I don't get it.".
> 
> It's a minor thing that I don't like to annoy people with, but it is important in the longer term to have our commit messages make sense to someone who is reading it in the future 1, 5 or 10 years from now because that's when it will be very important to be able to trace what changes were made and why exactly. Of course, specifically the whole tab size thing is unlikely to be some hot topic in the future that would ever need someone going through all the old commits and reconstructing history, but I'm making the point anyway because it's a good habit to have in general to pay attention to this principle about commit messages so that it just happens automatically and then it really becomes important when you are making high impact changes to the APIs or breaking changes between major version upgrades, etc.
> 
> Another great example to demonstrate the importance of these things is security related topics. There could be something in the public API surface that annoys/constraints what people can do in order to ensure security and then in the future people could propose to change that without knowing the negative impact on security (which could be devastating).
> 
> This got way too long so I'll stop talking and also cc some more people because this is definitely an important principle that I want well publicized. :-)
> 
> cc: @Leeyoungone @Jeff-Ushry @twiggins120 @awadhana => This is Monday's pro tip about software engineering. ;-)

Hey Peter,
Thank you for pointing out! I did not really think about the consequences. Anyways, I fixed it just now.
image
I will make it one of my habit in the future.

hxlaf:
@petermetz 
Maybe you have to restart the CI workflow again due to my update.

petermetz:
@hxlaf No worries at all, I re-ran it just now. Thank you for the update!

tools/docker/iroha-all-in-one/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```sh
+DOCKER_BUILDKIT=1 docker build ./tools/docker/iroha-all-in-one/ -t iaio
+```
+
+
+## Shell into a running container:
+
+```sh
+docker run -it --entrypoint bash hyperledger/cactus-iroha-all-in-one
```
petermetz, line 22: @hxlaf 

```suggestion
docker run -it --entrypoint bash iaio
```

tools/docker/iroha-all-in-one/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+## Logs of Iroha Ledger via shell:
+
+```sh
+docker logs cactus-iroha-all-in-one
```
petermetz, line 28: @hxlaf 

```suggestion
docker logs iaio
```

petermetz (changes requested):
@hxlaf Almost ready, just had two more nits but then it's prime time, thank you again!

petermetz (changes requested):
@hxlaf 

I added some more description in addition to the title, 
please also make sure to amend your commit so that it has the message in there as well in the message body, specifically this part:

```sh
1. Provides port randomization for host port binding
2. Possibility to launch from within test cases, no need for docker-compose
3. Can be linked to postgres container via the default host network interface instead of being forced to deal with docker networks and all the statefulness they bring into the picture
4. Contains a workaround to the FILE -> ENV configuration precedence of Iroha (opened an issue about this - PR is dropping soon but in the meantime we needed it to work anyway so that's why the entrypoint script has the jq trickery in there to update the configuration file based on what's in the environment variables for postgres credentials)
```

hxlaf:
> @hxlaf
> 
> I added some more description in addition to the title,
> please also make sure to amend your commit so that it has the message in there as well in the message body, specifically this part:
> 
> ```shell
> 1. Provides port randomization for host port binding
> 2. Possibility to launch from within test cases, no need for docker-compose
> 3. Can be linked to postgres container via the default host network interface instead of being forced to deal with docker networks and all the statefulness they bring into the picture
> 4. Contains a workaround to the FILE -> ENV configuration precedence of Iroha (opened an issue about this - PR is dropping soon but in the meantime we needed it to work anyway so that's why the entrypoint script has the jq trickery in there to update the configuration file based on what's in the environment variables for postgres credentials)
> ```

Done! Thank you! :D

petermetz (approved):

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
```diff
@@ -0,0 +1,352 @@
...
+   * @see https://github.com/hyperledger/iroha/blob/main/example/admin%40test.priv
+   */
+  public getGenesisAccountPrivKey(): string {
+    return "f101537e319568c765b2cc89698325604991dca57b9716b58016b253506cab70";
+  }
+
+  public async getNodeKeyPair(): Promise<IKeyPair> {
+    const fnTag = `IrohaTestLedger#getNodeKeyPair()`;
+    if (!this.container) {
+      throw new Error(`${fnTag} this.container cannot be falsy.`);
```
baziorek, line 156:
```suggestion
      throw new Error(`${fnTag} this.container cannot be false.`);
```

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
```diff
@@ -0,0 +1,352 @@
...
+        this.imageFqn,
+        [],
+        [],
+        {
+          ExposedPorts: {
+            [`${this.rpcToriiPort}/tcp`]: {}, // Iroha RPC - Torii
+          },
+          Env: this.envVars,
+          Healthcheck: {
+            Test: ["CMD-SHELL", "python3 /healthcheck.py admin@test 50051"],
```
baziorek, line 192: 1. Is the file in absolute directory `/`? I think it is better to use different one
2. I suggest to use `getGenesisAdminAccount()` instead of `admin@test`, the same with `50051`

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
```diff
@@ -0,0 +1,352 @@
...
+            Interval: 1000000000, // 1 second
+            Timeout: 3000000000, // 3 seconds
+            Retries: 299,
+            StartPeriod: 3000000000, // 1 second
```
baziorek, line 196: I suggest constant for 1 second - comments would be not necessarily.

BTW. I think here is not correct comment:
```
            StartPeriod: 3000000000, // 1 second
```

tools/docker/iroha-all-in-one/README.md
```diff
@@ -0,0 +1,29 @@
+# Iroha All in One Image
+
+An all in one Iroha + Postgres image that uses about 0.5 GB of RAM and one vCPU worth of hardware resources idling.
+This makes it much easier to pull up for tests that are about functionality rather than running a production grade ledger.
```
baziorek, line 4: I don't think that those information should be here.
1. I assume You didn't mean `RAM`?
2. Good docker instances are using as less memory as possible, so we don't add information about its, because it can change in future.
3. Instead of first line I think there should be information: what it is, what it is for, what it contains, etc.
4. I don't think that the second line is necessarily - user would see alternatives of getting docker container and choose one of them.

tools/docker/iroha-all-in-one/entrypoint.sh
```diff
@@ -0,0 +1,30 @@
...
+  set -- irohad "$@"
+fi
+
+if [ "$1" = 'irohad' ]; then
+  echo key=$KEY
+  echo $PWD
+  if [ -n "$IROHA_POSTGRES_HOST" ]; then
+    echo "NOTE: IROHA_POSTGRES_HOST should match 'host' option in config file"
+    PG_PORT=${IROHA_POSTGRES_PORT:-5432}
+    /wait-for-it.sh -h $IROHA_POSTGRES_HOST -p $PG_PORT -t 30 -- true
```
baziorek, line 22: is the path of the script correct: `/wait-for-it.sh ` - should it be here `/`?

tools/docker/iroha-all-in-one/healthcheck.py
```diff
@@ -0,0 +1,37 @@
+import os
+import sys
+from iroha import Iroha, IrohaCrypto, IrohaGrpc
+
+#setup of iroha client address
+IROHA_HOST_ADDR = os.getenv('IROHA_HOST_ADDR', '127.0.0.1')
+IROHA_PORT = os.getenv('IROHA_PORT', sys.argv[2])
```
baziorek, line 7: I suggest to read external arguments in consistent way: use environment variables, or arguments. In all other places You are using environment variables, so let's not do exception here.

What is more if user doesn't provide arguments - the program would crash without information about usage.

tools/docker/iroha-all-in-one/healthcheck.py
```diff
@@ -0,0 +1,37 @@
...
+net = IrohaGrpc(f'{IROHA_HOST_ADDR}:{IROHA_PORT}')
+#setup admin account
+ADMIN_ACCOUNT_ID = os.getenv('ADMIN_ACCOUNT_ID', 'admin@test')
+ADMIN_PRIVATE_KEY = os.getenv(
+    'ADMIN_PRIVATE_KEY', 'f101537e319568c765b2cc89698325604991dca57b9716b58016b253506cab70')
+iroha_admin = Iroha(ADMIN_ACCOUNT_ID)
+user_account = 'admin@test'
+iroha_user = Iroha(user_account)
+
+#health check function
```
baziorek, line 17: I suggest to use docstring inside the function instead of the comment

tools/docker/iroha-all-in-one/healthcheck.py
```diff
@@ -0,0 +1,37 @@
...
+def get_account(account:str):
+    query = iroha_user.query('GetAccount', account_id=account)
+    IrohaCrypto.sign_query(query, ADMIN_PRIVATE_KEY)
+    response = net.send_query(query)
+    #health check passes when response contains "admin@test"
+    if response.account_response.account.account_id == user_account:
+        print("Succss!")
```
baziorek, line 24:
```suggestion
        print("Success!")  
```

tools/docker/iroha-all-in-one/healthcheck.py
```diff
@@ -0,0 +1,37 @@
...
+iroha_admin = Iroha(ADMIN_ACCOUNT_ID)
+user_account = 'admin@test'
+iroha_user = Iroha(user_account)
+
+#health check function
+def get_account(account:str):
+    query = iroha_user.query('GetAccount', account_id=account)
+    IrohaCrypto.sign_query(query, ADMIN_PRIVATE_KEY)
+    response = net.send_query(query)
+    #health check passes when response contains "admin@test"
```
baziorek, line 22:
```suggestion
    # health check passes when response contains value of account parameter
```

tools/docker/iroha-all-in-one/healthcheck.py
```diff
@@ -0,0 +1,37 @@
...
+    if response.account_response.account.account_id == user_account:
+        print("Succss!")  
+        sys.exit(0)
+    #health check fails when response does not contain "admin@test"
+    else:
+        print("Connection success, but account does not exists!")
```
baziorek, line 28:
```suggestion
        print(f"Connection success, but account '{account}' does not exist!")
```

tools/docker/iroha-all-in-one/healthcheck.py
```diff
@@ -0,0 +1,37 @@
...
+        sys.exit(1)
+
+if __name__ == '__main__':
+    try:
+        get_account(sys.argv[1])
```
baziorek, line 33: I suggest checks if user provided required arguments, if not to print usage information

tools/docker/iroha-all-in-one/healthcheck.py
```diff
@@ -0,0 +1,37 @@
...
+    else:
+        print("Connection success, but account does not exists!")
+        sys.exit(1)
+
+if __name__ == '__main__':
+    try:
+        get_account(sys.argv[1])  
+    #health check fails when connection with Iroha client cannot establish      
+    except Exception as e:   
+        print("Connection failure!")
```
baziorek, line 36:
```suggestion
        print("Connection failure! Reason:", e)
```

tools/docker/iroha-all-in-one/healthcheck.py
```diff
@@ -0,0 +1,37 @@
...
+        sys.exit(1)
```
baziorek, line 37: I suggest to format all python files with: https://github.com/psf/black because python programmers are trying to be compatible with PEP programming style, the tool will reformat Your code.

tools/docker/iroha-all-in-one/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# irohad is the core of Iroha ledger
+COPY --from=builder /iroha/build/bin/irohad /usr/bin/irohad
+# copying iroha-cli optional; only copied for debugging purpose
+COPY --from=builder /iroha/build/bin/iroha-cli /usr/bin/iroha-cli
+# files below are necessary
+COPY --from=builder /iroha/example/ /opt/iroha_data/
+COPY --from=builder /iroha/docker/release/wait-for-it.sh /
+COPY genesis.block /opt/iroha_data/genesis.block
+COPY entrypoint.sh /
+COPY healthcheck.py /
```
baziorek, line 23:
```suggestion
COPY entrypoint.sh
COPY healthcheck.py
```

baziorek (commented):
I suggest more constants instead of raw text

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
hxlaf: 1. Yes. The file is in the absolute directory `/`. I put it here because Iroha team put their entrypoint.sh and wait-for-it.sh at the same absolute directory `/`.
2. Unfortunately, I do not think I can put any function or string literal at line 192. It won't compile.

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
hxlaf: I believe that 'falsy' is a common language used in the Javascript world.
[()](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)

tools/docker/iroha-all-in-one/healthcheck.py
hxlaf: Thank you for pointing out.
The idea of doing 
`python3 healthcheck useraccount port` here is because iroha client could potentially be generated on other ports other than 50051, so people could simply pass in a different port number for the healthcheck purpose.

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
baziorek: Ups, sorry, I didn't know, I though that it was a typo:D

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
baziorek: OK, thanks for explanation, so I'm making resolved

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
hxlaf: Thank you! Done!

tools/docker/iroha-all-in-one/healthcheck.py
hxlaf: Resolved. Thank you!

tools/docker/iroha-all-in-one/entrypoint.sh
hxlaf: Yes! The Iroha team copied wait-for-it.sh to absolute path `/`.
[Reference](https://github.com/hyperledger/iroha/blob/main/docker/release/Dockerfile#L14)

tools/docker/iroha-all-in-one/healthcheck.py
hxlaf: Done! Thank you!

tools/docker/iroha-all-in-one/healthcheck.py
hxlaf: Fixed! Thank you!

tools/docker/iroha-all-in-one/healthcheck.py
hxlaf: Done. Thank you! That is a great tool.

tools/docker/iroha-all-in-one/Dockerfile
hxlaf: Similar to above. The Iroha team copied wait-for-it.sh to absolute path `/`.
[Reference](https://github.com/hyperledger/iroha/blob/main/docker/release/Dockerfile#L14)
Decided to do `/` explicitly to make it obvious.

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
hxlaf: You are welcome. :)

tools/docker/iroha-all-in-one/entrypoint.sh
baziorek: OK, thanks for explaination. So the path is OK if iroha is doing the same thing:)

tools/docker/iroha-all-in-one/Dockerfile
baziorek: Sorry, I my suggestion was wrong. It is good that You didn't applied its:)

takeutak:
@hxlaf Please let me confirm one thing. A docker tool for Hyperledger Iroha have already been committed to Cactus as the following: <https://github.com/hyperledger/cactus/tree/main/tools/docker/iroha-testnet>
Is there any difference between the above one and your docker tool from the viewpoint of features?

hxlaf:
> @hxlaf Please let me confirm one thing. A docker tool for Hyperledger Iroha have already been committed to Cactus as the following: https://github.com/hyperledger/cactus/tree/main/tools/docker/iroha-testnet
> Is there any difference between the above one and your docker tool from the viewpoint of features?

Yes! 
I saw your testnet example and that really helped me to understand Iroha's architecture. The difference is that I believe the testnet features the use of docker-compose and seamless operation of 4 nodes, but this PR features a modified Iroha Dockerfile (including port randomization, no need to set up a dedicated docker network/volume, customized Iroha healthcheck). More importantly, the [Cactus Iroha connector plugin PR](https://github.com/hyperledger/cactus/pull/1169) needs the features (port randomization, no need for Iroha  network/volume, Iroha healthcheck) provided by this PR to run.

tools/docker/iroha-all-in-one/README.md
hxlaf: Thank you for pointing out the potential way to improve!
I have updated the `README.md` file according to what we discussed today.

baziorek (approved):
The PR looks much better. I found only few not important things to correct. You can correct them or no.

There is one more thing - I've noticed that files: `iroha-test-ledger.ts` and `postgres-test-container.ts` contains functions almost the same with single line difference. If You have time You can think about its how to avoid code duplication.

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
```diff
@@ -0,0 +1,352 @@
...
+    const aContainerInfo = await this.getContainerInfo();
+
+    if (aContainerInfo) {
+      const { NetworkSettings } = aContainerInfo;
+      const networkNames: string[] = Object.keys(NetworkSettings.Networks);
+      if (networkNames.length < 1) {
+        throw new Error(`${fnTag} container not connected to any networks`);
+      } else {
+        // return IP address of container on the first network that we found
+        // it connected to. Make this configurable?
```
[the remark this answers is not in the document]
hxlaf: The question exists in most packages, seems like a pending question not resolved at this moment.
image

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  /**
+   * Output is based on the standard Iroha admin user public key.
+   *
+   * @see https://github.com/hyperledger/iroha/blob/main/example/admin%40test.pub
+   * @see https://github.com/hyperledger/iroha/blob/main/example/genesis.block
+   */
+  public getGenesisAccountPubKey(): string {
+    return "313a07e6384776ed95447710d15e59148473ccfc052a681317a72a69f2a49910";
```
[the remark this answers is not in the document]
hxlaf: Thank you for pointing out! Fixed!

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
```diff
@@ -0,0 +1,352 @@
...
+    "IROHA_POSTGRES_PASSWORD=my-secret-password",
+    "KEY=node0",
+  ],
+});
+
+//can we go inside the docker container and then test within the docker container
+///initiated by the dockrode service
+/*
+ * Provides validations for Iroha container's options
+ */
```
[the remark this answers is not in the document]
hxlaf: Resolved! Thank you!

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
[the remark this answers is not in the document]
baziorek: OK, thanks for answer, so lets mark its resolved

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
petermetz: @hxlaf Sorry I'm late for the review party!

> Unfortunately, I do not think I can put any function or string literal at line 192. It won't compile.

Let's talk more about this, I'm pretty sure we can make it work in a way that Greg was asking for it (I also agree with the request)
Can you post here what the compiler error is when you change the code so that the credentials and the port are coming from a variable instead of being magic strings?

petermetz (changes requested):
Minor nit pick, but looks ready to go otherwise.

packages/cactus-test-tooling/src/main/typescript/iroha/iroha-test-ledger.ts
hxlaf: Sorry. I must admit that I made a dumb mistake! It works. Thank you for bringing up the issue again.
image
I confused this part with the HostConfig section. I said no because I tried numerous times to put a string literal in the HostConfig section but failed. I am doing a couple of more tests and will upload the fix once the tests complete.

hxlaf: Done with testing and squashing! Thank you!

petermetz (approved):

baziorek (approved):
Looks good for me (Hyperledger Iroha's site)

PR #1185, opened 2021-08-04
build(package.json): add reset script that git cleans prior to configure

petermetz:
Why though?
This is designed to eliminate possible build issues that arise because of leftover code files
somehow lurking around the project directory structure that are being ignored by git and
therefore lulling a contributor to thinking that since VSCode shows no index/staged changes
the project is in a clean slate and the build should pass all the same.
The reset script in essence is a softer version of going all out, deleting your entire
project directory and cloning it from scratch which is a lot more effort than just running
the reset script added here that has a good chance of fixing the same problems as the hard
resetting if re-clone would fix.


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #1191, opened 2021-08-09
feat(connector-quorum): remove hard dependency on keychain

### LINKED ISSUE #1161: feat(connector-quorum): remove hard dependency on keychain

TonyRowntree:
feat(quorum): remove hard dependency on keychain

Removed hard dependency on keychain for Quorum plugin.

Updated old test cases and added new test cases.

resolves #1161 

Signed-off-by: TonyRowntree <33454202+TonyRowntree@users.noreply.github.com>


packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -64,7 +67,7 @@ export interface IPluginLedgerConnectorQuorumOptions
   rpcApiHttpHost: string;
   logLevel?: LogLevelDesc;
   prometheusExporter?: PrometheusExporter;
-  pluginRegistry: PluginRegistry;
+  pluginRegistry?: PluginRegistry;
```
petermetz, line 70: @TonyRowntree The plugin registry itself does not need to be optional because the API server is guaranteed to inject it to all plugin constructor options anyway (even if the options interface of a plugin doesn't actually ask for it, it's always there at runtime).

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -100,7 +103,7 @@ export class PluginLedgerConnectorQuorum
     Checks.truthy(options, `${fnTag} arg options`);
     Checks.truthy(options.rpcApiHttpHost, `${fnTag} options.rpcApiHttpHost`);
     Checks.truthy(options.instanceId, `${fnTag} options.instanceId`);
-    Checks.truthy(options.pluginRegistry, `${fnTag} options.pluginRegistry`);
+    //Checks.truthy(options.pluginRegistry, `${fnTag} options.pluginRegistry`);
```
petermetz, line 106: @TonyRowntree The plugin registry itself does not need to be optional because the API server is guaranteed to inject it to all plugin constructor options anyway (even if the options interface of a plugin doesn't actually ask for it, it's always there at runtime).

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -111,7 +114,7 @@ export class PluginLedgerConnectorQuorum
     );
     this.web3 = new Web3(web3Provider);
     this.instanceId = options.instanceId;
-    this.pluginRegistry = options.pluginRegistry;
+    this.pluginRegistry = options.pluginRegistry as PluginRegistry;
```
petermetz, line 117: @TonyRowntree The plugin registry itself does not need to be optional because the API server is guaranteed to inject it to all plugin constructor options anyway (even if the options interface of a plugin doesn't actually ask for it, it's always there at runtime).

petermetz (changes requested):
@TonyRowntree A few nits, will have time to review more a little later hopefully!

packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
```diff
@@ -686,6 +872,40 @@
...
         },
+        "/api/v1/plugins/@hyperledger/cactus-plugin-ledger-connector-quorum/invoke-contract-no-keychain": {
+            "post": {
+                "x-hyperledger-cactus": {
+                    "http": {
+                        "verbLowerCase": "post",
+                        "path": "/api/v1/plugins/@hyperledger/cactus-plugin-ledger-connector-quorum/invoke-contract-no-keychain"
+                    }
+                },
+                "operationId": "invokeContractV1NoKeychain",
```
petermetz, line 883: @TonyRowntree The `V1` suffix should always be at the end of the operation ID

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -335,6 +451,172 @@ export class PluginLedgerConnectorQuorum
     }
   }
 
+  public async invokeContractNoKeychain(
```
petermetz, line 454: @TonyRowntree Please look into it if smaller portions of this method could be extracted into common methods and then those methods could be shared between this method and the invokeContract method as well.

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -268,6 +289,101 @@ export class PluginLedgerConnectorQuorum
         contractJSON.networks[networkId].address,
       );
       this.contracts[contractName] = contract;
+    } else if (
+      !isWeb3SigningCredentialCactusKeychainRef(req.signingCredential)
+    ) {
+      const networkId = await this.web3.eth.net.getId();
```
petermetz, line 295: @TonyRowntree Could we invoke here the newly added method below (the one that also serves the separate endpoint for the no keychain option)?

packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
```diff
@@ -89,9 +89,7 @@
                 "type": "object",
                 "required": [
                     "type",
-                    "ethAccount",
-                    "keychainId",
-                    "keychainEntryKey"
+                    "ethAccount"
```
petermetz, line 92: @TonyRowntree I don't get this removal here, could you elaborate?

petermetz (changes requested):
@TonyRowntree Let's talk more about the naming of the endpoints. What I'm thinking is that we should focus on the name to show what the endpoint is, vs. just saying what it isn't, meaning that we can have other types of credentials apart from keychain so it's not binary but instead one of many options and just saying "this is the no-keychain endpoint" doesn't convey the information of which one is it.

Decision to make: Do we make endpoints for each web3 signing credential type that we support? I'm starting to feel like we should.

Decision 2: Do we make endpoints for each contract providing mode (e..g contract is in the keychain, contract is in the request, contract is somewhere else that's coming in the future, etc.)

Decision 3: What do we do about the permutations if we say yes to both 1) and 2) will the number of endpoints start to blow up with a combinatoric explosion in the future if we need to add one more aspect to the requests that multiples the endpoints yet again.

packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
TonyRowntree: I also don't get this removal, must have been by accident. I will revert the change.

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
TonyRowntree: Note to self - Error handling doesnt need to be for each signing creditial, if the signing credential keychain then throw general error saying needs to be signed via keychain.

TonyRowntree: Next step - break up each signing endpoint into seperate endpoint, shovel out the endpoint code into common methods. Also removing the big web3 endpoint and invoke the smaller endpoints in the test cases instead.

petermetz:
@TonyRowntree Is this ready for review?

TonyRowntree:
> @TonyRowntree Is this ready for review?

is is ready now

petermetz (approved):
Waving this through for now to eliminate dependency for other PRs.
@TonyRowntree Will need to refactor according to comments in the next PR where the web3 signing credentials are handled via separate endpoints.

takeutak (approved):

petermetz:
@TonyRowntree Did you modify something on the branch permissions of your fork? I'm trying to update it but says

> Update branch attempt failed
Couldn't update "rm-keychain-dependancy-quorum": At least 1 approving review is required by reviewers with write access.

TonyRowntree:
> @TonyRowntree Did you modify something on the branch permissions of your fork? I'm trying to update it but says
> 
> > Update branch attempt failed
> > Couldn't update "rm-keychain-dependancy-quorum": At least 1 approving review is required by reviewers with write access.

I did add some branch protection. I have disabled it temporarily if you would like to try again

PR #1193, opened 2021-08-09
refactor: remap ts typings from types to lib

Leeyoungone:
This was necessary because if we import JSON files and
then re-use their schema as types, then the d.ts files
will need to have access to said JSON file which does
not happen if the typings and the actual .js code are
in separate directories because then the relative paths
are not working as they should.

Moving the lib and the types together means that now
the d.ts files can import with the same relative paths
that the orignial source code (.ts files) can as well.


petermetz (changes requested):
@Leeyoungone Apologies I just noticed that the new packages that were added to the main branch in the meantime do not have the changes necessary (I found them by searching for `dist/types`)
Could you please update the PR to change these as well?

dist-types-vs-lib-screenshot-from-2021-08-09 14-10-36

Leeyoungone:
It's all good, thanks for pointing out the new files. I'm planning to change the carbon-accounting-app-dummy-infrastructure.ts file in PR 1145 to make things organized. The other ts files have been changed for this PR!

petermetz (approved):

petermetz:
@Leeyoungone Thank you for the updates, LGTM!

takeutak (approved):

PR #1198, opened 2021-08-10
refactor(plugin-registry): getOneById to use instanceId #1197

### LINKED ISSUE #1197: refactor(plugin-registry): getOneById to use instanceId
### Description

Right now it uses the package name under the hood which is misleading.

Relevant code:

```typescript
  /**
   * The main difference between this method and `findOneByPackageName` is that this throws an Error if there was nothing to
   * return. It is recommended to use this method over `findOneByPackageName` if you have a hard dependency on a certain
   * plugin being loaded for your code.
   *
   * @param packageName The package name of the plugin that you are looking to obtain an instance of from the registry.
   * @throws If there is no plugin in the registry by the package name specificed.
   */
  public getOneById<T extends ICactusPlugin>(packageName: string): T {
    return this.findOneByPackageName(packageName).orElseThrow(
      () => new Error(`Plugin ${packageName} not present in registry`),
    ) as T;
  }

  public findOneByPackageName<T extends ICactusPlugin>(
    packageName: string,
  ): Optional<T> {
    const plugin = this.getPlugins().find(
      (p) => p.getPackageName() === packageName,
    );
    return Optional.ofNullable(plugin as T);
  }


```

petermetz:
The getOneById function will now use the instance ID
for lookups instead of the package name.

Also added a couple of extra convenience methods to make
them more consistent in general and increased the
test coverage of the plugin registry.

Fixes #1197


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #1201, opened 2021-08-10
fix(plugin-consortium-manual): drop repo constructor arg #1199

### LINKED ISSUE #1199: fix(plugin-consortium-manual): drop repo constructor arg
### Description

The constructor definitions of plugins must always contain an options object that is JSON serializable and the consortium repository is not. 
This leads to crashes when you instantiate the plugin through the API server's static configuration and then try to use it because in that scenario there is no consortium repo object in the options.
The second problem is that there was no validation added at the time either so the constructor does not check in the end if there is a repo object available or not.

petermetz:
Removes the non-serializable consortiumRepo argument from the
constructor of the consortium plugin manual class's options.

Refactors the constructor and the internals of the plugin  to initialize
the consortium repo from the consortium database at runtime
instead of expecting it passed in via the constructor.
Refactors the internal code previously using the options.consoritumRepo
object to use this.repo instead which is what gets initalized in
the constructor as explained above.

All this leads to equivalent functionality but less boilerplate and now
thanks to this the plugin can be (should be - more tests needed)
initialized by the API server purely based on the static configuration
file when necessary.

Fixes #1199

### DIFF
--- a/examples/cactus-example-supply-chain-backend/src/main/typescript/supply-chain-app.ts
+++ b/examples/cactus-example-supply-chain-backend/src/main/typescript/supply-chain-app.ts
@@ -15,7 +15,7 @@ import {
   LedgerType,
 } from "@hyperledger/cactus-core-api";
 
-import { PluginRegistry, ConsortiumRepository } from "@hyperledger/cactus-core";
+import { PluginRegistry } from "@hyperledger/cactus-core";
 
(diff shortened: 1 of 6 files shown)


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #1203, opened 2021-08-10
fix(prometheus): metrics.ts leaks to global registry #1202

### LINKED ISSUE #1202: fix(prometheus): metrics.ts leaks to global registry
**Describe the bug**

The metrics.ts file that we have for all the plugin's metrics constructs a new Metrics object at import time which then defaults to registering the metric in the global registry of `prom-client` by default.

This causes crashes when separate versions of the same metrics.ts file are imported in a scenario were the API server imports plugins from a different directory (this issue is coming from the branch where I'm working on plugin sandboxing via the live-plugin-manager).

**To Reproduce**

Run this test: packages/cactus-cmd-api-server/src/test/typescript/integration/remote-plugin-imports.test.ts
With different plugin versions being present and then watch it crash with prom-client complaining that you are duplicating the metrics in the global scope.

**Expected behavior**

No crash.

**Logs/Stack traces**

N/A

**Screenshots**

NA

Dev machine

NA 

0.7.0

None, just the keychain vault plugin.

**Additional context**

NA

petermetz:
fix(prometheus): metrics.ts leaks to global registry #1202
    
1. Specified a `register` property of the gauges as an empty
array so that it does not pollute the global namespace. This
is how this is supposed to be done as per the docs of prom-client.

2. Once the change from 1) took place, the issue became that
the metrics gathering code was still trying to hit up the
global scope for the metrics, e.g. calling the get metrics
methods directly on the promClient object instead of the
registry that we create for each prmoetheus exporter object
separately. So a little additional refactor ensued to fix this
as well by making sure that we grab a reference of the registry
object at construction time and then re-use that wherever needed
instead of going through the global promClient object.

3. Added missing .startMetricsCollection calls in the plugin
constructors to ensure that the prometheus exporter object
gets initialized properly (this is where the registry gets
created as well so without this there are crashes happening
when one tries to access the metrics through the registry)

Why though?
The problem was that the metrics.ts file that we have for all the
plugin's metrics constructs a new Metric (Gauge) object at import
time which then defaults to registering the metric in the global
registry of prom-client by default.

The latter was causing crashes when separate versions of the same
metrics.ts file are imported in a scenario were the API server
imports plugins from a different directory (this issue is coming
from the branch where I'm working on plugin sandboxing via the
live-plugin-manager).
    
Fixes #1202

### DIFF
--- a/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
+++ b/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
@@ -120,6 +120,7 @@ export class ApiServer {
         pollingIntervalInMin: 1,
       });
     }
+    this.prometheusExporter.startMetricsCollection();
     this.prometheusExporter.setTotalPluginImports(this.getPluginImportsCount());
 
     this.log = LoggerProvider.getOrCreate({
--- a/packages/cactus-cmd-api-server/src/main/typescript/prometheus-exporter/metrics.ts
+++ b/packages/cactus-cmd-api-server/src/main/typescript/prometheus-exporter/metrics.ts
@@ -4,6 +4,7 @@ export const K_CACTUS_API_SERVER_TOTAL_PLUGIN_IMPORTS =
   "cactus_api_server_total_plugin_imports";
 
 export const totalTxCount = new Gauge({
+  registers: [],
   name: K_CACTUS_API_SERVER_TOTAL_PLUGIN_IMPORTS,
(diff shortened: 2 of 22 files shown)


petermetz:
cc: @jagpreetsinghsasan

jagpreetsinghsasan:
LGTM

jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #1205, opened 2021-08-11
test(cmd-api-server): randomize port in config service test

petermetz:
If you don't specify port zero for the API and the cockpit then they
will default to 3000 and 4000 instead which is bad for
parallel test execution where we need guarantees not to
use potentially conflicting port numbers.


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #1219, opened 2021-08-13
test(tools): besu multi-party AIO container image

petermetz:
Works with a modified (forked) version of quorum-dev-quickstart
See https://github.com/ConsenSys/quorum-dev-quickstart/pull/72

On the fork we added an extra commit that removes the usage of
tput (which is a low impact change because all
it is being used for is to make text in the terminal bold for cosmetic
purposes).

This is published on the public registry as:
ghcr.io/hyperledger/cactus-besu-all-in-one-multi-party:2021-08-13--private-tx


petermetz:
cc: @travis-payne

jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #1220, opened 2021-08-14
ci: remove retries from ci.sh

petermetz:
The tests are still not 100% stable, but a lot more stable and
the retries might not be needed anymore.


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #1225, opened 2021-08-16
fix(examples): front-end packages missing browserify polyfills #1224

### LINKED ISSUE #1224: fix(examples): front-end packages missing browserify polyfills
**Describe the bug**

```sh
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "path": require.resolve("path-browserify") }'
        - install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "path": false }
```

petermetz:
The webpack build done by the Angular CLI (ng) was broken due to a few
different issues combined together:
1. After upgrading to Angular v12 IE 11 is no longer supported which
necessitated the explicit configuration in the browserlist rc config
file accordingly so that the compilation process does not try to support
a browser for which the build is broken to begin with.
2. polyfills had to be added via customizing the webpack configuration
used by the Angular CLI internally which is now being done by using a
custom builder as specified in the angular.json file. This is the new
method of dealing with these kind of issues after the ng eject command
was deprecated a while back. The webpack.config.overrides.js files
that are beind added in this commit contain only the overrides (as the
name suggests) not the entire webpack configuration that the Angular
CLI uses internally to build the front end packages.

Fixes #1224


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1243, opened 2021-08-18
feat(connector-fabric): add support for vault transit secret engine

### LINKED ISSUE #1212: proposal for fabric-secure-connector
## Background

One question that every person designing a HL fabric based solution ask is : **What the hack should I do with private key and certificates of clients?** . Although `fabric-sdk` provide two different kind of approaches.

1. storing client's private key and certificate together on some database (which implements wallet interface). For this approach a rough flow goes something like this
     - client send request to `org's application` server to `invoke` or `query` chaincode using a give `KEY`.
     - `org's application` upon receiving the request , it fetches `private key` and `certificate` corresponding to `KEY`.
     - uses the fetched private key to sign fabric `message` in order to query or invoke chaincode. <br>NOTE : developer of the application has to implement authentication for client's `KEY`

2. store private key in HSM , but in this approach `application` and `HSM` have to be present on same physical machine (i.e `HSM` should directly be connected with `machine` running the application). This approach is mostly used for developing client's desktop application, which will be responsible for communicating with fabric peers.

- Major Advantage in First approach : clients won't be responsible for communicating with fabric network
- Major Disadvantage in First approach : all private key are stored in single database which leads to `money pot` for hackers to exploit .
- Major Advantage in Second approach : client's private key are never exposed to `org's` application.
- Major Disadvantage in Second approach : clients are responsible for communicating with fabric network

**Now comes the `fabric-secure-connector`** which only brings the advantages from both the approaches. Secure Fabric connector provides a solution to the fabric organization for managing their client's private key such that the client's private key is never brought to `Node Server` for singing. Currently `fabric-secure-connector` provide a option of keeping client's private key into `Vault Transit Engine` or in browser `Extension` (TODO).

fabric-secure-connector

Broadly this package will support two kind of identity 
1. Key stored with vault transit engine like server. (which provide cryptography as service)
2. key stored with client's extension and signing will be done via various transport connection (ws , gRPC)

Another component , is browser `extension` for
- Storing private key (elliptical curve key of size 256 and 384)
- singing message digest sent to it via transport layer.

**NOTE : design of extension is TBD**

FAQ : 
1. Can this not be implemented in `cactus-plugin-ledger-connector-fabric` ?<br>Ans : Can be , but will lead to lot of breaking changes. developer can choose `cactus-plugin-ledger-connector-fabric` for the `first approach` and `fabric-secure-connector` to support multiple type of identity. 

2. Is their any use case of it ?<br>Ans: currently we are trying to implement `vault` and `ws` identity  for `blockchain-carbon-accounting` project (under hyperledger labs)
3. Any Blockage that can be faced during the implement `fabric-secure-connector`?
<del>Ans : currently `fabric-sdk-node` support sync singing of message , [here](https://github.com/hyperledger/fabric-sdk-node/blob/25dbf8a8050aa121be85a67554515e2aa3c6e9dd/fabric-common/types/index.d.ts#L64) </del>
```ts
export interface ICryptoSuite {
         //........
	sign(key: ICryptoKey, digest: Buffer): Buffer;
}
```
<del> but since for `Identity` supported by this package will require `sign` method to be as `async` , so current `fabric-sdk-node` cannot be directly used. To solve this I have already opened a PR to `fabric-node-sdk` [here](https://github.com/hyperledger/fabric-sdk-node/pull/479) . In meanwhile they merge the PR and release version `2.2`, I have pushed a npm package of my fork. </del> PR merged

4. Where can we find an implementation of it?<br>Ans: https://github.com/Zzocker/blockchain-carbon-accounting/tree/secureFabric/secure-fabric . this support `vault` identity for now.

Thank you

@petermetz @sichen1234 @brioux @knagware9 @udosson

Below are technical details of how I am planning to implement 
 
## Top Level exported `class` and `interface`

```ts
export enum IdentityProvidersType {
     // identity by default provided by fabric-sdk-node wherein private key and certificate
     // stored together (First approach)
     Default = 'Default-X.509'


     // vault identity provider wherein private key are store with vault transit engine
     // and certificate are store in certData store
     Vault = 'Vault-X.509',

     // [ TRANSPORT ] WebSocket identity provider wherein private key are store with client inside the extension
     // signing of digest are done by sending of data as websocket message through webSocket connection between
     // server and client extension
     // certificate are store in certData store
     WebSocket = 'WS-X.509',

     // [ TRANSPORT ] gRPC identity provider wherein private key are store with client inside the extension
     // signing of digest are done by sending of data as gRPC streaming via gRPC bi-directional connection 
     // between server and client extension
     // certificate are store in certData store
     gRPC = 'gRPC-X.509',
}
```

```ts
// every identity  type and certificate data will extends this interface
interface IIdentity{
     // key used for storing identity data with cert datastore
     key:string
     type:IdentityProvidersType
} 
```

```ts
// IVaultIdentity represents a vault client.
export interface IVaultIdentity extends IIdentity{
     keyName:string
     token:string
}

// IWebsocketIdentity represents a websocket client.
// message signing will be done by back and forth message between 
// server and client over weConn
export interface IWebsocketIdentity extends IIdentity{
     wsConn:connection
}

// IGRPCIdentity represents a bi-directional gRPC client.
// message signing will be done by back and forth message between 
// server and client over biConn
export interface IGRPCIdentity extends IIdentity{
     biConn:connection
}

// IIdentityData : data that will be stored with cert datastore
// with key as client's commonName (from X509 certificate) and value as following field
interface IIdentityData extends IIdentity{
     credentials: {
          certificate: string;
          // if identity type is IdentityProvidersType.Default
          privateKey?:string;
     };
     mspId: string;
}
```

```ts

export interface ISecureFabricConnector{
     // array of identity types that application is going to support
     // eg [IdentityProvidersType.Vault , IdentityProvidersType.Default IdentityProvidersType.WebSocket]
     // this will accept client request with vault , default or websocket for signing fabric messages
     supportedIdentity:IdentityProvidersType[]


     // vault server config if Vault identity is support 
     vaultOptions?:{endpoint:string,transitEngineMountPath:string}

     // for registering client : NOTE /register endpoint should not be exposed to client
     // rather this endpoint is for org's admin
     registrar : {
          certificate:string
          mspId:string
          // if privateKey is provided , this private will be used for signing
          privateKey?:string

          // if provided , application will use private key of registerer stored with vault
          // to register and revoke client
          vaultKey?:{token:string,keyName:string}
     }
     // usual field required in fabric-sdk-node's GatewayOptions
    connectionProfile:object
    tlsInfo?: {
        certificate: string;
        key: string;
    };
    discovery?: DiscoveryOptions;
    eventHandlerOptions?: DefaultEventHandlerOptions;
    queryHandlerOptions?: DefaultQueryHandlerOptions;
    'connection-options'?: any;
}
```
```ts
export class ISecureFabricConnector{
     constructor(ISecureFabricConnector)

     /**
      * @method transact invoke/query fabric chaincode
      * @param caller client identity
      * @param channel on which chaincode is committed
      * @param ccName : fabric chaincode name
      * @param method supported by chaincode
      * @param args , method specific arguments
      */
     transact(type:'query'|'invoke',caller:IIdentity,channel:string,ccName:string,method:string,...args:string[]):Promise<Buffer>

     /**
      * @method enroll a already registered client
      * @param caller client identity
      * @param request for enrollment
      */
     enroll(caller:IIdentity,request:{enrollmentID:string,enrollmentSecret:string}):Promise<void>

     /**
      * @method rotateKey will rotate client key and store newly enrolled certificate in cert datastore
      * 
      */
     rotateKey(caller:IIdentity):Promise<void>
}
```

Zzocker:
- Add support for signing fabric transactions using key stored with vault transit engine
- Add support for following methods inside `PluginLedgerConnectorFabric`  class 
    - enroll 
    - register
    - rotateKey
 - And extended `transact` methods to support both `default-X.509` and `vault-X.50` identity types

Resolve #1212 

@petermetz @sichen1234 @brioux @knagware9 @udosson


packages/cactus-plugin-keychain-vault/src/test/typescript/unit/transit-vault.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+    t.end();
+  });
+
+  t.test("rotateKey", async (t: Test) => {
+    const client = new PluginKeychainVault(options);
+
+    {
+      // rotate EC256 key
+      await client.transitRotateKey(testECP256);
```
petermetz, line 166: @Zzocker Is there any way to verify that the rotation has indeed happened? I'm thinking maybe we could perform some signing operation after the rotation and verify that the signature was done with the new key.
Please do let me know if this is a huge rabbit hole, then we can reevaluate it. If it's not too much trouble then it would be really nice to have some assertions like above.

packages/cactus-plugin-keychain-vault/src/test/typescript/unit/transit-vault.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    logLevel: "DEBUG",
+    keychainId: uuidv4(),
+    endpoint: vaultHost,
+    token: testToken,
+    transitSecretMountPath: mountPath,
+    instanceId: uuidv4(),
+  };
+
+  t.test("newKey", (t: Test) => {
+    const client = new PluginKeychainVault(options);
```
petermetz, line 63: @Zzocker Nit: Please name it something other than `client` because once we create the REST API endpoints for the same thing, then the tests will also have to be written here to use the API client object and at that point it would become confusing to see two client variables that are different slightly. My recommendation is to call it either `plugin` or `keychain` or `keychainPlugin` whichever you prefer.

packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
```diff
@@ -272,14 +305,134 @@ export class PluginKeychainVault implements IPluginWebService, IPluginKeychain {
...
+    try {
+      const resp = await this.transitBackend.read("keys/" + keyName);
+      this.log.debug(`Response from Vault: %o`, JSON.stringify(resp));
+      if (resp?.data?.latest_version && resp?.data?.keys) {
+        if (
+          !Object.values(VaultTransitKeyType).includes(
+            resp.data.type as VaultTransitKeyType,
+          )
+        ) {
+          throw new Error("invalid public key type");
```
petermetz, line 410: @Zzocker Please also include the `fnTag` in the error message. Rationale: If someone calls the API and this message will be the only thing they see in an error response, we should provide them with a little more context on where/what went wrong. In the same line of thinking I also highly recommend to include the `keyName` in the error message.

packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
```diff
@@ -272,14 +305,134 @@ export class PluginKeychainVault implements IPluginWebService, IPluginKeychain {
...
+   * @description will generate a a private key, new if key doesn't exists , and does nothing if key already exists
+   * @param keyName label of the key
+   * @param keyType
+   */
+  async transitNewKey(
+    keyName: string,
+    keyType: VaultTransitKeyType,
+  ): Promise<void> {
+    if (!this.transitBackend) {
+      throw new Error("transit vault secret engine not supported");
```
petermetz, line 381: @Zzocker Same here regarding `fnTag`, see below for details. (in my other comment)

packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
```diff
@@ -272,14 +305,134 @@ export class PluginKeychainVault implements IPluginWebService, IPluginKeychain {
...
+        const base64Sig = (resp.data.signature as string).split(":")[2];
+        return Buffer.from(base64Sig, "base64");
+      } else {
+        throw new Error(
+          `${fnTag}: Invalid response received from Vault. Expected "resp.data.signature" property chain to be truthy`,
+        );
+      }
+    } catch (error) {
+      if ((error as any)?.response?.statusCode === HttpStatus.NOT_FOUND) {
+        throw new Error("signing key not found");
```
petermetz, line 365: @Zzocker Include `keyName` and `fnTag` in the error message to provide more context. Further details provided below in my other comment.

packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
```diff
@@ -97,15 +109,30 @@ export class PluginKeychainVault implements IPluginWebService, IPluginKeychain {
     this.endpoint = this.opts.endpoint;
     this.apiVersion = this.opts.apiVersion || "v1";
 
-    this.kvSecretsMountPath =
-      opts.kvSecretsMountPath || K_DEFAULT_KV_SECRETS_MOUNT_PATH;
-    this.log.info(`this.kvSecretsMountPath=${this.kvSecretsMountPath}`);
+    const kvSupport = opts.kvSecretsMountPath !== undefined;
+    const transitSupport = opts.transitSecretMountPath !== undefined;
+    if (!kvSupport && !transitSupport) {
+      throw new Error("require at least kv mount path or transit mount path");
```
petermetz, line 115: @Zzocker Please include the `fnTag` in the error message as well

packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
```diff
@@ -97,15 +109,30 @@ export class PluginKeychainVault implements IPluginWebService, IPluginKeychain {
...
+    }
+    if (kvSupport) {
```
petermetz, line 117: @Zzocker Is it possible that these three if conditions could be refactored into a single `if {  } if else {} else { throw }` where by the end of it the `backend` field is either definitely assigned (e.g. no need for the `?` optional marker) OR we have thrown an exception if no transit and no kv support was provided.
I'm just asking this because I'm not sure if this would cause complications in the rest of the implementation or not. If it would, then let's discuss further. If it wouldn't then I think it would be a great simplification and also we could guarantee this way that if the vault plugin instance was created, then it definitely has a valid `backend` so no need to check for it in all the methods that use the `backend` variable.

petermetz (changes requested):

packages/cactus-plugin-keychain-vault/src/test/typescript/unit/transit-vault.test.ts
Zzocker: I have already added this case inside integration test. [here](https://github.com/Zzocker/cactus/blob/bb31bb73bc58126ad30657eb0281c95499efd822/packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts#L429)

```ts
      // check if signature is signed with new key
      const verify = createVerify("sha256");
      verify.update(digest);
      t.false(verify.verify(pub, sigRotated), "signature should not be valid");
```

This test is to check whether signature is signed with rotated key or previous key, This test works like once key has been rotated the generated signature should be valid against previous key  .

Zzocker:
@petermetz I played with `crypto` package in my machine for both version `v12` and `v15` and it seems this package has version dependent methods . so instead I tried to  use `jsrsasign` package which is very rich cryptography package in `js` . And also I would be using `jsrsasign` package for generating CSR and encoding signature into a fabric understandable format in order to support vault to sign fabric messages .

But here is error message which got while running `yarn run configure`
```ts
/home/pritam/go/src/github.com/hyperledger/cactus/packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts:32:19 - Unknown word (KJUR)
/home/pritam/go/src/github.com/hyperledger/cactus/packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts:32:32 - Unknown word (jsrsasign)
/home/pritam/go/src/github.com/hyperledger/cactus/packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts:395:49 - Unknown word (KJUR)
315/540 ./packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts 35.71ms X
/home/pritam/go/src/github.com/hyperledger/cactus/packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts:34:19 - Unknown word (KJUR)
/home/pritam/go/src/github.com/hyperledger/cactus/packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts:34:32 - Unknown word (jsrsasign)
/home/pritam/go/src/github.com/hyperledger/cactus/packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts:409:26 - Unknown word (KJUR)
/home/pritam/go/src/github.com/hyperledger/cactus/packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts:424:26 - Unknown word (KJUR)
/home/pritam/go/src/github.com/hyperledger/cactus/packages/cactus-plugin-keychain-vault/src/test/typescript/integration/plugin-keychain-vault.test.ts:432:26 - Unknown word (KJUR)
```

these error are from `CSpell` but name `KJUR` is a namespace inside `jsrsasign` package . How to solve this ??

packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
Zzocker: sounds reasonable , I am using two different backend (one for each `kv` and `transit`) because , I wanted to take advantage of `node-vault` package functionality of specifying montPath in `pathPrefix` filed while creating the `backend`. so that I won't have to implement my own function for creating `path` just like you did [here](https://github.com/hyperledger/cactus/blob/4cc6f2c5d3345a7af8cc04b9218c38d7670872a8/packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts#L219) .

But I think if adding one liner function for creating transitPath can simplify `if-else` in constructor, than I should add that function and  simplify the constructor

Zzocker:
@petermetz I am closing this and going to implement vault transit client inside fabric connector where I need. I think Having this logic inside keychain won't be adding any value to cactus.

petermetz:
@Zzocker Fine by me, it is hard to know in advance where a piece of code would fit best so no harm done if you replace it elsewhere.
Also note: You don't need to close this and open a new one, you could just force push your local changes with the refactor later, the PR can stay in place (and maybe rename it to `feat(connector-fabric): add support for vault transit secret engine` )

Zzocker:
@petermetz `feat(connector-fabric): add support for vault transit secret engine` incoming in two days :slightly_smiling_face:

takeutak (approved):

PR #1244, opened 2021-08-19
feat(besu): support besu v21.1.6

### LINKED ISSUE #982: feat(connector-besu): support Besu v21.1.6
### Description

The connector should support Besu v21.1.x. 
We need actual test coverage and the AIO image updated for said test coverage so that we can credibly claim said support instead of just saying "well it should because I read the release notes of Besu v21.1.6 for example and nothing struck me as a change that would mess with our connector's ability to work with it".

### Acceptance Criteria
1. Specific test case that is locked down to a 21.1.6 ledger (the AIO image) so that even if we add more things later on, this test remains frozen in time forever verifying the connector against 21.1.6. This will be extremely valuable in the long run in catching regressions.
2. Will probably have to make sure an updated version of the Besu AIO image is published

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo @petermetz @arnab-roy

AzaharaC:
Resolve #982


tools/docker/besu-all-in-one/Dockerfile_v21.1.x
```diff
@@ -0,0 +1,30 @@
+ARG BESU_VERSION=21.1.6
```
petermetz, line 1: @AzaharaC I know this file naming pattern has been precedent in the Fabric AIO image, but it is something that I wanted to steer us away from because VSCode doesn't recognize Dockerfile's with custom names (syntax helper).
Could you refactor this in a way that instead of the file being `Dockerfile_v21.1.x` there is a directory called `v21_1_x` and then three's a `Dockerfile` in there with the contents of this file that I'm commenting on. Sorry for the nit picking as usual.

petermetz (changes requested):

tools/docker/besu-all-in-one/Dockerfile_v21.1.x
AzaharaC: Don't worry Peter, I have already changed this.

petermetz (changes requested):
@AzaharaC LGTM, before merging, please publish the container image built from this branch onto the official repo with the fully qualified tag of
`ghcr.io/hyperledger/cactus-besu-21-1-6-all-in-one:2021-08-24--feat-1244`
and then update the test cases to use that as well.
Make sure to also update the `containerImageVersion` variables, leaving them as `latest` is going to destabilize the tests over time so it has to be the exact tag/version.

AzaharaC:
@petermetz I changed the tests with the new tag of the image. When I run the tests, I get an error 500.
I have checked the repository and cannot find this image tag. (https://github.com/orgs/hyperledger/packages?page=1)
Is it possible that you forgot to upload the image or that it is not accessible?

petermetz:
> Is it possible that you forgot to upload the image or that it is not visible?

@AzaharaC I didn't upload it, wanted to see first if you could do it because I realized the other day that the permissions are handled different on GHCR than on DockerHub (because Travis reported that he was able to publish official images himself)

petermetz (commented):
Comment above ^^

AzaharaC:
@petermetz You are right! I was able to upload the image myself. I have already updated the PR.

petermetz (approved):

takeutak (approved):

Issue #1271, opened 2021-08-25
fix(package.json): remove deprecated and no longer working build-dev scripts


m-courtin:
**Describe the bug**

Newest version of Cactus package.json is containing deprecated scripts which are no longer working as such.
Examples:

- build:dev:plugin-ledger-connector-besu
- build:dev:plugin-ledger-connector-quorum
-  etc.

**Expected behavior**

No longer working scripts need to be removed from package.json

PR #1273, opened 2021-08-25
fix(config): remove no longer working scripts from package.json

m-courtin:
Closes: #1271


petermetz (approved):
LGTM

Issue #11, opened 2019-11-03
 Avoid MVCC_READ_CONFLICT when adding signatures 


jonathan-m-hamilton:
**Current state** 
Signatures on Fabric are stored in a single key-value store and will cause MVCC_READ_CONFLICT if modified concurrently  

**Desired state** 
Change on-chain storage and according functions to avoid conflicts and allow for concurrent modification of different signatures.

denis-yu-glotov:
My attempts commits are here. https://github.com/denis-yu-glotov/blockchain-integration-framework/tree/denis/concurrency.
Attempting to test it, I hit https://github.com/hyperledger-labs/blockchain-integration-framework/issues/56#issuecomment-557647247 and was trying to fix it myself (actually, I discovered that issue a day later). 
Eventually, I tested Quorum network (both the version 2.1.1 that we are using now and latest 2.3.0) and I see that is unable to make transactions through websockets. So I agree with Peter, we need to switch back to http, because
* http is always used in the wild (I have never seen anyone using websockets on github projects),
* we will not have this `web3_timestamp_fix.patch` mess anymore,
* the reason why we used ws from the beginning was that our early developers believed ws is more advanced to http and it allows to pass more parameters in a single call. I argued that ws is never used in public. And now it is just broken.

PR #1274, opened 2021-08-26
feat(tools): substrate test ledger

RafaelAPB:
Substrate-based blockchains like Polkadot are becoming increasingly relevant. However, Cactus does not have tools to support developing Substrate-based connectors, such as a test ledger.

This draft PR provides a first attempt at defining a Substrate test ledger, that allows to programmatically instantiate the Substrate node template (https://github.com/substrate-developer-hub/substrate-node-template).
Thus, it contains a Dockerfile with the test ledger and the test-ledger file that contains the execution logic.

I would appreciate feedback on the tools/docker/substrate-all-in-one/Dockerfile and packages/cactus-test-tooling/src/main/typescript/substrate-test-ledger/substrate-test-ledger.ts files prior to merging.

Notes:
At the moment a bug occurs in the Dockerfile at RUN   rustup update nightly

info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: latest update on 2021-08-26, rust version 1.56.0-nightly (0afc20860 2021-08-25)
info: downloading component 'rust-std' for 'wasm32-unknown-unknown'
info: downloading component 'cargo'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: removing previous version of component 'rust-std' for 'wasm32-unknown-unknown'
info: rolling back changes
error: could not rename component file from '/usr/local/rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-unknown/lib' to '/usr/local/rustup/tmp/jxirtjwr7gf70a8r_dir/bk'
error: caused by: other os error
error: backtrace:
error:    0: error_chain::backtrace::imp::InternalBacktrace::new
   1: rustup::utils::utils::rename
   2: rustup::dist::component::transaction::Transaction::remove_dir
   3: rustup::dist::component::components::Component::uninstall
   4: rustup::dist::manifestation::Manifestation::update
   5: rustup::dist::dist::update_from_dist_
   6: rustup::install::InstallMethod::install
   7: rustup::toolchain::DistributableToolchain::install_from_dist
   8: rustup::cli::rustup_mode::update
   9: rustup::cli::rustup_mode::main
  10: rustup_init::main
  11: std::rt::lang_start_internal::{{closure}}::{{closure}}
             at rustc/c7087fe00d2ba919df1d813c040a5d47e43b0fe7/src/libstd/rt.rs:52
      std::sys_common::backtrace::__rust_begin_short_backtrace
             at rustc/c7087fe00d2ba919df1d813c040a5d47e43b0fe7/src/libstd/sys_common/backtrace.rs:130
  12: main
  13: __libc_start_main
  14: <unknown>


RafaelAPB:
@petermetz yes, the "Create Your First Substrate Chain" one!

petermetz:
@RafaelAPB Took a peek just now I think the issue might be that you didn't set the image on GHCR to public. By default it has a visibility setting of "private" so you need to update that.
I pushed a couple fixes to your branch to make it compile.
When I run the test it goes out with `server error - Head "https://ghcr.io/v2/rafaelapb/cactus-substrate-aio/manifests/2021-08-26-first": denied` which is what makes me think that it's a container repo visibility issue. Please double check the visibility and let me know and then we can take it from there.

RafaelAPB:
@petermetz the image has not been submitted, because the build process fails. I'm trying to fix it (meanwhile @CatarinaPedreira might take this task).

petermetz:
> @petermetz the image has not been submitted, because the build process fails. I'm trying to fix it (meanwhile @CatarinaPedreira might take this task).

@RafaelAPB Oh, okay, now I get the full picture! I'll take a peek at the image build, maybe I can help.

CatarinaPedreira:
Yes, I will try to take a look at it in a bit and will report later :)

petermetz:
@RafaelAPB @CatarinaPedreira FYI: You only need the wasm-pack tool if you plan on packaging some Rust code into Javascript WASM modules to be executed by NodeJS. If this is not the case then you could just remove that part from the image definition (Dockerfile)

CatarinaPedreira:
@petermetz  @RafaelAPB Sorry for taking so long, I only got to see this issue with full attention today.
I did some research on the error and found this similar issue https://github.com/rust-lang/rustup/issues/2542. 

It seems that the problem lies in the rename() function, more specifically in the fact that docker's filesystem behavior and rustup filesystem's behavior are not compatible. 
There is no good solution yet, but the most suggested solution in this and other similar issues is uninstalling and reinstalling rustup. 
But since this image is created from paritytech/ci-linux's image (https://github.com/paritytech/scripts/blob/master/dockerfiles/contracts-ci-linux/Dockerfile), which first installs nightly, how would I do this?
Having instructions to uninstall it and reinstall rustup it in our Dockerfile seems redundant, but I'm not sure what to do.

I will research a bit more and if I don't find any better solution I think I will do that. In the mean time, if you have any ideas, please say something :)

RafaelAPB:
@CatarinaPedreira  thanks for taking this task :) Perhaps using another image? Peter's image on the Rust container could provide a good starting point, but I don't know if it has all the dependencies for the test ledger.

In the meantime it might be a good idea to open an issue at the repo responsible for the image.

CatarinaPedreira:
@RafaelAPB no problem!
In the meantime, I opened an issue at paritytech/scripts (https://github.com/paritytech/scripts/issues/339), with no reply yet.
I found another possible parent image in docker hub, that is more specific for using and testing smart contracts (https://hub.docker.com/r/paritytech/contracts-ci-linux). I'm going to test and see if this one works for our image.

CatarinaPedreira:
I'm having a bit of difficulty with this new bug:

=> ERROR [ 8/10] RUN cargo build --release                                                                                                                                     1.9s
------                                                                                                                                                                               
 > [ 8/10] RUN cargo build --release:
#11 1.196 error: could not find `Cargo.toml` in `/` or any parent directory

The image is now based in https://github.com/paritytech/scripts/blob/master/dockerfiles/contracts-ci-linux/Dockerfile, and I did not find any reference to a Cargo.toml file here. 

Anyhow, here are some things I tried:
- Created a Cargo.toml file in substrate-all-in-one directory, and used the COPY instruction to copy it to the container. 
- Added a "RUN cargo new node-template" instruction to the Dockerfile so that it creates a Cargo.toml file in the container

That didn't work. It seems that docker is not being able to find the Cargo.toml.

Also, @RafaelAPB, your docker-compose.yml is based in this one, right? (https://github.com/substrate-developer-hub/substrate-node-template/blob/master/docker-compose.yml). 
It seems that they only use the docker compose file and no dockerfile. 

On that note, I also tried changing the Cargo.toml file in substrate-all-in-one directory to be equal to the substrate one (https://github.com/substrate-developer-hub/substrate-node-template/blob/master/Cargo.toml), executed "docker-compose up" and it found the Cargo.toml, but provided this error:

Attaching to node-template-cactus
node-template-cactus | error: failed to load manifest for workspace member `/var/www/node-template/node`
node-template-cactus | 
node-template-cactus | Caused by:
node-template-cactus |   failed to read `/var/www/node-template/node/Cargo.toml`
node-template-cactus | 
node-template-cactus | Caused by:
node-template-cactus |   No such file or directory (os error 2)
node-template-cactus exited with code 101

Which makes sense since the members are not created there. But I'm not sure if I understand why I would have to create them in the first place. 

Also, @RafaelAPB, two doubts:
- In line 6 of the docker-compose.yml, shouldn't it be the image created by our Dockerfile instead of the current one?
- The instruction "ENV CACTUS_CFG_PATH=/etc/hyperledger/cactus" is repeated in the Dockerfile. Is this a mistake?

Thanks, and sorry If I'm not making a lot of sense right now - I'm going to pause a little on this and be back in a few hours :)

petermetz:
> I'm a bit of a docker noob, but from my understanding, every time we build our docker image, docker builds the parent image as well, right? So I think this is not a problem. We can just remove that specific instruction since nightly will always be updated due to the parent image. Please let me know if I'm wrong.

@CatarinaPedreira I think the parent image doesn't get re-built, just pulled as is. E.g., the state it is in is determined at the time when whoever built that parent image did so. With that said if you are referencing a moving tag like `latest` for the version of the parent image then it might happen that the image behind the `latest` tag gets overwritten with a different build over time kind of resulting in the behavior that you described as it always being rebuilt.

> Thanks, and sorry If I'm not making a lot of sense right now - I'm going to pause a little on this and be back in a few hours :)

No worries at all, it's a difficult topic from what I can tell so far! What does help with helping/debugging is if you also push the changes you made to the dockerfile so that I can try and reproduce the issues locally so I encourage you to do that every time you post a comment with an error/question/etc. :-)

CatarinaPedreira:
@petermetz Ohh, ok, that makes sense. Thank you! :) 
Yes, tomorrow I'll push the changes to this PR.

CatarinaPedreira:
@petermetz  @RafaelAPB The changes I've made are here: https://github.com/CatarinaPedreira/cactus/tree/t-odap/tools/docker/substrate-all-in-one (the commit being this one https://github.com/CatarinaPedreira/cactus/commit/0e72cbefb596c47ea00342ff9d9a18c13080c67f).

Main changes:
* I removed instructions that were redundant from the Dockerfile, such as cargo build --release and ./target/release/node-template --dev --ws-external and left them only on the docker-compose.yml (these instructions were both in the Dockerfile and in the "command" instruction of the docker-compose.yaml (wich overrides the default command in Dockerfile));
* Changed the image used by docker-compose.yaml to be the one generated by our dockerfile (hyperledger/substrate-all-in-one:latest);
* Added a cargo.toml file

The dockerfile build works ok now (removed the rustup update instruction given that the parent image always strives to contain the latest nightly version as said in https://github.com/paritytech/scripts/tree/master/dockerfiles/contracts-ci-linux).

But this new cargo.toml file (same as https://github.com/substrate-developer-hub/substrate-node-template/blob/master/Cargo.toml) is the one causing errors right now ("no such file or directory" error in my last comment), when trying to build the container (docker-compose -f tools/docker/substrate-all-in-one/docker-compose.yaml up).

It makes sense to cause this error given that the members of the workspace, described in the Cargo.toml file, are not created in our substrate-all-in-one folder (as they are created in https://github.com/substrate-developer-hub/substrate-node-template). I've tried changing it into a simpler Cargo.toml, but I get an error stating that I must have the workspaces field.
I'm not sure what our Cargo.toml file should look like here. 
@RafaelAPB Did you create a cargo.toml but didn't commit it here?

CatarinaPedreira:
A few days ago Polkadot announced https://github.com/paritytech/substrate-contracts-node which is essentially a minimal working node (based on the node template we were using) already with smart contract support, which they recommend for smart contract development and testing and thus is better for this case.
I followed the install instructions into the Dockerfile and the substrate contracts node is correctly installed (the image builds with no errors). 

I still have a bug though, when starting the node template (it says that substrate-contracts-node is not found), I'm not sure why. I posted an issue in the repo (https://github.com/paritytech/substrate-contracts-node/issues/5) and asked the question in the Polkadot discord and Elemental (will get back to you once I have the answer).

Can you try to build the dockerfile and then run it with the docker-compose.yaml in your devices to try to reproduce the error there? @petermetz @RafaelAPB  

Thank you :)

CatarinaPedreira:
The container now works nicely if we run the container directly through the image (using "docker run -t -p 9944:9944 --name substrate-contracts-node hyperledger/substrate-all-in-one:latest" to run it other than using the docker compose file).

This is a temporary fix so that I can move forward with the work, but when I have time I can come back to fix the YAML file to make the substrate-all-in-one folder more coherent with the others.

RafaelAPB:
@petermetz this PR is ready for review. Thanks for all the brilliant contributions, @CatarinaPedreira

packages/cactus-test-tooling/src/main/typescript/substrate-test-ledger/substrate-test-ledger.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  constructor(public readonly opts: ISubstrateTestLedgerOptions) {
+    const fnTag = `${this.className}#constructor()`;
+    Checks.truthy(opts, `${fnTag} arg options`);
+
+    this.publishAllPorts = opts.publishAllPorts;
+    this._containerId = Optional.empty();
+    this.imageName =
+      opts.imageName || "ghcr.io/hyperledger/cactus-substrate-test-ledger";
+    this.imageTag = opts.imageTag || "0.0.1";
```
petermetz, line 57: @RafaelAPB @CatarinaPedreira Now it's safe to set the default image tag to the one I published it as (see my message on RocketChat in response to your message the other day)

petermetz (changes requested):

packages/cactus-test-tooling/src/main/typescript/substrate-test-ledger/substrate-test-ledger.ts
```diff
@@ -0,0 +1,202 @@
...
+        Retries: 10,
+        StartPeriod: 1000000000, // 1 second
+      },
+      ExposedPorts: {
+        "9944/tcp": {}, // OpenSSH Server - TCP
+      },
+      HostConfig: {
+        AutoRemove: true,
+        PublishAllPorts: this.publishAllPorts,
+        Privileged: true,
```
petermetz, line 111: @RafaelAPB @CatarinaPedreira Usually we need `Privileged` as `true` for containers that host other containers within, internally. Is this the case with the Substrate AIO image? If there's no specific need for it I'd set it to false to adhere to the principle of least privilege.

packages/cactus-test-tooling/src/main/typescript/substrate-test-ledger/substrate-test-ledger.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    this.log.debug(`Created instance of ${this.className} OK`);
+  }
+  public getContainerImageName(): string {
+    return `${this.imageName}:${this.imageTag}`;
+  }
+  public async start(omitPull = false): Promise<Container> {
+    const docker = new Docker();
+    if (this.containerId.isPresent()) {
+      this.log.warn(`Container ID provided. Will not start new one.`);
```
petermetz, line 78: @RafaelAPB @CatarinaPedreira If this is a feature then it should be logged on a level that is less severe than warning/error. A warning level error indicates that if you see that in the logs you should look into something *may* be a problem, but here it appears that if someone specified the container ID then that is one of the expected use-cases.

petermetz (changes requested):

packages/cactus-test-tooling/src/main/typescript/substrate-test-ledger/substrate-test-ledger.ts
```diff
@@ -0,0 +1,202 @@
...
+    if (aContainerInfo) {
+      return aContainerInfo;
+    } else {
+      throw new Error(`${fnTag} no image "${image}"`);
+    }
+  }
+
+  // ./scripts/docker_run.sh ./target/release/node-template purge-chain --dev
+  protected async purgeDevChain(): Promise<void> {
+    throw new Error("TODO");
```
petermetz, line 200: @RafaelAPB @CatarinaPedreira Is this still TODO in the scope of this PR or meant to be implemented in a future PR?

packages/cactus-test-tooling/src/test/typescript/integration/substrate/substrate-test-ledger-constructor.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  const pruning = pruneDockerAllIfGithubAction({ logLevel });
+  await t.doesNotReject(pruning, "Pruning didn't throw OK");
+  t.end();
+});
+
+test(testCase, async (t: Test) => {
+  const options = {
+    publishAllPorts: true,
+    logLevel: logLevel,
+    imageName: "ghcr.io/rafaelapb/cactus-substrate-aio",
```
petermetz, line 19: @RafaelAPB @CatarinaPedreira After you updated the default image in the test ledger class specifying the image tags here will no longer be necessary and I recommend either removing it here or setting it to the same value that will be the new default)

tools/docker/substrate-all-in-one/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[profile.release]
```
petermetz, line 1: @RafaelAPB @CatarinaPedreira Where is this file being used by the container? I could not find a reference to it in the `Dockerfile`

tools/docker/substrate-all-in-one/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Usage
+```sh
+docker run -t -p 9944:9944 --name substrate-contracts-node hyperledger/substrate-all-in-one:latest
+```
+
+## Build
+
+```sh
+DOCKER_BUILDKIT=1 docker build -f ./tools/docker/substrate-all-in-one/Dockerfile . --tag hyperledger/substrate-all-in-one
```
petermetz, line 19:
```suggestion
DOCKER_BUILDKIT=1 docker build -f ./tools/docker/substrate-all-in-one/Dockerfile . --tag saio
```

@RafaelAPB @CatarinaPedreira Recommend the default local build tag to be less ambiguous with the official tag (easier to not confuse to the this way while developing new versions of the same image)

tools/docker/substrate-all-in-one/docker-compose.yaml
```diff
@@ -0,0 +1,17 @@
+# From https://github.com/substrate-developer-hub/substrate-node-template
```
petermetz, line 1: @RafaelAPB @CatarinaPedreira Just double checking, is this still operational? Has it been tested manually with the latest published official docker image?

tools/docker/substrate-all-in-one/docker-compose.yaml
```diff
@@ -0,0 +1,17 @@
...
+version: "3.2"
+
+services:
+  dev:
+    container_name: contracts-node-template-cactus
+    image: hyperledger/substrate-all-in-one:latest
```
petermetz, line 7: @RafaelAPB @CatarinaPedreira Please update the image to the latest official that is published

petermetz (changes requested):

packages/cactus-test-tooling/src/main/typescript/substrate-test-ledger/substrate-test-ledger.ts
RafaelAPB: I believe this can be done at a later stage

packages/cactus-test-tooling/src/main/typescript/substrate-test-ledger/substrate-test-ledger.ts
CatarinaPedreira: You're right - I'm changing it to this.log.debug

packages/cactus-test-tooling/src/main/typescript/substrate-test-ledger/substrate-test-ledger.ts
CatarinaPedreira: Yes, I believe so too

tools/docker/substrate-all-in-one/Cargo.toml
CatarinaPedreira: This is actually not needed anymore, I'm removing it

tools/docker/substrate-all-in-one/docker-compose.yaml
CatarinaPedreira: @petermetz This file is not used anymore, we are now running the container only using the Dockerfile. I'm going to remove it. 
I tested this with the latest image, but the error remained (a system file error). As I needed to move forward and have time constraints I moved on without the yaml, but I can go back at a later stage to try and fix this for coherence with other test ledgers :)

CatarinaPedreira:
@RafaelAPB @petermetz I applied the requested changes (except that TO-DO which is out of scope for this PR) and updated cactus-test-tooling's public-api. Hopefully the build will pass now :crossed_fingers:

CatarinaPedreira:
@petermetz @RafaelAPB My latest commits fix the errors I had on the latest build. I'm hopeful everything will work now!
Sorry to ask you again, but can you approve the workflow run again Peter?

CatarinaPedreira:
@petermetz @RafaelAPB  My latest commits fix the errors I had on the latest build. I'm hopeful everything will work now!
Can you approve the workflow run again Peter?

petermetz:
@CatarinaPedreira Please also make sure to fix the DCO check!

CatarinaPedreira:
@petermetz Done :) Is this ok to merge? I believe the builds were passing

tools/docker/substrate-all-in-one/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM paritytech/ci-linux:production
+LABEL AUTHORS="Rafael Belchior, Catarina Pedreira" 
+LABEL VERSION="2021-09-10"
+LABEL org.opencontainers.image.source=https://github.com/CatarinaPedreira/cactus
```
petermetz, line 4: @CatarinaPedreira Please replace this with a link to the upstream repo

petermetz (changes requested):
@CatarinaPedreira Cool, thank you! Just have one nit-pick left (see comments inline) once you address that please:
1. squash the changes into a single commit
2. that one commit message created after teh squash needs to have the description that you typed into the PR description already (but it's completely missing in the git commit message) 
3. Then we are good to merge

CatarinaPedreira:
@petermetz @RafaelAPB All done :)

petermetz (approved):
@CatarinaPedreira Thank you very much, LGTM!

tools/docker/substrate-all-in-one/docker-compose.yaml
petermetz: @CatarinaPedreira Totally get it regarding the time constraints, will be happy to accept extra PRs with it in the future if you ever get around to adding it, but also zero pressure of course, this is great already as it is!

CatarinaPedreira: Thank you Peter :)

CatarinaPedreira:
Hi @takeutak :) Can you review this PR?

takeutak (approved):

takeutak:
@CatarinaPedreira Thanks for contributing! LGTM.

Issue #1277, opened 2021-08-30
fix(cactus-core-api): interface i-cactus-plugin in the function isICactusPlugin is not checking for required onPluginInit


m-courtin:
**Describe the bug**

The newest version of Cactus is missing a  check on the required 'onPluginInit' function in the 'isICactusPlugin' function defined in the interface i-cactus-plugin.

PR #1278, opened 2021-08-30
fix(cactus-core-api): integrate check for req function onPluginInit into isCactusPlugin function

m-courtin:
interface i-cactus-plugin in the function isICactusPlugin is not
checking for required onPluginInit

Closes: #1277


petermetz (approved):
LGTM

takeutak (approved):

PR #1284, opened 2021-08-31
docs(connector-iroha): add Iroha menu item in the ledger support matrix

hxlaf:
Depends on #1290


petermetz (changes requested):
@hxlaf LGTM, but will leave it in "Requesting changes" state until we can verify that the links works. Please make sure to double check it once the dependent PR has been merged (which will then lead to us having a tag of v0.9.0

petermetz (approved):

takeutak (approved):

Issue #1286, opened 2021-08-31
fix(plugin-ledger-connector-quorum): OpenAPI spec validation missing from plugin REST endpoints


elenaizaguirre:
**Describe the bug**

Right now the constraints we define for request/response objects are not checked when a request comes in.

**To Reproduce**

Send an invalid request to some plugin endpoint that is invalid in a certain way and watch how it crashes the plugin resulting in internal server error vs. bad request.

**Expected behavior**

Fail with bad request even before the plugin's back-end code is reached if a request is sent that does not comply with the open api specs defined for that endpoint.

**Logs/Stack traces**

N/A

**Screenshots**

N/A

**Cloud provider or hardware configuration:**

N/A

**Operating system name, version, build:**

N/A

**Hyperledger Cactus release version or commit (git rev-parse --short HEAD):**

main

**Hyperledger Cactus Plugins/Connectors Used**

Quorum

**Additional context**

The interface of the web service plugins should be extended with a getOpenApiSpecs() method that returns an OpenAPI v3 spec document (see the openapi-types package)
Then the api-server calls up this method when it's installing the web services of the plugin, obtains the open api specs and creates a validator instance from it.

elenaizaguirre:
Working on this!

PR #1287, opened 2021-08-31
fix: openapi-validation test for quorum plugin

elenaizaguirre:
Includes tests for endpoints deployContractSolBytecodeV1,
invokeContractV1 and runTransactionV1, each of them
with test cases:
  - Right request

Closes: #1286

Relationed with #847 


packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
```diff
@@ -710,6 +693,75 @@
                     }
                 }
             }
+        },
+        "/api/v1/plugins/@hyperledger/cactus-plugin-consortium-manual/consortium/jws": {
```
elenaizaguirre, line 697: @petermetz I had to add these endpoints to Quorum openapi document because test api-client-routing-node-to-node.test.ts from cactus-test-api-client didn't working. I think it was trying to call consortium-manual endpoints through Quorum API client and validation was failing.
Is there any way to do it better?

petermetz (changes requested):
@elenaizaguirre Same requests/questions here as in the other OpenAPI validation PR

packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
petermetz: @elenaizaguirre It's most likely being caused by the open API validator trying to look up the validation metadata of the consortium manual plugin and not finding it, showing screenshots below with how I debugged into it.
One more tip: If you want to see what request handlers are registered in an express app you can run `app._router.stack` on the debug console to print it like I did (see other screenshot)

TLDR solution: you might need to tie back together the consortium manual plugin PR and the quorum connector PR because of this dependency.

Screenshot 2021-09-01 141141
Screenshot 2021-09-01 140852
Screenshot 2021-09-01 140009

petermetz (approved):
LGTM

takeutak (approved):

PR #1289, opened 2021-08-31
fix: openapi-validation test for xdai plugin

elenaizaguirre:
Includes tests for endpoints deployContractV1,
invokeContractV1 and runTransactionV1, each of them
with test cases:
  - Right request

Closes: #1288

Relationed with #847 


packages/cactus-plugin-ledger-connector-xdai/src/test/typescript/integration/openapi/openapi-validation.test.ts
```diff
@@ -0,0 +1,405 @@
...
+        },
+        bytecode: HelloWorldContractJson.bytecode,
+        gas: 1000000,
+        fake: 4,
+      };
+      await apiClient.deployContractV1(
+        (parameters as any) as DeployContractV1Request,
+      );
+    } catch (e) {
+      t2.equal(e.response.status, 400, "Bad request");
```
petermetz, line 197: @elenaizaguirre The success message should say something like "fake 4 response.status === 400 OK"
This way if it fails in the error log we can exactly tell which assertion failed.

packages/cactus-plugin-ledger-connector-xdai/src/main/typescript/plugin-ledger-connector-xdai.ts
```diff
@@ -158,6 +166,36 @@ export class PluginLedgerConnectorXdai
 
   async registerWebServices(app: Express): Promise<IWebServiceEndpoint[]> {
     const webServices = await this.getOrCreateWebServices();
+    app.use(
+      OpenApiValidator.middleware({
+        apiSpec: this.getOpenApiSpecs(),
+        validateApiSpec: false,
+      }),
+    );
+    app.use(
```
petermetz, line 175: @elenaizaguirre Can you move this to the API server so that it hooks it up at plugin registration time? That way the plugins don't have to depend on the openapi validator package at all and we will avoid issues where different versions of it in different plugins are conflicting.

petermetz (changes requested):

petermetz (approved):

takeutak (approved):

PR #1296, opened 2021-09-01
fix: openapi-validation test for fabric plugin

elenaizaguirre:
Includes tests for endpoints deployContractGoSourceV1,
deployContractV1 and runTransactionV1, each of them with
test cases:
  - Right request

Closes: #1295

Relationed with #847 


elenaizaguirre:
@petermetz Validation is failing because FabricTestLedgerV1, in method getConnectionProfileOrg1, returns an object that does not match with openapi document (it has not description field). Should I remove this field from the required ones?

petermetz:
@elenaizaguirre Something doesn't sound right about that. The Fabric connection profile is something completely unrelated to OpenAPI specs/documents. I need some more context to understand the problem. Where in the code do you need to use the two in conjunction and how?

elenaizaguirre:
@petermetz Sorry, you're right, context is missing.

Test which failed was fabric-v2-2-x/run-transaction-endpoint-v1.test.ts. This test failed because calling runTransactionEndpointV1 (line 258) it uses, inside gatewayOptions parameter, the field 'connectionProfile'. The value of this field is obtained from Fabric Test Ledger (method getConnectionProfileOrg1), which does not include field 'description' inside 'connectionProfile', but this field is required by the endpoint specs in openapi document.

```
const res = await apiClient.runTransactionV1({
  gatewayOptions: {
    connectionProfile: {
      description: "some relevant description",
      ...
    },
    ...
  },
  ...
});
```

This happened when validation was inside each connector, so now I think that it won't happen because this test does not calling installOpenapiValidationMiddleware and does not check openapi specs.

On the other hand, I'm thinking about that, Should I force each old test to call installOpenapiValidationMiddleware to ensure that each test fulfill openapi specs?

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/openapi/openapi-validation.test.ts
```diff
@@ -0,0 +1,482 @@
...
+
+  test.onFailure(async () => {
+    await Containers.logDiagnostics({ logLevel });
+  });
+
+  const ledger = new FabricTestLedgerV1({
+    emitContainerLogs: true,
+    publishAllPorts: true,
+    imageName: "ghcr.io/hyperledger/cactus-fabric2-all-in-one",
+    imageVersion: "2021-04-20-nodejs",
```
petermetz, line 59: @elenaizaguirre Please update the image to the new one (see other tests on the main branch)

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/openapi/openapi-validation-go.test.ts
```diff
@@ -0,0 +1,350 @@
...
+test(testCase, async (t: Test) => {
+  test.onFailure(async () => {
+    await Containers.logDiagnostics({ logLevel });
+  });
+  const ledger = new FabricTestLedgerV1({
+    emitContainerLogs: true,
+    logLevel,
+    publishAllPorts: true,
+    imageName: "ghcr.io/hyperledger/cactus-fabric-all-in-one",
+    imageVersion: "2021-04-21-2016750",
```
petermetz, line 56: @elenaizaguirre Please update the image to the new one (see other tests on the main branch)

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/openapi/openapi-validation.test.ts
elenaizaguirre: Done!

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/openapi/openapi-validation-go.test.ts
elenaizaguirre: Done!

petermetz:
@elenaizaguirre Ah, everything makes perfect sense now, sorry for the time it took! To answer the original question: Yes, in this case I agree as well that the best we can do is remove the field from the list of required fields because the connection profile we get back from the test ledger class is the one Fabric gives us so if the field is not present there then it doesn't matter what the documentation says about it being mandatory or not, we must not validate it as mandatory.

> Should I remove this field from the required ones?

Yup, thank you for explaining it in detail!

petermetz (approved):

elenaizaguirre:
@petermetz I prepared two test files for this issue. One of them (packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/openapi/openapi-validation-go.test.ts) is continually giving a timeout. Is possible to increase the timeout for that file? I can skip the test which returns a 200 code to make this test able to pass checks

takeutak (approved):

PR #1304, opened 2021-09-06
fix(cmd-api-server): cockpit off by default #1239

### LINKED ISSUE #1239: fix(cmd-api-server): Cactus-Cockpit to be off by default
**Describe the bug**

After a first installation of latest cactus version the Cactus-Cockpit is not completely disabled / removed. In the documentation (build.md) you will see "After starting the API server, you will see in the logs that plugins were loaded and that the API is reachable on the port you specified (4000 by default) and the Web UI (Cockpit) is reachable through port on the port your config specified (3000 by default)."

Also the Cactus API-Server server shows in the logs "[2021-08-16T13:49:52.029Z] INFO (api-server): Cactus Cockpit reachable https://127.0.0.1:3000"

**To Reproduce**

- Install latest version of Cactus
-  Execute "npm run generate-api-server-config" as written in documentation
- Execute  "npm run start:api-server" as written in documentation

**Expected behaviour**

As the Cactus - Cockpit should be disabled / removed, Cactus API-Server logs should not show that Cockpit is reachable at endpoint xyz and documentation should be updated accordingly.

AzaharaC:
Resolve #1239

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -161,8 +161,9 @@ npm run start:api-server
 ```
 
 After starting the API server, you will see in the logs that plugins were loaded
-and that the API is reachable on the port you specified (4000 by default) and
-the Web UI (Cockpit) is reachable through port on the port your config
+and that the API is reachable on the port you specified (4000 by default). The Web UI (Cockpit)
+is disabled by default but can be enabled by changing the property value 'cockpitEnabled'
+to true and it is reachable through port on the port your config
 specified (3000 by default).
 
 > You may need to enable manually the CORS patterns in the configuration file.
--- a/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
+++ b/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
@@ -86,7 +86,7 @@ export class ApiServer {
   private readonly log: Logger;
   private pluginRegistry: PluginRegistry | undefined;
   private readonly httpServerApi: Server | SecureServer;
-  private readonly httpServerCockpit: Server | SecureServer;
+  private readonly httpServerCockpit?: Server | SecureServer;
   private readonly wsApi: SocketIoServer;
   private readonly grpcServer: GrpcServer;
   private readonly expressApi: Application;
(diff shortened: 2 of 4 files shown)


takeutak (approved):

petermetz (changes requested):
@AzaharaC With this change in place, it will still listen for requests, just won't print the link to it, no?

We also need a flag in the config that disables it so that it's not accessible, not just not being printed.

packages/cactus-cmd-api-server/src/main/typescript/config/config-service.ts
```diff
@@ -572,7 +571,7 @@ export class ConfigService {
       cockpitWwwRoot: (schema.cockpitWwwRoot as SchemaObj).default,
       cockpitCorsDomainCsv: (schema.cockpitCorsDomainCsv as SchemaObj).default,
       cockpitMtlsEnabled: false,
-      cockpitTlsEnabled,
+      cockpitTlsEnabled: false,
```
petermetz, line 574: @AzaharaC TLS enablement needs to come from the secure defaults by default (one of our design principles)

petermetz (changes requested):

packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
```diff
@@ -223,14 +228,6 @@ export class ApiServer {
         this.log.info(`Cactus API reachable ${httpUrl}`);
       }
 
-      {
-        const { cockpitHost: host } = this.options.config;
-        const { port } = addressInfoCockpit;
-        const protocol = cockpitTlsEnabled ? "https:" : "http:";
-        const httpUrl = `${protocol}//${host}:${port}`;
-        this.log.info(`Cactus Cockpit reachable ${httpUrl}`);
```
petermetz, line 231: @AzaharaC Could we still log this based on the `cockpitEnabled` flag being set to true?

BUILD.md
```diff
@@ -161,9 +161,7 @@ npm run start:api-server
 ```
 
 After starting the API server, you will see in the logs that plugins were loaded
-and that the API is reachable on the port you specified (4000 by default) and
-the Web UI (Cockpit) is reachable through port on the port your config
-specified (3000 by default).
+and that the API is reachable on the port you specified (4000 by default).
```
petermetz, line 164: @AzaharaC It would be worth documenting the functionality still and just mention that by default it is off and that it can be turned on with a specific  flag.

petermetz (changes requested):
@AzaharaC Added a few change requests.
Also have a manual testing request: Could you please verify that with the code as-is the supply chain app example still works fine if you run it locally the way it's readme describes in the section called `Running the Example Application Locally`

AzaharaC:
> @AzaharaC Added a few change requests. Also have a manual testing request: Could you please verify that with the code as-is the supply chain app example still works fine if you run it locally the way it's readme describes in the section called `Running the Example Application Locally`

@petermetz I have tried the supply chain example and it seems to continue to work with the changes.

petermetz (approved):
@AzaharaC Great, thank you very much for testing it out! LGTM
I see that the CI failed but it looks like a flake because it seems to have failed due to DockerHub refusing requests so I'd say a rebase to upstream/main + force push with lease should take care of that, fingers crossed.

Issue #1308, opened 2021-09-07
feature(indy-testnet-docker): auto-start server on container startup


izuru0:
**Description**

You need to log into validator container for indy and manually run python command to start the validator.

**Expected Solution**

You do not need to log into valdiator container after docker-compose up

PR #1309, opened 2021-09-07
feature(improve-indy-testnet-docker): auto-start server on container startup #1308

izuru0:
Modified Dockerfile so that validator program is automatically run from supervisord.

Also:
- fixed a problem found during this implementation: sandbox directory for indy_pool was not included in current version. Now startup script automatically initializes itself when it cannot find necessary files.
- lessened dependency to indy-sdk (some files are copied from indy-sdk to indy-testnet directory) by copying some files from indy-sdk to this directory. 
- Note: as of now, a container for testing server containers still depends on indy-sdk source tree. This needs to be fixed in future PR.

This PR corresponds to Issue #1308 

Signed-off-by: Izuru Sato <sato.izuru@fujitsu.com>


takeutak (approved):

petermetz:
@izuru0 Same recommendation here regarding the scope part of the issue title/commit message as in the other issue.
https://www.conventionalcommits.org/en/v1.0.0/

petermetz (approved):

PR #1335, opened 2021-09-15
ci: re-enable image caching, skip flaky tests #1334

### LINKED ISSUE #1334: ci: re-enable image caching, skip flaky tests
### Description 

1. Re-enable image caching so that we can run the CI on the self hosted mac provided by Hyperledger/LF
2. Aggressively comment out tests that are flaky until we have time to fix them properly.

### Previous (old) Description

The CI is flaky again because the RAM consumption finally reached the point where the 7.5 GB is not enough.
I've provisioned a 16 GB VPS which has enough RAM but hitting issues with DockerHub rate limiting. I assume this happens because with GHA runners the pulling IP is mostly a different, random one that successfully gets around the limits, but with static IP of the VPS (self hosted runner) now we are back to square one with the DockerHub rate limiting.

On the other hand image caching should help tremendously beacuse the machine is always the same we just need to make sure that the caching can actually work by migrating from Docker In Docker to Docker out of Docker.

http://tdongsi.github.io/blog/2017/04/23/docker-out-of-docker/

### Acceptance Criteria
1. CI is rock solid
2. All images are updated and pushed to ghcr.io where docker in docker was/is necessary.
3. Images are migrated to vanilla Ubuntu 20-04 instead of the DinD editions.

petermetz:
1. Disables the coverage check of tap.
2. Starts skipping failing tests
3. Re-enables container image caching

Fixes #1334


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #1339, opened 2021-09-15
ci: add DCI-Lint GitHub action

petermetz:
Also fixed the linter errors caught by the linter where it was
possible and set the ignore file up accordingly where it was
not feasible due to third party dependencies configuration schemas
for example. Hopefully in the future those projects adapt and then
we won't need the ignore file anymore or at least less of it.


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1342, opened 2021-09-16
ci: set the swap file size to 10 GB for GitHub runners

petermetz:
This will hopefully alleviate the issue of some tests causing
out of memory crashes on the 7.5GB RAM GitHub Action Runners.

Could of course have a domino effect where the swap kicks in
and slows everything down so much that the tests just start
timing out instead of crashing with OOM ¯\_(ツ)_/¯


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1343, opened 2021-09-16
build(yarn): reset script now wipes node modules, yarn lock

petermetz:
Found myself in need of this when strange errors started occurring
between branch switches where the gRPC tooling was not installed
properly despite the configure script running successfully.
Now the reset is script is much more robust and it closer
resembles the outcome of a fresh clone.

Also: Ran the reset script and it updated the lock file, including
that in the commit as well to see if in the future othe reset runs
will also generate these/similar changes to the lock file or not.


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1344, opened 2021-09-17
build: downgrade to yarn v1.18.0 due to upgrade-interactive bug

petermetz:
Depends on #1343

See: https://github.com/yarnpkg/yarn/issues/7807 for the root cause
and https://classic.yarnpkg.com/en/docs/cli/upgrade-interactive/
for more information about what yarn upgrade-interactive does.


takeutak (approved):

jonathan-m-hamilton (approved):
LGTM

PR #1347, opened 2021-09-17
fix: endpoints implementation in corda plugin

### LINKED ISSUE #1346: fix (cactus-plugin-ledger-connector-corda): add endpoints implementation
**Describe the bug**

We need to be able to call corda endpoints.

**To Reproduce**

To reproduce this bug, we must call corda endpoints hosted in an express server instead of call directly to corda container. We will receive an error with status 404.

**Expected behavior**

We need that these endpoints return us the same result as if we call directly to corda container.

 - Corda

elenaizaguirre:
Closes #1346

### DIFF
--- a/packages/cactus-plugin-ledger-connector-corda/src/main/typescript/web-services/deploy-contract-jars-endpoint.ts
+++ b/packages/cactus-plugin-ledger-connector-corda/src/main/typescript/web-services/deploy-contract-jars-endpoint.ts
@@ -1,13 +1,10 @@
-import fs from "fs";
-import path from "path";
-
 import { Express, Request, Response } from "express";
-import temp from "temp";
-import { NodeSSH, Config as SshConfig } from "node-ssh";
+import { Config as SshConfig } from "node-ssh";
 
 import {
   IWebServiceEndpoint,
   IExpressRequestHandler,
+  Configuration,
 } from "@hyperledger/cactus-core-api";
 
 import {
@@ -26,6 +23,7 @@ import {
 import { IEndpointAuthzOptions } from "@hyperledger/cactus-core-api";
 
 import {
+  DefaultApi,
   DeployContractJarsSuccessV1Response,
   DeployContractJarsV1Request,
 } from "../generated/openapi/typescript-axios/api";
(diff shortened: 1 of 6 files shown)


elenaizaguirre:
@petermetz Can you take a look at this, please? I am not sure if this is that you wanted me to do. I started with a remote adapter like plugin-keychain-vault-remote-adapter.ts but later I saw that it was simpler to use those endpoints like proxies to container endpoints (as Vault.client in plugin-keychain-vault.ts) because Corda endpoints are not used when we call container endpoints.

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/web-services/deploy-contract-jars-endpoint.ts
```diff
@@ -140,75 +142,12 @@ export class DeployContractJarsEndpoint implements IWebServiceEndpoint {
     }
   }
 
-  private async doDeploy(
```
elenaizaguirre, line 143: @petermetz I changed this code for a call to the container endpoint, do you prefer me to revert to the previous function?

takeutak (approved):

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/web-services/deploy-contract-jars-endpoint.ts
petermetz: @elenaizaguirre No, it's the right call to keep things in the JVM implementation as long as that works.

petermetz (changes requested):
@elenaizaguirre A few questions:

1. Is the remote adapter pattern not implemented in here? I couldn't find a class for it in the diff.
2. Can the OpenAPI validation be done in the JVM implementation? Right now the validation only happens (I think) if you proxied the requests through the NodeJS app but if you were to hit the JVM app directly it wouldn't perform the validations.

elenaizaguirre:
> @elenaizaguirre A few questions:
> 
> 1. Is the remote adapter pattern not implemented in here? I couldn't find a class for it in the diff.
> 2. Can the OpenAPI validation be done in the JVM implementation? Right now the validation only happens (I think) if you proxied the requests through the NodeJS app but if you were to hit the JVM app directly it wouldn't perform the validations.

@petermetz 
That pattern is not implemented because I used a proxy to access the internal container.
Validations will be directly in the JVM app

petermetz (approved):
@elenaizaguirre 

> That pattern is not implemented because I used a proxy to access the internal container.

Okay, that's fine for now but we'll have to move that to a remote adapter once we are adding support to Corda 5.x because that one will have support for RPC directly from JS code

PR #1391, opened 2021-09-30
docs: adding example for besu private transactions

travis-payne:
Closes: 1213


petermetz (changes requested):
@travis-payne The commit message says that it's a breaking change, but doesn't look like it.
It's important to not mark it as a breaking change unless it really is because this will show up in the release notes of the next version. It would also trigger/force a major release in the automation (once we have that fully figured out)

petermetz (changes requested):
@travis-payne The hashmark is missing from the issue comment where you say `closes ...`

travis-payne:
@petermetz fixed both of these issues!

petermetz (approved):
@travis-payne  LGTM

takeutak (approved):

PR #1392, opened 2021-09-30
fix(plugin-ledger-connector-quorum): no keychain endpoints

elenaizaguirre:
1. Endpoint invoke-contract without keychain was failing because,
on deploy, network information was saved inside the contractJson
object, which was keeping it on memory and for this reason it was
not failing calling the connector.

2. Now, instead of save the contracts addresses in memory indexed
by contractName, this parameter should be sent in the request to
avoid that there can only be a single instance for each contract
name.

3. Openapi specs were changed to consider these changes,
including remove fields that are inside the contractJSON object
and definition of this object structure.

4. Includes openapi validation test for new endpoints.


packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -230,7 +226,12 @@ export class PluginLedgerConnectorQuorum
     req: InvokeContractV1Request,
   ): Promise<any> {
     const fnTag = `${this.className}#invokeContract()`;
-    const contractName = req.contractName;
+    if (req.contractName == undefined) {
```
petermetz, line 229: @elenaizaguirre Check for  falsy-ness with the exclamation mark, it's more robust

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -630,18 +582,17 @@ export class PluginLedgerConnectorQuorum
...
     if (req.contractJSON != undefined) {
-      const networkId = await this.web3.eth.net.getId();
-
       const web3SigningCredential = req.web3SigningCredential as
         | Web3SigningCredentialGethKeychainPassword
         | Web3SigningCredentialPrivateKeyHex;
       const receipt = await this.transact({
         transactionConfig: {
-          data: `0x${req.bytecode}`,
+          data: `0x${(contractJSON as any).bytecode}`,
```
petermetz, line 595: @elenaizaguirre Can this be casted to the type you specified in the openapi.json instead of any?

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -283,90 +279,51 @@ export class PluginLedgerConnectorQuorum
...
-          throw new Error(
-            `${fnTag} Unsupported invocation type ${req.invocationType}`,
-          );
-        }
-      } else if (
-        req.contractJSON == undefined &&
-        req.contractAbi == undefined &&
-        req.contractAddress == undefined
-      ) {
+    if (req.contractJSON == undefined) {
```
petermetz, line 282: @elenaizaguirre Check falsy-ness with the exclamation mark, it's more robust.

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json-json-object.test.ts
```diff
@@ -153,8 +150,6 @@ test(testCase, async (t: Test) => {
     );
 
     const { callOutput: helloMsg } = await connector.getContractInfo({
-      contractName,
```
petermetz, line 156: @elenaizaguirre I was wondering while looking through this diff: Do we even need the contractAbi and bytecode arguments at this point or can they both be removed in favor of `contractJSON`?

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
elenaizaguirre: Done!

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
elenaizaguirre: Done!

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
elenaizaguirre: Done!

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json-json-object.test.ts
elenaizaguirre: You're right. I removed them in favor of contractJSON and keychainPlugin (which saves a contractJSON object too)

PR #1433, opened 2021-10-04
Lint fix 34

cotikor:


petermetz:
@cotikor Could you please double check the issue number that this one is fixing? 34 is an issue that is not related to linter warnings.

PR #1435, opened 2021-10-04
docs(faq): added how to use local test ledger images

jsladerman:
Added FAQ question about how to use locally-created test ledger images in tests/example applications


takeutak (approved):

FAQ.md
```diff
@@ -150,3 +151,16 @@ proceed via the UI/shortcuts of the debugger.
 
 Once you let the execution continue, requests that previously appeared to be
 hanging should finish successfully.
+
+## How do I run test cases with my own ledger images?
+
+Build and tag the image locally using a name that is non-ambiguous with the official tags.
```
petermetz, line 157: @jsladerman Please also mention that the image tag needs to be explicitly passed in for the test ledger classes as well (the one that's been built locally) and include that in the code example as well

petermetz (changes requested):

PR #1436, opened 2021-10-04
feat(besu-test-ledger): added omitPull parameter to besu test ledger

jsladerman:
The omitPull option was missing from the start method of the Besu test ledger class
Added it in the same way it's used in the fabric and quorum test ledger classes


takeutak (approved):

petermetz (approved):
@jsladerman LGTM

PR #1441, opened 2021-10-06
fix: openapi validation for keychain-google-sm plugin

elenaizaguirre:
Includes tests for endpoints setKeychainEntry, getKeychainEntryV1,
hasKeychainEntryV1 and deleteKeychainEntryV1, each of them with
test cases:
  - Right request

Modifies cactus-core-api definitions for GetKeychainEntryRequestV1
and DeleteKeychainEntryRequest adding to them the option
additionalProperties: false, and removes this option from
GetKeychainEntryResponseV1 and DeleteKeychainEntryResponseV1

Relationed with #847


elenaizaguirre:
@petermetz I made a mistake when I changed core-api openapi.json and I had to change it. I've referenced it with a tag in the [branch](https://github.com/elenaizaguirre/cactus/tree/core-api-schemas) where I have that change. It is also included in this PR to be merged at the same time that this validation for future releases.

Could you create a tag in Hyperledger repository please? Then I'll change the references.

petermetz:
@elenaizaguirre Let's do this:
1. You can create a separate PR that contains *only* the core-api openapi.json changes
2. We'll issue a release of that (`v0.11.0`) which will then also tag it
3. Then you can come back here, rebase, use the 0.11.0 tag

Why though?
This way we avoid the one-off workaround tags and instead keep things tidy with the tags being official releases not just dangling helpers. 
It is a bit slower, but it's best to get into these kind of habits because they'll save us some headache in the future.

petermetz (changes requested):
@elenaizaguirre See above comment ^^

Issue #1444, opened 2021-10-07
fix(api-server): Invoke onPluginInit() on every connector during instantiation


m-courtin:
**Describe the bug**

In the latest Cactus version the onPluginInit() functionality required by ICactusPlugin Interface is nowhere invoked / called. So the onPluginInit() functionality is never executed.

**Expected behavior**

The onPluginInit() needs to be invoked for each connector in api-server at the end of the instantiatePlugin function.

PR #1446, opened 2021-10-07
fix(api-server): fixes issue 1444 invoking the required onPluginInit

m-courtin:
In api server the instantiatePlugin function is now calling the required
onPluginInit for each corresponding connector

Closes: #1444


m-courtin:
@petermetz: Rebase and squash is done

PR #1463, opened 2021-10-19
feat(plugin-keychain-memory-wasm): add WebAssmebly PoC

### LINKED ISSUE #1281: feat(cmd-api-server): web assembly plugin modules
### Description

As a user of Cactus I want to be able to load plugins that are WASM modules, not just vanilla JS ones.

### Acceptance Criteria
1. A PoC of a Rust wasm-pack based plugin that can be loaded into the NodeJS process of the API server
2. Tests verify that the above works as intended at runtime

petermetz:
Adding a new keychain in-memory plugin that has it's implementation
written in Rust that is then compiled down to WebAssmebly via wasm-pack
and used by the wrapper Typescript code.

This is NOT meant for production because it stores everything in plain
text and also provides zero durability/persistence guarantees given that
it's only storing everything in memory.

The actual news here is that we have a plugin now written in Rust which
is the pre-cursor to us being able to do something similar with the
Weaver relay component as the next phase of a bigger PoC.

The reason why not the entire plugin is implemented in Rust is because
we are unable to hook up ExpressJS request handlers from the Rust code
as far as I could determine. See this link for further details on this:
https://rustwasm.github.io/book/reference/js-ffi.html#from-the-rust-side

Because of the above, the way it works is this:

        +--------+ API Request   +----------+
        | HTTP   |-------------->| ExpressJS|
        | Client |               +----------+
        +--------+               Method|
                                 Call  |
                                       v
        +---------+ Method Call  +----------+
        | Calling |------------->|JS Plugin |
        | Module  |              |Module    |
        +---------+              +----------+
                                   |get()
                                   |set()
                                   |has()
                                   |delete()
                +-------------+    |
                | Wasm Plugin |<---+
                | Module      |
                +-------------+
                        ^
                        |
                        |
                        v
           +------------------------+
           |Rust Native             |
           |HashMap<String, String> |
           +------------------------+

Resolves #1281


petermetz:
> @petermetz am I correct in assuming that this is a refactoring of existing plugin, not a totally new plugin?

@izuru0 Yup, that's right.

Issue #1265, opened 2021-08-21
fix(iroha-connector): make compareAndSetAccountDetail accept the check_empty argument


hxlaf:
**Describe the bug**

One argument for compareAndSetAccountDetail, check_empty, is not dealt with inside the plugin-ledger-connector-iroha.ts file. The run-transaction test case also fail to use this argument. Moreover, the old_value argument should be optional, but it is a mandatory argument at this moment.
Please refer to the official Iroha doc for [description of compareAndSetAccountDetail](https://iroha.readthedocs.io/en/main/develop/api/commands.html#compare-and-set-account-detail).

**To Reproduce**

Check `packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/plugin-ledger-connector-iroha.ts` file. Looking for the case of compareAndSetAccountDetail.

**Expected behavior**

Still runs, but it does not follow behaviors set by the official document.

**Logs/Stack traces**

N/A

**Screenshots**

![image.png](https://images.zenhubusercontent.com/611fe79ceb6eacb302da85ac/cf33e1fc-eef3-4230-b9fb-824cf432c04b)

**Cloud provider or hardware configuration:**

N/A

**Operating system name, version, build:**

N/A

**Hyperledger Cactus release version or commit (git rev-parse --short HEAD):**

0.9.0

**Hyperledger Cactus Plugins/Connectors Used**

Iroha

**Additional context**

PR #1484, opened 2021-10-27
feat(iroha): adding checkEmpty param to request

travis-payne:
Depends on #1957

Current functionality did not make use of the newer additional parameter in the Iroha command

Closes: #1265


petermetz:
> @petermetz Currently the request params are accepted using an array of params, instead of named, typed params, generated using the OpenAPI. What's your opinion on refactoring this as part of the PR, also makes the optionality of the `oldValue` param a lot cleaner.

@travis-payne I agree. If we don't already have a separate task (so not within the scope of this) for the params refactor, we should (IIRC we did have a task for it though).

petermetz (commented):
@travis-payne Were you able to run the tests successfully with these changes? (Right now the tests are being skipped by default so the CI passing doesn't guarantee correctness)

takeutak (approved):

petermetz:
Rebased this onto upstream/main.
Will still need to verify that it actually works with the tests, but for that we first need the Iroha tests to be passing again.

takeutak:
@petermetz This pull-request has been left unreviewed for a long time.  What would you like to do? (still wait / close this PR / etc.)
I guess that this pull request can no longer be merged to the current branch because it has been a long time.

petermetz:
> @petermetz This pull-request has been left unreviewed for a long time. What would you like to do? (still wait / close this PR / etc.)
> I guess that this pull request can no longer be merged to the current branch because it has been a long time.

@takeutak Agreed on all of those points. Let's close this down for now and if there's willingness for someone to take it over the finish line then we can re-open it anytime we want. Closing it for now to de-clutter the PR queue a little bit more.

petermetz (commented):
TODO: Manually verify that the tests are passing with these changes.

Issue #1210, opened 2021-08-12
feat(cmd-api-server): add ability to install specific versions of plugins


petermetz:
### Description
As an ops person running Cactus I want to be able to tell the API server exactly what version of a given plugin package to install so that I can do things like downgrading from the latest when it's necessary (security bugs in new versions for example one scenario where this would have to be done urgently in a production environment)

### Acceptance Criteria
1. Test coverage specifically addresses the new feature
2. Also look into adding support for Github URLs
3. Also look into adding support for local file system paths

2) and 3) might already be possible but it hasn't been tested so it needs coverage either way even if it's already working as intended.

PR #1527, opened 2021-11-09
fix: added a dummy package

elenaizaguirre:
added a dummy package for tests about install packages from github
sources required for #1210 

### DIFF
--- a/packages/cactus-cmd-api-server/src/test/resources/cactus-dummy-package/index.js
+++ b/packages/cactus-cmd-api-server/src/test/resources/cactus-dummy-package/index.js
@@ -1,0 +1,24 @@
+class DummyPlugin {
+  onPluginInit() {
+    console.log("DummyPlugin#onPluginInit() OK");
+    return Promise.resolve();
+  }
+  getInstanceId() {
+    console.log("DummyPlugin#getInstanceId() OK");
+    return "dummy_instance_id";
+  }
+  getPackageName() {
+    console.log("DummyPlugin#getPackageName() OK");
+    return "@hyperledger/cactus-dummy-package";
+  }
+}
+
+class DummyPluginFactory {
+  create() {
+    return Promise.resolve(new DummyPlugin());
+  }
+}
+
+module.exports.createPluginFactory = (options) => {
+  return Promise.resolve(new DummyPluginFactory(options));
+};
--- a/packages/cactus-cmd-api-server/src/test/resources/cactus-dummy-package/package.json
+++ b/packages/cactus-cmd-api-server/src/test/resources/cactus-dummy-package/package.json
@@ -1,0 +1,6 @@
+{
+    "name": "@hyperledger/cactus-dummy-package",
+    "version": "1.0.0-rc.2",
+    "description": "Dummy package for testing.",
+    "main": "index.js"
+  }


petermetz (commented):
@elenaizaguirre Quick question: Do we need the .d.ts files as well? And how about the openapi.json and the generated code files?
You see where this is going: I want to reduce the 32 files in the diff to the absolute minimum viable dummy package (which might be as low as two files, the package.json and an index.js) but I might be overlooking one (or many) things so that's why this is a question not necessarily a change request.

takeutak:
@elenaizaguirre Thanks for contributing. Would you add the corresponding issue on Cactus GitHub? (If there is no corresponding issue, would you post the issue?)

elenaizaguirre:
> @elenaizaguirre Quick question: Do we need the .d.ts files as well? And how about the openapi.json and the generated code files? You see where this is going: I want to reduce the 32 files in the diff to the absolute minimum viable dummy package (which might be as low as two files, the package.json and an index.js) but I might be overlooking one (or many) things so that's why this is a question not necessarily a change request.

@petermetz I have deleted *.d.ts files but I think openapi.json and generated code are necessary because the first one is required for some files and generated code is used because when we instantiate a plugin from the api server, it ensures that the plugin can call some methods:
```typescript
    const pluginPackage = require(/* webpackIgnore: true */ packagePath); // this one is our dummy package
    const createPluginFactory = pluginPackage.createPluginFactory as PluginFactoryFactory;
    const pluginFactoryOptions: IPluginFactoryOptions = {
      pluginImportType: pluginImport.type,
    };
    const pluginFactory = await createPluginFactory(pluginFactoryOptions);
    const plugin = await pluginFactory.create(pluginOptions);
    try {
      await plugin.onPluginInit();
```

It also calls `installOpenapiValidationMiddleware` with each endpoint of the plugin.

elenaizaguirre:
> @elenaizaguirre Thanks for contributing. Would you add the corresponding issue on Cactus GitHub? (If there is no corresponding issue, would you post the issue?)

@takeutak Is not an issue by itself, it is the first step for a Peter suggestion  https://github.com/hyperledger/cactus/pull/1456#discussion_r745012362. Anyway, I've added the original issue number to the description

takeutak (approved):

petermetz:
> @petermetz I have deleted *.d.ts files but I think openapi.json and generated code are necessary because the first one is required for some files and generated code is used because when we instantiate a plugin from the api server, it ensures that the plugin can call some methods:

@elenaizaguirre Could we just not have the plugin implement web services at all? If we limit our scope to the dummy package just being the testing instrument for verifying that the local file system path based plugin installation works then it doesn't matter if it's a web svc plugin or not IMO.

I mean something like this in an index.js file:
(pseudo-ish code, syntax might have small issues)

```js
class DummyPlugin {
  onPluginInit() {
    console.log("DummyPlugin#onPluginInit() OK");
    return Promise.resolve();
  }
}

class DummyPluginFactory {
  create() {
    return Promise.resolve(new DummyPlugin());
  }
}

module.exports.createPluginFactory = (options) => {
  return Promise.resolve(new DummyPluginFactory(options));
};
```

petermetz (commented):
^^ See above

elenaizaguirre:
> > @petermetz I have deleted *.d.ts files but I think openapi.json and generated code are necessary because the first one is required for some files and generated code is used because when we instantiate a plugin from the api server, it ensures that the plugin can call some methods:
> 
> @elenaizaguirre Could we just not have the plugin implement web services at all? If we limit our scope to the dummy package just being the testing instrument for verifying that the local file system path based plugin installation works then it doesn't matter if it's a web svc plugin or not IMO.
> 
> I mean something like this in an index.js file: (pseudo-ish code, syntax might have small issues, some more methods are probably missing but the point is that we only need to provide the simplest possible implementation that does not make it crash at install & instantiation time to verify that it can be installed from the local file-system without a crash)
> 
> ```js
> class DummyPlugin {
>   onPluginInit() {
>     console.log("DummyPlugin#onPluginInit() OK");
>     return Promise.resolve();
>   }
> }
> 
> class DummyPluginFactory {
>   create() {
>     return Promise.resolve(new DummyPlugin());
>   }
> }
> 
> module.exports.createPluginFactory = (options) => {
>   return Promise.resolve(new DummyPluginFactory(options));
> };
> ```

@petermetz Thanks, you were right, it was much simpler than I thought.

petermetz (approved):
@elenaizaguirre My pleasure and thank you very much for taking the time to address the nits as usual!

PR #1533, opened 2021-11-11
fix(deps): sibling package dependencies keychain-memory 0.6.0 #1532

### LINKED ISSUE #1532: fix(deps): sibling package dependencies keychain-memory 0.6.0
**Describe the bug**

We have an outdated dependency on one of our own packages. Usually lerna auto-updates them at release creation time but this one might've been skipped due to a bug or someone accidentally committed it after a difficult rebase and the reviewers missed it.

**To Reproduce**

Do a search for `"0.6.0"`

**Expected behavior**

All the sibling package dependencies should be locked down to the current release which at the moment of this writing is `1.0.0-rc.2`

**Screenshots**

`1.0.0-rc.2`

**Additional context**

This was discovered during work done on the jest migration co-existence branch.

petermetz:
This might be the cause of some very strange errors that the configure
script has been experiencing lately (also might be a completely benign
bug but we'll see, it needed to be fixed either way).

As a follow-up to this an automated check should be made in the CI which
verifies every dependency declaration in all package.json files to be
referring to the latest and greatest version when pulling in sibling
packages as dependencies.

Fixes #1532


jonathan-m-hamilton (approved):
LGTM

takeutak (approved):

PR #1535, opened 2021-11-12
ci: custom check to verify uniform sibling dependency versions #1534

### LINKED ISSUE #1534: ci: custom check to verify uniform sibling dependency versions
### Description
As a maintainer of a monorepo I want to have the CI ensure that we use the same version of sibling packages so that there's no strange, hard to trace issues (related to this at least) during the build.

(add more description and relevant links)

### Acceptance Criteria
1. A new custom check is developed within the `./tools/` directory
2. Said custom check is included in the existing npm script that the CI runs for custom checks
3. The custom check ensures that all monorepo packages who depend on each other are doing so using the latest released version number which at the time of this writing is `1.0.0-rc.2` but of course this has to dynamically change within the check itself according to each release.

petermetz:
With this, the `yarn custom-checks` script will go through all the
package.json files and validate the following things:
1. The top level "version" property in the package.json of the package is
set to a value that equals whatever is set as the project-wide version
of the monorepo as defined by the lerna.json file's "version" property.
(the latter gets automatically bumped by the relese process)
2. Verifies that all dependencies within each package.json pointing to
sibling packages are also using the correct version from lerna.json as
described above.
3. Same check as in 2) but for the "devDependencies" section of each
package.json file within the monorepo.

The upside of this custom check is that maintainers no longer need to
manually ensure that the version numbers are kept up to date which was
especially tricky when a new package is being added in a pull request
that has been open for a while and we issued one or more releases during
the lifetime of the pull request.

Fixes #1534

Depends on #1533

### DIFF
--- a/tools/custom-checks/check-sibling-dep-version-consistency.ts
+++ b/tools/custom-checks/check-sibling-dep-version-consistency.ts
@@ -1,0 +1,129 @@
+import fs from "fs-extra";
+import path from "path";
+import { fileURLToPath } from "url";
+import { globby, Options as GlobbyOptions } from "globby";
+import { RuntimeError } from "run-time-error";
+import { isStdLibRecord } from "./is-std-lib-record";
+
+export interface ICheckSiblingDepVersionConsistencyRequest {
+  readonly argv: string[];
+  readonly env: NodeJS.ProcessEnv;
+  /**
+   * The version that will be used as the correct one when checking the sibling
+   * package dependency version declarations.
+   * If you omit this (optional) parameter then the root package.json file
+   * will be parsed to obtain its value at runtime.
+   */
+  readonly version?: string;
+}
+
+/**
+ * Verifies that each sibling dependency is up to date with the latest version
+ * that was released at any given time.
+ * Note: This only checks dependency versions for the packages that are hosted
+ * within this monorepo (Hyperledger Cactus) not dependencies in general.
+ *
+ * For example if the cmd-api-server package depends on the common package and
+ * currently the project is on release v1.2.3 then the dependency declaration of
+ * the package.json file of the cmd-api-server package should also use this
+ * v1.2.3 version not something outdated such as v0.3.1 because that may cause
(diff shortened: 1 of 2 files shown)


petermetz:
Note to reviewers: The CI is failing because the check is in fact *working as intended* and it will start passing once the other pull request that this one depends on is merged (which contains the fixes needed to make the checks pass).

jonathan-m-hamilton (approved):
LGTM

takeutak:
@petermetz Please let me clarify one thing (It's just a question).  As your PR comment, this PR makes the "version" and "devDependencies" update automatically (= maintainers no longer need to manually ensure that the version numbers are kept up to date), is it right?

> The upside of this custom check is that maintainers no longer need to
manually ensure that the version numbers are kept up to date which was
especially tricky when a new package is being added in a pull request
that has been open for a while and we issued one or more releases during
the lifetime of the pull request.

As I see the codes, in my understanding the checking tool make error when the version is not equal and maintainers need to manually change the version number, is it right?
https://github.com/hyperledger/cactus/pull/1535/files#diff-c0007e67e9e4907418512e590dcc69adc07ddf51f574e00afaffe3a97a4c884eR94

petermetz:
@takeutak 

> As I see the codes, in my understanding the checking tool make error when the version is not equal, and maintainers need to manually change the version number, is it right?

Yep, 100% correct. The verification is automated, but if the version is out of sync for whatever reason then it errors out and a human has to update it manually (so the checking is no longer manual but in the event of a check failure a manual update is required)

Also note: Most of the time the version update is also automated except for the edge case when a pending pull request was submitted first (with a new sibling package dependency being added somewhere) and then a release was issued on the main branch via another pull request concurrently. In this edge case the author of the first PR is responsible for making sure that the version numbers in their new package.json file are updated accordingly (and this check will now make sure that they are aware of this)

takeutak (approved):

takeutak:
@petermetz Thanks. I understood.

PR #1538, opened 2021-11-12
test(fabric): update AIO image tags to 1.0.0-rc.2 to lower flakes #1537

### LINKED ISSUE #1537: test(fabric): update AIO image tags to 1.0.0-rc.2 to reduce flakes
**Describe the bug**

The problem is that some pull requests got merged with them using older image tags that are known to be flaky.

**To Reproduce**

Run the CI repeatedly and observe that sometimes certain tests fail.

**Expected behavior**

Tests are stable.

**Logs/Stack traces**

CI logs: [2021-11-12-cactus-ci-log-odap-fabric-timeout.log](https://github.com/hyperledger/cactus/files/7529637/2021-11-12-cactus-ci-log-odap-fabric-timeout.log)

**Screenshots**

Are you running the software on a dev machine or somewhere in the cloud?

Ubuntu 20.04

1.0.0-rc.2

Fabric

petermetz:
1. Published the 1.0.0-rc.2 container images for Fabric AIO (out of
band with regards to this pull request, e.g. you don't see anything
for that in the diff)
2. Updated the test ledger class to use the 1.0.0-rc.2 version by default
3. Removed the explicit tags from the test cases so that they all default
to whatever the test ledger is using which should keep things simpler
in the future because the regression we are having right now with someone
pushing a test that uses an old, flaky tag will not happen by default,
so overall this change should increase future stability while also fixing
the problem at hand that the issue was originally created for.

Fixes #1537


izuru0 (approved):
LGTM

jonathan-m-hamilton (approved):
LGTM

PR #1544, opened 2021-11-17
feat(core-api): add weaver protocol buffer definitions #1523

### LINKED ISSUE #1523: feat(core-api): add weaver protocol buffer definitions
### Acceptance Criteria

1. Commit the protocol buffer definitions of Weaver to the Cactus Core API package
2. Set up automatic publishing of a Rust crate so that any time a commit is merged to main a new version of the crate is pushed with the protocol buffer definitions - this step is dependent on whether we want to also publish the generated Rust code for the protocol buffers or just the .proto files. If the latter then there's no need to set up the crate publishing *just yet*
3. Ensure Rama has signed off on adding the files (verbal sign-off has been acquired but also on the git commit log)
4. Ensure Typescript definitions are generated from the .proto files - figure out if this should be in it's own package or in one of the existing ones such as `core` or `cmd-api-server` etc.

petermetz:
There's two commits in this PR on purpose, one is to hold the sign-off from Rama and the other one is the admin chores performed prior to that sign-off.
This is an important milestone for us (Weaver and Cactus maintainers) working together.

---

Commit 1
---

Updated Weaver protos README and imported latest schema additions

Fixes #1523

Signed-off-by: VRamakrishna <vramakr2@in.ibm.com>

---

Commit 2
---

feat(core-api): add weaver protocol buffer definitions #1523

Adding the .proto definitions from here as discussed earlier with @VRamakrishna

To-do: I skipped the fabric driver proto due to an error on account of the
Fabric peer definitions not being present (unresolved import error).

Fixes #1523


petermetz:
@VRamakrishna Thank you for your help and cooperation!

takeutak (approved):

izuru0:
@petermetz 
is this going to be part of V1?

packages/cactus-core-api/src/main/proto/weaver/common/protos/README.md
```diff
@@ -0,0 +1,8 @@
+<!--
+ Copyright IBM Corp. All Rights Reserved.
```
izuru0, line 2: @petermetz 
is this line about copyright in line with the project's policy?

VRamakrishna: @izuru0 @petermetz I discussed this with my team, and the copyright needs to be retained. Thanks!

petermetz: @izuru0 I think it is. My original intuition 2 years ago was also that maybe it is not, but then we had the same discussion about the copy right statements of Fujitsu [1] back then and it turned out that it was fine. I'm not a lawyer though.

[1] https://github.com/hyperledger/cactus/blob/0ae4ddb371febcb479d7749a19d98c14ec710d82/contribs/Fujitsu-ConnectionChain/connection-chain/environment/base/cc_env/servers/cooperation/ecSide/ec2_connector/lib/dependent/ServerPlugin.js#L2

petermetz:
> @petermetz is this going to be part of V1?

@izuru0 There's no implementation to it yet so in that sense the answer is an implicit no. If we do manage to get any parts of the implementation in prior to 1.0 as well, then my suggestion will be to mark those as beta/dev preview features, meaning that it would be accessible at runtime, but in the documentation marked as not yet production ready and then we can maybe roll it out at a 1.1 release or whenever we feel ready.

packages/cactus-core-api/src/main/proto/weaver/common/protos/README.md
izuru0: thanks for your comment and I think it is ok then

izuru0:
thanks!

packages/cactus-core-api/src/main/proto/weaver/common/protos/README.md
petermetz: @izuru0 @VRamakrishna  Thank you both, I'll get on with the merging then.

jonathan-m-hamilton (approved):
LGTM

Issue #238, opened 2020-07-24
test: migrate from test runner tap to jest


petermetz:
### Update 2021-08-09

This used to be about AVA but now is about Jest

### Update 2020-11-02

Switched the target to AVA from jest, details below in comments.

### Description
As a developer I want to be able to write unit tests for the universal packages that execute the same code both in NodeJS and in the browser so that I don't have to duplicate work in order to have test coverage for both platforms for the cross-platform packages.

While working on the stable signature generation, we've discovered that accidentally the wrong library was used initially (congratulations to Peter) which is called "tap" instead of "tape". They both do the same thing (test runners for TAP - `Test Anything Protocol`) but the `tap` library does not work in browsers while the `tape` library is, so to keep the build/test infrastructure of the project as simple as possible it would be the best to migrate to `tape` entirely since it anyway has support for `TAP` (the protocol)

https://www.npmjs.com/package/ava

### Acceptance Criteria
1. Complexity of build/test scripts has not increased
2. Tests for cross-platform packages such as `cactus-common` can be written once and ran "anywhere" (NodeJS, browser)
3. Developer workflow has not lengthened, did not increase in complexity (from the human perspective, not code-wise that was mentioned in 1))

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo

petermetz:
Good news discovered in the meantime: Since both `tap` and `tape` use the `Test Anything Protocol` under the hood, they can actually run each other's test cases without issues. E.g. you can do `npx tap some-test-case.test.ts` and if will "just work" regardless of the test case being written with `tap` or `tape`.
This is good news because it means we can gradually perform the migration without any sort of unstable transition required.

petermetz:
Refactoring this as per https://github.com/hyperledger/cactus/issues/299#issuecomment-720743950 to be about migrating to AVA not tape...

petermetz:
https://www.npmjs.com/package/ava

petermetz:
```javascript
module.exports = {
  preset: "ts-jest",
  testEnvironment: "node",
  setupFilesAfterEnv: ["jest-extended"],
  testTimeout: 60 * 60 * 1000,
  testMatch: [
    `**/cactus-*/src/test/typescript/unit,integration,benchmark}/**/*.{test,spec}.ts`,
  ],
};

```

```sh
./packages/cactus-core/src/test/typescript/integration/api-surface.test.ts
[...]

PR #1550, opened 2021-11-18
test: jestify v2.3.0 deploy contract from json test

awadhana:

File Path:
packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json.test.ts


takeutak (approved):

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json.test.ts
```diff
@@ -451,45 +435,9 @@ test(testCase, async (t: Test) => {
...
-      "Total Transaction Count of 5 recorded as expected. RESULT OK.",
-    );
-    t2.end();
+    expect(res);
+    expect(res.data);
+    expect(res.status).toEqual(200);
+    expect(res.data).toContain(promMetricsOutput);
   });
-
-  test("deploys contract via .json file with constructorArgs", async (t2: Test) => {
```
petermetz, line 464: @awadhana Maybe a rebase gone wrong here, this test seems to be missing from the Jestified version.

petermetz (changes requested):
@awadhana See above for more comments + before merging please make sure the PR description and the git commit are updated to not resolve the main jest issue.

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json.test.ts
awadhana: Yup, I'm going to have to re-review this one and add that one back. Thank you, I missed that!

petermetz: @awadhana No worries, similar things happen to me all the time too!

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json.test.ts
```diff
@@ -474,22 +455,11 @@ test(testCase, async (t: Test) => {
...
+    expect(deployOut).toBeTruthy();
+    expect(deployOut.transactionReceipt).toBeTruthy();
+    expect(deployOut.transactionReceipt.contractAddress).toBeTruthy();
 
     contractAddress = deployOut.transactionReceipt.contractAddress as string;
-    t2.ok(
-      typeof contractAddress === "string",
-      "contractAddress typeof string OK",
-    );
+    expect(typeof contractAddress).toBeString;
```
petermetz, line 463: @awadhana Two things:

1. You are asserting that the `typeof contractAddress` is a string, but that will always be true because the `typeof` keyword always returns a string (it returns the name of the actual type in a string format, like in this case it would literally return "string" indicating that `contractAddress` has a type of string. So you need to remove the `typeof` and that way the assertion will check for the type of `contractAddress` instead of the type of the type of `contractAddress` (very meta, I know)
2. You are referencing the function pointer itself with `toBeString` meaning that you are not actually calling the function and my guess is that nothing actually happens in this line of code. You want to use the parentheses syntax to call function. Whenever in doubt, you can double check this in the IDE directly (thanks to the Typescript type information). See below:
Screenshot from 2022-01-27 12-58-08

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json.test.ts
```diff
@@ -226,13 +231,9 @@ test(testCase, async (t: Test) => {
         },
         nonce: 2,
       });
-      t2.ifError(setNameOutInvalid.transactionReceipt);
+      fail("it should not reach here");
```
jagpreetsinghsasan, line 234: The fail message should be like, "getContractInfoKeychain method should fail for new name, but recieved" + setNameOutInvalid.transactionReceipt (same change at other similar places)

jagpreetsinghsasan (changes requested):

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json.test.ts
petermetz: > The fail message should be like, "getContractInfoKeychain method should fail for new name, but recieved" + setNameOutInvalid.transactionReceipt (same change at other similar places)

@jagpreetsinghsasan I applied the fixes in all 3 places where it was applicable. Marking as resolved.

petermetz (approved):
@awadhana LGTM

jagpreetsinghsasan (approved):

PR #1554, opened 2021-11-19
chore: clean package.json

RafaelAPB:
Removes resource script that is not present in any package;
webpack script does not need npm run all


takeutak (approved):

petermetz (approved):
@RafaelAPB Thank you very much, LGTM!

PR #1561, opened 2021-11-23
feat(connector-corda): enable Flow Database Access CorDapp

### LINKED ISSUE #1493: fix(connector-corda): samples-java AddTokenValueFlow does not work
**Describe the bug**

https://github.com/corda/samples-java/blob/master/Basic/flow-database-access/workflows/src/main/java/net/corda/samples/flowdb/AddTokenValueFlow.java

**To Reproduce**

Try to invoke the add token value flow and see it crash.

**Expected behavior**

Does not crash.

**Logs/Stack traces**

`N/A`

Dev machine

Ubuntu-20.04

1.0.0-rc.1

- Corda

elenaizaguirre:
Added new corda-all-in-one image for flow-database-access sample

Modified Corda main-server image to allow calls to invoke without
executing any transaction.

Modified Corda InvokeContractV1Response to add the new flowId
field to return the flow handle id instead of use the callOutput
field, which now is used for the data returned by the flow. The
transactionId is not required because we can call invoke without
executing any transaction.

Resolves #1493

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -40,6 +40,7 @@
         "execa",
         "faio",
         "fidm",
+        "flowdb",
         "fsouza",
         "GETHKEYCHAINPASSWORD",
         "ghcr",
--- a/packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
+++ b/packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
@@ -78,6 +78,49 @@ class ApiPluginLedgerConnectorCordaServiceImpl(
         val returnValue = flowHandle.returnValue.get(timeoutMs, TimeUnit.MILLISECONDS)
         val id = flowHandle.id
 
+        // allow returnValue to be something different to SignedTransaction
+        var callOutput: kotlin.Any = ""
+        var transactionId: kotlin.String? = null
+
+        if (returnValue is SignedTransaction) {
+            logger.trace("returnValue is SignedTransaction - using returnValue.id.toString() ...");
+            transactionId = returnValue.id.toString();
+
+            callOutput = mapOf(
+                "tx" to mapOf(
+                    "id" to returnValue.tx.id,
+                    "notary" to returnValue.tx.notary,
+                    "requiredSigningKeys" to returnValue.tx.requiredSigningKeys,
+                    "merkleTree" to returnValue.tx.merkleTree,
+                    "privacySalt" to returnValue.tx.privacySalt,
+                    "attachments" to returnValue.tx.attachments,
+                    "commands" to returnValue.tx.commands,
+                    // "digestService" to returnValue.tx.digestService,
+                    "inputs" to returnValue.tx.inputs,
+                    "networkParametersHash" to returnValue.tx.networkParametersHash,
+                    "references" to returnValue.tx.references,
+                    "timeWindow" to returnValue.tx.timeWindow
+                ),
+                "id" to returnValue.id,
+                "inputs" to returnValue.inputs,
+                "networkParametersHash" to returnValue.networkParametersHash,
+                "notary" to returnValue.notary,
+                "references" to returnValue.references,
+                "requiredSigningKeys" to returnValue.requiredSigningKeys,
+                "sigs" to returnValue.sigs
+            );
+          
+        } else if (returnValue != null) {
+            callOutput = try {
+                val returnValueJson = writer.writeValueAsString(returnValue);
+                logger.trace("returnValue JSON serialized OK, using returnValue ...");
+                returnValueJson;
+            } catch (ex: Exception) {
+                logger.trace("returnValue JSON serialized failed, using returnValue.toString() ...");
+                returnValue.toString();
+            }
+        }
+
         logger.info("Progress(${progress.size})={}", progress)
         logger.info("ReturnValue={}", returnValue)
         logger.info("Id=$id")
@@ -96,7 +139,7 @@ class ApiPluginLedgerConnectorCordaServiceImpl(
         // org.hyperledger.cactus.plugin.ledger.connector.corda.server.model.InvokeContractV1Response["returnValue"]->
         // net.corda.client.jackson.internal.StxJson["wire"]->net.corda.client.jackson.internal.WireTransactionJson["outputs"])]
         // with root cause
-        return InvokeContractV1Response(true, id.toString(), (returnValue as SignedTransaction).id.toString(), progress)
+        return InvokeContractV1Response(true, callOutput, id.toString(), transactionId, progress)
     }
 
     // FIXME - make it clear in the documentation that this deployment endpoint is not recommended for production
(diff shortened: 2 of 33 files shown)


packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
```diff
@@ -79,6 +78,25 @@ class ApiPluginLedgerConnectorCordaServiceImpl(
...
         val id = flowHandle.id
 
+        // allow returnValue to be something different to SignedTransaction
+        var callOutput: kotlin.Any? = null
+        var transactionId: kotlin.String? = null
+
+        if (returnValue is SignedTransaction) {
+            logger.trace("returnValue is SignedTransaction - using returnValue.id.toString() ...");
+            transactionId = returnValue.id.toString();
+            // returnValueChecked = returnValue.id.toString();
```
petermetz, line 88: @elenaizaguirre Would something like this work here so that the callOutput can stay mandatory?

```kotlin
callOutput = returnValue.tx.outputs;
```

packages/cactus-plugin-ledger-connector-corda/src/main/json/openapi.json
```diff
@@ -455,16 +455,16 @@
                 "type": "object",
                 "required": [
                     "success",
-                    "transactionId",
-                    "callOutput"
```
petermetz, line 459: @elenaizaguirre Left another comment in the service code, I'd love it if we could somehow keep the callOutput mandatory (had an idea in that other comment as well). Sorry for throwing around ideas instead of having concrete steps, the local test environment is not working on my machine right now for some reason.

petermetz (changes requested):

packages/cactus-test-tooling/src/main/typescript/corda/corda-test-ledger.ts
```diff
@@ -270,9 +270,11 @@ export class CordaTestLedger implements ITestLedger {
         hasDbMigrations: false,
       });
       this.log.debug(`Pulled jar (%o bytes) %o`, jarPath, jar.length);
+    } catch (e) {
```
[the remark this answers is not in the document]
elenaizaguirre: @petermetz I put this catch because I didn't want to throw an exception, each example has a different filename but I wanted it to go to the next one in case it couldn't find it. Now it is a switch case

tools/docker/corda-all-in-one/corda-v4_8-flowdb/Dockerfile
```diff
@@ -0,0 +1,139 @@
+FROM docker:20.10.2-dind
```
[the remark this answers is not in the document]
elenaizaguirre: Done!

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
elenaizaguirre: @petermetz As we have discussed before, I am returning other values

packages/cactus-plugin-ledger-connector-corda/src/main/json/openapi.json
elenaizaguirre: Done!

tools/docker/corda-all-in-one/corda-v4_8-flowdb/Dockerfile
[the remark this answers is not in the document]
petermetz: @elenaizaguirre Thank you very much!

tools/docker/corda-all-in-one/corda-v4_8-flowdb/build.gradle
```diff
@@ -0,0 +1,139 @@
+buildscript { //properties that you need to build the project
```
petermetz, line 1: @elenaizaguirre Could you please write a few sentences of explanation about why was there a need for a different gradle build file instead of the one that the official Corda kotlin samples repo ships with (if that one was broken/didn't work then that's a perfectly valid explanation but I want to make sure the information is there for future generations to have as well)

petermetz (changes requested):
@elenaizaguirre Sorry for the nits again.

tools/docker/corda-all-in-one/README.md
```diff
@@ -22,6 +22,28 @@ docker run --rm --privileged caio
...
+```
+
+# cactus-corda-4-8-all-in-one-flowdb
+
+> This docker image is for `testing` and `development` only.
+> Do NOT use in production!
+
+## Customization
+
+`build.gradle` file from this sample has defined a single node called PartyA. It was modified to deploy the same nodes as in the obligation sample to make it work with our CordaTestLedger:
```
elenaizaguirre, line 36: @petermetz I added this regarding your comment https://github.com/hyperledger/cactus/pull/1561#discussion_r762627108

petermetz: @elenaizaguirre Great, thank you very much!

.cspell.json
```diff
@@ -83,6 +84,7 @@
         "myroot",
         "mysecretpassword",
         "myvolume",
+        "nogotiation",
```
petermetz, line 87: @elenaizaguirre I think we can remove this. Didn't find any occurrence of it via free text search + looks like it has a typo in it.

packages/cactus-test-tooling/src/main/typescript/corda/corda-test-ledger.ts
```diff
@@ -259,30 +259,105 @@ export class CordaTestLedger implements ITestLedger {
...
-      const jarPath = path.join(cordappRootDir, jarRelativePath);
-      this.log.debug(`Pulling jar file from container at: %o`, jarPath);
-      const jar = await Containers.pullBinaryFile(container, jarPath);
-      jars.push({
-        contentBase64: jar.toString("base64"),
-        filename: `${sampleCordapp}_contracts-1.0.jar`,
-        hasDbMigrations: false,
-      });
-      this.log.debug(`Pulled jar (%o bytes) %o`, jarPath, jar.length);
+    switch (sampleCordapp) {
```
petermetz, line 262: @elenaizaguirre Could the additional information needed on a per case basis get defined in another enum instead of writing a switch statement here? (I'm guessing the difference is the jar file name's version numbers so you could create an enum that maps instances of `SampleCordappEnum` to jar file names and then retain the much simpler code here wihch would also then benefit from the compiler at development time)

petermetz (changes requested):

.cspell.json
elenaizaguirre: Done, I added it when I was doing different tests and forgot to remove it.

packages/cactus-test-tooling/src/main/typescript/corda/corda-test-ledger.ts
elenaizaguirre: @petermetz How about this? I have added this information to the one already available (sample root directory) in the enum in `sample-cordapp-enum.ts`. Do you prefer me to keep them separately or can we leave them in the same object?

petermetz: @elenaizaguirre Yeah that's perfect too, thank you very much for putting up with the nit-picks as usual.

petermetz (approved):
@elenaizaguirre LGTM

.cspell.json
petermetz: Cheers @elenaizaguirre

takeutak (approved):

Issue #1559, opened 2021-11-23
fix(connector-corda): codegen missing prior removal of files


petermetz:
**Describe the bug**

Based on this discussion [1] with @elenaizaguirre it seems like the files are not being deleted prior to them being regenerated which most likely lead to the strange situation we've found ourselves where files that are no longer being generated by the generator are still hanging around in/under version control simply because they were never deleted.

[1]: https://github.com/hyperledger/cactus/pull/1294#discussion_r748034881

**To Reproduce**

Delete the codegen kotlin code, run a build and observe that there is a git diff which there should be none if this was working fine.

**Expected behavior**

Deleting generated code and then running the build should always produce equivalent, consistent code.

**Hyperledger Cactus release version or commit (git rev-parse --short HEAD):**

`1.0.0-rc.2`

**Hyperledger Cactus Plugins/Connectors Used**

- Corda

**Additional context**

Came up during: https://github.com/hyperledger/cactus/pull/1294
more specifically
https://github.com/hyperledger/cactus/pull/1294#discussion_r748034881

PR #1562, opened 2021-11-23
fix(connector-corda): add script to remove files before generate them

elenaizaguirre:
generate-server script calls this new script before call
openapi-generator-cli

Closes: #1559


packages/cactus-plugin-ledger-connector-corda/package.json
```diff
@@ -13,7 +13,8 @@
   ],
   "scripts": {
     "generate-sdk": "openapi-generator-cli generate -i ./src/main/json/openapi.json -g typescript-axios -o ./src/main/typescript/generated/openapi/typescript-axios/",
-    "generate-server": "openapi-generator-cli generate -i ./src/main/json/openapi.json -g kotlin-spring -o ./src/main-server/kotlin/gen/kotlin-spring/ -c ./src/main-server/openapi-generator-config.yaml",
+    "clean-server": "del-cli 'src/main-server/kotlin/gen/kotlin-spring/src/**/{model,api}/*'",
```
petermetz, line 16: @elenaizaguirre Sorry for the nit pick as usual. You could add this deletion pattern to the root `clean` script and then we don't need to deviate as much from how the build system works in general:
`'src/main-server/kotlin/gen/kotlin-spring/src/**/{model,api}/*'`

(there's already a list of patterns in that clean script and I'm concerned that it's getting longer and longer as well, but at least if we have the pattern there then all the patterns are at the same place for newcomers it should be easier to discover/understand what gets deleted and when during the build process)

petermetz (changes requested):

petermetz (approved):
@elenaizaguirre Thank you, LGTM

takeutak (approved):

PR #1564, opened 2021-11-23
fix(cmd-api-server): build occasionally broken - protoc-gen-ts #1563

### LINKED ISSUE #1563: fix(cmd-api-server): build occasionally broken - protoc-gen-ts
**Describe the bug**

Sometimes when I switch to the main branch and do a `yarn configure` it ends 

**To Reproduce**

Switch branches, do rebuilds (a lot) because of working on a bunch of different things simultaneously + reviewing PRs and then it will probably happen around once a week that the build fails because the codegen script bugged out and forgot to generate the Typescript implementation files from the protocol buffers of the API server and you are getting errors from the tsc script because it's missing the TS source code it needs to compile everything. 

**Expected behavior**

Build is stable.

`1.0.0-rc.2`

N/A

**Additional context**

This has been around for a while and I'm suspicious that it's a bug with `protoc-gen-ts` itself so the first step would be to try and upgrade that and see if it fixes the problem.

petermetz:
Upgrading protoc-gen-ts from 0.4.0 to 0.6.0 instantly fixed the issue
while I had it in a reproduced state (e.g. my build was broken on main)

This is not necessarily evidence that the fix is legit, but it's most likely
a step in the right direction and there's also the possibility that this
is the actual fix. It would take too much time to properly debug if
it really is the fix, so I'll just give it the benefit of the doubt
for now and assume that it is.

Fixes #1563


jonathan-m-hamilton (approved):
LGTM

PR #1578, opened 2021-11-24
test: jestify replace-all test

awadhana:

File Path:
packages/cactus-common/src/test/typescript/unit/strings/replace-all.test.ts


petermetz (approved):
@awadhana LGTM, thank you!

takeutak (approved):

petermetz:
@awadhana Different error message here but might be the same problem actually (and now I'm thinking it might be that it just needed an update, let's see how the CI does this time around)

```sh
Unable to find Jest's global expect.
Please check you have added jest-extended correctly to your jest configuration.
See https://github.com/jest-community/jest-extended#setup for help.
/home/runner/work/cactus/cactus/packages/cactus-common/src/test/typescript/unit/strings/drop-non-printable.test.ts:2
import "jest-extended";
                                                                                                                                                                                                                                                                                                                                              ^
ReferenceError: test is not defined
```

petermetz:
@awadhana I think the issue is that in the taprc you changed the ignore pattern, somehow the drop printable test get into the mix but this PR is about the replace all test. The checks fail because of the ignore patterns being the way they are leads to the test being run by the wrong test runner (Jest vs. tap)

awadhana:
> @awadhana I think the issue is that in the taprc you changed the ignore pattern, somehow the drop printable test get into the mix but this PR is about the replace all test. The checks fail because of the ignore patterns being the way they are leads to the test being run by the wrong test runner (Jest vs. tap)

@petermetz  weird, but I've just amended and pushed :) Hopefully when everything is done running it'll be good for a merge. Thank you!

petermetz (changes requested):
@awadhana Squash is still missing

petermetz (approved):
@awadhana LGTM

PR #1587, opened 2021-11-24
test: jestify bools test

awadhana:

File Path:
packages/cactus-common/src/test/typescript/unit/bools.test.ts


petermetz:
@awadhana This one looks like it's already been migrated. I think we could just close this one, what are your thoughts?

takeutak (approved):

petermetz (commented):
@awadhana This one looks like it's already been migrated. I think we could just close this one, what are your thoughts?

awadhana:
> @awadhana This one looks like it's already been migrated. I think we could just close this one, what are your thoughts?

@petermetz  That sounds good to me :) I've also just rebased.

Issue #1596, opened 2021-11-29
refactor(cmd-socketio-server): Circular dependency fix and minor fixes for verifier tests


outSH:
Relates to #1573
More fixes are needed to finish all the verifier tests:

- Fixed circular dependency between routing-interface and VerifierFactory
- Added default validator config used by ValidatorAuthentication,
- type fixes in Verifier,
- format some files according to standard.

PR #1597, opened 2021-11-29
refactor(cmd-socketio-server): circular dependency fix and minor fixes for verifier tests

outSH:
Fix circular dependency between routing-interface and VerifierFactory,
add default validator config used by ValidatorAuthentication, type fixes
in Verifier, format some files according to standard.

Closes: #1596


takeutak (approved):

petermetz (approved):
@outSH Thank you, LGTM

PR #1615, opened 2021-12-01
feat(connector-xdai): remove hard dependency on keychain

### LINKED ISSUE #1162: feat(connector-xdai): remove hard dependency on keychain

elenaizaguirre:
Resolves #1162 


takeutak (approved):

packages/cactus-plugin-ledger-connector-xdai/src/main/json/openapi.json
```diff
@@ -446,6 +487,51 @@
                         "minimum": 0,
                         "default": 60000,
                         "nullable": false
+                    },
+                    "contractJSON": {
+                        "type": "object",
+                        "description": "For use when not using keychain, pass the contract in as this variable",
```
petermetz, line 493: @elenaizaguirre Is it contradictory to have a contractJSON and a mandatory keychain ID parameter in the same request object definition?

elenaizaguirre: @petermetz 
We have this definition (like Quorum endpoints) because in the plugin connector the request object is used to generate the encoded ABI in both situations, when we use keychain and when we use directly the object to call the ws. 
The difference is that in the keychain endpoint this field is not required because is filled with the data retrieved from the memory keychain, but we make sure that we keep the structure and do not need conversions to type any

petermetz: @elenaizaguirre It is very possible that I failed to review the quorum connector's changes at similar detail at the time so apologies if I'm asking questions that should've been asked back then.

If the reason for having an optional parameter added is to avoid having to do a cast, that could be solved with a common supertype for the two request types which then has no need to have optional parameters (if I'm understanding it correctly, but feel free to grab me for some pair programming too if this is not making sense)

The underlying principle of what I'm saying is that it's better to have more types and some inheritance in this case than having less types but some with optional parameters (because the optional parameters end up needing checks and error handling of their own so the complexity is bleeding deeper into the codebase instead of getting nipped in the bud at the initial validation phase) 

So for example I would favor type A, B and C where B and C are sub-types of A and none of the 3 types need optional parameters compared to having only 2 types X and Y but then one or both of those need optional parameters because we could not express exactly the needed shapes for the different use-cases.

petermetz: @elenaizaguirre Sorry for the slow communication on this one. Please eliminate the `any` types as we discussed in today's call and then the change should be good to go.

petermetz (changes requested):
@elenaizaguirre Sorry for the slow communication on this one. Please eliminate the any types as we discussed in today's call and then the change should be good to go.

packages/cactus-plugin-ledger-connector-xdai/src/main/json/openapi.json
elenaizaguirre: @petermetz Do you want me to create a new issue to make this change also in the other plugins?

elenaizaguirre:
> @elenaizaguirre Sorry for the slow communication on this one. Please eliminate the any types as we discussed in today's call and then the change should be good to go.

@petermetz Done! Thanks for your help

packages/cactus-plugin-ledger-connector-xdai/src/main/typescript/plugin-ledger-connector-xdai.ts
```diff
@@ -322,6 +297,100 @@ export class PluginLedgerConnectorXdai
...
+    }
+    const keychainPlugin = this.pluginRegistry.findOneByKeychainId(keychainId);
+    if (!keychainPlugin.has(contractName)) {
+      throw new Error(
+        `${fnTag} Cannot invoke the contract because contractName is not in the keychainPlugin`,
+      );
+    }
+    const contractStr = await keychainPlugin.get(contractName);
+    const contractJSON = JSON.parse(contractStr);
+    (req as any).contractJSON = contractJSON; // remove
```
petermetz, line 321: @elenaizaguirre This line can be removed now, right?

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-xdai/src/main/typescript/plugin-ledger-connector-xdai.ts
elenaizaguirre: @petermetz Sure, sorry for the confusion

petermetz (approved):
> @petermetz Sure, sorry for the confusion 

@elenaizaguirre No worries, thank you again, LGTM!

packages/cactus-plugin-ledger-connector-xdai/src/main/json/openapi.json
petermetz: @elenaizaguirre Yes, that would be great! Thank you for the pro-activity! I recommend putting a direct link to this PR/specific conversations we had and some more explanation in those issue descriptions as well.

PR #1595, opened 2021-11-29
test: jestify plugin-registry test

awadhana:
Migrated test from Tap to Jest.

File Path:
packages/cactus-core/src/test/typescript/unit/plugin-registry.test.ts

This is a PARTIAL resolution to issue #238

Signed-off-by: awadhana <awadhana0825@gmail.com>

PR #1616, opened 2021-12-01
test: jestify transfer-initiation test

awadhana:

File Path:
packages/cactus-plugin-odap-hermes/src/test/typescript/integration/odap/transfer-initiation.test.ts


awadhana:
@petermetz these have a .not.toThrow() method without the message. I would wait a minute before merging to make sure this is the way you want them. Emailed you about one of the PR's I did a message with but still have errors with.

petermetz:
> @petermetz Another PR with .toThrow() and .not.toThrow() (in place of doesNotThrow) that I'm having an issue keeping the message with. I tried the Jest file that you've mentioned as well as https://jestjs.io/docs/expect#tothrowerror which has a few different ways to throw an error depending on how strict the equal is for the error message. Similar to the issue in #1595

@awadhana Could you put here in the comments the exact code that you tried that didn't work and the error logs they produced? That should help me provide more specific guidance (hopefully :-) )

petermetz (approved):
@awadhana Just realized this one is also like the other one meaning that you are good to go because it should be fine as-is on account of the lack of throwing is being asserted not the other way around so if the tests are passing then it's LGTM all the way.

PR #1618, opened 2021-12-01
test: jestify transfer-commence test

awadhana:

File Path:
packages/cactus-plugin-odap-hermes/src/test/typescript/integration/odap/transfer-commence.test.ts

This is a PARTIAL resolution to #238


awadhana:
@petermetz these have a .not.toThrow() method without the message. I would wait a minute before merging to make sure this is the way you want them. Emailed you about one of the PR's I did a message with but still have errors with.

takeutak:
@awadhana I think this is LGTM, but I'd like to wait the approval according to your above comment.  After the above trouble is fixed, I will review it again.

awadhana:
> @awadhana I think this is LGTM, but I'd like to wait the approval according to your above comment. After the above trouble is fixed, I will review it again. If you have any inconvenience, please contact me.

@takeutak, thank you for the offer! It's no problem at all, I can tag you when it's ready for review again, if you like, but I'll be circling back to this later.

petermetz (approved):
@awadhana Took a more detailed look finally: You don't need to keep the message for the negative cases (when it is being asserted that there is NO throw). So this is good to go as-is IMO. Thank you! :-)

takeutak (approved):

awadhana:
> @awadhana Took a more detailed look finally: You don't need to keep the message for the negative cases (when it is being asserted that there is NO throw). So this is good to go as-is IMO. Thank you! :-)

Good to know, thank you @petermetz!

Issue #1619, opened 2021-12-02
fix(api-server): shutdown hook configuration is using wrong config key


m-courtin:
**Describe the bug**

In the api-server constructor the evaluation for the activation / deactivation of the shutdown hook is based on the wrong configuration property so that it always is falling back to the default value and pre-configured value is not taken.

Depends on #1649

m-courtin:
@petermetz: Can you please assign this issue onto my name as discussed yesterday

petermetz:
@m-courtin You got it, thank you for reporting!

PR #1621, opened 2021-12-03
fix: shutdown hook configuration is using wrong config key

m-courtin:
In the api-server constructor the evaluation for the activation /
deactivation of the shutdown hook is based on the wrong configuration
property so that it always is falling back to the default value and
pre-configured value is not taken.

Closes: #1619


takeutak (approved):

packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
```diff
@@ -109,8 +109,10 @@ export class ApiServer {
       throw new Error(`ApiServer#ctor options.config was falsy`);
     }
 
-    this.enableShutdownHook = Bools.isBooleanStrict(options.enableShutdownHook)
-      ? (options.enableShutdownHook as boolean)
+    this.enableShutdownHook = Bools.isBooleanStrict(
```
petermetz, line 112: @m-courtin Could you also add a unit test written in Jest (not tap/tape-promise) that asserts for both the `true` and `false` cases that after the instantiation of the API server, the config value ends up as one would expect it to?

petermetz (changes requested):

packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
m-courtin: @petermetz Yes, sure thing. I will add a corresponding UT

m-courtin:
@petermetz: UTs are added and during debugging of the new UTs 2 additional bugs came up -> I will create corresponding bugs

packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
petermetz: @m-courtin Great, thank you very much!

petermetz (approved):
@m-courtin Cheers, LGTM, I will merge this once the tests are passing (once the prerequisite PR was merged)

PR #1623, opened 2021-12-03
fix(supply-chain-app): enable cockpit in supply-chain

### LINKED ISSUE #1622: example supply-chain does not work running or debugging locally
**Describe the bug**

When you are running or debugging supply-chain-app locally, http://localhost:3200/bamboo-harvest-list cannot be accessed

**To Reproduce**

1. Run or debug supply-chain-app as described in readme file.
2. Try to access http://localhost:3200/bamboo-harvest-list from browser
3. Localhost rejects the connection

**Expected behavior**

supply-chain application should be displayed in the browser

elenaizaguirre:
Resolves #1622


takeutak (approved):

petermetz (approved):

Could you please work together with @AzaharaC and also figure out/document why the manual tests didn't catch this at the time when we made the change [1]? The best would be if we had end to end test coverage for this otherwise the examples will keep breaking all the time.

[1] https://github.com/hyperledger/cactus/pull/1304#issuecomment-969077301

AzaharaC:
@petermetz In this case, the error was not detected because in the comment [(#1304 comment)](https://github.com/hyperledger/cactus/pull/1304#issuecomment-969077301) only said to test the part of the section `Running the Example Application Locally` and in this way, the application continued to work correctly.

petermetz:
@AzaharaC Do we know why? And once we do know the reason, can a test be written to prevent this problem next time?

AzaharaC:
@petermetz last week I was trying to find the reason for this with @elenaizaguirre. We run the application and it worked for me but not for Elena.
I ran a clean to make sure it was not using old code, but application still worked for me.
We thought that maybe, it could be some environment variable that I had in my local, but this was not the case.
Checking the application code, it makes sense that it should fail when the cockpit is deactivated, as in Elena's case. But in my case, the application works and we could not find the reason.
The value of the variable `cockpitEnabled` for Elena is `false` and this does not change, but it takes the value `true` in my case. I have not been able to determine in which exact part it takes this value.

petermetz:
@AzaharaC @elenaizaguirre Alright, thank you for putting the extra time in to investigate at least now we know that there's something going on and that more effort is needed to figure out the underlying issues. I opened a bug to track this task in the future. Thank you again.

I opened the issue here: https://github.com/hyperledger/cactus/issues/1653

Issue #1648, opened 2021-12-10
fix(config-service): faulty shutdownHook definition in the Config-Schema definition


m-courtin:
**Describe the bug**

In the config-service of the api-server the configuration value for "shutdownHook" is not getting applied due to an issue in the reducer of the newExampleConfigEnv(). The reducer is not populating the shutdownHook configuration based on a faulty Config-Schema definition for the shutdownHook where "arg" and "env" are missing.

m-courtin:
@petermetz: Can you please assign this issue onto my name

petermetz:
@m-courtin You got it!

PR #1649, opened 2021-12-10
fix(config-service): faulty shutdownHook definition in the Config-Schema

m-courtin:
In the config-service of the api-server the configuration value for
"shutdownHook" is not getting applied due to an issue in the reducer of
the newExampleConfigEnv(). The reducer is not populating the
shutdownHook configuration based on a faulty Config-Schema definition
for the shutdownHook where "arg" and "env" are missing.

This PR needs to be approved first as it is enabling PR1621 to execute the unit tests in correct way

Closes: #1648


m-courtin:
@petermetz: Can you please review / approve this PR before continuing with PR 1621 as PR 1621 needs to get this PR first to successfully execute the api-server config UTs

takeutak (approved):

petermetz (approved):
@m-courtin Thank you very much, LGTM

PR #1652, opened 2021-12-13
fix: config-service validator throws warnings

m-courtin:
In multiple unit tests the config-validator of the config service throws
warnings as the validate function is finding properties on the instances
of plugins and the corresponding schema is NOT providing the related
schema definition required by the plugin interface.


petermetz (changes requested):
@m-courtin I agree with the direction, but the tests don't seem to be working (and it doesn't look like a flake either)

m-courtin:
@petermetz: This PR is only about a fix of the schema validator so that during the UTs no further warnings are spamed into the logs regarding validation. The PR is not changing any test functionality

m-courtin:
@petermetz: yes, totally agree. I will check why the tests are failing

m-courtin:
@petermetz: Found the issue why the tests were failing and fixed it

petermetz (approved):
@m-courtin Thank you, LGTM! Fingers crossed that the flaky tests won't get in the way with the following executions.

takeutak (approved):

m-courtin:
@petermetz: Test execution went fine :) Can you please merge the PR into main

petermetz:
@m-courtin Yup, done now, thank you again ;-)

PR #1655, opened 2021-12-14
fix(cmd-api-server): disable validateKeyPairMatch

elenaizaguirre:
Disable validateKeyPairMatch call from config-service create method
because it fails with supply-chain default configuration and is going to
be removed when cleaning parameters


petermetz (approved):

jagpreetsinghsasan (approved):
LGTM

Issue #1602, opened 2021-11-30
refactor(cmd-api-server): add socket.io connection to each ledger-connector


takeutak:
**Is your feature request related to a problem? Please describe.**
On the current codes, cmd-api-server does not have socket.io connection to each ledger-connector.  This communication channel is needed for developing the APIs of sendSyncRequest, sendAsyncRequest, and startMonitor as these are developed in cmd-socket-server via socket.io connection to each ledger.
This feature was already developed in cmd-socket-server, but in case of merging those codes as #1502, we need to develop it.

**Describe the solution you'd like**
Add the socket.io connection to cmd-api-server

**EPIC**
#1605

petermetz:
@takeutak Could you please expand on this part? `On the current codes, cmd-api-server does not have socket.io connection to each ledger-connector.` Support for socket.io was added a long time ago.

takeutak:
@petermetz Thanks for your advice!
As I received your advice, I looked the codes again.

Let me clarify one thing, is the following figure right for calling your mentioned codes?
 - figure: https://github.com/hyperledger/cactus/blob/main/packages/cactus-plugin-ledger-connector-besu/docs/architecture/images/run-transaction-endpoint.png
 - API client: https://github.com/hyperledger/cactus/blob/d5e1708e84ab5192d07410d5120e144d477ef6ce/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/api-client/besu-api-client.ts
 - APi server: Is it right? > https://github.com/hyperledger/cactus/blob/main/packages/cactus-cmd-api-server/src/main/typescript/api-client/api-server-api-client.ts
 - Ledger connector: https://github.com/hyperledger/cactus/blob/9022064e245a92f71d2d303d77bfdaf64d1b1678/packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts

According to the figure, BLP sends requests to API clients on each ledger connector via socket.io connection (*1), and API clients sends requests to ledger connector via socket.io connection (*2), is it right?
 (*1) BLP <--(socket.io)--> API clients on each ledger connector 
 (*2) API clients on each ledger connector <--(socket.io)--> ledger connector

I think your implementation is good, but it makes more useful to prepare the following ways:
 - abstraction of calling API client on each ledgers when (*1) is executed.  I think Verifier structure resolve the feature.
[...]

PR #1660, opened 2021-12-15
feat(cactus-api-client): add support for plain socketio validators in api-server and api-client

outSH:
Full description of changes and planning log available in #1602. In
general, this commit adds two new components - one for communicating
with plain socketio validators using socketio interface, and verifier
that implements similar features as socketio-server verifier.

Closes: #1602


packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
```diff
@@ -0,0 +1,345 @@
...
+          this.log.error("####Error:", err);
+          this.socket.disconnect();
+          reject(err);
+        });
+        this.socket.on("error", (err: Record<string, unknown>) => {
+          this.log.error("####Error:", err);
+          this.socket.disconnect();
+          reject(err);
+        });
+        this.socket.on("response", (result: any) => {
```
petermetz, line 164: @outSH How do different types of responses get disambiguted here? Is this assumption baked in that the entire system only supports one type of response for a single use-case pertaining to a single endpoint/backend logic?

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
```diff
@@ -0,0 +1,345 @@
...
+  ): void {
+    try {
+      const requestData = {
+        contract: contract,
+        method: method,
+        args: args,
+      };
+
+      this.log.debug("sendAsyncRequest() Request:", requestData);
+      this.socket.emit("request2", requestData);
```
petermetz, line 118: @outSH How do different types of requests get disambiguted here? Is this assumption baked in that the entire system only supports one type of request for a single use-case pertaining to a single endpoint/backend logic?

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
```diff
@@ -0,0 +1,345 @@
...
+};
+
+export class SocketLedgerEvent {
+  id = "";
+  verifierId = "";
+  data: Record<string, unknown> | null = null;
+}
+
+// TODO - To entire class - analyze and handle broken connection / socket disconnected scenarios
+export class SocketIOApiClient implements ISocketApiClient<SocketLedgerEvent> {
```
petermetz, line 63: @outSH Bringing this over from the other thread, how about `LegacyApiClient` for the name?

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
```diff
@@ -0,0 +1,345 @@
...
+import { Observable, ReplaySubject } from "rxjs";
+import { finalize } from "rxjs/operators";
+
+// TODO - Remove any from here
+export function verifyValidatorJwt(
+  keyPath: string,
+  targetData: string,
+): Promise<any> {
+  return new Promise((resolve, reject) => {
+    const publicKey = readFileSync(resolvePath(__dirname, keyPath));
```
petermetz, line 30: @outSH Could this function benefit from being made asynchronous instead of blocking?

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
```diff
@@ -0,0 +1,345 @@
...
+
+const defaultMaxCounterRequestID = 100;
+const defaultSyncFunctionTimeoutMillisecond = 5000;
+
+import { Logger, Checks } from "@hyperledger/cactus-common";
+import { LogLevelDesc, LoggerProvider } from "@hyperledger/cactus-common";
+import { ISocketApiClient } from "@hyperledger/cactus-core-api";
+
+import { Socket, SocketOptions, ManagerOptions, io } from "socket.io-client";
+import { readFileSync } from "fs";
```
petermetz, line 18: @outSH How will this work in a browser environment? Or is this class entirely only designed for the back-end?

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
```diff
@@ -0,0 +1,345 @@
...
+        this.log.debug("set timeout");
+
+        // Time-out setting
+        const timeoutMilliseconds =
+          this.options.syncFunctionTimeoutMillisecond ||
+          defaultSyncFunctionTimeoutMillisecond;
+        setTimeout(() => {
+          if (responseFlag === false) {
+            this.log.debug("requestTimeout reqID:", reqID);
+            resolve({ status: 504, amount: 0 });
```
petermetz, line 209: @outSH What does `amount` refer to here? Is that something specific to a certain use-case?

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
```diff
@@ -0,0 +1,345 @@
...
+        this.log.error(`##Error: startMonitor, ${err}`);
+        this.monitorSubject.error(err);
+      }
+
+      return this.monitorSubject.pipe(
+        finalize(() => {
+          if (this.monitorSubject && !this.monitorSubject.observed) {
+            // Last observer finished
+            this.log.debug("##emit: stopMonitor");
+            this.socket.emit("stopMonitor");
```
petermetz, line 327: @outSH How do these string constants get disambiguiated on the back-end when there are multiple different endpoints that have different async use-cases?

packages/cactus-api-client/src/main/typescript/verifier.ts
```diff
@@ -0,0 +1,93 @@
...
+// TODO - Document classes and methods
+
+import { Subscription } from "rxjs";
+
+import { Logger } from "@hyperledger/cactus-common";
+import { LogLevelDesc, LoggerProvider } from "@hyperledger/cactus-common";
+import { ISocketApiClient } from "@hyperledger/cactus-core-api";
+
+export interface VerifierEventListener<BlockType> {
+  onEvent(ledgerEvent: BlockType): void;
```
petermetz, line 17: @outSH Would it make sense to have an onError handler here so that it can used by the callsite when something bad happens (I saw somewhere else in the code that right now it just logs errors out and then moves along without doing anything else)

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: @petermetz socketio-server At the moment Verifier is meant to be used in a backend only, I've adjusted webpack imports to reflect that (so browser package will not contain these functions). Also, I've mentioned in the github issue that I plan adding browser support in the future, but definitely in separate PR.

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: @petermetz No typecheck here, it's assumed that the caller knows what is in the response. This is a lift-and-shift implementation from socketio-server, can be improved in the future (along with fixing other loose-typing artifacts like any/object)

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: @petermetz Same as above, only backend support for now.

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: @petermetz There's only one single endpoint per "legacy" socketio validator, and all validators stop when they receive "stopMonitor", so there is no ambiguity (or I misunderstood your question)

packages/cactus-api-client/src/main/typescript/verifier.ts
outSH: Yes, I've even included this in the TODO comment below in this file, I wanted to add a required onError callback in the following refactor PR. Now, when I think of it, it's more reasonable to add onError optional callback, which can be added now and will not trigger changes in multiple places through the code :)

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: I've changed this to `readFile()`, and also fixed fixed all `any` types in this function.

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: @petermetz I consulted japanese colleagues and it seems this amount field is obsolete. I've removed this in most recent version.

packages/cactus-api-client/src/main/typescript/verifier.ts
outSH: Added optional onError callback and it's unit test in the most recent version.

packages/cactus-test-api-client/src/test/typescript/integration/verifier-integration-with-openapi-connectors.test.ts
```diff
@@ -0,0 +1,235 @@
+// Besu setup code based on:
```
outSH, line 1: Added integration test for verifier -> BesuApiClient integration

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
izuru0: I don't want jump on to your idea to change the name to LegacyApiClient. We may have some discussion to decide what name is the best one. But for this PR, can we focus on what is described in the commit message?

outSH:
@petermetz PR is complete from my side, I've also answered all your questions from draft phase, please review

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
petermetz: @outSH Do you mean that you'd open a brand new connection for each request/response exchange? What I mean is how would the same connection get shared between different request/response cycles?

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
petermetz: @outSH What I mean is how would something like this be possible on the same socket connection?
Would you do routing based on the payload of the message with some lookup at runtime for the correct handler?

```typescript

const someLogic = () => {
// do something
}

const someOtherDifferentLogic = () => {
// do something else
}

socket.on("request", someLogic);
socket.on("request", someOtherDifferentLogic);
```

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
petermetz: @izuru0 Sure, I am.

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
petermetz: @outSH Got it. When do plan on adding browser support in there?

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
petermetz: @outSH What I'm getting at is that I would assume that there are multiple endpoints if this is to be the `generic abstraction` as mentioned in the other comments of the originating issue and that causes ambiguity. You said in the other comment that this is a lift-and-shift which makes much more sense to me instead of describing it as a generic abstraction with a value add it's more like a lift-and-shift as-is. Which I am okay with as long as we call it what it is because this then feeds into later discussions about release management as well (which is why I'm having so many questions here - not to be annoying with small details).

petermetz:
> @petermetz PR is complete from my side, I've also answered all your questions from draft phase, please review

@outSH I responded to each of them, but it's really just the same conversation in different threads. The bottom line is here IMO so we could just have it there: https://github.com/hyperledger/cactus/pull/1660#discussion_r772653517

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: @petermetz It's hard to estimate at the moment, I create a new issue for this problem

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: I don't recall calling `SocketIOApiClient` a generic abstraction, it is a component to communicate with socketio connectors with validator-style interface, as it is used in `cmd-socketio-server` at the moment, with as little change as possible. There's only one request<->response channel per connection, so there's no need to differentiate responses. Creating / deciding the common socketio interface has to be done in the near future (in other place), at the moment we want to support current validators as the way they worked so far. 

`ISocketApi` is the generic abstraction ensuring certain methods are available. How they are handled is managed by specific implementation (connection sharing and stuff), each ledger can implement different logic as long as they provide specified methods.

petermetz: > I don't recall calling SocketIOApiClient a generic abstraction,

@outSH IIRC it was in one of Takuma's comments describing the change in general.


> it is a component to communicate with socketio connectors with validator-style interface, as it is used in cmd-socketio-server at the moment, with as little change as possible. There's only one request<->response channel per connection, so there's no need to differentiate responses. Creating / deciding the common socketio interface has to be done in the near future (in other place), at the moment we want to support current validators as the way they worked so far.

> ISocketApi is the generic abstraction ensuring certain methods are available. How they are handled is managed by specific implementation (connection sharing and stuff), each ledger can implement different logic as long as they provide specified methods.

I understand. I feel like we are talking past each other though. I'll try from a different angle: Do you expect this (in it's current form) to be part of the 1.0 GA release early next year?

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
petermetz: @outSH Cool, thank you. If you have the issue opened please send a link here as to that one as well (to establish the link/reference)

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: > Do you expect this (in it's current form) to be part of the 1.0 GA release early next year?

The goal of my Japanese colleagues is to communicate `api-server` with current socketio-validator connectors, **without changing the current validator interface**, and also implement connector-agnostic `Verifier`. Both of these are done, one thing left is to upgrade socketio-validator connectors to socketio 4 so they can actually talk with `api-server` (which uses socketio 4), I plan to do this in separate PR by the end of this year. Beside that, all the goals are already met by this PR and from my point of view should be included in the GA.

Most of your comments were about interface between `SocketIoApiServer` and socketio-validator connector, which we can't change in near future, but I will create a separate github issue where I'd like to introduce few proposals, and start a discussion regarding the common socketio interface with all connectors. This also relates to naming `SocketIoApiServer`.

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: @petermetz Done - https://github.com/hyperledger/cactus/issues/1675

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
petermetz: @outSH Okay, cool. :-)

outSH:
@petermetz 

As mentioned in one of the comments, we'd like to have a discussion regarding socket.io API. I've created a github issue with proposal here - #1703. I see all discussion was resolved in this PR, is there anything else to do or clarify before the merge?

packages/cactus-core-api/src/main/typescript/plugin/ledger-connector/i-socket-api-client.ts
```diff
@@ -0,0 +1,25 @@
+import { Observable } from "rxjs";
```
petermetz, line 1: @outSH

```suggestion
import type { Observable } from "rxjs";
```

This way the entire rxjs library won't end up in the bundle of the core-api package because the compiler will throw out the import at compilation time (we want to keep the core-api package as light as possible and decoupled from the implementation bundles).

This will also allow you to move `rxjs` into the `devDependency` section of the package.json, I'll make another comment about that.

packages/cactus-core-api/package.json
```diff
@@ -81,6 +81,7 @@
   },
   "dependencies": {
     "@hyperledger/cactus-common": "1.0.0-rc.3",
-    "axios": "0.21.4"
+    "axios": "0.21.4",
+    "rxjs": "7.3.0"
```
petermetz, line 85: @outSH Please see my other comment with the `import type` suggestion for details. TLDR here: you can move `rxjs` to the `devDependency` section to reduce bundle size for the core-api package.

Also: apologies for not spotting this sooner.

petermetz (changes requested):
@outSH ^^Last two nits above, then we are good to go (sorry for the last minute CRs)

packages/cactus-core-api/src/main/typescript/plugin/ledger-connector/i-socket-api-client.ts
outSH: @petermetz Cool, I didn't like including the full package but wasn't aware you can import just the typedefs, thanks! Already changed

packages/cactus-core-api/package.json
outSH: Sure, done

petermetz (approved):
@outSH Thank you for the quick turnaround! LGTM

izuru0 (approved):
LGTM

PR #1661, opened 2021-12-16
fix: remove jade dependencies

StepniowskiP:
Removed jade dependencies in multiple package json files because of reported security flaw


StepniowskiP:
Closed because I forgot to create Issue for this PR

Issue #1647, opened 2021-12-09
refactor(cmd-socketio-server): include cmd-socketio-server in common monorepo setup


outSH:
- cmd-socketio doesn't integrate well with the rest of the monorepo structure (yet), it uses local npm_modules (not top level modules hoisted by yarn) and different, not inherited tsconfig (with newer typescript version).
- Running jest from the root dir assumes global, common config which causes existing tests to fail (jest treats all the files as single project and fails due to some module conflicts). That's why there's separate jest config inside cmd-socketio-server which must be removed after implementing this fix.
- Similar work was already done by Peter in #719. From initial check I see there are significant changes in the structure itself since that commit, and a some changes are not strictly related to this issue, but overall the PR will probably be helpful when implementing this so use it as a reference.

PR #719, opened 2021-03-24
refactor: dangling package.json inside packages directory #682

petermetz:
## Dependencies

Depends on #656 

## Commit to review

Changes
-----------

1. Moved the dangling folders under the package that was created
2. Fixed linter errors (1000+)
3. Also migrated the cartrade example to import via the package

TODO:

1. Other examples to be converted into packages that are part of the build
2. Examples made to work with the build system as well

Fixes #682

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

PR #1670, opened 2021-12-21
build: include cmd-socketio-server and socketio validators in the monorepo setup

outSH:
Multiple changes, but they are tightly related to same issue so I introduced them in single PR, hope it's okay :)

- Added two build steps in common packages.json to facilitate more complex build process of packages related to this PR. I will write another comment in code to this.
- Added standalone projects to the global monorepo setup, namely:
```
./packages/cactus-cmd-socketio-server
./packages/cactus-plugin-ledger-connector-fabric-socketio
./packages/cactus-plugin-ledger-connector-go-ethereum-socketio
./packages/cactus-plugin-ledger-connector-sawtooth-socketio
```
- All changed packages extends `tsconfig.base.json`, but most overwrites strict flag to false. This is the task for another PR. After fixing strict warnings we can merge verifier tests to the main jest config.
    - Related issue - https://github.com/hyperledger/cactus/issues/1671
- All projects builds in typescript 4 set by the root config.
- Updated the README files form connectors and examples to describe new (monorepo) build process. Also converted tabs to spaces.
- I've lost a while trying to debug socketio errors when I forgot to run the init step in example app (I was sure it was my changes that caused that). I don't see for now any benefit of running two separate commands, so I've merged the init-* step to the build process itself so it's always executed.
- Rewritten socketio and go-eheterum dockerfiles. Not strictly necessary for this PR, but now they are much smaller both in dockerfile lines and resulting container size. Tested with electricty-trade - works the same as before.
- **yarn.lock update** - Changes from previous commits (I think it wasn't regenerated for a while) and from removing redundant and duplicated packages in this issue.
- Fixed bug in Verifier that caused cartrade to fail compilation - for now just restored the previous implementation. Will be part of fixes to enable strict flag in this package (another task).
- Examples were not included in monorepo setup, since it'd be tricky at the moment. It's better to containerize them instead of doing some fixes now.
- Projects use `../../.build-cache` like the other ones, so manual `npm run build` in package dir can do nothing. Do builds from root dir or clear the cache.

Closes: #1647


package.json
```diff
@@ -39,9 +39,11 @@
     "build:prod": "npm-run-all build:prod:frontend",
     "build:prod:frontend": "lerna run build:prod:frontend",
     "build:dev": "npm-run-all build:dev:backend webpack:dev:web build:dev:frontend",
-    "build:dev:backend": "npm-run-all lint clean codegen tsc",
+    "build:dev:backend": "npm-run-all lint clean codegen build:dev:backend:prebuild tsc build:dev:backend:postbuild",
     "build:dev:frontend": "lerna run build:dev:frontend --scope=\"@hyperledger/cactus-example-supply-chain-frontend\" && lerna run build:dev:frontend --scope=\"@hyperledger/cactus-example-carbon-accounting-frontend\"",
     "build:dev:common": "lerna exec --stream --scope '*/*common' -- 'del-cli dist/** && tsc --project ./tsconfig.json && webpack --env=dev --target=node --config ../../webpack.config.js'",
+    "build:dev:backend:prebuild": "lerna run build:dev:backend:prebuild",
+    "build:dev:backend:postbuild": "lerna run build:dev:backend:postbuild",
```
outSH, line 46: @petermetz 
Projects changed in this PR needs some logic executed before and after it's build (to copy specific files in most cases). This can be simplified in the future, especially prebuild step could be removed I suppose, but for now I think adding pre/post build steps is the most clean solution. On the other hand, maybe it would be reasonable to replace plain `tsc` call with some lerna target present in all projects that would build the project? It's possible that in the future other projects may have more complicated build process, and this way we'd encapsulate all build logic in given package "build" npm script, instead of relying that they all will just work with `tsc`? What's your opinion on this?

petermetz: @outSH Been there, done that. :-) I called the script `resources` at the time: https://github.com/hyperledger/cactus/blob/0a13eff5b1bc3b8f45960d6ca59daf155ae446a0/package.json#L24 
back in https://github.com/hyperledger/cactus/pull/719
You are doing the same thing that I was back then, just with a different name for the post build script. (Which I'm fine with, whatever. My reasoning for using shorter script names where possible was that they are easier to type out when you are running them manually from a terminal but I also see the benefit if fully qualified names like the ones you gave them)

For the prebuild script, you could just use the `codegen` script instead of introducing a new one called `build:dev:backend:prebuild` which would have the performance benefit of running less lerna invocations (which are project wide). Even if the expression `codegen` does not exactly match what you'll end up doing in it inside the customized packages (but IMO it kinda does too) I think it would be descriptive enough compared to something generic like pre build.

Making `tsc` a lerna target could turn out to be a rabbit hole because whatever we do in there needs to play nice with the watch scripts and the IDE's own processes that powers auto-complete, syntax highlighting/etc. => 
The `tsc` script used to be a lerna target, but that meant that you had to exit and launch VSCode again every time you re-built the project because it would get out of sync and all the imports would go red in VSCode even though on the terminal the compiler was giving you no errors. I'm not saying that this is a reason to not to do it/look into it, just that there's a history there which we'd need to avoid repeating because the developer experience was quite bad then (not that it couldn't be improved now, but still even worse)

TLDR: I'm fine with this, up to you if you want to refactor it based on the above information.

package.json
```diff
@@ -28,7 +28,7 @@
     "start:example-supply-chain": "yarn build:dev && cd ./examples/supply-chain-app/ && yarn --no-lockfile && yarn start",
     "start:example-carbon-accounting": "CONFIG_FILE=examples/cactus-example-carbon-accounting-backend/example-config.json node examples/cactus-example-carbon-accounting-backend/dist/lib/main/typescript/carbon-accounting-app-cli.js",
     "purge-build-cache": "del-cli .build-cache/*",
-    "clean": "npm run purge-build-cache && del-cli \"./{packages,examples,extensions}/cactus-*/{dist,.nyc_output,src/main/proto/generated/*,src/main/typescript/generated/proto/protoc-gen-ts/*,src/main/typescript/generated/openapi/typescript-axios/*,src/main-server/kotlin/gen/kotlin-spring/src/**/{model,api}/*}\"",
+    "clean": "npm run purge-build-cache && del-cli \"./{packages,examples,extensions}/cactus-*/{dist,.nyc_output,src/main/proto/generated/*,src/main/typescript/generated/proto/protoc-gen-ts/*,src/main/typescript/generated/openapi/typescript-axios/*,src/main-server/kotlin/gen/kotlin-spring/src/**/{model,api}/*}\" && del-cli \"./dist/*\"",
```
petermetz, line 31: @outSH `&& del-cli \"./dist/*\"` Will this wipe every directory everywhere recursively that's called `dist`? Could it cause problems with third party dependencies within `node_modules/` that also have `dist` directories after installation (that we don't want to delete)?

package.json
outSH: @petermetz Thank you very much for such a detailed explanation :)

> For the prebuild script, you could just use the codegen script instead of introducing a new one called `build:dev:backend:prebuild`

Agree, changed `prebuild` to `codegen`, didn't notice this before. I've left postbuild as is, I don't think this should be called in separation from the whole build, can be changed to something simpler in the future if I'm wrong.

package.json
outSH: @petermetz It's meant to clean only top level dist (i.e. `cactus/dist/*`) where `cmd-socketio-server` put's it's build artifacts. I've tested it once again and it did not remove ./dist content in child directories - checked with `./dist` in `./packages/cactus-plugin-ledger-connector-fabric-socketio` and `./packages/cactus-plugin-ledger-connector-sawtooth-socketio`

petermetz: @outSH Understood, thank you very much for double checking!

package.json
petermetz: @outSH You are welcome, glad it was at least somewhat helpful. 

> Agree, changed prebuild to codegen, didn't notice this before. I've left postbuild as is, I don't think this should be called in separation from the whole build, can be changed to something simpler in the future if I'm wrong.

Sounds good to me.

petermetz:
@outSH One more tip: There's a 're request review' button on GH that you can use to pass the ball back once you made requested changes on a PR (or when you addressed feedback/questions in general)

https://github.blog/changelog/2019-02-21-re-request-review-on-a-pull-request/

outSH:
> @outSH One more tip: There's a 're request review' button on GH that you can use to pass the ball back once you made requested changes on a PR (or when you addressed feedback/questions in general)
> 
> https://github.blog/changelog/2019-02-21-re-request-review-on-a-pull-request/

@petermetz Thanks, I'll use it from now on :) Is it PR ready for your approval?

petermetz (approved):
@outSH Yup, we are good to go as far as I  can tell, thank you!

izuru0 (approved):
LGTM

Issue #1579, opened 2021-11-24
docs(examples): add authN/authZ for supply chain app example 


petermetz:
### Description

We want to be reasonably sure that we haven't forgotten anything basic from the JWT support code in the API server and the best way to find out is to try and make one of the end to end examples work with JWT authorization so that we can put ourselves in the shoves of someone who is developing an application with the framework leveraging the plugins and the API server.
While doing the above, likely some issues will come to light that otherwise would've been hidden until we go to 1.0.0 and people start using it so this way we can get ahead of that.
If any issues arise, please make sure to create new, separate tasks for fixing those that can also be addressed by separate PRs so that we avoid one huge PR that updates the examples and also updates core/API-server/core-api package code at the same time.
By the end of it in the PR resolving this issue we should only have the changes that are touching only the examples folder code because everything else necessary (prerequisites) was done already.

### Acceptance Criteria
1. The supply chain example is locked down by default with authorization, e.g. JWT validation is enabled
2. To keep it user friendly the example tutorials need to end with valid JWTs printed to the command line at the very end so that end users can see/use them with the least amount of hassle possible
3. Do not mix changes between examples and core code (further explained above but wanted to underline it here too)
4. Make sure the documentation is updated accordingly as well
5. Publish the container image to GHCR and also make sure to update the root README.md and the inner supply-chain-app README.md with those images.

elenaizaguirre:
@petermetz In point 5 of Acceptance Criteria you say "_Publish the container image to GHCR_" but I cannot do it before merge changes because I need the npm version, so I think the right order is this:

1. Create the PR for these changes
2. Publish new npm packages (I don´t have permissions)
3. Create and publish the container image (I have permissions for this)
4. Create another PR to update README with newest container image

petermetz:
> @petermetz In point 5 of Acceptance Criteria you say "_Publish the container image to GHCR_" but I cannot do it before merge changes because I need the npm version, so I think the right order is this:
> 
> 1. Create the PR for these changes
> 2. Publish new npm packages (I don´t have permissions)
> 3. Create and publish the container image (I have permissions for this)
> 4. Create another PR to update README with newest container image

@elenaizaguirre Oops, you are right. We can break this issue down into two separate ones to model the reality of the dependent steps.

PR #1676, opened 2021-12-22
fix: add optional auth token to api-client and consortium-manual

elenaizaguirre:
fix: add optional auth token to api-client and consortium-manual

relationed with #1579


petermetz (changes requested):
@elenaizaguirre The consortium database information is meant to be public at least among the consortium participants but tokens are secrets that should not be shared with anyone except who is using it, so we cannot put the tokens on the consortium definition. I'd recommend adding a generic `ctorArgs: Record<string, unkown>` parameter to the ofLedger method instead which can be passed in from wherever the token is actually available (cramming more parameters into the ofLedger function is also not my favorite idea but it's the lesser evil within this context IMO).

packages/cactus-api-client/src/main/typescript/api-client.ts
```diff
@@ -130,9 +132,19 @@ export class ApiClient extends BaseAPI {
 
     const randomNode = nodes[randomIdx];
 
-    const configuration = new Configuration({
-      basePath: randomNode.nodeApiHost,
-    });
+    let configuration: Configuration;
+    if (ctorArgs.authToken) {
```
petermetz, line 136: @elenaizaguirre It would be more flexible and less complicated to pass in the ctor args to the `Configuration` constructor as-is after having done an object merge on it with the defaults. That way you don't need the if condition at all and the caller will have total flexibility over what to pass in to the `Configuration` class' constructor.

The caller will have slightly more complicated code because they'll have to assemble the config object with the authz headers themselves, but that's a good trade-off for flexibility IMO.

packages/cactus-plugin-consortium-manual/src/main/typescript/plugin-consortium-manual.ts
```diff
@@ -217,9 +218,23 @@ export class PluginConsortiumManual
...
     const nodes = this.repo.allNodes;
 
+    const { authToken } = this.options;
+
     const requests = nodes
       .map((cnm) => cnm.nodeApiHost)
-      .map((host) => new Configuration({ basePath: host }))
+      .map(function (host) {
+        // if node has token, add to the configuration
+        if (authToken) {
```
petermetz, line 227: @elenaizaguirre Similar thoughts here as in my other comment. The if condition is not necessary if we allow the caller to specify the ctor args entirely (with the defaults being provided as before)

petermetz (changes requested):

elenaizaguirre:
@petermetz I have implemented changes you requested but CI fails because of an unrelated test (packages/cactus-test-tooling/src/test/typescript/integration/rustc-container/rustc-container-target-nodejs.test.ts) as in other PRs.

petermetz (approved):
@elenaizaguirre Thank you very much for putting up with my change requests.
I pushed a fix for the `packages/cactus-test-tooling/src/test/typescript/integration/rustc-container/rustc-container-target-nodejs.test.ts` test so hopefully now we can get this PR of yours merged soon as well.

izuru0 (approved):
LGTM

Issue #1654, opened 2021-12-14
refactor: use new node protocol imports


petermetz:
Migrate the entire codebase onto the new syntax as explained here:
https://2ality.com/2021/12/node-protocol-imports.html

Not a particularly important thing, but it definitely has benefits to it and therefore we should at least track this in the backlog even if it's low priority.

Also note: We should do the migration on a package per package basis instead of the entire codebase at once in a huge PR because then it will be very difficult to review properly.

**Important**: It must be verified that this does not cause any backward compatibility issues with older versions of NodeJS (if it does, then this needs to be pushed back until the active LTS version supports it by default without any breakage)

---

From the article:

Previously:

    import * as fs from 'fs/promises';

Now:

    import * as fs from 'node:fs/promises';

Leeyoungone:
Hello! May I take this issue up please?

petermetz:
@Leeyoungone Yes, sure thing! First please verify that it is workable on Node 14 as well (because if not then we need to put this on ice until either v14 goes end-of-life or until the community backports this feature to v14 with a feature release.

Leeyoungone:
@petermetz Based on the article that's [linked](https://2ality.com/2021/12/node-protocol-imports.html), I believe that this is workable for `Node 14` and even `Node 12`


<img width="605" alt="Screen Shot 2021-12-20 at 3 06 39 PM" src="https://user-images.githubusercontent.com/85588157/146832880-e544ac79-a630-4069-a7b4-31471b5b360e.png">

petermetz:
@Leeyoungone I think it does answer the question, but the answer it gives is that you'll have to send a separate PR first upgrading the versions in the CI which will be a prerequisite for all other changes ;-) (let's call it PR zero)

Here[1] you see that we are currently using below Node `v14.18.0` so the CommonJS requires can work (that's what Typescript compiles the `import x from y` statements down to in the end)
[1] https://github.com/hyperledger/cactus/blob/main/.github/workflows/ci.yml#L56

> When working per package, were we thinking about one commit per this PR be for one package or opening up 16 different PRs one for each package and linking them all to this issue?

Let's start small and do it on a per package basis first, starting with the smallest/simplest package, e.g. PR one (after PR zero was merged) could be about maybe the common package or the core-api package (whichever is easier, I'm not sure).

---

And here I'd like to doubly call out the most important acceptance criteria again:

[...]

PR #1677, opened 2021-12-23
refactor: upgrade the version of node in CI

Leeyoungone:
This is `PR 0` that is the prerequisite for an upcoming PR that will fix issue #1654. 

What this PR does is upgrade the version in the CI from Node `v14.18.0` to Node `v16.0.0`.


petermetz:
@Leeyoungone Hold on a second, I think I didn't explain it in detail enough, apologies! You want to do upgrades *within* the major versions such as
`14.x.y` => `14.x2.y2` 
and
`16.n.m` => `16.n2.m2` 
so that 
`x2 > x` and `n2 > n`

(in short it's a minor version bump not a major one according to the rules of semantic versioning [1] )

[1] https://semver.org/

Leeyoungone:
@petermetz Thank you so much for the clear explanations, it's my first time hearing and learning about this! I am learning so much from you and I really appreciate it! 🙇

Leeyoungone:
>Important: It must be verified that this does not cause any backward compatibility issues with older versions of NodeJS (if it does, then this needs to be pushed back until the active LTS version supports it by default without any breakage)

@petermetz This doesn't cause any _major_ backwards compatibility issues with older versions of NodeJS but for Node version 16, there are certain _minor_ versions it doesn't work for. When the CI runs for `build-node-16` it is stuck on the `yarn run custom-checks` command. I have replicated this problem by changing the node versions on my machine. Here are some results below

Right now the CI *work* for node versions: 
- `16.11.1`
- `16.11.0`
- `16.10.0`
- `16.1.0`
- `16.0.0`

CI does *not work* for:
- `16.13.1`
- `16.13.0`
- `16.12.0` 

I have't replicated this problem for all the versions of `node v16`(ex: gap between `16.10.0` and `16.1.0`). But I suspect that something occurred within Node's `v16.12.0` minor update that's been messing with the newer, following versions.

I currently have the node (for the CI) running on `16.11.1` as it is the latest and the one that passes the CI. 

Some questions: 

1) Should I start with alternating the packages to follow the new protocol programming format or do you recommend I wait? 
2) I haven't been as thorough with checking the minor changes within `14.x.x` because the CI runs fine with the newest version of node for `node v14` do. you reccomend I do the same as I did for `node v16`?

petermetz:
> @petermetz Thank you so much for the clear explanations, it's my first time hearing and learning about this! I am learning so much from you and I really appreciate it! bow

@Leeyoungone No worries, I'm very glad to hear that. :-)

> Should I start with alternating the packages to follow the new protocol programming format?

No, unfortunately the test results you described above mean that we have to back away from using the node import syntax for now. If we didn't, then we'd end up with a bunch of bugs that would only manifest on very specific node versions (like you described for v16 and I just don't want to take the chance with that so I'm pushing the whole thing back until v16 will get phased out of LTS (which will be a few years from now).

> I haven't been as thorough with checking the minor changes within 14.x.x because the CI runs fine with the newest version of node for node v14 do. you reccomend I do the same as I did for node v16?

No need because of my paragraph above, Node v14 is just a moot point because of the v16 issues.

PR #1686, opened 2021-12-28
test: jestify js-object-signer test

awadhana:

File Path:
packages/cactus-common/src/test/typescript/
unit/js-object-signer.test.ts

### DIFF
--- a/packages/cactus-common/src/test/typescript/unit/js-object-signer.test.ts
+++ b/packages/cactus-common/src/test/typescript/unit/js-object-signer.test.ts
@@ -1,5 +1,4 @@
-import test, { Test } from "tape";
-
+import "jest-extended";
 import {
   JsObjectSigner,
   IJsObjectSignerOptions,
@@ -37,7 +36,7 @@ const verifySignFunction = (
   );
 };
 
-test("Simple JSON Test", async (assert: Test) => {
+test("Simple JSON Test", async () => {
   const jsObjectSignerOptions: IJsObjectSignerOptions = {
     privateKey: keyPairs.privateKey,
     logLevel: "debug",
@@ -50,11 +49,10 @@ test("Simple JSON Test", async (assert: Test) => {
   const payload2 = { field3: 13, field2: "test12", field1: "test11" };
   const sign2 = jsObjectSigner.sign(payload2);
 
-  assert.equals(sign1.toString, sign2.toString);
-  assert.end();
+  expect(sign1.toString).toEqual(sign2.toString);
 });
 
-test("Simple Nested JSON Test", async (assert: Test) => {
+test("Simple Nested JSON Test", async () => {
   const jsObjectSignerOptions: IJsObjectSignerOptions = {
     privateKey: keyPairs.privateKey,
     logLevel: "debug",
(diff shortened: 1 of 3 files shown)


takeutak (approved):

packages/cactus-common/src/test/typescript/unit/js-object-signer.test.ts
```diff
@@ -200,21 +191,17 @@ test("Test optional hash function", async (assert: Test) => {
...
     const pkey: unknown = undefined;
     const jsObjectSignerOptions: IJsObjectSignerOptions = {
       privateKey: pkey as Uint8Array,
     };
     new JsObjectSigner(jsObjectSignerOptions);
-  } catch (e) {
-    if (e instanceof Error) {
-      assert.equal(e.message, "JsObjectSigner#ctor options.privateKey falsy.");
-    }
+  } catch (e: any) {
```
petermetz, line 204: @awadhana Instead of `any` please qualify it as `unkown` and then you'll need the type narrowing if condition back too to avoid compilation errors.
Plus you could also add an else block that fails the test to fix the original issue [1] (which I assume you were fixing by removing the if condition entirely)

[1] => If the exception thrown ends up not being an instance of `Error` then the assertion will not happen at all, that was a problem with the original code, which is not related to the jestification but it's great that you fixed that part of it.

petermetz (changes requested):

packages/cactus-common/src/test/typescript/unit/js-object-signer.test.ts
awadhana: @petermetz, Tony and I tried something different, hopefully the tests pass through :)

packages/cactus-common/src/test/typescript/unit/js-object-signer.test.ts
```diff
@@ -1,5 +1,5 @@
-import test, { Test } from "tape";
-
+import "jest-extended";
+// import { RuntimeError } from "run-time-error";
```
petermetz, line 2: @awadhana Sorry for the nit-picking as usual. Please remove the commented line if it's no longer needed. Looks good otherwise though so we are almost there!

petermetz (changes requested):
@awadhana Almost LGTM, please don't forget to squash the commits in the end, I see that right now there's 2 of them ;-)

packages/cactus-common/src/test/typescript/unit/js-object-signer.test.ts
awadhana: @petermetz not a problem at all :) Clean code is the best code.

packages/cactus-common/src/test/typescript/unit/js-object-signer.test.ts
petermetz: @awadhana Perfect, thank you for making the change!

packages/cactus-common/src/test/typescript/unit/js-object-signer.test.ts
petermetz: @awadhana Yes, now we are cooking!

petermetz (approved):
@awadhana LGTM, thank you!

Issue #1679, opened 2021-12-27
build(socket.io, socket.io-client) Upgrade to common socket.io version


outSH:
Some packages use socket.io and socket.io-client version 2, while other use 4. In order for these components to communicate, we must use common socket.io package version in all cactus.

Depends on #1670

petermetz:
@outSH Thank you very much for opening this!

PR #1696, opened 2021-12-28
build(socket.io, socket.io-client): upgrade to common socket.io version

outSH:
Some packages use socket.io and socket.io-client V2, while other use V4.
In order for these components to communicate, we must use common
socket.io version in all cactus packages (V4.1.3). This commit introduce
some other related changes, like merging socketio unit tests to common
setup (to assert changes are correct) and fixes some strict-flag
warnings and minor bugs. It also updates python packages requirements
and updates readme when it was necessary.

Closes: #1679

Additional changes in **cactus-cmd-socketio-server**:
- Removed local jest config, test will be run from the root dir now.
- Fixed several strict flag warnings from verifier component.
- Removed nohoist pragma, changed package to public and added license info.
- Added monitor_error handle to fix few unit tests (cherry-picked from #1660)
- Used common socketio test helpers (cherry-picked from #1660)

**Depends on #1670**


petermetz:
@outSH Thank you very much for addressing this so quickly, I'm very grateful for the quick turnaround!
Right now the PR has a single commit that includes the full diff of the PR that this one is dependent on, there's two options we can do to make it easy to review:
1. Seperate out the commits so that the diffs can be reviewed separately
2. Or just wait until #1670 gets merged and then after a rebase this will have the correct diff to be reviewed.

I'm fine either way. Regarless of which one you choose, this is looking great at first glance.

packages/cactus-common/src/main/typescript/socketio-test-setup-helpers.ts
```diff
@@ -2,24 +2,23 @@
...
-import ioServer from "socket.io";
-import io from "socket.io-client";
+import { Server, Socket as ServerSocket } from "socket.io";
+import { io, Socket as ClientSocket } from "socket.io-client";
 import { createServer } from "http";
 
-export { ioServer };
+export { Server, ServerSocket, ClientSocket };
 
 /**
```
petermetz, line 11: @outSH Apologies for the nit picking: Could you move this file to the `cactus-test-tooling` package instead of `cactus-common`? 
Reasoning: If it's specifically meant for testing then I'd rather not have it increase the bundle size of the common package. I know that the `Servers` utility class looks similarly out of place, but that one was originally meant to be the utility that the `cmd-api-server` uses when doing graceful HTTP server shutdowns so that's why that one had to be put here in the `cactus-common` package. The fact that 99% of its use-cases for the `Servers` class is actually in the test automation is just a happy coincidence.

petermetz (changes requested):

outSH:
Depends on #1660

packages/cactus-common/src/main/typescript/socketio-test-setup-helpers.ts
outSH: @petermetz Sure, no problem, any tips are welcomed :) I was in fact mimicking `Servers` class and thus put it in the same folder, but I agree with your reasoning. I've moved `socketio-test-setup-helpers` to `cactus-test-tooling` both in this PR, and in #1660 (from which this change originated).

outSH:
@petermetz 
I've changed this PR to draft, after #1660 and #1670 are merged I will double check and do some test to ensure everything is fine, and we'll proceed :) In the future I'll probably just merge dependent commits to clarify the diffs

petermetz:
@outSH Understood, sounds like a plan! :-)

outSH:
Dependent PR were merged, I've done tests after rebasing to main and all seems fine.

@petermetz @izuru0 Please review the changes :)

izuru0 (approved):
LGTM

petermetz (approved):
@outSH LGTM, thank you very much.

Issue #1702, opened 2021-12-29
feat(cactus-common): add createRuntimeErrorWithCause() & newRex()


m-courtin:
**Describe the bug**

In multiple places in the code within a catch block the ` .message` and ` .stack` is directly getting accessed without any checks. As possibly all things can be thrown in TypeScript / JavaScript (even undefined or null etc.) and not only Error objects. Therefore it needs to be validated first if ` .message` respective ` .stack` are existing before they are getting accessed. Furthermore a fallback should be implemented in case the thrown objects are no regular Error objects and not providing the requested properties or the stringify of the ` .stack` for example is failing due to hierarchy depth etc.

petermetz:
@m-courtin FYI: Here is a related issue #1129 and the PR corresponding to said issue that @Leeyoungone has been working on for some time: #1384

Leeyoungone:
@petermetz Thanks for the ping and letting me know. Do you happen to have an example of

> needs to be validated first if ` .message` respective ` .stack` are existing before they are getting accessed.

Do you happen to have any code examples that validate the `.message` and `.stack` before they are getting accessed?

m-courtin:
@petermetz: Thank you very much for the heads-up about the PR @Leeyoungone is working on. So I will only include the basic functionality into my PR that @Leeyoungone can use it in her PR

petermetz:
> Do you happen to have any code examples that validate the `.message` and `.stack` before they are getting accessed?

@Leeyoungone  Yup, this is it:

```typescript
if (ex instanceof Error) {
// ex has .message and .stack
}
```

Leeyoungone:
@petermetz Ah cool cool! thank you :)

PR #1707, opened 2022-01-04
feat(cactus-common): add createRuntimeErrorWithCause() & newRex()

m-courtin:
Utility functions to conveniently re-throw excpetions typed as unknown
by their catch block (which is the default since Typescript v4.4).

Example usage can and much more documentation can be seen here:

packages/cactus-common/src/main/typescript/exception/create-runtime-error-with-cause.ts
and here
packages/cactus-common/src/test/typescript/unit/exception/create-runtime-error-with-cause.test.ts

Closes: #1702

[skip ci]

### DIFF
--- a/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
+++ b/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
@@ -48,6 +48,7 @@ import {
   Bools,
   Logger,
   LoggerProvider,
+  newRex,
   Servers,
 } from "@hyperledger/cactus-common";
 
@@ -248,17 +249,17 @@ export class ApiServer {
       }
 
       return { addressInfoCockpit, addressInfoApi, addressInfoGrpc };
-    } catch (ex) {
-      const errorMessage = `Failed to start ApiServer: ${ex.stack}`;
-      this.log.error(errorMessage);
+    } catch (ex1: unknown) {
+      const context = "Failed to start ApiServer";
+      this.log.error(context, ex1);
       this.log.error(`Attempting shutdown...`);
       try {
         await this.shutdown();
         this.log.info(`Server shut down after crash OK`);
-      } catch (ex) {
-        this.log.error(ApiServer.E_POST_CRASH_SHUTDOWN, ex);
+      } catch (ex2: unknown) {
+        this.log.error(ApiServer.E_POST_CRASH_SHUTDOWN, ex2);
       }
-      throw new Error(errorMessage);
+      throw newRex(context, ex1);
     }
   }
 
@@ -304,11 +305,11 @@ export class ApiServer {
         await this.getPluginImportsCount(),
       );
       return this.pluginRegistry;
-    } catch (e) {
+    } catch (ex: unknown) {
       this.pluginRegistry = new PluginRegistry({ plugins: [] });
-      const errorMessage = `Failed init PluginRegistry: ${e.stack}`;
-      this.log.error(errorMessage);
-      throw new Error(errorMessage);
+      const context = "Failed to init PluginRegistry";
+      this.log.debug(context, ex);
+      throw newRex(context, ex);
     }
   }
 
@@ -368,15 +369,10 @@ export class ApiServer {
       await plugin.onPluginInit();
 
       return plugin;
-    } catch (error) {
-      const errorMessage = `${fnTag} failed instantiating plugin '${packageName}' with the instanceId '${options.instanceId}'`;
-      this.log.error(errorMessage, error);
-
-      if (error instanceof Error) {
-        throw new RuntimeError(errorMessage, error);
-      } else {
-        throw new RuntimeError(errorMessage, JSON.stringify(error));
-      }
+    } catch (ex: unknown) {
+      const context = `${fnTag} failed instantiating plugin '${packageName}' with the instanceId '${options.instanceId}'`;
+      this.log.debug(context, ex);
+      throw newRex(context, ex);
     }
   }
 
@@ -397,10 +393,10 @@ export class ApiServer {
     try {
       await fs.mkdirp(pluginPackageDir);
       this.log.debug(`${pkgName} plugin package dir: %o`, pluginPackageDir);
-    } catch (ex) {
-      const errorMessage =
+    } catch (ex: unknown) {
+      const context =
         "Could not create plugin installation directory, check the file-system permissions.";
-      throw new RuntimeError(errorMessage, ex);
+      throw newRex(context, ex);
     }
     try {
       lmify.setPackageManager("npm");
@@ -418,19 +414,15 @@ export class ApiServer {
         // "--ignore-workspace-root-check",
       ]);
       this.log.debug("%o install result: %o", pkgName, out);
-      if (out.exitCode !== 0) {
-        throw new RuntimeError("Non-zero exit code: ", JSON.stringify(out));
+      if (out?.exitCode && out.exitCode !== 0) {
+        const eMsg = "Non-zero exit code returned by lmify.install() indicating that the underlying npm install OS process had encountered a problem:";
+        throw newRex(eMsg, out);
       }
       this.log.info(`Installed ${pkgName} OK`);
-    } catch (ex) {
-      const errorMessage = `${fnTag} failed installing plugin '${pkgName}`;
-      this.log.error(errorMessage, ex);
-
-      if (ex instanceof Error) {
-        throw new RuntimeError(errorMessage, ex);
-      } else {
-        throw new RuntimeError(errorMessage, JSON.stringify(ex));
-      }
+    } catch (ex: unknown) {
+      const context = `${fnTag} failed installing plugin '${pkgName}`;
+      this.log.debug(ex, context);
+      throw newRex(context, ex);
     }
   }
 
@@ -451,24 +443,25 @@ export class ApiServer {
     this.log.info(`Stopped ${webServicesShutdown.length} WS plugin(s) OK`);
 
     if (this.httpServerApi?.listening) {
-      this.log.info(`Closing HTTP server of the API...`);
+      this.log.info(`Closing Cacti HTTP server of the API...`);
       await Servers.shutdown(this.httpServerApi);
       this.log.info(`Close HTTP server of the API OK`);
     }
 
     if (this.httpServerCockpit?.listening) {
-      this.log.info(`Closing HTTP server of the cockpit ...`);
+      this.log.info(`Closing Cacti HTTP server of the cockpit ...`);
       await Servers.shutdown(this.httpServerCockpit);
       this.log.info(`Close HTTP server of the cockpit OK`);
     }
 
     if (this.grpcServer) {
-      this.log.info(`Closing gRPC server ...`);
+      this.log.info(`Closing Cacti gRPC server ...`);
       await new Promise<void>((resolve, reject) => {
         this.grpcServer.tryShutdown((ex?: Error) => {
           if (ex) {
-            this.log.error("Failed to shut down gRPC server: ", ex);
-            reject(ex);
+            const eMsg = "Failed to shut down gRPC server of the Cacti API server.";
+            this.log.debug(eMsg, ex);
+            reject(newRex(eMsg, ex));
           } else {
             resolve();
           }
--- a/packages/cactus-common/package.json
+++ b/packages/cactus-common/package.json
@@ -33,6 +33,11 @@
       "name": "Peter Somogyvari",
       "email": "peter.somogyvari@accenture.com",
       "url": "https://accenture.com"
+    },
+    {
+      "name": "Michael Courtin",
+      "email": "michael.courtin@accenture.com",
+      "url": "https://accenture.com"
     }
   ],
   "main": "dist/lib/main/typescript/index.js",
--- a/packages/cactus-common/src/main/typescript/exception/coerce-unknown-to-error.ts
+++ b/packages/cactus-common/src/main/typescript/exception/coerce-unknown-to-error.ts
@@ -1,0 +1,46 @@
+import stringify from "fast-safe-stringify";
+import { ErrorFromUnknownThrowable } from "./error-from-unknown-throwable";
+import { ErrorFromSymbol } from "./error-from-symbol";
+
+/**
+ * Safely converts `unknown` to an `Error` with doing a best effort to ensure
+ * that root cause analysis information is not lost. The idea here is to help
+ * people who are reading logs of errors while trying to figure out what went
+ * wrong after a crash.
+ *
+ * Often in Javascript this is much harder than it could be due to lack of
+ * runtime checks by the JSVM (Javascript Virtual Machine) on the values/objects
+ * that are being thrown.
+ *
+ * @param x The value/object whose type information is completely unknown at
+ * compile time, such as the input parameter of a catch block (which could
+ * be anything because the JS runtime has no enforcement on it at all, e.g.
+ * you can throw null, undefined, empty strings of whatever else you'd like.)
+ * @returns An `Error` object that is the original `x` if it was an `Error`
+ * instance to begin with or a stringified JSON representation of `x` otherwise.
+ */
+export function coerceUnknownToError(x: unknown): Error {
+  if (typeof x === "symbol") {
+    const symbolAsStr = x.toString();
+    return new ErrorFromSymbol(symbolAsStr);
+  } else if (x instanceof Error) {
+    return x;
+  } else {
+    const xAsJson = stringify(x, (_, value) =>
+      typeof value === "bigint" ? value.toString() + "n" : value,
+    );
+    return new ErrorFromUnknownThrowable(xAsJson);
+  }
+}
+
+/**
+ * This is an alias to `coerceUnknownToError(x: unknown)`.
+ *
+ * The shorter name allows for different style choices to be made by the person
+ * writing the error handling code.
+ *
+ * @see #coerceUnknownToError
+ */
+export function asError(x: unknown): Error {
+  return coerceUnknownToError(x);
+}
--- a/packages/cactus-common/src/main/typescript/exception/create-runtime-error-with-cause.ts
+++ b/packages/cactus-common/src/main/typescript/exception/create-runtime-error-with-cause.ts
@@ -1,0 +1,104 @@
+import { RuntimeError } from "run-time-error";
+import { coerceUnknownToError } from "./coerce-unknown-to-error";
+
+/**
+ * ### STANDARD EXCEPTION HANDLING - EXAMPLE WITH RE-THROW:
+ *
+ * Use the this utility function and pass in any throwable of whatever type and format
+ * The underlying implementation will take care of determining if it's a valid
+ * `Error` instance or not and act accordingly with avoding information loss
+ * being the number one priority.
+ *
+ * You can perform a fast-fail re-throw with additional context like the snippet
+ * below.
+ * Notice that we log on the debug level inside the catch block to make sure that
+ * if somebody higher up in the callstack ends up handling this exception then
+ * it will never get logged on the error level which is good because if it did
+ * that would be a false-positive, annoying system administrators who have to
+ * figure out which errors in their production logs need to be ignored and which
+ * ones are legitimate.
+ * The trade-off with the above is trust: Specifically, we are trusting the
+ * person above us in the callstack to either correctly handle the exception
+ * or make sure that it does get logged on the error level. If they fail to do
+ * either one of those, then we'll have silent failures on our hand that will
+ * be hard to debug.
+ * Lack of the above kind of trust is usually what pushes people to just go for
+ * it and log their caught exceptions on the error level but this most likely
+ * a mistake in library code where there just isn't enough context to know if
+ * an error is legitimate or not most of the time. If you are writing application
+ * logic then it's usually a simpler decision with more information at your
+ * disposal.
+ * 
+ * The underlying concept is that if you log something on an error level, you
+ * indicate that another human should fix a bug that is in the code. E.g., 
+ * when they see the error logs, they should go and fix something.
+ *
+ * ```typescript
+ * public doSomething(): void {
+ *   try {
+ *     someSubTaskToExecute();
+ *   } catch (ex) {
+ *     const eMsg = "Failed to run **someSubTask** while doing **something**:"
+ *     this.log.debug(eMsg, ex);
+ *     throw createRuntimeErrorWithCause(eMsg, ex);
+ * }
+ * ```
+ *
+ * ### EXCEPTION HANDLING WITH CONDITIONAL HANDLING AND RE-THROW - EXAMPLE:
+ *
+ * In case you need to do a conditional exception-handling:
+ *  - Use the RuntimeError to re-throw and
+ * provide the previous exception as cause in the new RuntimeError to retain 
+ * the information and distinguish between an exception you can handle and
+ * recover from and one you can't
+ *
+ * ```typescript
+ * public async doSomething(): Promise<number> {
+ *    try {
+ *      await doSubTaskThatsAPartOfDoingSomething();
+ *    } catch (ex) {
+ *      if (ex instanceof MyErrorThatICanHandleAndRecoverFrom) {
+ *        // An exception with a fixable scenario we can recover from thru an additional handling
+ *        // do something here to handle and fix the issue
+ *        // where "fixing" means that the we end up recovering
+ *        // OK instead of having to crash. Recovery means that
+ *        // we are confident that the second sub-task is safe to proceed with
+ *        // despite of the error that was caught here
+ *        this.log.debug("We've got an failure in 'doSubTaskThatsAPartOfDoingSomething()' but we could fix it and recover to continue".);
+ *      } else {
+ *        // An "unexpected exception" where we want to fail immediately
+ *        // to avoid follow-up problems
+ *        const context = "We got an severe failure in 'doSubTaskThatsAPartOfDoingSomething()' and need to stop directly here to avoid follow-up problems";
+ *        this.log.erorr(context, ex);
+ *        throw newRex(context, ex);
+ *      }
+ *    }
+ *    const result = await doSecondAndFinalSubTask();
+ *    return result; // 42
+ *  }
+ * ```
+ *
+ * @param message The contextual information that will be passed into the 
+ * constructor of the returned {@link RuntimeError} instance.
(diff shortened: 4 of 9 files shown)


izuru0 (approved):
LGTM

m-courtin:
@izuru0: Thank you very much for reviewing this PR.

m-courtin:
As an example on how to integrate / use the new generic functions the cactus-cmd-api-server was taken as a show case

petermetz (changes requested):
@m-courtin Can you make it support these (and similar) cases of incorrect error throwing as well? Right now these would all result in the loss of the information that the thrower was trying to convey (which would be their fault for not throwing an exception, but since the language allows for it we have to assume it will happen...)

```typescript
  try {
    throw "Something very important about what went wrong";
  } catch (ex) {
    const theStack = LogHelper.getExceptionStack(ex);
    const theMessage = LogHelper.getExceptionMessage(ex);
    console.log(theStack);
    console.log(theMessage);
  }

  try {
    throw { error: "Something very important about what went wrong" };
  } catch (ex) {
    const theStack = LogHelper.getExceptionStack(ex);
    const theMessage = LogHelper.getExceptionMessage(ex);
    console.log(theStack);
    console.log(theMessage);
  }

```

The other thing I'd ask is to check for the input being a specific instance of `RuntimeError` so that this also does not lose information:

```typescript
  const exA = new RuntimeError("A");
  const exB = new RuntimeError("B", exA);
  const exC = new RuntimeError("C", exB);

  // The nested exceptions here are swallowed, the only stack that is shown
  // is the last one (for "C")
  const x = LogHelper.getExceptionStack(exC);
  console.log(x);
```

JSON stringifying the `RuntimeError` above will preserve the full chain[1] of nested exceptions, but with the current code it only preserves the last one.

[1]:

```json
{
    "name": "RuntimeError",
    "stack": "RuntimeError: C\n    at /..../cactus-1/packages/cactus-common/src/test/typescript/unit/logging/log-helper.test.ts:9:15\n    at _dispatchDescribe (/..../cactus-1/node_modules/jest-circus/build/index.js:97:26)\n    at describe (/..../cactus-1/node_modules/jest-circus/build/index.js:60:5)\n    at Object.<anonymous> (/..../cactus-1/packages/cactus-common/src/test/typescript/unit/logging/log-helper.test.ts:6:1)\n    at Runtime._execModule (/..../cactus-1/node_modules/jest-runtime/build/index.js:1646:24)\n    at Runtime._loadModule (/..../cactus-1/node_modules/jest-runtime/build/index.js:1185:12)\n    at Runtime.requireModule (/..../cactus-1/node_modules/jest-runtime/build/index.js:1009:12)\n    at jestAdapter (/..../cactus-1/node_modules/jest-circus/build/legacy-code-todo-rewrite/jestAdapter.js:79:13)\n    at runTestInternal (/..../cactus-1/node_modules/jest-runner/build/runTest.js:389:16)\n    at runTest (/..../cactus-1/node_modules/jest-runner/build/runTest.js:475:34)",
    "message": "C",
    "cause": {
        "name": "RuntimeError",
        "stack": "RuntimeError: B\n    at /..../cactus-1/packages/cactus-common/src/test/typescript/unit/logging/log-helper.test.ts:8:15\n    at _dispatchDescribe (/..../cactus-1/node_modules/jest-circus/build/index.js:97:26)\n    at describe (/..../cactus-1/node_modules/jest-circus/build/index.js:60:5)\n    at Object.<anonymous> (/..../cactus-1/packages/cactus-common/src/test/typescript/unit/logging/log-helper.test.ts:6:1)\n    at Runtime._execModule (/..../cactus-1/node_modules/jest-runtime/build/index.js:1646:24)\n    at Runtime._loadModule (/..../cactus-1/node_modules/jest-runtime/build/index.js:1185:12)\n    at Runtime.requireModule (/..../cactus-1/node_modules/jest-runtime/build/index.js:1009:12)\n    at jestAdapter (/..../cactus-1/node_modules/jest-circus/build/legacy-code-todo-rewrite/jestAdapter.js:79:13)\n    at runTestInternal (/..../cactus-1/node_modules/jest-runner/build/runTest.js:389:16)\n    at runTest (/..../cactus-1/node_modules/jest-runner/build/runTest.js:475:34)",
        "message": "B",
        "cause": {
            "name": "RuntimeError",
            "stack": "RuntimeError: A\n    at /..../cactus-1/packages/cactus-common/src/test/typescript/unit/logging/log-helper.test.ts:7:15\n    at _dispatchDescribe (/..../cactus-1/node_modules/jest-circus/build/index.js:97:26)\n    at describe (/..../cactus-1/node_modules/jest-circus/build/index.js:60:5)\n    at Object.<anonymous> (/..../cactus-1/packages/cactus-common/src/test/typescript/unit/logging/log-helper.test.ts:6:1)\n    at Runtime._execModule (/..../cactus-1/node_modules/jest-runtime/build/index.js:1646:24)\n    at Runtime._loadModule (/..../cactus-1/node_modules/jest-runtime/build/index.js:1185:12)\n    at Runtime.requireModule (/..../cactus-1/node_modules/jest-runtime/build/index.js:1009:12)\n    at jestAdapter (/..../cactus-1/node_modules/jest-circus/build/legacy-code-todo-rewrite/jestAdapter.js:79:13)\n    at runTestInternal (/..../cactus-1/node_modules/jest-runner/build/runTest.js:389:16)\n    at runTest (/..../cactus-1/node_modules/jest-runner/build/runTest.js:475:34)",
            "message": "A",
            "cause": null
        }
    }
}
```

m-courtin:
@petermetz: Yes, good point to support similar "false" exceptions as well and try to get their message / load. I will extend the generic handling accordingly

petermetz:
@m-courtin Also how about something like this for the stack extraction? I came to realize that the `stack` property contains the `message` property for valid `Error` objects anyway.

```typescript
export const K_FALLBACK_STACK = "NO_STACK_INFORMATION_INCLUDED_IN_EXCEPTION";

export class LogHelper {
  public static getExceptionStack(exception: unknown): string {
    if (exception instanceof RuntimeError) {
      return stringify(exception);
    } else if (exception instanceof Error) {
      return stringify(exception.stack);
    } else if (exception) {
      return stringify(exception);
    } else {
      return K_FALLBACK_STACK;
    }
  }

```

packages/cactus-common/src/main/typescript/logging/log-helper.ts
```diff
@@ -0,0 +1,100 @@
+import { RuntimeError } from "run-time-error";
+
+export class LogHelper {
+  public static getExceptionStack(exception: unknown): string {
```
petermetz, line 4: @m-courtin Could you please put a few code examples in the JSDoc of the methods and the class? Examples that cover how to: 
1) catch + log the error + re-throw with more context (with the usage of `RuntimeError`) 
2) catch + log the error + re-throw as-is
3) catch + conditionally handle it OR re-throw with more context (with the usage of `RuntimeError`)

m-courtin:
@petermetz: Regarding the stack handling I would prefer to use the `typeof exception === "object"` part in combination with the  call for hasOwnProperty stack over the `instance of error` to also include custom build errors / exceptions which are having a stack property not based on Error class. The other case of stringify I will include to also cover other stringify-able exceptions and not loosing their stack info

packages/cactus-common/src/main/typescript/logging/log-helper.ts
```diff
@@ -0,0 +1,100 @@
...
+          message =
+            customExceptionPrefix &&
+            this.safeJsonStringify(exception, invalidException);
+        }
+      }
+    }
+    return message;
+  }
+
+  private static safeJsonStringify(
```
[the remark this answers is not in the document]
m-courtin: @petermetz: You mentioned some time ago that the cactus-common module is a critical one as it is getting used nearly everywhere. Therefore I was a bit conservative and not including any additional external libraries into this. But for me definitely fine to integrate the "fast-safe-stringify" library.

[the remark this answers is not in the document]
petermetz: > @petermetz: You mentioned some time ago that the cactus-common module is a critical one as it is getting used nearly everywhere. Therefore I was a bit conservative and not including any additional external libraries into this. But for me definitely fine to integrate the "fast-safe-stringify" library.

@m-courtin Yeah, I also did not make that recommendation easily for the same reason. I did take a look at the package size before and while the 40 KBs it takes up in space is not ideal (I feel like it could be done in a few KBs) it's also not that bad[1] for something as crucial as this (I have nightmares about debugging production issues with log messages of `[Object object]` in them or with crashes within the catch blocks, etc.)

[1]: Unfortunately most packages nowadays are well above a megabyte even when their code itself seems very small.

petermetz:
> @petermetz: Regarding the stack handling I would prefer to use the `typeof exception === "object"` part in combination with the call for hasOwnProperty stack over the `instance of error` to also include custom build errors / exceptions which are having a stack property not based on Error class. The other case of stringify I will include to also cover other stringify-able exceptions and not loosing their stack info

@m-courtin I can live with that as long as the arrow programming gets eliminated from it. Right now the nesting depth is 4 if conditions and most of them do not have else blocks so that's why I have the urge to reduce/flatten the depth to 1 instead. Even if it ends up being slightly more repetitive, less if conditions and without nesting will almost always make it easier to read/debug IMO.

m-courtin:
> > @petermetz: Regarding the stack handling I would prefer to use the `typeof exception === "object"` part in combination with the call for hasOwnProperty stack over the `instance of error` to also include custom build errors / exceptions which are having a stack property not based on Error class. The other case of stringify I will include to also cover other stringify-able exceptions and not loosing their stack info
> 
> @m-courtin I can live with that as long as the arrow programming gets eliminated from it. Right now the nesting depth is 4 if conditions and most of them do not have else blocks so that's why I have the urge to reduce/flatten the depth to 1 instead. Even if it ends up being slightly more repetitive, less if conditions and without nesting will almost always make it easier to read/debug IMO.

@petermetz: Yes, reading code like flow-chart diagrams with the different nodes and branches might not be the easiest. So I changed the code style to create an easier readable code.

packages/cactus-common/src/main/typescript/logging/logger.ts
```diff
@@ -21,6 +22,61 @@ export interface ILoggerOptions {
   level?: LogLevelDesc;
 }
 
+/**
+ * STANDARD EXCEPTION HANDLING - EXAMPLE:
+ * ====================================
+ * public doSomething(): void {
```
petermetz, line 28: @m-courtin The re-throw here is missing. I want to encourage the reader to abide by the fail-fast principle instead of "log it and forget it".
https://en.wikipedia.org/wiki/Fail-fast

packages/cactus-common/src/main/typescript/logging/logger.ts
```diff
@@ -21,6 +22,61 @@ export interface ILoggerOptions {
...
+ *
+ *
+ *
+ * EXCEPTION HANDLING WITH CONTEXT - EXAMPLE:
+ * ==========================================
+ * public doSomething(): void {
+ *   try {
+ *     executeSomething();
+ *   } catch (ex) {
+ *     this.log.exception(ex, "Some additional information / context about the scenario");
```
petermetz, line 43: @m-courtin The re-throw here is missing, similar thoughts as above.

packages/cactus-common/src/main/typescript/logging/logger.ts
```diff
@@ -21,6 +22,61 @@ export interface ILoggerOptions {
...
+ *       executeSomething();
+ *     } catch (ex) {
+ *       this.log.exception(
+ *          ex,
+ *         "Some additional information / context about the scenario",
+ *        );
+ *
+ *       // maybe need to do a shut-down of some component here before re-throwing
+ *
+ *        throw ExceptionHelper.getNewRuntimeErrorFromException(ex, "ExecuteSomething failed");
```
petermetz, line 66: @m-courtin There is a re-throw here, but for the third one I wanted to have a conditional handling + re-throw on the else block like

```typescript
class X {
  public async doSomething(): Promise<number> {
    try {
      await doSubTaskThatsAPartOfDoingSomething();
    } catch (ex) {
      if (ex instanceof MyErrorThatICanHandleAndRecoverFrom) {
        // An "expected exception"
        // do something here to handle the error
        // where "handling" means that the we end up recovering
        // OK instead of having to crash. Recovery means that
        // we are confident that the second sub-task is safe to proceed with
        // despite of the error that was caught here
      } else {
        // An "unexpected exception" where we want to fail immediately
        // to avoid cascading problems
        throw new RuntimeError(
          "Doing $SOMETHING failed with: ",
          ExceptionHelper.getExceptionStack(ex),
        );
      }
    }
    const result = await doSecondAndFinalSubTask();
    return result; // 42
  }
}

```

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
```diff
@@ -0,0 +1,327 @@
...
+    );
+
+    // log to destination
+    logger.error(...exceptionContent);
+  }
+
+  /**
+   * Creates a new RuntimeError based on the incoming exception and context to preserve the complete hierarchy of exception information
+   *
+   * @param exception as an arbitrary exception of whatever type and format it might be
```
petermetz, line 30: @m-courtin Typos

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
```diff
@@ -0,0 +1,327 @@
...
+   * @param context providing context about the exception scenario
+   * @returns a new RuntimeError created from exception to preserve the complete hierarchy of exception information
+   */
+  public static getNewRuntimeErrorFromException(
+    exception: unknown,
+    context: string,
+  ): RuntimeError {
+    const exceptionMessageInfo = this.getExceptionMessageInfo(exception);
```
petermetz, line 38: @m-courtin Could this be shortened so that it checks for `exception` being an `instanceof Error` and if yes, use it as the cause, otherwise coerce it to string with the `getExceptionStack` method? (that way it's just a single if condition but with equivalent output AFAICT)

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  /**
+   * Creates a new RuntimeError based on the incoming exception and context to preserve the complete hierarchy of exception information
+   *
+   * @param exception as an arbitrary exception of whatever type and format it might be
+   * @param context providing context about the exception scenario
+   * @returns a new RuntimeError created from exception to preserve the complete hierarchy of exception information
+   */
+  public static getNewRuntimeErrorFromException(
```
petermetz, line 34: @m-courtin I recommend removing the `get` from the name because it instantiates/creates a new object instead of retrieving an existing one. Also recommend dropping `FromException` because we accept the `exception` input as anything (`unknown`)

petermetz (changes requested):

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
petermetz: @m-courtin Oops, typos were my own, nvm.

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
```diff
@@ -0,0 +1,327 @@
...
+    exception: unknown,
+    context: string,
+  ): RuntimeError {
+    const exceptionMessageInfo = this.getExceptionMessageInfo(exception);
+    const exceptionStackInfo = this.getExceptionStackInfo(exception);
+    let content = "";
+    let runtimeError = new RuntimeError(context, exceptionMessageInfo.message);
+
+    if (exception instanceof RuntimeError || exception instanceof Error) {
```
petermetz, line 43: @m-courtin No need to check for the sub-class separately in Javascript. Sub-classes are `instanceof` of their parent classes as well.

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
m-courtin: @petermetz: the `getExceptionMessageInfo` is handling 4 different scenarios with 4 different outcomes:

- Scenario 1+2: Handling any exception / error / RuntimeError containing a message property
      - Scenario 1: message property is of string type (so no stringify is needed) and is directly taken as returned message
      - Scenario 2: message property is not of type string and therefore needs to be stringified

- Scenario 3: A plain string only was thrown and therefore the returned message gets a PREFIX attached mentioning that a custom exception was thrown (therefore not all expected properties like stack etc might be available)

- Scenario 4: It is an arbitrary exception not containing a message property. So the whole exception with all the included properties will be stringified and a PREFIX attached mentioning that a custom exception was thrown (therefore not all expected properties like stack etc might be available)

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
m-courtin: @petermetz: I'm fine with the renaming

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
m-courtin: @petermetz: Removed the sub-class check

packages/cactus-common/src/main/typescript/logging/logger.ts
m-courtin: @petermetz: Totally agree on the fail-fast principle. Added the re-throw in the documentation of the "Standard Exception Handling" as well as the "Exception Handling with Context"

packages/cactus-common/src/main/typescript/logging/logger.ts
m-courtin: @petermetz: Totally agree on the fail-fast principle. Added the re-throw in the documentation of the "Standard Exception Handling" as well as the "Exception Handling with Context"

packages/cactus-common/src/main/typescript/logging/logger.ts
m-courtin: @petermetz: Re-structured the documentation on the cascaded / conditional handling according to your suggestion

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
m-courtin: @petermetz: 👍

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
petermetz: @m-courtin Sorry I meant the entire `newRuntimeError` function's code, not the code of the `getExceptionMessageInfo` function which is being invoked on the line I'm commenting on. Apologies for the confusion!

petermetz: @m-courtin Sorry I meant the entire `newRuntimeError` function's code, not the code of the `getExceptionMessageInfo` function which is being invoked on the line I'm commenting on. Apologies for the confusion!

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
petermetz: @m-courtin Cheers!

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
petermetz: @m-courtin Thank you!

petermetz (changes requested):
@m-courtin Sorry I meant the entire newRuntimeError function's code, not the code of the getExceptionMessageInfo function which is being invoked on the line I'm commenting on. Apologies for the confusion!

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
m-courtin: @petermetz: The `newRuntimeError` function is containing a distinction in the else code to prevent duplicated parts of information in the returned message. I added some lines of further detail documentation to the function to describe it further.

petermetz: > prevent duplicated parts of information in the returned message

@m-courtin Right now `ExceptionHelper.newRuntimeError("Oops", "Doh").cause` evaluates to `"A CUSTOM EXCEPTION WAS THROWN: Oops - A CUSTOM EXCEPTION WAS THROWN: \"Oops\""` anyway (I know it's the cause not the message) which I don't have a problem with (and the same goes for `message` not just `cause`), but I'd much rather us tighten the budget on the number of `if` conditions and only use them when information could be lost unless we check for something.

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
```diff
@@ -0,0 +1,330 @@
...
+   * @returns a new RuntimeError created from exception to preserve the complete hierarchy of exception information
+   */
+  public static newRuntimeError(
+    exception: unknown,
+    context: string,
+  ): RuntimeError {
+    const exceptionMessageInfo = this.getExceptionMessageInfo(exception);
+    const exceptionStackInfo = this.getExceptionStackInfo(exception);
+    let content = "";
+    let runtimeError = new RuntimeError(context, exceptionMessageInfo.message);
```
petermetz, line 41: @m-courtin AFAICT here you always create a `RuntimeError` that then will ***always*** get overwritten by subsequent assignments down the road in the if conditions below, so I recommend removing the extra work that's being done on this line.

petermetz (changes requested):

packages/cactus-common/src/main/typescript/logging/logger.ts
```diff
@@ -46,6 +118,24 @@ export class Logger {
     this.backend.info("Shut down logger OK.");
   }
 
+  /**
```
petermetz, line 121: @m-courtin Please clarify in the docs the value add of this method compared to `logger.error(ex);` (e.g. one is human readable [1] the other is machine readable)

I'd also make sure to let people know that it's a wrapper around `logger.error` so that they don't have to wonder what level it will get logged at or if it gets logged at all since `exception` (as-in the method name itself) is a noun not a verb. 

[1]

```sh
console.error
    [2022-02-07T23:02:51.905Z] ERROR (TestX): r [RuntimeError]: Doh
        at Function.newRuntimeError (/home/peter/a/blockchain/cactus/packages/cactus-common/src/main/typescript/exception/exception-helper.ts:45:22)
        at /home/peter/a/blockchain/cactus/packages/cactus-common/src/test/typescript/unit/exception/exception-helper.test.ts:340:32
        at _dispatchDescribe (/home/peter/a/blockchain/cactus/node_modules/jest-circus/build/index.js:97:26)
        at describe (/home/peter/a/blockchain/cactus/node_modules/jest-circus/build/index.js:60:5)
        at /home/peter/a/blockchain/cactus/packages/cactus-common/src/test/typescript/unit/exception/exception-helper.test.ts:336:3
        at _dispatchDescribe (/home/peter/a/blockchain/cactus/node_modules/jest-circus/build/index.js:97:26)
        at describe (/home/peter/a/blockchain/cactus/node_modules/jest-circus/build/index.js:60:5)
        at Object.<anonymous> (/home/peter/a/blockchain/cactus/packages/cactus-common/src/test/typescript/unit/exception/exception-helper.test.ts:7:1)
        at Runtime._execModule (/home/peter/a/blockchain/cactus/node_modules/jest-runtime/build/index.js:1646:24)
        at Runtime._loadModule (/home/peter/a/blockchain/cactus/node_modules/jest-runtime/build/index.js:1185:12) {
      cause: r [RuntimeError]: Oops
          at /home/peter/a/blockchain/cactus/packages/cactus-common/src/test/typescript/unit/exception/exception-helper.test.ts:341:7
          at _dispatchDescribe (/home/peter/a/blockchain/cactus/node_modules/jest-circus/build/index.js:97:26)
          at describe (/home/peter/a/blockchain/cactus/node_modules/jest-circus/build/index.js:60:5)
          at /home/peter/a/blockchain/cactus/packages/cactus-common/src/test/typescript/unit/exception/exception-helper.test.ts:336:3
          at _dispatchDescribe (/home/peter/a/blockchain/cactus/node_modules/jest-circus/build/index.js:97:26)
          at describe (/home/peter/a/blockchain/cactus/node_modules/jest-circus/build/index.js:60:5)
          at Object.<anonymous> (/home/peter/a/blockchain/cactus/packages/cactus-common/src/test/typescript/unit/exception/exception-helper.test.ts:7:1)
          at Runtime._execModule (/home/peter/a/blockchain/cactus/node_modules/jest-runtime/build/index.js:1646:24)
          at Runtime._loadModule (/home/peter/a/blockchain/cactus/node_modules/jest-runtime/build/index.js:1185:12)
          at Runtime.requireModule (/home/peter/a/blockchain/cactus/node_modules/jest-runtime/build/index.js:1009:12) {
        cause: 'Snap!'
      }
    }
```

petermetz (changes requested):

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
m-courtin: @petermetz: Now I got your saying: The `ExceptionHelper.newRuntimeError("Oops", "Doh").cause` was evaluating to `"A CUSTOM EXCEPTION WAS THROWN: Oops - A CUSTOM EXCEPTION WAS THROWN: \"Oops\""` which is exactly what should be prevented -> So the distinction in the else-case made no sense in this scenario you tested yesterday. YES !!

So I did a re-check on your test-scenario with the "Oops", "Doh" and found the root cause why it still showed the duplication: In the function `getExceptionMessageInfo()` in. scenario 3 the thrown exception as in your example is a string only type and was missing `containsCompleteStringifiedException = true` as this scenario handles exception of string only type and there is no other property so it is per se completely stringified. With this fix the distinction in the else case is now working as expected and no longer leading to duplication.

packages/cactus-common/src/main/typescript/exception/exception-helper.ts
m-courtin: For the first topic please see comment above.

Regarding the created RuntimeError the initialisation is now done with an undefined as it will be overwritten as you already mentioned

packages/cactus-common/src/main/typescript/logging/logger.ts
m-courtin: @petermetz: I added some documentation to describe that the exception function is retrieving exception content in safe and secure way whatever exception is thrown and that there is no further need to check on stack trace, dealing with different exception types etc in code.

petermetz (approved):

PR #1708, opened 2022-01-06
test(test-plugin-htlc-eth-besu): fix neg test case

### LINKED ISSUE #1697: test(plugin-htlc-eth-besu): fix assert mixup in refund-endpoint-invalid
In the bottom of the file:
packages/cactus-test-plugin-htlc-eth-besu/src/test/typescript/integration/plugin-htlc-eth-besu/refund-endpoint-invalid.test.ts
There is an assertion of a negative test case that verifies that incorrect
input correctly leads to a failure of the request.
It could be done by using a simple `rejects` assertion of Jest, but instead
a try catch block is used and there's some dead code in the form of some assertion
in there as well because the operation is never meant to succeed.

The problem is not just that it's dead code but also that it's making the test case
not work at all because if the operation ends up succeeding we are passing the test
anyway with the code as it is today.

```typescript
    try {
      const refundRequest: RefundReq = {
        id,
        web3SigningCredential,
        connectorId,
        keychainId,
      };
      const refundResponse = await api.refundV1(refundRequest);
      expect(refundResponse.status).toEqual(200);
    } catch (error: any) {
      expect(error.response.status).toEqual(500);
    }
```

### LINKED ISSUE #1698: test(test-plugin-htlc-eth-besu): fix asserts new-contract-endpoint-invalid
In the bottom of the file:
packages/cactus-test-plugin-htlc-eth-besu/src/test/typescript/integration/plugin-htlc-eth-besu/new-contract-endpoint-invalid.test.ts
There is an assertion of a negative test case that verifies that incorrect
input correctly leads to a failure of the request.
It could be done by using a simple rejects assertion of Jest, but instead
a try catch block is used and there's some dead code in the form of some assertion
in there as well because the operation is never meant to succeed.

The problem is not just that it's dead code but also that it's making the test case
not work at all because if the operation ends up succeeding we are passing the test
anyway with the code as it is today.

```typescript
    try {
      const bodyObj: NewContractObj = {
        contractAddress: hashTimeLockAddress,
        inputAmount: 10,
        outputAmount: 0x04,
        expiration: DataTest.expiration,
        hashLock: DataTest.hashLock,
        receiver: DataTest.receiver,
        outputNetwork: "BTC",
        outputAddress: "1AcVYm7M3kkJQH28FXAvyBFQzFRL6xPKu8",
        connectorId: connectorId,
        web3SigningCredential,
        keychainId: "",
        gas: DataTest.estimated_gas,
      };
      const resp = await api.newContractV1(bodyObj);
      expect(resp).toBeTruthy();
      expect(resp.status).toEqual(200);
    } catch (error: any) {
      expect(error.response.status).toEqual(500);
    }
```

Leeyoungone:
Fixes #1697 


packages/cactus-test-plugin-htlc-eth-besu/src/test/typescript/integration/plugin-htlc-eth-besu/refund-endpoint-invalid.test.ts
```diff
@@ -210,17 +210,17 @@ describe(testCase, () => {
...
         connectorId,
         keychainId,
       };
-      const refundResponse = await api.refundV1(refundRequest);
-      expect(refundResponse.status).toEqual(200);
-    } catch (error: any) {
-      expect(error.response.status).toEqual(500);
+      await api.refundV1(refundRequest);
+    } catch (error: unknown) {
+      expect(error).toReject;
```
petermetz, line 223: @Leeyoungone I'd refactor this the same way as the one above with the `.rejects.toHaveProperty(["response", "status"], 500);` way.

petermetz (changes requested):

packages/cactus-test-plugin-htlc-eth-besu/src/test/typescript/integration/plugin-htlc-eth-besu/refund-endpoint-invalid.test.ts
Leeyoungone: Sorry, I think I altered an extra file on accident here. Let me fix this real quick.

petermetz (approved):
> Sorry, I think I altered an extra file on accident here. Let me fix this real quick.  

@Leeyoungone No worries at all, now it looks perfect! Thank you very much for the fix.

izuru0 (approved):
LGTM

PR #1757, opened 2022-01-11
fix: test plugin htlc eth besu erc20 exception handling and upgrade TS

### LINKED ISSUE #1745: build(deps): upgrade to TS 4.4 test-plugin-htlc-eth-besu-erc20

Leeyoungone:
Fixes #1745 
Note: Strange redundant code in the `refund-endpoint-test.ts` and `withdraw-endpoint.test.ts` within the `try-catch` block?


petermetz (changes requested):
@Leeyoungone Please mark this as a dependent to the parent PR (or just rebase it to main if the dependency was by accident)

Strange redundant code: Do you mean the part that asserts the status code being 400 at the end of the try blocks?

Leeyoungone:
@petermetz Yeah, I thought it was a little strange why the same code is written within both the `try` and the `catch` block if it happens either way. 

```typescript 
expect(res.status).toEqual(400);
```

On a similar note (in relation to this epic's PRs). Some of these PR don't require the parent PR (for example, this one). I've rebased onto main. Is that all I needed to do?

petermetz:
> @petermetz Yeah, I thought it was a little strange why the same code is written within both the `try` and the `catch` block if it happens either way.
> 
> ```ts
> expect(res.status).toEqual(400);
> ```

@Leeyoungone Yeah we need to refactor that whole assertion after the jestification went through to just assert that a rejection happened instead of the logic with the try-catch block. (Later, when this PR gets re-opened in the future - just wanted to leave this information here in the meantime)

Leeyoungone:
@petermetz Sounds good, thanks for answering !

Issue #833, opened 2021-04-22
docs(cactus-core-api): Create README file


AzaharaC:
**Description.**
As a potential user of Cactus I want to have documentation of what and how this package is used.

**Describe the solution you'd like.**
Add a README.md file in the packages/cactus-core-api folder with this package usage information.

**Acceptance criteria.**
The README file exists and contains the package information.

PR #1759, opened 2022-01-11
feat(core-api): add weaver protobuf codegen  #1556

### LINKED ISSUE #1556: feat(core-api): add weaver protobuf codegen
https://github.com/hyperledger/cactus/issues/1523#issuecomment-971929781

> VRamakrishna commented
> @petermetz I think we should import the compiled versions (i.e., language-specific bindings) too. Presently, Weaver has these for JS, Go, and Java. Rust is being worked on.

petermetz:
Rust code now gets generated from the Weaver protocol buffer definitions
Next step is to have the automated publishing set up.

Fixes #1556


petermetz:
@VRamakrishna What do you think of this? If it looks good I'll set up the auto-publishing github workflow for a crate.

packages/cactus-core-api/build.rs
```diff
@@ -0,0 +1,23 @@
...
+    tonic_build::configure()
+        .type_attribute(".", "#[derive(serde::Serialize, serde::Deserialize)]")
+        .compile(
+            &[
+                "./src/main/proto/weaver/common/protos/relay/datatransfer.proto",
+                "./src/main/proto/weaver/common/protos/networks/networks.proto",
+                "./src/main/proto/weaver/common/protos/driver/driver.proto",
+            ],
+            &[
+                "./src/main/proto/weaver/common/protos/",
```
VRamakrishna, line 16: Is this line now enough? Do we need to explicitly compile the protos within subfolders? (I'm not that familiar with the syntax myself.)

packages/cactus-core-api/src/main/rust/generated/proto-rs/common.access_control.rs
```diff
@@ -0,0 +1,21 @@
+/// AccessControlPolicy specifies a set of data that can be accessed by some
```
VRamakrishna, line 1: This file doesn't seem to be a relevant dependency. Why is it getting built?

packages/cactus-core-api/build.rs
petermetz: @VRamakrishna Oops, I checked it and it works with this as well, I'll change it accordingly.

```rust
            &[
                "./src/main/proto/weaver/common/protos/relay/datatransfer.proto",
                "./src/main/proto/weaver/common/protos/networks/networks.proto",
                "./src/main/proto/weaver/common/protos/driver/driver.proto",
            ],
            &[
                "./src/main/proto/weaver/common/protos/",
            ],
```

packages/cactus-core-api/src/main/rust/generated/proto-rs/common.access_control.rs
petermetz: @VRamakrishna My bad once again, it was a leftover from the various different tries I executed while I was trying to get the compilation to work in the initial phase of the work.

petermetz:
> How do we now build the relay as a standalone component using the code generated from the protos?
> 
> Also, can you add a `Makefile` and a short `README.md` to the folder?

Adding the readme in the scope of another issue that we had open for a while, stay tuned for the follow-up PR with that: https://github.com/hyperledger/cactus/issues/833

petermetz:
https://github.com/hyperledger/cactus/pull/1794

takeutak (approved):

VRamakrishna (approved):

jagpreetsinghsasan (approved):

Issue #789, opened 2021-04-13
feat(connector-corda): containerize plugin


jordigiam:
**Is your feature request related to a problem? Please describe.**

Cloud has became the de-facto way of running applications on computers.
Containers are rapidly becoming (have become?) the de-facto way of running applications in the cloud.
We therefore need to make sure that Cactus is not just built and distributed as the language native artifacts (e.g. npm packages/webpack bundles) but that it's also distributed as containers.
This doesn't just mean containerizing the API server that can host the plugins, but also means containerizing each plugin individually as well OR providing a generic image of the API server that can be configured to run any and all of the available Cactus plugins on-demand.

**Describe the solution you'd like**

I want to be able to whisper things like this to my terminal and expect to have a fully functional connector for Ledger X up and running in a few seconds and with the config.json file mounted at the specified path/volume that gets picked up by the API server running inside the container as well.

```sh
echo "{ \"some-configuration\": 42 } " > config.json

docker run \
  --rm \
  --volume ./config.json:/opt/hyperledger/cactus/cactus-plugin-ledger-connector-corda/config.json \ 
  hyperledger/cactus-plugin-ledger-connector-corda
```

**Acceptance criteria**

1. Dockerfile for generate de docker image with the plugin connector and the api server running.

elenaizaguirre:
I will continue this issue!

petermetz:
@elenaizaguirre Are you still working on this?

elenaizaguirre:
> @elenaizaguirre Are you still working on this? And if yes, are you working directly on Jordi's old branch (#805) or did you start a new one?

@petermetz Sorry, I completely forgot this issue. I can work on it when I finish the one I'm working on.

petermetz:
@elenaizaguirre No worries, it's easy to lose track. Recently, I've been using the VSCode GitHub extension to keep track of assigned issues/old PRs and it's been much easier through that compared to the web UI so I recommend you give it a try too if you are interested.

petermetz:
Closing this down for now because it's for Corda v4 but we are focusing on Corda v5.

PR #1761, opened 2022-01-12
feat(connector-corda): read privateKey from filesystem

elenaizaguirre:
Allow plugin-ledger-connector-corda to read privateKey from filesystem
instead of obtain it directly from corda-aio container

Relationed with #789


packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
```diff
@@ -54,6 +56,7 @@ export interface IPluginLedgerConnectorCordaOptions
   cordaStartCmd?: string;
   cordaStopCmd?: string;
   apiUrl?: string;
+  privateKeyPath?: string;
```
petermetz, line 59: @elenaizaguirre Please write some JSdoc code documentation for the property specifically here explaining what this does, make it clear that if specified, it will override the value specified in `sshConfigAdminShell`. Also mention that it's optional for the reason that it's not recommended using it.

In the meantime I'll try to do some research into how we could avoid the whole problem with some bash sorcery.

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
```diff
@@ -146,6 +151,14 @@ export class PluginLedgerConnectorCorda
     return webServices;
   }
 
+  private readPrivateKey(): void {
```
petermetz, line 154: @elenaizaguirre Please make the name a little more descriptive. At first glance it might not be clear what kind of private key the method is operating with so a quick sentence for the documentation would be cool and maybe a slighly longer name like `readSshPrivateKey()` or something.

petermetz:
@elenaizaguirre 

Would something like this work (need to substitute the container ID variable)?

```sh
export CORDA_AIO_IMAGE_KEY=$(docker exec $CONTAINER_ID cat /etc/hyperledger/cactus/corda-aio-image.key)
```

then in the compose file you just use the environment variable as usual:

https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers

https://docs.docker.com/compose/environment-variables/#substitute-environment-variables-in-compose-files

And finally what I wanted to double check is the specific reason for using docker compose. Is it to simplify manual testing or are you somehow embedding the docker compose parts into a test case?

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
jagpreetsinghsasan: Yes, we should be avoiding general names like `privateKey` , `publicPrivatekeyPair`, and methods related to that etc, and replace them with a bit more understandable naming schema like the one @petermetz mentioned (`sshPrivateKey`)

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
elenaizaguirre: Done

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/plugin-ledger-connector-corda.ts
elenaizaguirre: Done

elenaizaguirre:
> @elenaizaguirre
> 
> Would something like this work (need to substitute the container ID variable)?
> 
> ```shell
> export CORDA_AIO_IMAGE_KEY=$(docker exec $CONTAINER_ID cat /etc/hyperledger/cactus/corda-aio-image.key)
> ```
> 
> then in the compose file you just use the environment variable as usual:
> 
> https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers
> 
> https://docs.docker.com/compose/environment-variables/#substitute-environment-variables-in-compose-files
> 
> And finally what I wanted to double check is the specific reason for using docker compose. Is it to simplify manual testing or are you somehow embedding the docker compose parts into a test case?

@petermetz 

I am using docker compose because I think it improves the user experience by deploying all containers with a single command, but I'm having a lot of problems because of it.

If you want to see what I am doing it is available in [this branch](https://github.com/elenaizaguirre/cactus/tree/containerize-corda).

Now I am not sure how to deal with this issue, I think there are three ways to do it:

1. Forget docker compose and do it with independent containers

2. Create the keys from the host and also send them to corda-aio as environment variables (we would have to modify the image so I don't think this is the best option).

3. Merge this PR, I now have the key located inside the second container so it would not be a problem. I used your command to export the key to a host file and then pass it as a secret to the second container (it is not possible to do it in this case with environment variables because they are evaluated before starting the first container).

petermetz (approved):
@elenaizaguirre Got it, thank you for explaining. I am definitely in favor of merging this PR so that option 3) becomes feasible ASAP and we can think about the rest of them later as well. Thank you for addressing my change requests.

izuru0 (approved):
LGTM

PR #1771, opened 2022-01-13
fix: plugin ledger connector sawtooth socketio exception handling and upgrade TS

### LINKED ISSUE #1739: build(deps): upgrade to TS 4.4 plugin-ledger-connector-sawtooth-socketio

Leeyoungone:
Fixes #1739

### DIFF
--- a/packages/cactus-plugin-ledger-connector-sawtooth-socketio/src/main/typescript/connector/ServerPlugin.ts
+++ b/packages/cactus-plugin-ledger-connector-sawtooth-socketio/src/main/typescript/connector/ServerPlugin.ts
@@ -24,6 +24,9 @@ import { ValidatorAuthentication } from "./ValidatorAuthentication";
 // Load libraries, SDKs, etc. according to specifications of endchains as needed
 const Web3 = require("web3");
 
+import axios from "axios";
+import RunTimeError, { RuntimeError } from "run-time-error";
+
 /*
  * ServerPlugin
  * Class definition for server plugins
@@ -106,21 +109,25 @@ export class ServerPlugin {
         }
         logger.debug(`##getNumericBalance: retObj: ${JSON.stringify(retObj)}`);
         return resolve(retObj);
-      } catch (e) {
-        const emsg = e.toString().replace(/Error: /g, "");
-        logger.error(emsg);
-        retObj = {
-          resObj: {
-            status: 504,
-            errorDetail: emsg,
-          },
-        };
-        logger.debug("##getNumericBalance: add reqID");
-        if (reqID !== undefined) {
-          retObj["id"] = reqID;
+      } catch (e: unknown) {
+        if (axios.isAxiosError(e)) {
+          const emsg = e.toString().replace(/Error: /g, "");
+          logger.error(emsg);
+          retObj = {
+            resObj: {
+              status: 504,
+              errorDetail: emsg,
+            },
+          };
+          logger.debug("##getNumericBalance: add reqID");
+          if (reqID !== undefined) {
+            retObj["id"] = reqID;
+          }
+          logger.debug(`##getNumericBalance: retObj: ${JSON.stringify(retObj)}`);
+          return reject(retObj); 
+        } else {
+          throw new RuntimeError("expected an axios error, got something else");
         }
-        logger.debug(`##getNumericBalance: retObj: ${JSON.stringify(retObj)}`);
-        return reject(retObj);
       }
     });
   }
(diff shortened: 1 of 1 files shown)


packages/cactus-plugin-ledger-connector-sawtooth-socketio/src/main/typescript/connector/ServerPlugin.ts
```diff
@@ -413,21 +433,25 @@ export class ServerPlugin {
...
+            },
+          };
+          logger.debug("##contract: add reqID");
+          if (reqID !== undefined) {
+            retObj["id"] = reqID;
+          }
+          logger.debug(`##contract: retObj: ${JSON.stringify(retObj)}`);
+          return reject(retObj);
+        } else {
+          throw new RuntimeError("expected an axios error, got something else");
```
petermetz, line 453: @Leeyoungone Why did we expect an axios error here though? I also don't know what should happen if the thrown error is not what the code expects it to be, but we can ask @outSH about that.

@outSH: Do you know the reason why the `Error: ` string here needs to be removed and what should happen if the thrown exception was not something with a `.toString()` function?

outSH: @petermetz @Leeyoungone Sorry for late response. Well, first of all, as far as I know sawtooth connector supports only block monitoring at the moment, and this code is probably a leftover from copying go-ethereum structure (when it was initially developed), so I think it could be removed completely. I'm not very familiar with sawtooth DLT but I'm pretty sure you don't use web3 to interact with it :) @izuru0  - can you confirm this?

As for the exceptions (since exact same code is in go-ethereum):
- We don't expect `AxiosError` but an `Error(...)` exception, which are also thrown by web3.
- Removing of `Error: ` is just a convention. Message type is already indicated in response key (`errorDetail`), and we can prevent from printing repeated Error strings in client-side logs (like `Error(cmd-socketio) - Error: fooo`).
- All exceptions caught here should be `Error`, if they are not we could return 504 with a message like "Received unknown error", and also log some more details for future diagnostics.

petermetz: @outSH No worries, thank you for the response. For now (temporarily) the PR is closed, but we can continue the discussion in the meantime because the plan is to re-open it later.

> so I think it could be removed completely. I'm not very familiar with sawtooth DLT but I'm pretty sure you don't use web3 to interact with it :) @izuru0 - can you confirm this?

Okay, so if @izuru0 can confirm it then we can just remove it and the whole conversation is solved that way. :-)

---

> Removing of Error: is just a convention. Message type is already indicated in response key (errorDetail), and we can prevent from printing repeated Error strings in client-side logs (like Error(cmd-socketio) - Error: fooo).

> All exceptions caught here should be Error, if they are not we could return 504 with a message like "Received unknown error", and also log some more details for future diagnostics.

This might become an irrelevant point if the whole code just gets removed (see above) but if not, here's what I'd recommend: Remove the additional logic of processing the error in the form of removing the string. IMO, the upside is not worth the potential downside. You give the person reading the logs a tiny bit of extra convenience, which is nice, but you are risking that they won't get any useful information at all if your catch block code crashes due to the assumptions you are making about the thrown object (which in JS can be literally anything and therefore we must assume that libraries that are being used could do things like `throw null;` or `throw "Some Explanation of an Exception";` or `throw { problem: "Some explanation."};`

petermetz:
@Leeyoungone I'll re-open this while we are finishing up that conversation in the comments above (this way we won't lose it)

packages/cactus-plugin-ledger-connector-sawtooth-socketio/src/main/typescript/connector/ServerPlugin.ts
outSH: @petermetz 

> Okay, so if @izuru0 can confirm it then we can just remove it and the whole conversation is solved that way. :-)

The same thing applies to other socketio only validators, that's why I wrote down some explanations :)

> .. but you are risking that they won't get any useful information at all if your catch block code crashes due to the assumptions you are making about the thrown object

That's why I like the idea of checking the type of the exception first, then doing the processing if applicable. Unknown exceptions should not crash the program, but trigger generic error response (like proposed earlier `504 - Received unknown error` or anything similar). If the exception can be `null` or `5` you can't really assume that it's string representation has any meaning for the validator user, so returning it doesn't make much sense. And in that case, when we're sure we operate on `Error`, stripping off the `Error: ` part is convenient and semantically correct with entire response body (so we don't return `error: "Error error error"`).

Either way, I don't think this topic is worth a lengthy discussion, do what you think is best :)

takeutak (dismissed):

packages/cactus-plugin-ledger-connector-sawtooth-socketio/src/main/typescript/connector/ServerPlugin.ts
izuru0: sorry for slow response. we will check this very soon.

petermetz: @outSH 

> That's why I like the idea of checking the type of the exception first

I agree. This was my point. 

> If the exception can be null or 5 you can't really assume that it's string representation has any meaning for the validator user, so returning it doesn't make much sense.

I'd say you can't really assume that it does NOT have any meaning either. See my above examples (some of which you've covered, but not the ones where a non-Error throw could carry useful information like a POJO).
My motivation in these scenarios is selfish, meaning that the validator user may find no meaning in some obscure looking information that we returned, but I still could if they sent me the logs.  

> Either way, I don't think this topic is worth a lengthy discussion, do what you think is best :)

Understood, thank you!

outSH: > My motivation in these scenarios is selfish, meaning that the validator user may find no meaning in some obscure looking information that we returned, but I still could if they sent me the logs.

@petermetz We still have to return a string error here, so a lot of information can be lost, that's why I've suggested generic error message and error log entry for future investigation :)

BTW I've talked with Izuru-san few days ago, I'll double check and remove this `ServerPlugin.ts` if possible from here, I'll also check other validator codes and fix returned errors as discussed.

petermetz: @outSH 

> We still have to return a string error here

I wasn't advocating for not returning a string error. (or at least I didn't mean to).

I see that the file is now gone, sorry for the slow responses especially apologies to @Leeyoungone but based on this news we can just close this PR down since the improvements were made to a file that has been deleted in the meantime.

petermetz:
Closing because the file that the diff is for has been deleted in another change in parallel. Apologies @Leeyoungone this was my fault for taking too long to respond to the conversation.

Issue #1381, opened 2021-09-24
feat(core): add shared error handling tools for web services


petermetz:

RafaelAPB:
This issue will be considered if the depending package remains, as part of the cleanup initiative: https://github.com/hyperledger-cacti/cacti/issues/4037

mn-ram:
Hi @petermetz, @RafaelAPB — proposing LFDT mentorship scope (Issue #62 cleanup).

Status:
- Core error utilities exist
- Missing docs + inconsistent usage (`res.status(400)` still used)

Plan:
1. Add `REST-ERROR-HANDLING.md` (decision tree)
2. Migrate endpoints (start: cmd-api-server → SATP-Hermes)
3. Continue post #4025

Question:
Start docs now and pause migrations, or wait for #4025?

Tracking: EPIC #4036  
Context: #4161, #4168, #4172

PR #1772, opened 2022-01-13
fix: plugin ledger connector quorum exception handling and upgrade TS

### LINKED ISSUE #1738: build(deps): upgrade to TS 4.4 plugin-ledger-connector-quorum

Leeyoungone:
Depends on \#1707 

Depends on #1381

Fixes \#1738

### DIFF
--- a/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
+++ b/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
@@ -43,6 +43,7 @@ import {
   Bools,
   Logger,
   LoggerProvider,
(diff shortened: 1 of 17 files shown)


petermetz:
@Leeyoungone Let's wait with this one until the dependencies are resolved, it's tricky to review before then because correctness depends on how the dependencies will get resolved exactly.

petermetz (changes requested):
@Leeyoungone Let's wait with this one until the dependencies are resolved, it's tricky to review before then because correctness depends on how the dependencies will get resolved exactly.

PR #1779, opened 2022-01-14
style: 2021-09-20 linter warnings batch 23/26

zondervancalvez:
fixes: #1371


petermetz (approved):
@zondervancalvez Thank you very much, LGTM!

petermetz (approved):
@zondervancalvez LGTM, thank you!

takeutak (approved):

petermetz (changes requested):
@zondervancalvez Sorry, I just realized that the commit message you have is way too generic.

Could you make the commit message mirror the issue title instead?
`style: 2021-09-20 linter warnings batch 23/26`

And also include the `Fixes #???` syntax in the commit message body so that we can trace it back later directly from the git logs instead of via the github PR description only.

Issue #1799, opened 2022-01-19
fix(plugin-ledger-connector-fabric-socketio): upgrade Fabric due to jsrsasign


petermetz:
**Describe the bug**

We have a critical severity vulnerability put back in the code by a recent PR: https://github.com/hyperledger/cactus/pull/1754

packages/cactus-plugin-ledger-connector-fabric-socketio/package.json

```json
    "fabric-ca-client": "~1.4.0",
    "fabric-client": "~1.4.0",
    "fabric-network": "~1.4.0",
```

**To Reproduce**

`npm ls jsrsasign` 

```sh
$ npm ls jsrsasign
@hyperledger/cactus@ /home/peter/a/blockchain/cactus
├─┬ @hyperledger/cactus-cmd-socket-server@1.0.0-rc.3 -> ./packages/cactus-cmd-socketio-server
│ ├─┬ fabric-ca-client@2.2.10
│ │ ├─┬ fabric-common@2.2.10
│ │ │ └── jsrsasign@10.5.0 deduped
│ │ └── jsrsasign@10.5.0
│ └─┬ fabric-network@2.2.10
│   └─┬ fabric-common@2.2.10
│     └── jsrsasign@10.5.0 deduped
├─┬ @hyperledger/cactus-plugin-ledger-connector-fabric-socketio@1.0.0-rc.3 -> ./packages/cactus-plugin-ledger-connector-fabric-socketio
│ ├─┬ fabric-ca-client@1.4.18
│ │ └── jsrsasign@8.0.24
│ └─┬ fabric-client@1.4.18
│   └── jsrsasign@8.0.24
├─┬ @hyperledger/cactus-plugin-ledger-connector-fabric@1.0.0-rc.3 -> ./packages/cactus-plugin-ledger-connector-fabric
│ ├─┬ fabric-ca-client@2.3.0-snapshot.62
│ │ └── jsrsasign@10.5.0
│ ├─┬ fabric-common@2.3.0-snapshot.63
│ │ └── jsrsasign@10.5.0 deduped
│ ├── jsrsasign@10.4.0
│ └─┬ ws-wallet@1.1.5
│   └── jsrsasign@10.4.1
└─┬ @hyperledger/cactus-test-tooling@1.0.0-rc.3 -> ./packages/cactus-test-tooling
  └─┬ fabric-ca-client@2.2.10
    ├─┬ fabric-common@2.2.10
    │ └── jsrsasign@10.5.0 deduped
[...]

PR #1754, opened 2022-01-11
refactor(connector-fabric-socketio): migrate to fabric-all-in-one from fabric14-fabcar-testnet

outSH:
\- Remove obsolete fabric14\-fabcar\-testnet fabric14\-testnet, use only fabric\-all\-in\-one for all purposes for all test / sample app purposes.
\- Add helper scripts from fabric\-sample that are used during cartrade setup phase \(I've decided it's better to keep them in repo than increasing the size of fabric\-all\-in\-one image\).
\- Upgrade of fabcar network \(from 1.4.1 to 1.4.8\) triggered multiple changes in fabric\-socketio validator, since fabcar nodes \(peers, orderer\) use TLS and require different setup, additionaly done some refactors and fixes in the validator.
\- Update the cartrade and discounted\-cartrade sample apps to work with the new setup, refreshed scripts can be used to ease up the setup process.

Closes: \#1753
Signed\-off\-by: Michal Bajer <michal.bajer@fujitsu.com>

Please note that I'll work on containerizing fabric\-socketio validator and cartrade sample, for this reason I didn't put too much effort in documenting changes from this commit \(since most of it will change very soon\).
Also, docker startup scripts will probably be replaced by sample\-specific docker\-compose.

Depends on #1696

PR #1800, opened 2022-01-20
fix(plugin-ledger-connector-fabric-socketio): upgrade Fabric due to jsrsasign

outSH:
Revert fabric sdk package change from PR #1754

Closes: #1799

After change:
```
@hyperledger/cactus@ /home/outsh/cactus
├─┬ @hyperledger/cactus-cmd-socket-server@1.0.0-rc.3 -> ./packages/cactus-cmd-socketio-server
│ └─┬ fabric-ca-client@1.4.19
│   └── jsrsasign@10.5.0
├─┬ @hyperledger/cactus-example-carbon-accounting-backend@1.0.0-rc.3 -> ./examples/cactus-example-carbon-accounting-backend
│ └─┬ fabric-network@2.2.10
│   └─┬ fabric-common@2.2.10
│     └── jsrsasign@10.5.0 deduped
├─┬ @hyperledger/cactus-plugin-ledger-connector-fabric-socketio@1.0.0-rc.3 -> ./packages/cactus-plugin-ledger-connector-fabric-socketio
│ └─┬ fabric-client@1.4.19
│   └── jsrsasign@10.5.0 deduped
├─┬ @hyperledger/cactus-plugin-ledger-connector-fabric@1.0.0-rc.3 -> ./packages/cactus-plugin-ledger-connector-fabric
│ ├─┬ fabric-ca-client@2.3.0-snapshot.62
│ │ └── jsrsasign@10.5.0
│ ├─┬ fabric-common@2.3.0-snapshot.63
│ │ └── jsrsasign@10.5.0 deduped
│ ├── jsrsasign@10.4.0
│ └─┬ ws-wallet@1.1.5
│   └── jsrsasign@10.4.1
└─┬ @hyperledger/cactus-test-tooling@1.0.0-rc.3 -> ./packages/cactus-test-tooling
  └─┬ fabric-ca-client@2.2.10
    ├─┬ fabric-common@2.2.10
    │ └── jsrsasign@10.5.0 deduped
    └── jsrsasign@10.5.0 deduped
```


examples/cartrade/package.json
```diff
@@ -20,8 +20,8 @@
     "ethereumjs-common": "^1.5.1",
     "ethereumjs-tx": "^2.1.2",
     "express": "~4.16.1",
-    "fabric-ca-client": "~1.4.0",
-    "fabric-network": "~1.4.0",
+    "fabric-ca-client": "^1.4.0",
```
petermetz, line 23: @outSH Thank you for the quick fix for the CVE! Could you lock it down to a specific version (in this case it would be `1.4.19`) by any chance? The reason I ask is because having the exact versions specified prevents auto-upgrades and it would get us (closer) to the utopia of [reproducible builds](https://reproducible-builds.org/)

(I'll just comment on this one file of the diff but the same question goes for all the other package.json files of course)

petermetz (changes requested):

examples/cartrade/package.json
outSH: @petermetz Builds are already reproducible due to `yarn.lock`, or am I missing something?? In which scenario range definitions are getting in your way?

petermetz: @outSH It would go like this (at least that's how I imagine it, willing to be wrong of course)

1. The example works today with 1.4.19, we all go home happy
2. Next week 1.4.20 gets released that introduces a bug that breaks the example
3. New prospective contributors/users come in, try the example and it's broken because their npm cache did not have 1.4.19 and the range made it so that they got 1.4.20 because they ran the examples from scratch post-release of the 1.4.20 version of the package.

outSH: @petermetz

> they got 1.4.20 because they ran the examples from scratch post-release of the 1.4.20 version of the package.

But, as long as they use `yarn run configure` as documented, they will still get 1.4.19 even after an update was released. I understand having precise version would be safer for users not familiar with package pinning, but I'm afraid this could be problematic in future update management. I've already encountered problem when trying to upgrade just `fabric-ca-client` with `~1.4.0` specifier, probably due to it's dependencies (didn't investigate it further, though), I'm afraid having plain `1.4.19` could make it worse.

Either way, if you really sure this is the best way to go, I can change that, let me know

petermetz: @outSH 

> But, as long as they use yarn run configure as documented

The cartrade example (and many others) are not part of the monorepo builds (the configure script). I had provided avenue to remedy this in an earlier PR but my work was deleted, so as of right now, it's just dangling outside of the lerna build, same as before.
If you look at the cartrade example readme there is no mention of using yarn configure, it is telling you to switch directories into the example and then run npm install yourself.

> I've already encountered problem when trying to upgrade just fabric-ca-client with ~1.4.0 specifier, probably due to it's dependencies (didn't investigate it further, though)

Hard to tell without further details.

> Either way, if you really sure this is the best way to go, I can change that, let me know

Thank you! I vote for exact versions, regardless of it being more effort to manually bump them later or not.

outSH: @petermetz Oh, right! I'm currently working on putting them into containers and monorepo build right now, and for some reasons assumed it's already done in my mind :)

> Thank you! I vote for exact versions, regardless of it being more effort to manually bump them later or not.

OK. Shall I switch to exact versions for every dependency in `package.json`? Only fabric one? Shall I do this for all `package.json` affected by this PR?

outSH: @petermetz nevermind, I've changed only fabric packages to exact versions, since only them are related to this commit, if you wish to continue this process please open new issue describing the scope :)

izuru0 (approved):
LGTM

outSH:
@petermetz Can you approve and merge this please? It's blocking PR next in line :)

petermetz:
> @petermetz Can you approve and merge this please? It's blocking PR next in line :)  

@outSH Thank you for the information. I'm working on the reviews as fast as I can. :)

petermetz (approved):
> @petermetz Oh, right! I'm currently working on putting them into containers and monorepo build right now, and for some reasons assumed it's already done in my mind :)

> Thank you! I vote for exact versions, regardless of it being more effort to manually bump them later or not.

@outSH No worries, I mix up my own PRs all the time. Extending on my earlier answer: the yarn.lock wouldn't save us in some edge cases either and because of that I always ask people to specify exact versions instead of ranges (so I would've asked the same thing even if the packages were part of the monorepo already)

> OK. Shall I switch to exact versions for every dependency in `package.json`? Only fabric one? Shall I do this for all `package.json` affected by this PR? 

That can be part of another PR, I dont' want to complicate this one any further.

outSH:
```
Thank you for the information. I'm working on the reviews as fast as I can. :)
```
Sure, not rushing, I just wanted to put this PR to the front ;) Thank's for quick handling other PR as well, btw

Issue #1805, opened 2022-01-21
refactor(cartrade): dockerize catrade and fabric validator


outSH:
Full cartrade example should be put into container.

PR #1806, opened 2022-01-21
refactor(docs): dockerize catrade and fabric validator

outSH:
- Cartrade example will be setup in docker-compose, updated the readme with new instructions.
- Fabric validator runs in container as well.
- Removed obsolete scripts, changed them where needed.
- Build artifacts are not put directly in root `dist/` dir anymore (it's still used by apps not dockerized yet, though, but will be changed in future commits).
- Validator configuration is made cartrade-specific, this will be done in the same way for other sample apps in future commits.
- Some setup changes were necessary in other samples as well, all should work like before.
- Added small patch in fabric-all-in-one to support cartrade scenario (endorsment from peer from single org).
- Updated indy clientbase base image because it didn't work with `ppa:deadsnakes` anymore.

Closes: #1805

### TODO (in future PR)
- Clean exports from `@hyperledger/cactus-cmd-socket-server` (will not use this long fragile path in the future)

Depends on #1800


outSH:
@izuru0 @takeutak @petermetz - Please review

tools/docker/fabric-all-in-one/run-fabric-network.sh
```diff
@@ -49,6 +49,12 @@ function main()
     /bootstrap.sh ${FABRIC_VERSION} ${CA_VERSION} -b -s
     # Major version is 1.x or earlier (assumption is 1.4.x only)
     cd /fabric-samples/fabcar/
+
+    # PATCH
+    # This will change endorsment policy on fabcar chaincode.
+    # cactus-plugin-ledger-connector-fabric-socketio supports only single peer endorsment at the moment.
+    sed -i "s/AND('Org1MSP.member','Org2MSP.member')/OR('Org1MSP.member','Org2MSP.member')/g" ./startFabric.sh
```
petermetz, line 56: @outSH Could you please put this behind an `CACTUS_TEST_LOOSE_MEMBERSHIP_CHECK=ON/OFF` environment variable that is off by default? 
My concern is that the existing tests using the image could get compromised because of the altered ledger behavior. With the env var you can have it enabled when you need to but otherwise the change is backward compatible by default.

petermetz (changes requested):

tools/docker/fabric-all-in-one/run-fabric-network.sh
outSH: @petermetz I thought there are no tests using faio 1.4 (since it didn't work for few months), but that's a good suggestion either way, will do 👍

petermetz: @outSH The tests are pinned to certain know-to-work versions of the AIO images for this exact reason. With that said, my memories are foggy about which ones are being skipped at present due to flakiness, so on top of what I just said it is still possible that the 1.4 test has not been part of the CI in recent months (but even if this is the case they'll be brought back on as soon as we can get rid of the flakes)

Thank you very much for the update, cheers!

petermetz (approved):
@outSH LGTM, thank you again for the updates!

izuru0 (approved):
LGTM

Issue #1804, opened 2022-01-21
test(api-client): fix flaky socketio-api-client.test


petermetz:
**Describe the bug**

The `packages/cactus-api-client/src/test/typescript/unit/socketio-api-client.test.ts` appears to be failing intermittently.

**To Reproduce**

Run it through the CI enough times to make it fail. Observe that it works sometimes and not other times.

**Expected behavior**

Tests should be stable, e.g. always failing or always succeeding with a given revision of the code.

**Logs/Stack traces**

```sh
Summary of all failing tests
FAIL packages/cactus-api-client/src/test/typescript/unit/socketio-api-client.test.ts (15.777 s)
SocketIOApiClient Tests › Sync Request Tests › Returns decoded response from the validator

thrown: "Exceeded timeout of 3000 ms for a hook.
Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

    276 |
    277 |   describe("Sync Request Tests", function () {
> 278 |     beforeEach(async () => {
        |     ^
    279 |       // Connect client and server sockets
    280 |       testServer.on("connection", (socket) => {
    281 |         log.debug("Server socket connected", socket.id);

    at packages/cactus-api-client/src/test/typescript/unit/socketio-api-client.test.ts:278:5
    at packages/cactus-api-client/src/test/typescript/unit/socketio-api-client.test.ts:277:3
    at Object.<anonymous> (packages/cactus-api-client/src/test/typescript/unit/socketio-api-client.test.ts:196:1)
```

**Cloud provider or hardware configuration:**

GitHub Action Runner for the Cactus CI

**Operating system name, version, build:**

Ubuntu-20.04

**Hyperledger Cactus release version or commit (git rev-parse --short HEAD):**

`1.0.0-rc.3`

@outSH

outSH:
@petermetz 
I've just executed this test over 800 times on local VM and didn't find any errors, I assume this is probably due to periodic low performance of CI environment. I've increased the timeouts for all tests (3s -> 5s), and separate setup timeout (3s -> 1 minute). Let's see if this solves the issue :)

petermetz:
@outSH Great! Thank you very much! I agree, I'm pretty sure that the free VMs we get allocated for the GitHub CI jobs are heavily throttled in the hypervisor level somehow so they are probably pretty "weak" even compared to a regular laptop.

PR #1807, opened 2022-01-21
test(api-client): fix flaky socketio-api-client.test

outSH:
Increase timeout on all tests, add extra timeout for setup operations to solve jest timeout errors in CI

Closes: #1804


packages/cactus-api-client/src/test/typescript/unit/socketio-api-client.test.ts
```diff
@@ -72,7 +74,7 @@ import {
 // Test Timeout
 //////////////////////////////////
 
-jest.setTimeout(3000);
+jest.setTimeout(testTimeout);
```
petermetz, line 77: @outSH This will override the global default, from within a single test case which makes the configuration distributed among multiple files instead of having it in a centralized location. Could you define the increased timeout in the jest config file instead or if it's not possible there ( but I'm really hoping that it is) then some other central location like the way they are recommending it in the Jest documentation [1] of this method (they also do not recommend using it in individual test cases)

[1] https://jestjs.io/docs/jest-object#jestsettimeouttimeout

petermetz (changes requested):

packages/cactus-api-client/src/test/typescript/unit/socketio-api-client.test.ts
outSH: @petermetz 

- This is set locally since timeout in global config is ridiculously long (1 hour), and due to async nature of socketio tests, there's no way to detect negative result directly, so timeout is critical in detecting failing tests. I've set it to shorter in this file, so that failing test will not freeze entire CI.
- I don't know why the global timeout is so long, but I didn't want to change that to avoid breaking some other tests that may actually depend on it.
- Also, since our jest test are very diverse, some are fully-fledged integration tests with docker setup etc., some are socketio tests, and some are possibly plain-old unit tests, it's hard to define one common timeout for all these tests. I think that solving this outside of scope of this PR :)

petermetz: @outSH 

> This is set locally

My bad, I need to learn to read. Sorry about the false alarm. I'm good now.

>  timeout in global config is ridiculously long (1 hour)

Yeah, that's like that because some of the container images we use for the tests are quite large (1GB+) in size and there are some flakes that were happening due to some bandwidth throttling when downloading the images for a test case would take like 20 minutes and then another 10 for the integration test itself to actually run, so that's how we ended up with the one hour.

---

Something I still don't understand (but this is just an off-topic question besides the review itself) is why do you need the test itself to time out for a negative test case when you are verifying that timeouts do happen? What's wrong with using setTimeout within the test case itself and then writing assertions there, etc. Why does the test case specifically have to time out? I feel like I'm missing something basic here.

petermetz (approved):
@outSH Thank you, LGTM

izuru0 (approved):
LGTM

packages/cactus-api-client/src/test/typescript/unit/socketio-api-client.test.ts
outSH: @petermetz Well, I'd have to write separate `setTimeout` in every test in this file, since each test could fail due to  socketio connection hang, error in our code, wrong socketio api usage after upgrade etc.. (which would freeze entire execution). This could be messy and problematic when someone'd add a new test and forget to call this own `setTimeout`, so I've used a jest timeout built-in to handle all these cases :)

petermetz: @outSH 

> each test could fail due to socketio connection hang, error in our code, wrong socketio api usage after upgrade etc.. (which would freeze entire execution).

You mentioned earlier "negative results" which made me think that you are talking about these being negative tests, but now it does not sound like that anymore, so I'm going to take a step back and ask this: Are these tests where you increased the timeout are positive test cases or negative ones?

petermetz: @outSH Marking the thread as resolved because otherwise it won't let me press 'merge' but I am still interested in fully understanding what's going (see my comment above)

outSH: @petermetz Sorry, I mislead you with unclear statement, there are no negative tests as far as I remember :)

When I said `there's no way to detect negative result directly` I meant detecting that test case failed, for instance:
- I test if request is send from ApiClient to a server, and a response is received.
- Something went wrong, unhandled exception crashes the server, the server doesn't respond.
- Client code (in unit-test) awaits a reponse but doesn't get it, it timeouts
- Test case crashes (thanks to timeout)
- Timeout is shorter so we don't wait an hour to get this information that something is broken :)

PR #1814, opened 2022-01-26
test: fix htlc eth besu initialize-endpoint-invalid

### LINKED ISSUE #1684: test: fix htlc eth besu initialize-endpoint-invalid
The mentioned test case has this assertion on the bottom that does not seem logical:
it does something, asserts that it succeeded (doing all of this in a try block)
and then in the catch block there's another assertion that if an exception was thrown
all it does is asserts that the exception is truthy which seems like the test does not
actually verify anything at all about the behavior of the code.

My guess is that it is a negative test case, e.g. it verifies that some failure
does happen with invalid parameters (based on the file name) and so what is
the illogical part is actually the assertion of success within the try block.

So with all that explained the solution I'd like here is to see it refactored in 
a way that it uses promise rejection assertions of Jest instead of the complicated
try/catch block and the phony success assertion.

Path to the test case: 

`packages/cactus-test-plugin-htlc-eth-besu/src/test/typescript/integration/plugin-htlc-eth-besu/initialize-endpoint-invalid.test.ts`

elenaizaguirre:
Closes #1684


petermetz (approved):

petermetz (approved):

izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):

Issue #1829, opened 2022-02-02
refactor(socketio-validators): remove dead-code, fix error handling


outSH:
- Remove unused code from socketio validators.
- Fix error handling as discussed in #1771

### Validators checked:
- `cactus-plugin-ledger-connector-fabric-socketio/`
- `cactus-plugin-ledger-connector-go-ethereum-socketio/`
- `cactus-plugin-ledger-connector-sawtooth-socketio/`

PR #1830, opened 2022-02-02
refactor(socketio-validators): remove dead-code, fix error handling

outSH:
- Remove unused code from socketio validators
- Fix error handling so that thrown object doesn't need `toString()` method defined (we convert to string representation with `String()` now)

Closes: #1829


packages/cactus-plugin-ledger-connector-fabric-socketio/src/main/typescript/connector/ServerMonitorPlugin.ts
```diff
@@ -144,7 +144,7 @@ export class ServerMonitorPlugin {
           logger.error(err);
           const errObj = {
             status: 504,
-            errorDetail: err,
+            errorDetail: String(err),
```
petermetz, line 147: @outSH I highly recommend using something like https://www.npmjs.com/package/fast\-safe\-stringify instead of \`String\(...\)\`

because of this:

\`\`\`javascript
const ex = \{ error: "Something very important that must be known to the preson who'll read the error logs later in order to provide a fix to the problem." \};

const errorDetail = String\(ex\);

// Now all we have about what happened is this:
// "\[object Object\]"
\`\`\`

This is a recommendation not a change request.

petermetz (approved):
@outSH LGTM, but please see my recommendation above^^

packages/cactus-plugin-ledger-connector-fabric-socketio/src/main/typescript/connector/ServerMonitorPlugin.ts
outSH: @petermetz Thanks, I didn't know this library before, it also solved problem of `String([Symbol("foo")])`, which is highly unlikely to be thrown, but would crash the whole application in this odd scenario. I'll replace it :)

petermetz (approved):
@outSH LGTM

packages/cactus-plugin-ledger-connector-fabric-socketio/src/main/typescript/connector/ServerMonitorPlugin.ts
petermetz: @outSH Awesome, I'm very glad it was helpful and thank you for going the extra mile! :-)

izuru0 (approved):
LGTM

PR #1839, opened 2022-02-05
ci: fix CodeQL security scans

### LINKED ISSUE #1836: ci: fix CodeQL security scans
We have had a workflow file for CodeQL in the codebase for a long time now,
but it was placed in the wrong folder and therefore the GitHub Actions framework
did not pick it up for execution, ever. /facepalm and pat on the back for Peter
once again...

This is the file path at the time of this writing:
`./codeql-analysis.yml`

And it should be under `.github/workflows/` path instead.

petermetz:
- Moved the file to the correct directory
- Pinned the runner version to Ubuntu 20.04

Fixes #1836

Depends on #1912 
Depends on #1910 
Depends on #1908
 
### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     strategy:
       fail-fast: false


takeutak (approved):
@petermetz I approved because this is urgent, but I'd like to confirm one thing.
In my understanding, the codes on "Unchanged files with check annotations" have not changed by this PR.  Has it right?

petermetz:
> @petermetz I approved because this is urgent, but I'd like to confirm one thing. In my understanding, the codes on "Unchanged files with check annotations" have not changed by this PR. Has it right?

@takeutak That is correct. Those files are being changed by the other PRs (the ones that this has listed as dependencies).
This PR only adds the check (which will keep failing until those other PRs are merged and then it will start succeeding)

jagpreetsinghsasan (approved):
LGTM

Issue #1855, opened 2022-02-11
refactor(discounted-cartrade): replace fabcar with asset-transfer-basic


outSH:
- Fabcar has been deprecated in fabric 2.x
- Refactor `discounted-cartrade` to use `asset-transfer-basic` CC instead.

`discounted-cartrade` will be renamed to `discounted-asset-trade` in a future PR.

PR #1850, opened 2022-02-10
chore(cartrade): remove cartrade sample

outSH:
There's no benefit from having cartrade and more complete
discounted-cartrade at the same time, remove cartrade sample app from
the repo.

Closes: https://github.com/hyperledger/cactus/issues/1849
Signed-off-by: Michal Bajer <michal.bajer@fujitsu.com>

Depends on #1843

PR #1856, opened 2022-02-11
refactor(discounted-cartrade): replace fabcar with asset-transfer-basic

outSH:
- Use basic CC instead of fabcar.
- Add debug overwrite for discounted-cartrade docker compose (will export nodejs debug port).
- Remove faio 1.4 docker-compose (not used anymore).
- Remove fabcar setup from faio 2.x (not used).
- Change Faio 2.x healtcheck from init transaction to query call - more suited for this role. InitLedger will be executed once during faio startup.
- Rename fabcar-cli-1.4 and adjust to basic asset transfer, they will be removed in the future (after solving config management in socketio validators).
- Fix sending unsigned requests to fabric in discounted-cartrade BLP.
- Remove `discounted-cartrade\validatorDriver` - not working anymore and duplicated in fabric-socketio validator code. - Remove `get-eth-balance` app, not used anymore. 
- Query ledgers with BLP requests, not separate tools.

Closes: https://github.com/hyperledger/cactus/issues/1855

Depends on https://github.com/hyperledger/cactus/pull/1850


.vscode/template.launch.json
```diff
@@ -53,6 +53,15 @@
       "console": "integratedTerminal",
       "sourceMaps": true,
       "cwd": "${workspaceFolder}/examples/supply-chain-app/"
+    },
+    {
```
petermetz, line 57: @outSH Is there some documentation about how this should work? I tried it just now but it didn't appear to launc the VSCode debugger or any terminal processes so I couldn't figure out what it was doing.

outSH: It only works when you already `docker-compose up` a sample app with debug yaml file, without that it does nothing. I want to add more utility for starting the sample apps in the future (after including them in monorepo etc...), didn't want to loose this config until then :)

petermetz (approved):
> @petermetz It only works when you already `docker-compose up` a sample app with debug yaml file, without that it does nothing. I want to add more utility for starting the sample apps in the future (after including them in monorepo etc...), didn't want to loose this config until then :) 

@outSH OK, please add some documentation when you get the chance. I can also recommend looking into the `preLaunchTask` section of https://code.visualstudio.com/docs/editor/debugging with which you could probably make it self-contained (haven't tested this myself either, so this is just an idea)

izuru0 (approved):
LGTM

Issue #1878, opened 2022-02-23
feat(cactus-api-client): common verifier-factory


outSH:
This is continuation of https://github.com/hyperledger/cactus/issues/1602

It turns out that we need a full `VerifierFactory` API that is compatible with both types of connectors/validators (both plain socketio and socketio/openapi), since it's already used by someone.

VerifierFactory will:
- Read validator configuration (possibly from fs)
- Create verifier only from validator id (defined in config). Optionally, there'll be argument for validator type for type safety.
- Verifier will allow communicating with both validators in unified way.

We'd like this to be included in 1.0

PR #1879, opened 2022-02-23
feat(cactus-api-client): common verifier-factory

outSH:
- Adjust SocketIOApiClient and common Verifier behavior to previous one in cmd-socketio, fix related tests.
- Extract Verifier interfaces to common location to ensure interface compatibility between old and new verifier / verifier factory (both should implement same interface).
- Create new package cactus-verifier-client for common verifier related stuff, to prevent circular dependencies.
- Add verifier-factory to create verifiers by supplying it's ID only, based on initial configuration. Configuration is set in ctor, but can be read from a file as in cmd-socketio scenarious. VerifierFactory config should be compatible with existing ledgerPluginInfo.


outSH:
@petermetz 
As described in https://github.com/hyperledger/cactus/issues/1878, we need to include common "VerifierFactory" for release 1.0. I'm very sorry for such a short notice for feature like that, but it turned out to be needed just recently.

I'm still working on this task, but the core design proposal is complete, so please review in spare time and let me know what to clarify or improve :)

The interesting parts are:
- Implementation: [packages/cactus-api-client/src/main/typescript/verifier-factory.ts](https://github.com/hyperledger/cactus/pull/1879/files#diff-7464ad832e1a0f3580a109987fe8c084b16ba8e4c6a242b194761ded537b9c0c)
- Use sample: [packages/cactus-test-api-client/src/test/typescript/integration/verifier-factory.test.ts](https://github.com/hyperledger/cactus/pull/1879/files#diff-b5ae8042e05c1a2237c36a3232e2bce699f132360fceb8a8f14105563acfd5cb)

I will add more tests tomorrow.

packages/cactus-api-client/package.json
```diff
@@ -64,6 +64,8 @@
     "@hyperledger/cactus-core": "1.0.0-rc.3",
     "@hyperledger/cactus-core-api": "1.0.0-rc.3",
     "@hyperledger/cactus-plugin-consortium-manual": "1.0.0-rc.3",
+    "@hyperledger/cactus-plugin-ledger-connector-besu": "1.0.0-rc.3",
```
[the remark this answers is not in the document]
petermetz: @outSH Please make it a separate package then so that we reduce the probability of running into circular dependencies in the future or just the bloat that you also mentioned.

petermetz (changes requested):

packages/cactus-api-client/package.json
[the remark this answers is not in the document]
outSH: @petermetz Done

outSH:
There's another branch where I've adjusted electricity-trade to use new verifier factory (sort of sanity test). It breaks dockerization though, so it's not included here (need a simpler way to use local cactus packages, without pushing them to npm first). To start the sample just follow the readme as usual, open another shell, `cp config/* /etc/cactus/` and `node ./dist/www.js` to start BLP in this shell.

Branch - https://github.com/outSH/cactus/tree/electricity_trade_common_verifier

outSH:
@izuru0 @takeutak Please review

packages/cactus-verifier-client/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@hyperledger/cactus-verifier-client",
```
petermetz, line 2: @outSH May I recommend using a more specific package name like `cactus-verifier-client-besu-socketio`?

petermetz (changes requested):

packages/cactus-verifier-client/package.json
outSH: @petermetz This is a common interface for all ApiClients, not only besu one. For now it only supports besu and our "legacy-socketio" fabric, go-eth and sawtooth validators, but in near future we'll add support for quorum and possibly other validators, and we want to make them compatible with common verifier interface (introduced in cmd-socketio package).

petermetz: @outSH Understood, `cactus-verifier-client-legacy-socketio` then?

petermetz (changes requested):

packages/cactus-verifier-client/package.json
outSH: @petermetz Once again, this is not an interface for `legacy-socketio` (*-socketio* connectors), but in the end to all the connectors, including openapi ones. This is the verifier component documented in the whitepaper, but modified to support all connectors / validators.

petermetz:
> @petermetz Once again, this is not an interface for `legacy-socketio` (*-socketio* connectors), but in the end to all the connectors, including openapi ones. This is the verifier component documented in the whitepaper, but modified to support all connectors / validators. 

@outSH OK, cool.

The following is a not a change request, just something to ponder: Making a package that is intended to re-export literally every connector package (of which the number is to grow infinitely by design) will have a bundle size trending to infinity as well. This is especially important for browser based applications where the end user experience depends heavily on bundle size.

packages/cactus-verifier-client/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "composite": true,
+    "outDir": "./dist",
```
petermetz, line 5: @outSH Any chance that you could match the compiler options here used by the rest of the packages?

petermetz (changes requested):

outSH:
> Making a package that is intended to re-export literally every connector package (of which the number is to grow infinitely by design) will have a bundle size trending to infinity as well. This is especially important for browser based applications where the end user experience depends heavily on bundle size.

@petermetz Yes, that was my concern few messages above as well, but it's a requirement from one of the teams already using verifier interface. In the follow up PR I want to investigate possibility of importing apiclient modules dynamically, but this is not critical now when we only import 2 or 3 clients.

izuru0 (approved):
LGTM

packages/cactus-verifier-client/tsconfig.json
petermetz: > @petermetz Done, I've also added a single README just now.

@outSH Great, thank you very much for the adjustment!

petermetz:
> > Making a package that is intended to re-export literally every connector package (of which the number is to grow infinitely by design) will have a bundle size trending to infinity as well. This is especially important for browser based applications where the end user experience depends heavily on bundle size.
> 
> @petermetz Yes, that was my concern few messages above as well, but it's a requirement from one of the teams already using verifier interface. In the follow up PR I want to investigate possibility of importing apiclient modules dynamically, but this is not critical now when we only import 2 or 3 clients.

@outSH I guess in an ideal situation they would understand that using pre-1.0.0 software means that there will be breaking changes, but I know that's not how you roll over there so that's totally fair.

petermetz (approved):
@outSH LGTM, thank you for the updates and putting up with my nit-picking as usual!

PR #1892, opened 2022-03-03
test: jestify jwt-unprotected-endpoint-authz-ops-confirm

awadhana:
Migrated test from Tap to Jest

File Path:
packages/cactus-cmd-api-server/src/test/typescript/
integration/jwt-unprotected-endpoint-authz-ops-confirm.test.ts

This is a PARTIAL resolution to issue hyperledger#238

Signed-off-by: awadhana <awadhana2021@gmail.com>


petermetz (approved):
@awadhana LGTM

FYI: I updated the assertion to use pattern matching like this:

```typescript
    await expect(startPromise).rejects.toThrow(
      ApiServer.E_NON_EXEMPT_UNPROTECTED_ENDPOINTS,
    );
```

PR #1903, opened 2022-03-08
test(cmd-api-server): fix flaky runtime-plugin-imports-test

### LINKED ISSUE #1667: test(cmd-api-server): fix flaky runtime-plugin-imports test
This is a flaky test case meaning that it cannot be reliably reproduced, it only happens
every now and then which of course makes it harder to fix as well.

There is a clue I found based on the error message printed by Jest though:
https://stackoverflow.com/a/50793993

## Logs

[logs_12177.zip](https://github.com/hyperledger/cactus/files/7736558/logs_12177.zip)

```sh
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From packages/cactus-test-cmd-api-server/src/test/typescript/integration/runtime-plugin-imports.test.ts.

      346 |
      347 |       // eslint-disable-next-line @typescript-eslint/no-var-requires
    > 348 |       const pluginPackage = require(/* webpackIgnore: true */ packagePath);
          |                             ^
      349 |       const createPluginFactory = pluginPackage.createPluginFactory as PluginFactoryFactory;
      350 |       const pluginFactoryOptions: IPluginFactoryOptions = {
      351 |         pluginImportType: pluginImport.type,

      at ApiServer.instantiatePlugin (packages/cactus-cmd-api-server/src/main/typescript/api-server.ts:348:29)
          at runMicrotasks (<anonymous>)
      at ApiServer.initPluginRegistry (packages/cactus-cmd-api-server/src/main/typescript/api-server.ts:313:22)
      at ApiServer.getOrInitPluginRegistry (packages/cactus-cmd-api-server/src/main/typescript/api-server.ts:289:33)
```

Leeyoungone:
This is a flaky test case that can't be reliably reproduce. I've ran the test individually 40 times and all 40 times it has passed successfully. 

This error, however, seems to pop up only when the `yarn jest` command is run (when I ran the test individually). 

Based on the linked ["hint"](https://stackoverflow.com/questions/50793885/referenceerror-you-are-trying-to-import-a-file-after-the-jest-environment-has/50793993#50793993) from the issue, it seems like using `jest.useFakeTimers()` a good solution. However, using the `jest.useFakeTimers()` with`async/await` was not recommended.

However, I found a [different source](https://gist.github.com/apieceofbart/e6dea8d884d29cf88cdb54ef14ddbcc4#file-test-js-L58) that showed examples of how to use the `jest.useFakeTimers()`. Our original problem, however, is probably because the after the test environment is torn down, the second test file is trying to import files from the first environment, triggering the error ([explained here](https://github.com/facebook/jest/issues/11438#issuecomment-916711164)). So I've come to the solution of installing another node package: [node-cleanup](https://www.npmjs.com/package/node-cleanup) and running that within the afterAll at the very end of the test. 

My laptop can't take running the `yarn jest` script too much so I haven't seen it reproduce the error yet. (WOT ruhrow) 

_______________________edit below _________________________

So it turns out that this flaky test due to libraries trying to third libraries trying to run after the test runs is a simple fix that was documented [here](https://testing-library.com/docs/using-fake-timers/) where we just add the `useRealTimer` in the `afterEach`. Lol so sorry!! 

Fixes #1667

### DIFF
--- a/packages/cactus-test-cmd-api-server/src/test/typescript/integration/runtime-plugin-imports.test.ts
+++ b/packages/cactus-test-cmd-api-server/src/test/typescript/integration/runtime-plugin-imports.test.ts
@@ -16,6 +16,11 @@ const logLevel: LogLevelDesc = "TRACE";
 const testCase = "can import plugins at runtime (CLI)";
 describe(testCase, () => {
   let apiServer: ApiServer;
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
   test(testCase, async () => {
     const pluginsPath = path.join(
       __dirname, // start at the current file's path
@@ -67,5 +72,7 @@ describe(testCase, () => {
     ];
     await expect(apiServer.start()).not.toReject();
   });
-  afterAll(async () => await apiServer.shutdown());
+  afterAll(async () => {
+    await apiServer.shutdown();
+  });
 });


takeutak (approved):

petermetz (changes requested):
@Leeyoungone Great investigation I'm pretty sure you are on the right track. I read the documentation of node-cleanup and the way you are using it now seems to be a no-op to me because the purpose of the library is to provide hooks where the user (who is the person writing the Cactus code in this case) can do the "cleanup" logic themselves if they know that some allocated resource (TCP socket, file handle, database lock, etc.) is still being held onto. The point is, the node-cleanup library doesn't magically know what to clean up, it just gives you the opportunity to do so yourself (and right now what you technically do when you invoke the library's function is just saying that you want do the cleanup, but then you aren't).
So in that sense, I think this warrants a little more investigation, but you are definitely on the right track!

Leeyoungone:
@petermetz Thank you so much for the advice! I've dug into it a little more and I think that adding the `useRealTimers` within the `afterEach` should fix this flaky issue (as mentioned [here](https://testing-library.com/docs/using-fake-timers/)). My laptop with 8 GB of RAM isn't strong enough to withstand running all the `jest` tests to replicate this flaky issue. I might just have to Hail Mary this PR and hope this solves it.

package.json
```diff
@@ -120,6 +120,7 @@
     "lerna": "4.0.0",
     "lint-staged": "11.1.2",
     "make-dir-cli": "3.0.0",
+    "node-cleanup": "2.1.2",
```
petermetz, line 123: @Leeyoungone You are welcome and I'm glad the advice helped!
This diff is looking much better and it's mostly LGTM, but of course I have a nit about the forgotten `node-cleanup` dependency declaration that needs to be cleaned up (haha, what a great pun, I know).
TLDR: Remove this line from the diff and also make sure to do the same to the diff in the yarn.lock file

petermetz (changes requested):

package.json
Leeyoungone: lol the pun is so bad it's borderline good, I just got rid of the dependency so we should be good to go!!

petermetz: > lol the pun is so bad it's borderline good, I just got rid of the dependency so we should be good to go!!

@Leeyoungone Haha, thank you for taking care of it!

yarn.lock
```diff
@@ -16603,6 +16612,11 @@ node-addon-api@^4.3.0:
   resolved "https://registry.yarnpkg.com/node-addon-api/-/node-addon-api-4.3.0.tgz#52a1a0b475193e0928e98e0426a0d1254782b77f"
   integrity sha512-73sE9+3UaLYYFmDsFZnqCInzPyh3MqIwZO9cw58yIqAZhONrrabrYyYe3TuIqtIiOuTXVhsGau8hcrhhwSsDIQ==
 
+node-cleanup@2.1.2:
```
petermetz, line 16615: @Leeyoungone Please also throw away the changes you made to the yarn.lock file (these changes are made by yarn automatically when you update the package.json file but then if you revert your changes in the package.json file you must manually get rid of the corresponding automated changes in the lock file)

No need to go through it's diff line by line or anything just take it out of the diff completely since we know that dependencies were not changed at all by the PR.

petermetz (changes requested):

yarn.lock
Leeyoungone: Got it!! So sorry for all of it, I just fixed it :))

petermetz (changes requested):
@Leeyoungone Hmm, yarn.lock is still in diff if you look here: https://github.com/hyperledger/cactus/pull/1903/files
Did you forget to push your changes maybe?

petermetz (approved):
@Leeyoungone LGTM, thank you!

PR #1905, opened 2022-03-09
docs(whitepaper): stop using third party uml servers

### LINKED ISSUE #164: docs(whitepaper): stop using 3rd party Plant UML servers
**Describe the bug**

The remote, third party PlantUML servers should not be used when embedding images in the whitepaper, instead we should use locally pointing, relative URLs that reference image files.

The trade off is that now we'll have to store image files under version control which is not very good, but on the other hand it avoids the issue of the whitepaper having links embedded on (not so much but still) random third party domains.

**To Reproduce**

Open the `whitepaper.md` file and see the remote links.

**Expected behavior**

Local images are used instead.

cc: @jonathan-m-hamilton  @sfuji822  @takeutak

Leeyoungone:
Naming convention of the image files might not be consistent. I can change that but I'm not sure how nit picky we are going to be! 

Fixes #164


petermetz (approved):
> I'm not sure how nit picky we are going to be!

@Leeyoungone I'll nit pick as usual, but this time in a way that this can move along at the same time. ;-) 

The file names should not have uppercase characters in them, instead they all should be dash-case-file-names so that case sensitivity on file-systems does not come have a chance of hurting us. With that said, please create a follow-up issue to fix this later, the current PR I'm typing this on is good to go from my side.
Thank you :-)

takeutak:
@Leeyoungone Replacing third party uml server to png files is good idea.  But would you add their original puml files on the same directory?  If someone want to enhance the diagram, the puml files is helpful for them  (The png files themselves cannot be edited).

jagpreetsinghsasan:
@Leeyoungone as mentioned above, hosting the puml files in the same folder alongside the png files (you can refer https://github.com/hyperledger/cactus/tree/main/packages/cactus-plugin-ledger-connector-fabric/docs/architecture) is helpful in long run

Leeyoungone:
> With that said, please create a follow-up issue to fix this later, the current PR I'm typing this on is good to go from my side.

@petermetz Hey Peter, Thanks for the feedback! With the other comments below, I went ahead and altered the naming structure to match the hyphen usage. 

> But would you add their original puml files on the same directory?

@takeutak I apologize if this is an basic question, but is there a way to refer to the original puml code from the hyperlink? Here's what the whitepaper was previously filled in with: 
Screen Shot 2022-03-22 at 5 55 50 PM

As a general question, should I move the png files to a different file directory?

takeutak:
@Leeyoungone I thought it would be easy, so I asked.  But when I think again, it is difficult to restore a puml from a hyperlink.  I'm sorry to trouble you.
I'll approve your pull-request.  Anyway, I remember that the puml should be restored later.

@jagpreetsinghsasan Is it OK for you?  If you have some solution, I appreciate you share it with us.

takeutak (approved):

Issue #1925, opened 2022-03-14
fix(api-server): allow no authorization on socketio endpoints


outSH:
**Describe the bug**

`AuthorizationProtocol.NONE` is ignored for socketio endpoints, requests must always contain token which is problematic for tests.

**To Reproduce**
Run Besu connector as apiServer plugin and try use watchBlocks(), it will throw:
```
client/lib/transports/polling-xhr.js:231:10) {
      data: {
        message: 'no token provided',
        code: 'credentials_required',
        type: 'UnauthorizedError'
      }
    }
```

I will provide PR very soon.

PR #1926, opened 2022-03-14
fix(api-server): allow no authorization on socketio endpoints

outSH:
authorizeSocket middleware will be installed only if AuthorizationProtocol different than NONE was provided.

Closes:  #1925

Note
- This change work on my other local branch (not PR ready yet).
- This can be refactored, so that all authorization middleware creation would be handled by `AuthorizerFactory.createMiddleware`, but I didn't want to touch it without test coverage. I will probably introduce some tests with of socketio authorization as part of other PR, we can think about improvements then.

### DIFF
--- a/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
+++ b/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
@@ -732,14 +732,15 @@ export class ApiServer {
 
     this.wsApi.attach(this.httpServerApi, wsOptions);
 
-    const socketIoAuthorizer = authorizeSocket({
-      ...authzConf.socketIoJwtOptions,
-      onAuthentication: (decodedToken) => {
-        this.log.debug("Socket authorized OK: %o", decodedToken);
-      },
-    });
-
-    this.wsApi.use(socketIoAuthorizer as never);
+    if (authorizerO.isPresent()) {
+      const socketIoAuthorizer = authorizeSocket({
+        ...authzConf.socketIoJwtOptions,
+        onAuthentication: (decodedToken) => {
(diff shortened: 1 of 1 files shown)


outSH:
@petermetz @izuru0 @takeutak Please review

izuru0 (approved):
LGTM

packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
```diff
@@ -732,14 +733,15 @@ export class ApiServer {
...
 
-    const socketIoAuthorizer = authorizeSocket({
-      ...authzConf.socketIoJwtOptions,
-      onAuthentication: (decodedToken) => {
-        this.log.debug("Socket authorized OK: %o", decodedToken);
-      },
-    });
-
-    this.wsApi.use(socketIoAuthorizer as never);
+    if (authzProtocol != AuthorizationProtocol.NONE) {
```
petermetz, line 736: @outSH Thank you for the fix. One small nit is to swap the if condition with the same logic that the HTTP endpoints have to keep their behavior consistent in edge-cases

```suggestion
if (authorizerO.isPresent()) {
```

The presentness/emptiness of the optional there is driven by the logic within the authorizer factory which has the desired strict behavior that if you pass in some garbage value as authorization protocol it won't let you get away with it (it will throw) and therefore undefined behavior is less likely compared to us just saying that we'll carry on with whatever authorization protocol as long as it was not equal to `NONE`.

petermetz (changes requested):

packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
outSH: @petermetz Done

Issue #1927, opened 2022-03-14
refactor(quorum-multi-party-all-in-one) - enhance multiparty image


outSH:
- Current Quorum-MP WS endpoints are not available outside of docker, fix it.
- Use upstream npm setup package instead of contributor branch
- Test all members in healthcheck.
- Update readme.
- Add test tooling to setup multi party ledger in functional tests.
- Cleanup and simplify dockerfile
- Publish MP image to ghcr

I will provide PR soon

Issue #1881, opened 2022-02-24
ci: fix failing ghcr image publish jobs


petermetz:
**Describe the bug**

The problem is described in the email I'm pasting below:

---
On Thu, Dec 30, 2021 at 09:18 Somogyvari, Peter <[peter.somogyvari@accenture.com](mailto:peter.somogyvari@accenture.com)> wrote:
Hi Community Architects,

I'm at the finish line with migrating Cactus away from DockerHub to GHCR and hit on something that I think might be about org level permissions somehow (just a guess). Below is my attempt to clarify the symptoms, could you please advise if you think it's possible that some of our images are configure differently (in GH org settings that I cannot see with my permissions)
1.	The same workflow yaml files are working fine on the hyperledger-cicd org (where I am an admin)
2.	On the "production" org (e.g. [github.com/hyperledger](https://urldefense.proofpoint.com/v2/url?u=http-3A__github.com_hyperledger&d=DwMFaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=7LpfyaMR44exzQCWvxiAAfpAoMMDWWRNgSJjU7B7o4c&m=GNHUH_O2KisLywj9sWAXXkJJ27gcPyYxIYbwr2jKqRbFaWarX9-Lla7CKJ5Sf3DE&s=uyE8m9MsBrHSs5hxDwUs9F-CfmognOoPQzYhA1cEMg8&e=)) the same yaml files that are publishing containers are only partially working
a.	So yes, some of them work fine, but others fail to publish the built image with permission problems
3.	The error for the ones that fail the git push step is this: denied: permission_denied: write_package
4.	An example of a failing one is this image: [ghcr.io/hyperledger/cactus-besu-all-in-one](https://urldefense.proofpoint.com/v2/url?u=http-3A__ghcr.io_hyperledger_cactus-2Dbesu-2Dall-2Din-2Done&d=DwMFaQ&c=eIGjsITfXP_y-DLLX0uEHXJvU8nOHrUK8IrwNKOtkVU&r=7LpfyaMR44exzQCWvxiAAfpAoMMDWWRNgSJjU7B7o4c&m=GNHUH_O2KisLywj9sWAXXkJJ27gcPyYxIYbwr2jKqRbFaWarX9-Lla7CKJ5Sf3DE&s=wfRgrt62VYuc-N5re6qWcT_HIUV1V3BIqC9iZdSRG_E&e=)
[...]

PR #1928, opened 2022-03-14
refactor(quorum-multi-party-all-in-one): enhance quorum-mp image

outSH:
Current Quorum-MP WS endpoints are not available outside of docker, fix
it. Use upstream npm setup package instead of contributor branch. Test
all members in healthcheck. Update readme. Add test tooling to setup
multi party ledger in functional tests. Cleanup and simplify dockerfile.
Publish MP image to ghcr.

Closes: #1927

Notes
- I've added a new image publish in this PR, but there's ongoing issue with package publish - https://github.com/hyperledger/cactus/issues/1881
- Since I'm not sure when the image will be published, I've used my personal ghcr for now, since I'm already using it in some functional tests on another branch. I will change it to hyperledger ghcr when the publish completes correctly.

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/quorum/quorum-mp-test-ledger.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/quorum/quorum-mp-test-ledger.ts
@@ -1,0 +1,255 @@
+import { EventEmitter } from "events";
+import Docker, { Container, ContainerCreateOptions } from "dockerode";
+import {
+  Bools,
+  Logger,
+  LoggerProvider,
+  LogLevelDesc,
+} from "@hyperledger/cactus-common";
+import { ITestLedger } from "../i-test-ledger";
+import { Containers } from "../common/containers";
+
+export interface IQuorumMultiPartyTestLedgerOptions {
+  readonly containerImageName?: string;
+  readonly containerImageVersion?: string;
+  readonly logLevel?: LogLevelDesc;
+  readonly emitContainerLogs?: boolean;
+  readonly envVars?: string[];
+  // For test development, attach to ledger that is already running, don't spin up new one
+  readonly useRunningLedger?: boolean;
+}
+
+export class QuorumMultiPartyTestLedger implements ITestLedger {
+  public readonly containerImageName: string;
+  public readonly containerImageVersion: string;
+  private readonly logLevel: LogLevelDesc;
+  private readonly emitContainerLogs: boolean;
+  private readonly useRunningLedger: boolean;
+  private readonly envVars: string[];
+
+  private readonly log: Logger;
+  public container: Container | undefined;
+  public containerId: string | undefined;
+
+  constructor(public readonly options: IQuorumMultiPartyTestLedgerOptions) {
+    // @todo Replace with hyperledger ghcr link when available
+    this.containerImageName =
+      options?.containerImageName ||
+      "ghcr.io/hyperledger/cactus-quorum-multi-party-all-in-one";
+
+    this.containerImageVersion =
+      options?.containerImageVersion || "2022-03-30--1928";
+
+    this.logLevel = options?.logLevel || "info";
+
+    this.emitContainerLogs = Bools.isBooleanStrict(options.emitContainerLogs)
+      ? (options.emitContainerLogs as boolean)
+      : true;
+
+    this.useRunningLedger = Bools.isBooleanStrict(options.useRunningLedger)
+      ? (options.useRunningLedger as boolean)
+      : false;
+
+    this.envVars = options?.envVars || [];
+
+    this.log = LoggerProvider.getOrCreate({
+      level: this.logLevel,
+      label: "quorum-multi-party-test-ledger",
+    });
+  }
+
+  public get fullContainerImageName(): string {
+    return [this.containerImageName, this.containerImageVersion].join(":");
+  }
+
+  public async start(omitPull = false): Promise<Container> {
+    if (this.useRunningLedger) {
+      this.log.info(
+        "Search for already running Quorum Test Ledger because 'useRunningLedger' flag is enabled.",
+      );
+      this.log.info(
+        "Search criteria - image name: ",
+        this.fullContainerImageName,
+        ", state: running",
+      );
+      const containerInfo = await Containers.getByPredicate(
+        (ci) =>
+          ci.Image === this.fullContainerImageName && ci.State === "running",
+      );
+      const docker = new Docker();
+      this.containerId = containerInfo.Id;
+      this.container = docker.getContainer(this.containerId);
+      return this.container;
+    }
+
+    if (this.container) {
+      await this.container.stop();
+      await this.container.remove();
+      this.container = undefined;
+      this.containerId = undefined;
+    }
+
+    if (!omitPull) {
+      await Containers.pullImage(
+        this.fullContainerImageName,
+        {},
+        this.logLevel,
+      );
+    }
+
+    const createOptions: ContainerCreateOptions = {
+      ExposedPorts: {
+        "8545/tcp": {}, // HTTP RPC
+        "8546/tcp": {}, // WS RPC
+        "20000/tcp": {}, // Member1 HTTP RPC
+        "20001/tcp": {}, // Member1 WS RPC
+        "9081/tcp": {}, // Member1 Tessera
+        "20002/tcp": {}, // Member2 HTTP RPC
+        "20003/tcp": {}, // Member2 WS RPC
+        "9082/tcp": {}, // Member2 Tessera
+        "20004/tcp": {}, // Member3 HTTP RPC
+        "20005/tcp": {}, // Member3 WS RPC
+        "9083/tcp": {}, // Member3 Tessera
+      },
+
+      Env: this.envVars,
+
+      HostConfig: {
+        PublishAllPorts: true,
+        Privileged: true,
+      },
+    };
+
+    return new Promise<Container>((resolve, reject) => {
+      const docker = new Docker();
+      const eventEmitter: EventEmitter = docker.run(
+        this.fullContainerImageName,
+        [],
+        [],
+        createOptions,
+        {},
+        (err: any) => {
+          if (err) {
+            reject(err);
+          }
+        },
+      );
+
+      eventEmitter.once("start", async (container: Container) => {
+        this.container = container;
+        this.containerId = container.id;
+
+        if (this.emitContainerLogs) {
+          const fnTag = `[${this.fullContainerImageName}]`;
+          await Containers.streamLogs({
+            container: this.container,
+            tag: fnTag,
+            log: this.log,
+          });
+        }
+
+        try {
+          await Containers.waitForHealthCheck(this.containerId);
+          resolve(container);
+        } catch (ex) {
+          this.log.error(ex);
+          reject(ex);
+        }
+      });
+    });
+  }
+
+  public stop(): Promise<unknown> {
+    if (this.useRunningLedger) {
+      this.log.info("Ignore stop request because useRunningLedger is enabled.");
+      return Promise.resolve();
+    } else if (this.container) {
+      return Containers.stop(this.container);
+    } else {
+      return Promise.reject(
+        new Error(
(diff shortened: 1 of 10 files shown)


outSH:
@petermetz In the meantime I've added `useRunningLedger` flag to test-tooling, that make test development and experimentation much faster.

packages/cactus-test-tooling/src/main/typescript/quorum/quorum-mp-test-ledger.ts
```diff
@@ -0,0 +1,255 @@
...
+  constructor(public readonly options?: IQuorumMultiPartyTestLedgerOptions) {
+    // @todo Replace with hyperledger ghcr link when available
+    this.containerImageName =
+      options?.containerImageName ||
+      "ghcr.io/outsh/cactus-quorum-multi-party-all-in-one";
+    this.containerImageVersion =
+      options?.containerImageVersion || "2022-03-14-65afd20";
+    this.logLevel = options?.logLevel || "info";
+
+    if (options?.emitContainerLogs != undefined) {
```
petermetz, line 42: @outSH Relying here only on the `!= undefined` check will let `this.emitContainerLogs` take up fun values like `{}` or `[]` or `0` or `"0"` (and many more, yaay Javascript...)
I recommend leveraging the strict boolean check from the `Bools` class similar to how it's done in the API server (and other places)
https://github.com/hyperledger/cactus/blob/release-v1.0.0/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts#L113

The above will ensure that at least if garbage values come in as the constructor arg, we still default to either true or false (depending on what you want).
Another option would be (which is preferred if you don't mind making your parameter mandatory) to do an assertion with the `Checks` class that throws if the input parameter is anything but a `true` or a `false` with something like `Checks.truthy(Bools.isBooleanStrict(options.emitContainerLogs))` but this only works if all the parameters are mandatory so that's why the other places in the code mostly don't use this.

packages/cactus-test-tooling/src/main/typescript/quorum/quorum-mp-test-ledger.ts
```diff
@@ -0,0 +1,255 @@
...
+      this.emitContainerLogs = options.emitContainerLogs;
+    } else {
+      this.emitContainerLogs = true;
+    }
+
+    if (options?.useRunningLedger != undefined) {
```
petermetz, line 48: @outSH Relying here only on the `!= undefined` check will let `this.emitContainerLogs` take up fun values like `{}` or `[]` or `0` or `"0"` (and many more, yaay Javascript...)
I recommend leveraging the strict boolean check from the `Bools` class similar to how it's done in the API server (and other places)
https://github.com/hyperledger/cactus/blob/release-v1.0.0/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts#L113

The above will ensure that at least if garbage values come in as the constructor arg, we still default to either true or false (depending on what you want).
Another option would be (which is preferred if you don't mind making your parameter mandatory) to do an assertion with the `Checks` class that throws if the input parameter is anything but a `true` or a `false` with something like `Checks.truthy(Bools.isBooleanStrict(options.emitContainerLogs))` but this only works if all the parameters are mandatory so that's why the other places in the code mostly don't use this.

packages/cactus-test-tooling/src/main/typescript/quorum/quorum-mp-test-ledger.ts
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  public get fullContainerImageName(): string {
+    return [this.containerImageName, this.containerImageVersion].join(":");
+  }
+
+  public async start(omitPull = false): Promise<Container> {
+    if (this.useRunningLedger) {
+      this.log.info(
+        "Search for already running Quorum Test Ledger because 'useRunningLedger' flag is enabled.",
```
petermetz, line 69: @outSH Just a recommendation: It could help people debug what went wrong here if we included in the log message the fully qualified container image name and the status===running predicate so that if someone tries to figure out from the error message what went wrong they won't only see that the predicate did not yield any results but also what the predicate was looking for (which would then allow the person looking at the logs to understand that they need to start the container manually and make sure that it actually runs)

packages/cactus-test-tooling/src/main/typescript/quorum/quorum-mp-test-ledger.ts
```diff
@@ -0,0 +1,255 @@
...
+        (ci) =>
+          ci.Image === this.fullContainerImageName && ci.State === "running",
+      );
+      const docker = new Docker();
+      this.containerId = containerInfo.Id;
+      this.container = docker.getContainer(this.containerId);
+      return this.container;
+    }
+
+    this.log.info(
```
petermetz, line 81: @outSH Continuing on my comment above, this message is good in that it prints the image name, but it only reaches this point in the execution on the happy path not when people would need to figure out what's going wrong.

tools/docker/quorum-multi-party-all-in-one/Dockerfile
```diff
@@ -1,116 +1,54 @@
...
-RUN augtool 'set /files/etc/ssh/sshd_config/AuthorizedKeysFile ".ssh/authorized_keys /etc/authorized_keys/%u"'
-RUN augtool 'set /files/etc/ssh/sshd_config/PermitRootLogin yes'
-RUN augtool 'set /files/etc/ssh/sshd_config/PasswordAuthentication no'
-RUN augtool 'set /files/etc/ssh/sshd_config/PermitEmptyPasswords no'
-RUN augtool 'set /files/etc/ssh/sshd_config/Port 22'
-RUN augtool 'set /files/etc/ssh/sshd_config/LogLevel DEBUG2'
-RUN augtool 'set /files/etc/ssh/sshd_config/LoginGraceTime 10'
-# Create the server's key - without this sshd will refuse to start
-RUN ssh-keygen -A
+################################
```
petermetz, line 1: @outSH The new image supresses the logs from the inner containres that print out this very useful list of hosts & ports for debugging & development purposes, could you please make sure that the output is still there in the new image. If you run the old one, it prints these at the end when the booting of the network has succeeded, I'm including an example of it here:

```sh
*************************************
Quorum Dev Quickstart 
*************************************
2022-03-30 19:42:08,192 DEBG 'quorum-network' stdout output:
*************************************
Quorum Dev Quickstart 
*************************************

----------------------------------
List endpoints and services
----------------------------------
2022-03-30 19:42:10,623 DEBG 'quorum-network' stdout output:
----------------------------------
List endpoints and services
----------------------------------

JSON-RPC HTTP service endpoint                 : http://localhost:8545
JSON-RPC WebSocket service endpoint            : ws://localhost:8546
Web block explorer address                     : http://localhost:25000/
2022-03-30 19:42:10,624 DEBG 'quorum-network' stdout output:
JSON-RPC HTTP service endpoint                 : http://localhost:8545
JSON-RPC WebSocket service endpoint            : ws://localhost:8546
Web block explorer address                     : http://localhost:25000/

Prometheus address                             : http://localhost:9090/graph
2022-03-30 19:42:11,910 DEBG 'quorum-network' stdout output:
Prometheus address                             : http://localhost:9090/graph

Cakeshop toolkit address                       : http://localhost:8999
2022-03-30 19:42:13,285 DEBG 'quorum-network' stdout output:
Cakeshop toolkit address                       : http://localhost:8999

Grafana address                                : http://localhost:3000/d/a1lVy7ycin9Yv/goquorum-overview?orgId=1&refresh=10s&from=now-30m&to=now&var-system=All

For more information on the endpoints and services, refer to README.md in the installation directory.
****************************************************************
2022-03-30 19:42:14,553 DEBG 'quorum-network' stdout output:
Grafana address                                : http://localhost:3000/d/a1lVy7ycin9Yv/goquorum-overview?orgId=1&refresh=10s&from=now-30m&to=now&var-system=All

For more information on the endpoints and services, refer to README.md in the installation directory.
****************************************************************

```

petermetz (changes requested):

packages/cactus-test-tooling/src/main/typescript/quorum/quorum-mp-test-ledger.ts
outSH: @petermetz Done.

packages/cactus-test-tooling/src/main/typescript/quorum/quorum-mp-test-ledger.ts
outSH: @petermetz Done.

tools/docker/quorum-multi-party-all-in-one/Dockerfile
outSH: @petermetz These messages come directly from `Quorum Dev Quickstart` which is started the same way as before, I've double checked and the logs are still there.

packages/cactus-test-tooling/src/main/typescript/quorum/quorum-mp-test-ledger.ts
```diff
@@ -0,0 +1,255 @@
...
+  private readonly emitContainerLogs: boolean;
+  private readonly useRunningLedger: boolean;
+  private readonly envVars: string[];
+
+  private readonly log: Logger;
+  public container: Container | undefined;
+  public containerId: string | undefined;
+
+  constructor(public readonly options?: IQuorumMultiPartyTestLedgerOptions) {
+    // @todo Replace with hyperledger ghcr link when available
```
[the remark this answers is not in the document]
outSH: @petermetz There were no changes needed in Dockerfile, so I've already replaced my ghcr in sources with image you've published.

tools/docker/quorum-multi-party-all-in-one/Dockerfile
petermetz: @outSH Weird stuff, let's try to figure it out together then because I triple checked it and it's not there. Here's the command I ran and the log file. I did a full text search in the log file and the endpoint information is not in there. 

What command did you run to build the image? What command to run it? Can you also attach the logs so we can compare?
Maybe I didn't build the image right/same way you did or something.

`docker run --rm --privileged ghcr.io/hyperledger/cactus-quorum-multi-party-all-in-one:2022-03-30--1928 &> container.log`

[container.log](https://github.com/hyperledger/cactus/files/8399720/container.log)

petermetz (changes requested):
^^ Still debugging the endpoint info logs

tools/docker/quorum-multi-party-all-in-one/Dockerfile
outSH: @petermetz Ooh, I see, this is weird indeed...

Basically the logs appear again when you add `-ti` flag, without this it complains about `No value for $TERM and no -T specified`. I think it might be caused by the fact that I use newer sample app (from npm, not git). One difference I see is that the new script writes it's messages to stderr instead of stdout. Nevertheless, it should work the same, both streams should be piped through supervisord and it didn't cause any problems before, so I have no idea what's happening now, or how to solve it without digging into quorum code :/

petermetz: @outSH A workaround is fine (instead of fixing it upstream in the quorum repo itself) but we need to make sure that people see the endpoint list when using the image as documented/instructed. 

From what you are saying my guess is that you could get away with 
1. Updating readme of the image so that every mention of the `docker run ...` command in there contains the `--tty` flag (leave out the --interactive flag in my opinion it's all about having `--tty` on, the `-i` flag makes no difference in this case so less confusing if not included in the examples)
2. Add a little note in there explaining that the `--tty` is required by the underlying software that we are running inside the container, without the `--tty` flag the endpoint list will not be printed in the logs of the container. --- This should preemptively take care of people getting confused or struggling to work with the image.

petermetz (changes requested):
^^ Added a comment above.

tools/docker/quorum-multi-party-all-in-one/Dockerfile
outSH: @petermetz Done, I've updated the README

petermetz (approved):
> @petermetz Done, I've updated the README  

@outSH Great, thank you so much for the updates! LGTM

Issue #1897, opened 2022-03-04
Building docker images with dependencies to local monorepo packages question


outSH:
**Is your feature request related to a problem? Please describe.**
I've noticed that most Dockerfiles are using already published npm packages, instead of using binaries built from local repository. I see the benefit for "production" images, but this seem problematic during development. For instance, what shall I do when I've made some changes to `packages/A` and `packages/B`, and I want to use them in `examples/electricity-trade` container? I don't see any solution that isn't laborious or intrusive..

**Describe alternatives you've considered**
- Copy dependent monorepo packages `dist` and `packages.json` to the container? This could work but `/dist` is removed from docker context by root level dockerignore (for obvious performance reasons).
- Copy only dependent monorepo package sources, build them inside the container, use multistage build to copy relevant files to destination container?
- I though webpack could produce a single standalone app with all the dependencies included, but apparently it `node-config` package doesn't work with webpack to well :/
- Maybe introduce an image that would build an entire monorepo (i.e. `npm run configure`), and then use this container as base in multistage build?

@petermetz Could you please share how would you produce an image that has dependencies to development versions of local packages? Any suggestions? Haven't work with monorepo all that much :)

RafaelAPB:
Wouldn't locally running the changed plugin do the trick?

outSH:
@RafaelAPB Well, not if I want to test the docker setup itself :) For instance, `cactus-plugin-ledger-connector-quorum` doesn't work at the moment because of a missing package. This could not be tested until the release of @hyperledger/cactus-plugin-ledger-connector-quorum to npm, which is waaaaay to late for catching these kind of errors.

Also, I'd like to use docker during the development phase, not after. For electricity trade or any other sample app, it's convenient to work on docker-compose to setup all the components. Also, when working on a feature, I want to be able to share an image with feature proposal to a team requesting it for test, and not make them clone, build entire repository, manually setup the component, etc..

outSH:
@petermetz Can you share your opinion on this?

outSH:
I guess we have to use released packages in docker images, closing.

PR #1933, opened 2022-03-17
feat(quorum-connector): implement validator interface on go-quorum-connector

outSH:
- Add three new endpoints to quorum ledger connector achieve legacy verifier compatibility.
  - `InvokeRawWeb3EthContractEndpoint` can be used to form any call to deployed contract.
  - `InvokeRawWeb3EthMethodEndpoint` can be used to call any web3.eth function. Both are marked as low-level functions, should be used only when there's no designated endpoint for given functionality yet.
  - `WatchBlocksV1Endpoint` can be used to monitor new block headers / data from the ledger. Type of the output is determined from input option flag.
- Extend `QuorumApiClient` to support Verifier interface, that is: block monitoring, and sending sync/async requests. Sending requests is marked as deprecated, because user can use direct REST calls from generated ApiClient, nevertheless this API was requested by one of the teams.
- Added functional tests for two new, request based endpoints.
- Moved verifier-besu integration test to besu-test package.
- Added verifier-quorum integration test, it supplements direct endpoint tests and provides a reference for API usage.
- Added support for `QuorumApiClient` in Verifier.

Closes: https://github.com/hyperledger/cactus/issues/1604

### Notes
- I can't make CodeQL to cooperate, it reports dynamic call errors, even though I explicitly check if methods are defined on a contract / object (also assert they type is a function). Tried different methods to make the scanner notice it but did not succeed. Please double check and mark as false positive if possible.

### DIFF
--- a/packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
+++ b/packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
@@ -147,8 +147,10 @@
             "EthContractInvocationType": {
                 "type": "string",
                 "enum": [
-                    "SEND",
-                    "CALL"
+                    "send",
+                    "call",
+                    "encodeABI",
+                    "estimateGas"
                 ]
             },
             "SolidityContractJsonArtifact": {
@@ -507,7 +509,7 @@
                     }
                 }
             },
-            "DeployContractSolidityBytecodeJsonObjectV1Request": { 
+            "DeployContractSolidityBytecodeJsonObjectV1Request": {
                 "type": "object",
                 "required": [
                     "web3SigningCredential",
@@ -534,7 +536,7 @@
                         "default": 60000,
                         "nullable": false
                     },
-                    "contractJSON": { 
+                    "contractJSON": {
                         "$ref": "#/components/schemas/ContractJSON",
                         "description": "For use when not using keychain, pass the contract in as this variable",
                         "nullable": false
@@ -713,7 +715,7 @@
                         "default": 60000,
                         "nullable": false
                     },
-                    "contractJSON": { 
+                    "contractJSON": {
                         "$ref": "#/components/schemas/ContractJSON",
                         "description": "For use when not using keychain, pass the contract in as this variable",
                         "nullable": false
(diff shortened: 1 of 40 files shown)


outSH:
I've updated quorum ledger plugin readme as requested by @izuru0:
- Added section on `QuorumApiClient` shortly describing it's capabilities.
- Updated running container commands.

Running containers doesn't work though because of missing ` run-time-error` package. I've added it to `package.json`, can't test it because of reasons described in https://github.com/hyperledger/cactus/issues/1897.

I've also changed dockerfile to use upstream `ghcr.io/hyperledger/cactus-cmd-api-server:v1.0.0` instead of local `cactus-api-server:latest`

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -654,4 +710,80 @@ export class PluginLedgerConnectorQuorum
...
+        Object.prototype.hasOwnProperty.call(looseWeb3Eth, args.methodName) &&
+        typeof looseWeb3Eth[args.methodName] === "function"
+      ) {
+        const web3Method = looseWeb3Eth[args.methodName];
+
+        const web3Response: Promise<unknown> = args.params
+          ? web3Method(...args.params)
+          : web3Method();
+
+        web3Response
```
petermetz, line 738: @outSH Could this be made easier to read with an `await` and a try/catch?

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -654,4 +710,80 @@ export class PluginLedgerConnectorQuorum
...
+          .then((result: any) => {
+            this.log.debug("invokeRawWeb3EthMethod response (OK):", result);
+            return resolve(result);
+          })
+          .catch((err) => {
+            this.log.warn("invokeRawWeb3EthMethod response (ERROR):", err);
+            rejects(err);
+          });
+      } else {
+        rejects(new Error(`No method "${args.methodName}" in web3.eth`));
```
petermetz, line 748: @outSH Instead of returning two promises, could this be simplified to use the implicitly returned promise and then this `rejects` could be just a throw and there would be less identation/code in general?

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -654,4 +710,80 @@ export class PluginLedgerConnectorQuorum
...
+    return new Promise((resolve, rejects) => {
+      this.log.debug("invokeRawWeb3EthMethod input:", JSON.stringify(args));
+
+      Checks.nonBlankString(
+        args.methodName,
+        "web3.eth method string must not be empty",
+      );
+
+      const looseWeb3Eth = this.web3.eth as any;
+      if (
```
petermetz, line 728: @outSH If you turn around the if condition's check, you could throw in the negative case and the rest of the logic does not have to be indented at all while the behavior remains the same. (less indentation helps with having less line wraps which in turn increases readability IMO)

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
```diff
@@ -937,6 +1315,74 @@
...
+        },
+        "/api/v1/plugins/@hyperledger/cactus-plugin-ledger-connector-quorum/invoke-raw-web3eth-method": {
+            "post": {
+                "x-hyperledger-cactus": {
+                    "http": {
+                        "verbLowerCase": "post",
+                        "path": "/api/v1/plugins/@hyperledger/cactus-plugin-ledger-connector-quorum/invoke-raw-web3eth-method"
+                    }
+                },
+                "operationId": "invokeWeb3EthMethodV1",
```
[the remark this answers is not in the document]
izuru0: @petermetz isn't it already camelCase?

izuru0 (approved):
LGTM

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
outSH: @petermetz Yeah, while writing the function I was trying to "trick" CodeQL into  seeing that this execution path runs only when method is verified, but it didn't help :( 

I've refactored this function into smaller and less convoluted one.

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
outSH: @petermetz See above

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
outSH: @petermetz See above

outSH:
@petermetz Thanks for the detailed review, appreciate it :) I've added fixes in a separate commit for easier review [fix: fixing code review issues](https://github.com/hyperledger/cactus/pull/1933/commits/675a83e869a9a6186e8b876663f031402c39ac42), will squash it once the PR is ready to merge.

packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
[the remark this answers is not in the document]
petermetz: @izuru0 @outSH My bad, apologies. Sleep depravation might be making me hallucinate... :-)

packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
```diff
@@ -937,6 +1315,74 @@
...
+        },
+        "/api/v1/plugins/@hyperledger/cactus-plugin-ledger-connector-quorum/invoke-raw-web3eth-contract": {
+            "post": {
+                "x-hyperledger-cactus": {
+                    "http": {
+                        "verbLowerCase": "post",
+                        "path": "/api/v1/plugins/@hyperledger/cactus-plugin-ledger-connector-quorum/invoke-raw-web3eth-contract"
+                    }
+                },
+                "operationId": "InvokeRawWeb3EthContractV1",
```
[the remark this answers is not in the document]
petermetz: > @petermetz Good catch, thanks! Fixed

@outSH No worries, thank you for the fix!

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
petermetz: > @petermetz Yeah, while writing the function I was trying to "trick" CodeQL into seeing that this execution path runs only when method is verified, but it didn't help :(
> 
> I've refactored this function into smaller and less convoluted one.

@outSH Great, thank you so much!

packages/cactus-plugin-ledger-connector-quorum/src/main/json/openapi.json
```diff
@@ -147,8 +147,10 @@
             "EthContractInvocationType": {
                 "type": "string",
                 "enum": [
-                    "SEND",
-                    "CALL"
+                    "send",
```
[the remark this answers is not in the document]
petermetz: @outSH Thank you! FYI: I created a label for marking breaking changes, so we can still create an issue to track a change like this and then make sure that we get that change done as part of the v2.0.0 release in the future so not all is lost, but definitely a little more complicated now that we issued the 1.0.0 release.
The label that you can use for what I mentioned above: https://github.com/hyperledger/cactus/labels/Breaking_V1

petermetz:
> @petermetz Thanks for the detailed review, appreciate it :) I've added fixes in a separate commit for easier review [fix: fixing code review issues](https://github.com/hyperledger/cactus/pull/1933/commits/675a83e869a9a6186e8b876663f031402c39ac42), will squash it once the PR is ready to merge.

@outSH My pleasure and thank you for the separate commits, that does make it easier! It's LGTM now so please squash+rebase  prior merging.
If you need any help with addressing CodeQL errors just let me know we can work together to figure it out.

petermetz (approved):
@outSH LGTM

outSH:
@petermetz Thanks! :)

As for CodeQL, I've wrote an explanation in notes already that I don't think I can solve them :( I've seen on my branch that similar issue already occurs here and in some other modules, so I assume it was ignored there. The problem is most probably false positive at this point, since all execution paths are checked (please review the affected functions in the connector to be sure)

outSH:
@petermetz OK, I've rebased and cleared the commit history. Can you review the CodeQL related message I wrote earlier and merge if possible?

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -654,4 +711,67 @@ export class PluginLedgerConnectorQuorum
...
+      contract,
+      args.contractMethod,
+    );
+    if (!isSafeToCall) {
+      throw new RuntimeError(
+        `Invalid method name provided in request. ${args.contractMethod} does not exist on the Web3 contract object's "methods" property.`,
+      );
+    }
+
+    return contract.methods[args.contractMethod](...contractMethodArgs)[
```
petermetz, line 773: @outSH I'll dismiss this one as a false-positive because the `isSafeToCallContractMethod` call is right above it so AFAICT this is all good.

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -654,4 +711,67 @@ export class PluginLedgerConnectorQuorum
...
+    args: InvokeRawWeb3EthMethodV1Request,
+  ): Promise<any> {
+    this.log.debug("invokeRawWeb3EthMethod input:", JSON.stringify(args));
+
+    Checks.nonBlankString(
+      args.methodName,
+      "web3.eth method string must not be empty",
+    );
+
+    const looseWeb3Eth = this.web3.eth as any;
```
petermetz, line 727: @outSH Use the `isSafeToCallContractMethod` here as well and then no need to cast into any. Once that's done, I can dismiss the warning (if still present) as a false-positive. If you want more comprehensive checks within the `isSafeToCallContractMethod` function, feel free to add more code in there.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
outSH: @petermetz `isSafeToCallContractMethod` is meant to work with `Contract` instances only (i.e. it checks `methods` member field), and in `invokeRawWeb3EthMethod` I check existance of a method on a `web3.Eth` object. Also, casting to any is necessary here in order to get to a method with string indexing later on. It's not necessary with contracts, since `contract.methods` is already type of `any` :) `web3.Eth` instance that is used here doesn't define string indexing, so without a cast it will report an error.

I've noticed there's duplicated logic of checking method on object though, so I've extracted new function - `isSafeToCallObjectMethod`, that is used by `invokeRawWeb3EthMethod` and `isSafeToCallContractMethod`, so the duplication is eliminated.

petermetz: @outSH Ah, my bad, sorry. Should've been reading the code more carefully. Cool, thanks for the fix, I'll dismiss it as a false positive then and we are good to go.

petermetz (approved):
@outSH LGTM, thank you!

PR #1938, opened 2022-03-20
feat(plugin-odap): addition of client endpoints

### LINKED ISSUE #1939: feat(odap-plugin): implementation of client endpoints
In order to fully implement the Open Digital Asset Protocol (ODAP) and its crash recovery, there is the need to implement the client endpoints. At the moment, only the server endpoints are defined and the client side runs as a single function. In case of crash, there is no possibility of resuming the execution of the protocol in a specific point of the protocol without the definition of endpoints on both sides.

Definition of the client endpoints.

AndreAugusto11:
Implementation of client endpoints in ODAP plugin.

The changes can be summarized as follows:
- Implementation of client endpoints
- Adapted and added more tests (now divided into unit and integration tests)
- Addition of more checks regarding the validity of the messages received on the server side

closes #1939


takeutak (approved):

RafaelAPB:
LGTM.

RafaelAPB:
@AndreAugusto11  we should also think of providing a way to persist session data in a similar way to publishOdapProof (a flag is given to storeSessionData or set in the gateway constructor to use IPFS or the local store for log saving). This can be made by configuring the logging profile; to make things simple the log profile can still be a string, stating the support (IPFS vs local).

petermetz:
> @AndreAugusto11 we should also think of providing a way to persist session data in a similar way to publishOdapProof (a flag is given to storeSessionData or set in the gateway constructor to use IPFS or the local store for log saving). This can be made by configuring the logging profile; to make things simple the log profile can still be a string, stating the support (IPFS vs local).

@AndreAugusto11  @RafaelAPB 

Putting my recommendation down here for KnexJS and SQLite3. I believe it to have close to optimal trade-offs compared to rolling your own format/storage/etc.

See the relevant examples of their documentation below. If you go with this then you also have the upside of me being able to help much more compared to other options ;-)
Source: http://knexjs.org/

---

Note: When you use the SQLite3 or Better-SQLite3 adapter, there is a filename required, not a network connection. For example:

```
const knex = require('knex')({
  client: 'sqlite3', // or 'better-sqlite3'
  connection: {
    filename: "./mydb.sqlite"
  }
});
```
Note: You can also run either SQLite3 or Better-SQLite3 with an in-memory database by providing :memory: as the filename. For example:

```
const knex = require('knex')({
  client: 'sqlite3', // or 'better-sqlite3'
  connection: {
    filename: ":memory:"
  }
});
```

Note: When you use the SQLite3 adapter, you can set flags used to open the connection. For example:

```
const knex = require('knex')({
  client: 'sqlite3',
  connection: {
    filename: "file:memDb1?mode=memory&cache=shared",
    flags: ['OPEN_URI', 'OPEN_SHAREDCACHE']
  }
});
```

petermetz (approved):
@AndreAugusto11 LGTM, thank you very much for the work you put in!

PR #1940, opened 2022-03-22
feat(keychain-aws-sm): bootstrap readme.md

### LINKED ISSUE #968: feat(keychain-aws-sm): bootstrap README.md
### Description
As a developer, I want to update the README.md describing the plugin and examples for its usage.

### Acceptance Criteria
1. README.md present at packages/cactus-plugin-keychain-aws-sm/README.md should be updated with the plugin description and example usage.

jagpreetsinghsasan:
## Commit to be reviewed

feat(keychain-aws-sm): bootstrap readme.md

	Primary Changes
	-----------------------
	1. Updated the README.md located at packages/cactus-plugin-keychain-aws-sm/README.md
	2. The aws secret manager plugin now includes the prometheus metrics exporter integration
	3. OpenApi spec now has api endpoint for getting the prometheus metrics

Fixes #968


takeutak (approved):

packages/cactus-plugin-keychain-aws-sm/docs/architecture/delete-keychain-entry-endpoint.puml
```diff
@@ -0,0 +1,31 @@
...
+box "Hyperledger Cactus" #LightGray
+entity "API Client" as apic
+entity "API Server" as apis
+end box
+
+box "AWS SM Connector" #LightGreen
+database "AWS SM" as awssm
+end box
+
+a --> apic : Tx DeleteKeychainEntryV1
```
petermetz, line 23: @jagpreetsinghsasan Should we call this something else other than `Tx` (implies transaction to me). I'm wondering because if it's a transaction then it would mean that it can be rolled back for example, but I think the keychain implementation right now is closer to "fire and forget" from the perspective of the HTTP request sender because they cannot combine their request with other requests to achieve some sort of "all or nothing" set of operations. They just send their request and then the deletion will either succeed or fail and that's it.
The same question applies to the other 3 diagrams as well. Please let me know your thoughts!

packages/cactus-plugin-keychain-aws-sm/README.md
```diff
@@ -1,3 +1,164 @@
...
+  - [1. Usage](#1-usage)
+    - [1.1. Installation](#11-installation)
+    - [1.2. Using as a Library](#12-using-as-a-library)
+    - [1.3. Using via the API Client](#13-using-via-the-api-client)
+  - [2. Architecture](#2-architecture)
+    - [2.1. set-keychain-entry-endpoint](#21-set-keychain-entry-endpoint)
+    - [2.2. get-keychain-entry-endpoint](#22-get-keychain-entry-endpoint)
+    - [2.3. has-keychain-entry-endpoint](#23-has-keychain-entry-endpoint)
+    - [2.4. delete-keychain-entry-endpoint](#24-delete-keychain-entry-endpoint)
+  - [3. Prometheus Exporter](#3-prometheus-exporter)
```
petermetz, line 13: @jagpreetsinghsasan Please put the Promehetus related sections one level lower, e.g. there should be a top level `Monitoring` section and then everything about the prometheus monitoring should be under that (and later if we have other monitoring topics to discuss they can fit in there)

petermetz (changes requested):

packages/cactus-plugin-keychain-aws-sm/docs/architecture/delete-keychain-entry-endpoint.puml
jagpreetsinghsasan: I think this should be referred as Tx (transaction) itself.
My point of view on this is that ledger functions ultimately boils down to similar read/write/get/delete sort of operations.
From rollback point, it can be assumed that we aren't using the correct version of these keychains or they themselves dont support transaction rollback.
A very example to support my view is the kv engine 2 for Hashicorp Vault (which do support rollback functionality due to version-control of the keys stored)
https://www.vaultproject.io/docs/secrets/kv
image

jagpreetsinghsasan: @petermetz what is your opinion on this?

petermetz (approved):
> I think this should be referred as Tx (transaction) itself.
My point of view on this is that ledger functions ultimately boils down to similar read/write/get/delete sort of operations.
From rollback point, it can be assumed that we aren't using the correct version of these keychains or they themselves dont support transaction rollback.
A very example to support my view is the kv engine 2 for Hashicorp Vault (which do support rollback functionality due to version-control of the keys stored)
https://www.vaultproject.io/docs/secrets/kv
image
 

@jagpreetsinghsasan Yeah, works for me, I was on the fence but you also make a good case.

packages/cactus-plugin-keychain-aws-sm/docs/architecture/delete-keychain-entry-endpoint.puml
petermetz: > @petermetz what is your opinion on this?

@jagpreetsinghsasan Yup, works for me, thanks for expanding on it

Issue #1610, opened 2021-11-30
refactor(ledger-connector-corda): add startMonitor features


takeutak:
**Is your feature request related to a problem? Please describe.**
On the current codes, some ledger-connectors (besu, quorum, corda, ...) does not have common API for monitoring block data on connected ledgers.
Corda has no block structure, but if some numbers of transactions are assumed as blocks, this feature can be developed.

This feature was already developed in cmd-socket-server and related ledger-connectors, but in case of merging those codes, we need to enhance the range.

**Describe the solution you'd like**
Add startMonitor feature to the above ledger-connector

**EPIC**
#1611

petermetz:
Applies to Corda v4.8

outSH:
@petermetz I have some questions regarding corda connector and possible monitoring implementation, could you answer them please?
- Does BLP (apiClient) ever need to call kotlin server directly? There're tests for that, but my assumption is that on production kotlin server could be strictly separated from the rest of the network.
- Why there's a single API for both kotlin and typescript services? They'll either diverge along the way, or one of the services will be obsolete, or am I missing something?
- Can we add `watchBlocks` endpoint to typescript component only, and different monitoring implementation endpoints on kotlin server (startMonitor, stopMonitor, getBlocks)?

petermetz:
> Does BLP (apiClient) ever need to call kotlin server directly?

If the application is designed that way then yes. It is possible to design the application in such a way that the answer is no, but that limits certain options of course.

> There're tests for that, but my assumption is that on production kotlin server could be strictly separated from the rest of the network.

You could deploy it on a separate network if you wanted to, yes.

> Why there's a single API for both kotlin and typescript services? They'll either diverge along the way, or one of the services will be obsolete, or am I missing something?

The Typescript code is being redone to be the code for Corda v5. You just cannot see this on main yet but there's been a PR open for quite some time that we need to finalize and then merge (and then if you read the code on `main` it will make much more sense)

> Can we add watchBlocks endpoint to typescript component only, and different monitoring implementation endpoints on kotlin server (startMonitor, stopMonitor, getBlocks)?

If you'd like to, sure thing. Just keep in mind that Kotlin server is for Corda v4 and the Typescript code of the connector is for Corda v5 (in the near future as soon as we can finalize the PR for it)

outSH:
@petermetz The PR you have in mind is https://github.com/hyperledger/cactus/pull/1860?

PR #1956, opened 2022-03-31
feat(corda4): implement monitoring of state changes

outSH:
- Add new endpoints to corda kotlin server used to start and stop monitoring, and get/clean state changes from its internal buffer.
- Add reactive watchBlocksV1 that polls kotlin server and reports new transactions asynchronously.
- Add CordaApiClient support to VerifierClient.
- Add functional test for both monitoring interfaces.
- Update corda setup in corda-all-in-one to newer version.

Closes: #1610


packages/cactus-plugin-ledger-connector-corda/src/test/typescript/integration/monitor-transactions-v4.8.test.ts
```diff
@@ -0,0 +1,760 @@
+/**
+ * Test low-level state change monitoring interface in Kotlin Corda v4 component.
+ */
+
+// Contants: Log Levels
+const testLogLevel: LogLevelDesc = "debug";
+const sutLogLevel: LogLevelDesc = "info";
+
+// Contants: Test ledger
+const ledgerImageName = "ghcr.io/outsh/cactus-corda-4-8-all-in-one-obligation";
```
outSH, line 10: @petermetz Please upload image from this branch to cactus ghcr, and I will replace this reference.

packages/cactus-plugin-ledger-connector-corda/src/test/typescript/integration/monitor-transactions-v4.8.test.ts
```diff
@@ -0,0 +1,760 @@
...
+const ledgerImageVersion = "2022-03-31-2dd619d";
+const partyARpcUsername = "user1";
+const partyARpcPassword = "password";
+const partyBRpcUsername = partyARpcUsername;
+const partyBRpcPassword = partyARpcPassword;
+const stateToMonitor = "net.corda.samples.example.states.IOUState";
+const flowToInvoke = "net.corda.samples.example.flows.ExampleFlow$Initiator";
+
+// Contants: Kotlin connector server
+const kotlinServerImageName = "ghcr.io/outsh/cactus-connector-corda-server";
```
outSH, line 20: @petermetz Same as above

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
```diff
@@ -40,6 +43,9 @@ class ApiPluginLedgerConnectorCordaServiceImpl(
     val rpc: NodeRPCConnection
 ) : ApiPluginLedgerConnectorCordaService {
 
+    val watchedTransactionsBuffer = mutableMapOf<String, MutableList<GetMonitorTransactionsV1ResponseTx>>()
```
petermetz, line 46: @outSH How will this work if the connector is deployed behind a load balancer in an auto-scaling group? Can it be made work with sticky sessions for example?

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/api-client/corda-api-client.ts
```diff
@@ -0,0 +1,152 @@
...
+  Checks,
+} from "@hyperledger/cactus-common";
+import { ISocketApiClient } from "@hyperledger/cactus-core-api";
+import {
+  DefaultApi,
+  GetMonitorTransactionsV1ResponseTx,
+} from "../generated/openapi/typescript-axios";
+import { Configuration } from "../generated/openapi/typescript-axios/configuration";
+
+const DEFAULT_POOL_RATE_MS = 5000;
```
petermetz, line 16: @outSH Is it pooling or polling? Or is it both, e.g. polling the pool?

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/api-client/corda-api-client.ts
```diff
@@ -0,0 +1,152 @@
...
+        } else {
+          this.log.info(`Monitoring for ${stateName} transactions started.`);
+        }
+      })
+      .catch((err) => {
+        reportError(err);
+      });
+  }
+
+  private async poolTransactionsLogin(
```
petermetz, line 79: @outSH Is it pooling or polling? Or is it both, e.g. polling the pool?

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
```diff
@@ -346,4 +352,108 @@ class ApiPluginLedgerConnectorCordaServiceImpl(
...
+    override fun stopMonitorV1(req: StopMonitorV1Request?): StopMonitorV1Response {
+        val stateName = req?.stateFullClassName
+
+        if (stateName.isNullOrEmpty()) {
+            // Missing input class name
+            logger.info("Request rejected because missing state class name")
+            return StopMonitorV1Response(false)
+        }
+
+        this.watchedTransactionsSubs[stateName]?.unsubscribe()
```
petermetz, line 452: @outSH Is there some logic in here that gracefully frees up resources if the client didn't do so itself? 

Example: 
1. Browser client starts monitoring
2. Server creates a subscription here 
3. Browser client disappears (phone battery dies, laptop falls into a river, etc.)
4. Server keeps on chugging along with the subscription or does it automatically cleans up at some point?

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
```diff
@@ -346,4 +352,108 @@ class ApiPluginLedgerConnectorCordaServiceImpl(
         logger.info("Returning {} NodeInfo elements in response.", nodeInfoList.size)
         return nodeInfoList
     }
+
+    /**
+     * Start monitoring state changes for stateClass specified in request body.
+     */
+    override fun startMonitorV1(req: StartMonitorV1Request?): StartMonitorV1Response {
```
petermetz, line 359: @outSH Recommendation: I'd try to make it work with JVM SocketIO instead of HTTP polling for these reasons:
1. Polling uses more CPU/bandwidth
2. Keep it consistent with the other plugin implementations that we have

petermetz:
> @petermetz Please upload image from this branch to cactus ghcr, and I will replace this reference. 

@outSH Here it is: `ghcr.io/hyperledger/cactus-corda-4-8-all-in-one-obligation:2022-03-31-28f0cbf--1956`

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
outSH: @petermetz What do you mean, isn't sticky session configured on a load-balancer only? How can we make an application work with that?

I will add multi-user support soon to this PR, but even then a client will have to call a single server I'm afraid.

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/api-client/corda-api-client.ts
outSH: @petermetz Fixed this and couple other misspells :)

packages/cactus-plugin-ledger-connector-corda/src/main/typescript/api-client/corda-api-client.ts
outSH: @petermetz Same as above

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
outSH: @petermetz And what if BLP looses a connection for few minutes due to network error and we trash the events he was expecting to receive to proceed with it's logic? It could be devastating from business point of view. To have this at least once delivery behavior we need some sort of queue on the connector. We can clear it from time to time to prevent memory draining, put the queue in redis or regular DB, but the more improvements in this area we do, the more we're reinventing the wheel of a message broker. In current implementation I've assumed long-lived clients only (BLP) that would sooner or later reconnect, and would like to fetch any state changes that occurred during it's blackout.

This whole problem comes down to a question - **what are expected message delivery guarantees** for block, or any other monitored async data?  We're right now clarifying exact requirements for message delivery guarantees we need internally, **we would like to know your point of view on this as well**.
- Current socketio monitoring (for instance, in besu) is not reliable at all, since blocks can be easily missed, which might be OK in some cases, but in others (like finance area where corda majors) is definitely not.
- The more guarantees we need, the more backend implementation and communication complexity we need.
- At some point it will be better to just use an actual message broker like kafka or AMQP based solution.
- Another thing is, that actual enterprise-grade message broker or anything near it will not work well with your idea ApiClients in browser, mobile phones, toaster ovens talking to the connector directly :)
- So, please share the requirements and possibly scenarios you must support, and then we can try to come to some conclusion. I will implement the required changes in this PR then.

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
outSH: Yeah, that was my initial plan too (mainly because of #2), but the `socket.io-server-java` setup looked overly complicated for this scenario and time frame, especially for someone like me who did not work at all with spring boot before. It doesn't seem very popular either, I would be very cautious before basing entire monitoring on that. `netty-socketio` was easier to setup and more popular, but unfortunately it doesn't support socketio v4. 

Side note - as described in previous review comment, current polling application is less likely to skip/loose events than socketio implementation. Similar reliability features could be implemented on socketio as well of course, but I'm afraid the implementation would be much more complex than polling one. Final implementation to be decided soon, but I'd rather not use socket.io here for reasons mentioned above.

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
petermetz: @outSH Delivery guarantees are a different rabbit hole. I just want to know if in its current state it's leaking memory or not. :-)

For the delivery guarantees: I largely agree. My point of view is that we can probably support different transports for the same endpoints, meaning that pushing the data down the initiating web socket would be one transport, appending to a queue could be another, AMQP, JMS or whatever we can actually support.

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
petermetz: @outSH Cool. Would you mind including this information in the commit message as well? If we try to retrace our steps later on it will be a very useful reminder to people like me who are having trouble remembering simple things like what they had for breakfast. (joking, but not joking)
And I know it's on GH as well, but if we ever migrate, then we'll most likely only be able to take the git commit history not whatever is stored in GH's private database as comments.

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
petermetz: @outSH My question was not phrased very well, apologies. I'll try differently: 
Is there (or is there going to be) some recovery mechanism if the server instance goes away or is the client expected to implement this logic for now?

`Client -> LB -> [A, B, C, D]`

If the client has the subscription with `A` but then it goes away and the `LB` migrates the sticky session to `B` and then `B` would start returning an empty list in the GetMonitorTransactionsV1 responses (if I read the code right, but I'm not a Kotlin pro myself either)

petermetz (changes requested):
^^ See above

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
outSH: @petermetz Sure, agree, I've updated the commit message to clarify the monitoring design.

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
outSH: @petermetz Yeah, that make sense. The initial implementation assumed single connector and single client calling it (our current requirements for this feature are very basic), so the case you've described was not supported. I've done a major rework and reorganization of kotlin code. One of the changes is support for multiple clients. Fetching transactions from server on which we didn't explicitly start the monitoring first will end up with an error now, so i think the scenario you've described is solved now (the caller will be informed that the monitoring has broke and it needs to be restarted, possibly on new connector node).

.cspell.json
```diff
@@ -132,7 +132,8 @@
         "vscc",
         "wasm",
         "Xdai",
-        "goquorum"
+        "goquorum",
+        "outsh"
```
izuru0, line 136: @outsh I think you don't need this because image name in test code is updated to one that doesn't include this word.

outSH: @petermetz There are 2 images, I'm still waiting for `ghcr.io/outsh/cactus-connector-corda-server` upload to hyperledger ghcr

izuru0: @outSH I see. thanks!

packages/cactus-core-api/src/main/typescript/plugin/ledger-connector/i-socket-api-client.ts
```diff
@@ -21,5 +21,5 @@ export interface ISocketApiClient<BlockType> {
 
   watchBlocksV1?(
```
petermetz, line 22: @outSH Wouldn't this be a breaking change though?

I'm thinking it breaks (crashes) call sites at runtime (if we switch to returning a promise) that look like this:

```typescript
const anObservable = socketApiClient.watchBlocksV1({ ... });
anObservalbe.subscribe(...);
```

Maybe we can add a brand new method instead called `watchBlocksAsyncV1`? That way it's an addition, not a breaking change of the existing API that was locked down when we tagged `1.0.0`.

That would also have the advantage that `watchBlocksAsyncV1` does not need to return `Observable<BlockType> | Promise<Observable<BlockType>>` it can return specifically `Promise<Observable<BlockType>>` and this way the callers will have one less thing to worry about.

petermetz (changes requested):

packages/cactus-core-api/src/main/typescript/plugin/ledger-connector/i-socket-api-client.ts
outSH: @petermetz Well, I did not consider this to be used by any client code, only as apiClient contract and inside verifier API. Nevertheless, it is exported as public so yeah, should be treated like public API I suppose. Also, I like the idea of separate naming, this can prevent possible bugs along the way. I've added `watchBlocksAsyncV1` method and adjusted Verifier accordingly.

petermetz: @outSH Yeah, that's the thing, if it's public then someone will probably end up using it somehow even if it wasn't meant to be. Thanks very much for the update!

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
petermetz: @outSH Awesome, marking this as resolved, thank you for the updates!

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
petermetz: @outSH Cheers, resolving this.

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/impl/ApiPluginLedgerConnectorCordaServiceImpl.kt
petermetz: @outSH 

> @petermetz Yeah, the question about delivery guarantees was not directly related to this review but to whole PR, I wanted to clarify expectation for the messaging implementation :)

Fair enough. We could open a dedicated issue in the tracker, link to this conversation for more context and then hash it out further if needed (not much to say form my side though, we appear to be on the same page)

> @petermetz I'd personally prefer to settle on single communication mechanism (at least for async), I'm afraid having multitude of pluggable technologies can turn into matinanace nightmare as soon as we have any production workloads using them. Nevertheless, our teams don't have any specific requirements at the moment, so I think we can go with free-for-all approach at the moment.

I also don't want to maintain a several different transports, but I do want to be able to say `"Yes"` to someone who comes and asks us: `Hey, is there a way to add support for a different transport without having to majorly refactor the entire codebase?` Then they would have to implement & maintain that for themselves on account of their own production workloads (them = the OSS community)
This is also not a change request just me dumping thoughts.

petermetz: @outSH 

> @petermetz As for memory leak, in current version each client that started a monitoring on the connector is checked periodically, and clients who has no activity for specified amount of time are removed and they resources are freed. Timeout is configurable.

Thank you very much, marking it as resolved!

.cspell.json
petermetz: > @izuru0 There are 2 images, I'm still waiting for `ghcr.io/outsh/cactus-connector-corda-server` upload to hyperledger ghcr

@outSH @izuru0 Apologies for the wait, I've just uploaded `ghcr.io/hyperledger/cactus-connector-corda-server:2022-04-18-a8a7ed1--1956`

packages/cactus-plugin-ledger-connector-corda/src/test/typescript/integration/monitor-transactions-v4.8.test.ts
petermetz: > @petermetz Same as above

@outSH Here's the new tag, please update the code and then we are good to merge IMO
`ghcr.io/hyperledger/cactus-connector-corda-server:2022-04-18-a8a7ed1--1956`

petermetz (commented):
@outSH Here's the new tag, please update the code and then we are good to merge IMO
ghcr.io/hyperledger/cactus-connector-corda-server:2022-04-18-a8a7ed1--1956

petermetz (changes requested):
@outSH Here's the new tag, please update the code and then we are good to merge IMO
`ghcr.io/hyperledger/cactus-connector-corda-server:2022-04-18-a8a7ed1--1956`

petermetz (changes requested):
@outSH Please also rebase + resolve the conflicts

outSH:
> @outSH Please also rebase + resolve the conflicts

@petermetz Done

petermetz (changes requested):
@outSH Here's the new tag, please update the code and then we are good to merge IMO
`ghcr.io/hyperledger/cactus-connector-corda-server:2022-04-18-a8a7ed1--1956`

outSH:
> @outSH Here's the new tag, please update the code and then we are good to merge IMO `ghcr.io/hyperledger/cactus-connector-corda-server:2022-04-18-a8a7ed1--1956`

@petermetz I've replaced the image as well, also removed my nick from cspell exception

petermetz (approved):

petermetz:
@outSH  LGTM

izuru0 (approved):
LGTM

PR #1958, opened 2022-04-04
ci(connector-fabric): include endpoints in the fuzzer tests 

### LINKED ISSUE #1952: ci(connector-fabric): include endpoints in the fuzzer tests
This is a follow-up/improvement on top of what we added for fixing https://github.com/hyperledger/cactus/issues/495

The current fuzzing only verifies that the API server's own endpoints are secure
but we should do the same thing for each REST endpoint that is provided by each
of the plugins.
In the scope of this task the Fabric connector plugin's endpoint's should be covered.

The way to do this is to update the current fuzzing test to install the Fabric
connector into the API server and then also update the fuzzer configuration to
include the URL paths of those endpoints so that they get tested by the fuzzer.

See the workflow yaml file `.github/workflows/.dast-nuclei-cmd-api-server.yaml`
on the details about how the fuzzing test works. This is the file where you'll
have to add the URL paths of the Connector connector plugin as well.
The above file (`.github/workflows/.dast-nuclei-cmd-api-server.yaml`) is also the
place where you can modify the configuration of the API server instance to include
the Fabric connector plugin.

jagpreetsinghsasan:
# Commit to be reviewed
ci(connector-fabric): include endpoints in the fuzzer tests

    Primary changes
    ----------------------
    1. Updated the gitworkflow to include the fabric connector plugin endpoint
    2. Updated the fabric connector plugin README.md for minor fixes

Fixes #1952

### DIFF
--- a/.github/workflows/.dast-nuclei-cmd-api-server.yaml
+++ b/.github/workflows/.dast-nuclei-cmd-api-server.yaml
@@ -41,6 +41,11 @@ jobs:
           echo https://localhost:4000/api/v1/plugins/@hyperledger/cactus-plugin-consortium-manual/node/jws >> urls.txt
           echo https://localhost:4000/api/v1/plugins/@hyperledger/cactus-plugin-consortium-manual/consortium/jws >> urls.txt
           echo https://localhost:4000/api/v1/plugins/@hyperledger/cactus-plugin-consortium-manual/get-prometheus-exporter-metrics >> urls.txt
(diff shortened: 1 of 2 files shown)


takeutak (approved):

petermetz (approved):

jagpreetsinghsasan:
The semantic pull request check seems to be stuck.
@petermetz do I need to make a force push again for it to re-do?

petermetz:
> The semantic pull request check seems to be stuck. @petermetz do I need to make a force push again for it to re-do?

@jagpreetsinghsasan I'm working on a fix. In the meantime just to unblock us I disabled the need for the semantic PR check to pass.

PR #1942, opened 2022-03-22
replaced rocketchat with discord

Ejashba:
Replaced the old rocketchat link with the new discord link

PR #1964, opened 2022-04-10
Rocketchat to Discord

### LINKED ISSUE #1937: docs(readme): update chat from rocketchat to discord
Right now the readme talks about the old (and now obsolete) RocketChat channel that we used to have.
We need to replace that with discord.

AbhinavMir:
Fixes #1937

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Can be found here: [ROADMAP.md](./ROADMAP.md)
 
 ## Contact
 * mailing list: [cactus@lists.hyperledger.org](mailto:cactus@lists.hyperledger.org)
-* rocketchat channel: [https://chat.hyperledger.org/channel/cactus](https://chat.hyperledger.org/channel/cactus).
+* Discord: [https://discord.com/invite/hyperledger](https://discord.com/invite/hyperledger)
 
 ## Build/Development Flow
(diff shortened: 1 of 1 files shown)


izuru0 (approved):
LGTM

takeutak:
@petermetz The contents of this PR itself looks good to me, but this is duplicated with the PR #1942 .
If we prioritize the pull-request that was posted earlier, we should fix the semantic error in this PR #1942.
How do you think, @petermetz ?

AbhinavMir:
My bad! I did not know there was another active PR  to the same issue. Happy to do a re-commit with the right semantics if needed.

petermetz:
> My bad! I did not know there was another active PR  to the same issue. Happy to do a re-commit with the right semantics if needed. 

@AbhinavMir Let's see how the other PR turns out. The author of that one said he'd fix its issues today. If that comes to pass then we'll merge that one on account of the first come first served PR acceptance policy and this will have to be closed as a duplicate. On the other hand if the other PR experiences more delays then we might just move ahead with this one to make sure that the change gets in as soon as possible.

AbhinavMir:
There seems to be a delay. Happy to resolve it now.

PR #1967, opened 2022-04-12
ci: remove nodejs v12 exclusion logic from run custom checks

### LINKED ISSUE #1961: ci(custom-checks): remove NodeJS v12 exclusion logic
While coming up with the fix for https://github.com/hyperledger/cactus/issues/1809
I realized that after upgrading `ts-node` to the latest available version,
the custom-checks script now can run just fine on NodeJS v12 as well.

Based on the above, we can remove the logic both from `tools/ci.sh` and 
`tools/custom-checks/run-custom-checks.ts` which stops the custom checks
execution if NodeJS v12 is detected.

The other reason to clean up this from the code is because NodeJS v12 is end of
life now and we no longer include it in the CI at all anyway. So this is one 
of those tasks that adds value by removing code instead of adding it.

Leeyoungone:
Fixes #1961

### DIFF
--- a/tools/custom-checks/run-custom-checks.ts
+++ b/tools/custom-checks/run-custom-checks.ts
@@ -12,12 +12,7 @@ export async function runCustomChecks(
   let overallSuccess = true;
   let overallErrors: string[] = [];
 
-  const [majorVersion] = version.split(".");
-  const nodeV12OrOlder = parseInt(majorVersion) <= 12;
-  if (nodeV12OrOlder) {
-    console.log(`${TAG} Checks skipped due to old NodeJS (v${version}) OK.`);
-    return;
-  }
+  console.log("${TAG} Current NodeJS version is v", version);
 
   {
     const [success, errors] = await checkOpenApiJsonSpecs({ argv, env });
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -117,16 +117,6 @@ function mainTask()
 
   yarn run configure
 
-  # Obtains the major NodeJS version such as "12" from "v12.14.1"
-  # We only run the custom checks above v12 because the globby dependency's
-  # latest version is forcing us to use Ecmascript Modules which do not work
-  # on NodeJS 12 even with the additional flags passed in.
-  nodejs_version=`node --version | awk -v range=1 '{print substr($0,range+1,2)}'`
-  if [ "$nodejs_version" -gt "12" ]; then
-    echo "$(date +%FT%T%z) [CI] NodeJS is newer than v12, running custom checks..."
-    yarn run custom-checks
-  fi
-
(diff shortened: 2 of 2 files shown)


takeutak (approved):

tools/custom-checks/run-custom-checks.ts
```diff
@@ -6,19 +6,11 @@ import { checkSiblingDepVersionConsistency } from "./check-sibling-dep-version-c
 export async function runCustomChecks(
   argv: string[],
   env: NodeJS.ProcessEnv,
-  version: string,
```
petermetz, line 9: @Leeyoungone I wouldn't remove the version parameter itself because later on we might again find ourselves in a situation where we need to do some custom logic based on it (let's say Node v18 comes out and something doesn't work on it or something stops working on Node v14 which right now is perfectly fine, etc.)
If you removed it because the linter complained that it's not being used for anything then my trick for that usually is to just log it out so that a) we provide more information in the logs and b) the linter chills out

package.json
```diff
@@ -136,7 +136,7 @@
     "tape-promise": "4.0.0",
     "ts-jest": "27.0.3",
     "ts-loader": "9.2.5",
-    "ts-node": "10.2.0",
+    "ts-node": "10.7.0",
```
petermetz, line 139: @Leeyoungone I remember this being done in another pull request that's unrelated to this one so you might want to mark this as a dependent of that PR if it otherwise doesn't work.

petermetz (changes requested):

package.json
Leeyoungone: ooo neat neat! thank you, I think I might have changed the `ts-node`without meaning to so I'll change it back

tools/custom-checks/run-custom-checks.ts
Leeyoungone: Sounds good! Thank you for the cool trick 👓 !

petermetz (changes requested):
@Leeyoungone Could you please also remove the same logic from `./tools/ci.sh` ?

```sh
  # Obtains the major NodeJS version such as "12" from "v12.14.1"
  # We only run the custom checks above v12 because the globby dependency's
  # latest version is forcing us to use Ecmascript Modules which do not work
  # on NodeJS 12 even with the additional flags passed in.
  nodejs_version=`node --version | awk -v range=1 '{print substr($0,range+1,2)}'`
  if [ "$nodejs_version" -gt "12" ]; then
    echo "$(date +%FT%T%z) [CI] NodeJS is newer than v12, running custom checks..."
    yarn run custom-checks
  fi
```

petermetz (approved):
@Leeyoungone LGTM, thank you for the updates!

Issue #1957, opened 2022-04-01
test(connector-iroha): fix and unskip run-transaction-endpoint-v1.test


petermetz:
## Description

Currently this  test case is being skipped due to issues with the Iroha AIO
image:
`packages/cactus-plugin-ledger-connector-iroha/src/test/typescript/integration/run-transaction-endpoint-v1.test.ts`

## Acceptance criteria

1. The mentioned test is once again part of the CI and it passes consistently.

Leeyoungone:
Hey @petermetz, just wanted to let you know that I've ran the test case at least 20 times (without any changes except not skipping the test cases anymore) and it has passed every single one of them. I've put up an issue mentioning this "fix" where I got rid of the `.skip`, let me know your thoughts

petermetz:
> Hey @petermetz, just wanted to let you know that I've ran the test case at least 20 times (without any changes except not skipping the test cases anymore) and it has passed every single one of them. I've put up an issue mentioning this "fix" where I got rid of the `.skip`, let me know your thoughts

@Leeyoungone Nice, let's see how it goes. I don't remember for sure if we skipped it because it wasn't working at all or if it was flaky. We can start not skipping it and monitor the CI runs on unrelated PRs for failures due to this test case (if it starts being flaky we'll have to reverse your change and reopen this issue until we figure out the reason for the flaky execution of the test case)

Leeyoungone:
@petermetz sounds good! thanks for your thoughts :))

PR #1968, opened 2022-04-12
test(connector-iroha): tested previously skipped test

Leeyoungone:
All I did was change from `test.skip` to just `test` and it turns out that the test works just fine. So for I've ran the test 20 times and it has passed every single time. This issue is a "fix" to #1957 ?


takeutak (approved):

petermetz (changes requested):
> All I did was change from `test.skip` to just `test` and it turns out that the test works just fine. So for I've ran the test 20 times and it has passed every single time. This issue is a "fix" to #1957 ?

> Signed-off-by: Youngone Lee <youngone.lee@accenture.com> 

@Leeyoungone Please put this also in the commit message and also the information I wrote in the issue comment so that people looking at the git log see the full context. Once that's done it's LGTM

Leeyoungone:
@petermetz Sounds good! Just fixed it :))

petermetz (approved):
@Leeyoungone LGTM, thank you!
Note to my future self: if this gets flaky again on the CI, we need to start skipping it again.
Note to Wen: Please remove  commented code that you no longer need ;-)

PR #1971, opened 2022-04-13
chore(deps): upgrade angular in examples/carbon accounting

### LINKED ISSUE #1651: chore(deps): upgrade angular in examples/carbon accounting
At the time of this writing there is no published fix for the issue unfortunately
but we should track the problem regardless since this is a high severity vulnerability
according to the advisory pasted below.

It is only present in the example code of the carbon accounting app's frontend
so at the end of the day it's not a huge deal because this would never get 
deployed to production but regardless of that we should take care of it with a 
high priority.

https://github.com/advisories/GHSA-whgm-jr23-g3j9

    CVE-2021-23424
    high severity
    Vulnerable versions: <= 0.0.7
    Patched version: No fix
    This affects all versions of package ansi-html. 
    If an attacker provides a malicious string, 
    it will get stuck processing the input for an extremely long time.

aldousalvarez:
Fixes #1651

Depends on https://github.com/hyperledger/cactus/issues/1980


takeutak (approved):

petermetz:
@aldousalvarez 

Depends on #1980

petermetz (approved):
@aldousalvarez  Can you confirm that you performed the upgrade with the angular CLI tool?

aldousalvarez:
This is the documentation that I read about angular cli and in the update command there is a link that goes on about the steps on how to upgrade Angular version.  https://angular.io/cli

image

ng update - Updates your application and its dependencies. See https://update.angular.io/

and I followed the steps on the guide that was given and used this command to update the angular of carbon accounting frontend
- npx @angular/cli@13 update @angular/core@13 @angular/cli@13

petermetz:
> 

@aldousalvarez Yep, that's the one I was looking for, thank you for confirming!

aldousalvarez:
Hello @petermetz while I was upgrading the angular on the supply chain app #2002 and used the angular cli tool there are newer versions that I have found. My question is should I update also the angular on this carbon accounting to its latest? Thanks

Issue #1976, opened 2022-04-15
test(connector-fabric-socketio): add functional test, bug fix


outSH:
Add a functional test to `connector-fabric-socketio` so it's operations can be checked in CI. Refactor the connector to improve testability, fix any bugs that are revealed by the tests.

PR #1975, opened 2022-04-15
refactor(connector-fabric-socketio): fix strict flag warnings

outSH:
cactus-plugin-ledger-connector-fabric-socketio will compile with global strict flag.

Related issue: #1671

Signed-off-by: Michal Bajer <michal.bajer@fujitsu.com>

PR #1977, opened 2022-04-15
test(connector-fabric-socketio): add functional test, bug fix

outSH:
Added functional jest test `fabric-socketio-connector.test` that can be run during CI process. It checks evaluate/sending transactions and monitoring for new events. Connector had to be refactored to be testable, tests also discovered some bugs that had to be fixed in order to pass.

### SocketIOApiClient refactors:
- Added option for supplying `validatorKeyValue` instead of `validatorKeyPath`.
- JWT validation function works with key value now (instead of reading the key).
- Validator can now return messages that are not encrypted (it throwed error previously).
- Adjusted unit tests.

### connector-fabric-socketio refactors:
- Connector can be run both as a standalone app and loaded as a module `www.js`. Caller can use exported `startFabricSocketIOConnector` function to run the connector. Configuration must be supplied in file or in env variable like it's done in functional test.
- All cryptographic data (keys, certificates, etc…) can now be supplied as a value (previously it supported only path to a file).
- `sendSignedTransaction` can now be called synchronously (it had wrong response format before).
- Fixed a bug introduced during my last changes in this component, which caused fabric-client session to be disconnected but still reused by follow-up requests. I didn't know that gateway disconnects client it operates on.
- Increased JWT expiration to 15 minutes to prevent constant JWT expiration error (I'm pretty sure 15 minutes is still secure period).
- Minor improvements (logging, formatting, etc…)

### fabric-test-ledger-v1 changes:
- Added `adminCredentials()` to have programatic access to admin credentials on currently used ledger
 (unlikely, but can change in the future).

Depends on: https://github.com/hyperledger/cactus/pull/1975

Closes: https://github.com/hyperledger/cactus/issues/1976

### DIFF
--- a/packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
+++ b/packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
@@ -17,7 +17,7 @@ import {
 import { ISocketApiClient } from "@hyperledger/cactus-core-api";
 
 import { Socket, SocketOptions, ManagerOptions, io } from "socket.io-client";
-import { readFile } from "fs";
+import { readFileSync } from "fs";
 import { resolve as resolvePath } from "path";
 import { verify, VerifyOptions, VerifyErrors, JwtPayload } from "jsonwebtoken";
 import { Observable, ReplaySubject } from "rxjs";
@@ -26,40 +26,31 @@ import { finalize } from "rxjs/operators";
 /**
  * Default logic for validating responses from socketio connector (validator).
  * Assumes that message is JWT signed with validator private key.
- * @param keyPath - Absolute or relative path to validator public key.
+ * @param publicKey - Validator public key.
  * @param targetData - Signed JWT message to be decoded.
  * @returns Promise resolving to decoded JwtPayload.
  */
 export function verifyValidatorJwt(
-  keyPath: string,
+  publicKey: string,
   targetData: string,
 ): Promise<JwtPayload> {
   return new Promise((resolve, reject) => {
-    readFile(
-      resolvePath(__dirname, keyPath),
-      (fileError: Error | null, publicKey: Buffer) => {
-        if (fileError) {
-          reject(fileError);
+    const option: VerifyOptions = {
+      algorithms: ["ES256"],
+    };
+
+    verify(
+      targetData,
+      publicKey,
+      option,
+      (err: VerifyErrors | null, decoded: JwtPayload | undefined) => {
+        if (err) {
+          reject(err);
+        } else if (decoded === undefined) {
+          reject(Error("Decoded message is undefined"));
+        } else {
+          resolve(decoded);
         }
-
-        const option: VerifyOptions = {
-          algorithms: ["ES256"],
-        };
-
-        verify(
-          targetData,
-          publicKey,
-          option,
-          (err: VerifyErrors | null, decoded: JwtPayload | undefined) => {
-            if (err) {
-              reject(err);
-            } else if (decoded === undefined) {
-              reject(Error("Decoded message is undefined"));
-            } else {
-              resolve(decoded);
-            }
-          },
-        );
       },
     );
   });
@@ -71,7 +62,8 @@ export function verifyValidatorJwt(
 export type SocketIOApiClientOptions = {
   readonly validatorID: string;
   readonly validatorURL: string;
-  readonly validatorKeyPath: string;
+  readonly validatorKeyValue?: string;
+  readonly validatorKeyPath?: string;
   readonly logLevel?: LogLevelDesc;
   readonly maxCounterRequestID?: number;
   readonly syncFunctionTimeoutMillisecond?: number;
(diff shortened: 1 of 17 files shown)


outSH:
@petermetz @izuru0 Please review

izuru0 (approved):
LGTM

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
```diff
@@ -26,40 +26,31 @@ import { finalize } from "rxjs/operators";
...
  * Default logic for validating responses from socketio connector (validator).
  * Assumes that message is JWT signed with validator private key.
- * @param keyPath - Absolute or relative path to validator public key.
+ * @param publicKey - Validator public key.
  * @param targetData - Signed JWT message to be decoded.
  * @returns Promise resolving to decoded JwtPayload.
  */
 export function verifyValidatorJwt(
-  keyPath: string,
+  publicKey: string,
```
petermetz, line 34: @outSH Does this break code that uses the 1.0.0 version with a file path instead of the public key as a string?

outSH: @petermetz `verifyValidatorJwt` is exported just for unit-testing purposes, I don't think there's a use case where a client would be concerned with this function.

petermetz: > verifyValidatorJwt is exported just for unit-testing purposes

@outSH Can you (eventually) move the whole thing into one of the test packages then?

> I don't think there's a use case where a client would be concerned with this function

I would still recommend practicing good sem-ver hygiene just to get into the habit. My assumption is that there's a tiny effort associated to this. If that assumption is incorrect, then it's a different story of course. Let me know.

petermetz (changes requested):

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: > Can you (eventually) move the whole thing into one of the test packages then?

@petermetz It's not a test-util. It's exported separately to simplify unit test of message verification functionality. It doesn't make sense outside of `SocketIOApiClient`, but testing it through an ApiClient would greatly convolute the test.

> I would still recommend practicing good sem-ver hygiene just to get into the habit. My assumption is that there's a tiny effort associated to this. If that assumption is incorrect, then it's a different story of course. Let me know.

There's very little effort to just add a new function and leave the old as is, but it will be pure dead code, not sure which one is better.

outSH: @jagpreetsinghsasan Could you take over the review of this PR, please?

A side note regarding `verifyValidatorJwt` that is a matter of discussion here, I want to extract it into separate package in a commit I'm working on right now, to keep the validator auth logic in one place. Would be useful to decide whether it's OK to treat it as a private interface (which is true from functional perspective). Please check discussion earlier for details :)

jagpreetsinghsasan: @outSH Definitely I can have my PR reviews as well. The process of a PR acceptance and merge involves PR review from all the reviewers.
To make the process more manageable, I personally, try to give sequential reviews prior/later to the ongoing reviews, so that the common reviews or clashing reviews don't create further troubles/misunderstanding for the contributor as the involved reviewers.

petermetz: @outSH 

> @petermetz It's not a test-util. It's exported separately to simplify unit test of message verification functionality.

`to simplify unit test of ...` is the very definition of a test utility IMO.

> but testing it through an ApiClient would greatly convolute the test.

I might be missing something. Wouldn't the change be a one liner for updating the import statement within the test case that pulls in the function?

> but it will be pure dead code, not sure which one is better.

We can call the dead code by its pretty name (deprecated) and open an issue to remove it in v2.0.0 (and label the issue with the breaking change label so we know not to work on it prior to 2.0.0 so my vote goes for that.

petermetz (changes requested):
Sorry for the slow turnaround. https://github.com/hyperledger/cactus/pull/1977#discussion_r882150679

packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
outSH: @petermetz 

Separation of component to make it injectable is a _test utility_ for you? OK :)

Could you please clarify what is your definition of a public interface? This function is not exported outside of this package (i.e. not exported in `public-api`, which as the name suggest, should be only public interface definition IMO). It's exported from this file-module just to simplify injection in unit test phase (internally, within a package). Do we want to keep static everything with export word anywhere in the project?

petermetz (approved):
> Separation of component to make it injectable is a _test utility_ for you? OK :)

I'm going off of what you told me earlier. No worries.

> This function is not exported outside of this package

You could've lead with that. If it's not on the public API then LGTM

PR #1989, opened 2022-04-25
ci: add yaml workflow file for semantic-pull-request action

### LINKED ISSUE #1987: ci: add yaml workflow file for semantic-pull-request action
**Describe the bug**

Right now we have the out of the box version of the action coming straight from the marketplace with whatever default configuration that uses.
The issue with this is that it is flaky: it randomly gets stuck in a way that a perfectly good PR is blocked from being merged just because the semantic PR check is not executing for whatever reason. This is annoying as a maintainer because it reduces our velocity & PR throughput (which is already something we need to improve upon)

**To Reproduce**

Unclear. It comes and goes.

**Expected behavior**

The check is stable, executes every time reliably without issues.

We should use an explicit yaml workflow file for configuration and the latest version of the action from here:
https://github.com/marketplace/actions/semantic-pull-request

**Logs/Stack traces**

N/A

**Screenshots**

Screenshot from 2022-04-22 10-52-49

GitHub Hosted Action Runners / Marketplace

N/A

N/A

N/A

**Additional context**

This is a problem that is affecting multiple pull requests at the moment and it's not the first time that the issue randomly popped up. It probably would disappear again in a few days but I'd rather us try to go for a stable fix so that it's never coming back.

petermetz:
Fixes #1987

### DIFF
--- a/.github/workflows/semantic-pull-request.yaml
+++ b/.github/workflows/semantic-pull-request.yaml
@@ -1,0 +1,17 @@
+name: "Lint PR"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: amannn/action-semantic-pull-request@v4.4.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


petermetz:
The check won't execute until it is merged onto `main` so we have to fly blind on this one in the sense that either the configuration will work or not. Once it's merged onto `main` then the check will start executing for all PRs

takeutak (approved):

jagpreetsinghsasan (approved):
LGTM

PR #1991, opened 2022-04-26
docs(maintainers): add Rama and Sandeep

petermetz:
Adding Rama and Sandeep as maintainers.

Also updated the chat column for Takuma, Izuru, Jagpreet and Peter
so that their Discord IDs are visible there instead of the former
RocketChat user handles which are no longer in existence anyway since
the Hyperledger RocketChat got shut down in early 2022.


takeutak (approved):

izuru0 (approved):

jagpreetsinghsasan (approved):
My discord ID is Jag#2402, can that be updated as well?

PR #1992, opened 2022-04-26
ci(connector-besu): include endpoints in fuzzer tests

### LINKED ISSUE #1949: ci: fuzzer tests to include besu connector endpoints
This is a follow-up/improvement on top of what we added for fixing https://github.com/hyperledger/cactus/issues/495

The current fuzzing only verifies that the API server's own endpoints are secure
but we should do the same thing for each REST endpoint that is provided by each
of the plugins.
In the scope of this task the Besu connector plugin's endpoint's should be covered.

The way to do this is to update the current fuzzing test to install the besu
connector into the API server and then also update the fuzzer configuration to
include the URL paths of those endpoints so that they get tested by the fuzzer.

See the workflow yaml file `.github/workflows/.dast-nuclei-cmd-api-server.yaml`
on the details about how the fuzzing test works. This is the file where you'll
have to add the URL paths of the Besu connector plugin as well.
The above file (`.github/workflows/.dast-nuclei-cmd-api-server.yaml`) is also the
place where you can modify the configuration of the API server instance to include
the Besu connector plugin.

jagpreetsinghsasan:
# Commit to be reviewed

ci(connector-besu): include endpoints in fuzzer tests 

    Primary Changes
    --------------
    1. Updated the git workflow to include the besu connector plugin endpoint
    2. Added the task to run Besu all-in-one image
       as a valid ws endpoint is required for connector to initialize
    3. Updated the besu connector plugin README.md for minor fixes

Fixes #1949


petermetz (approved):

takeutak (approved):

petermetz (changes requested):
@jagpreetsinghsasan Please resolve the conflicts and then re-request review!

PR #1999, opened 2022-05-02
feat(odap-plugin): backup gateway implementation

AndreAugusto11:
* Addition of backup gateways allowed field
* New test where a backup gateway is allowed to replace one that crashed

### DIFF
--- a/packages/cactus-plugin-odap-hermes/README.md
+++ b/packages/cactus-plugin-odap-hermes/README.md
@@ -52,7 +52,7 @@ Firstly let us identify the different entities involved in the protocol and what
 
 The sequence diagram of ODAP is pictured below.
 
-![odap-sequence-diagram](https://imgur.com/a/NN62LyL)
+![odap-sequence-diagram](https://i.imgur.com/AFNvL9v.png)
 
 ### API Endpoints
 This plugin uses OpenAPI to generate the API paths.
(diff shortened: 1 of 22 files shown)


packages/cactus-plugin-odap-hermes/src/test/typescript/integration/backup-gateway-after-client-crash.test.ts
```diff
@@ -0,0 +1,907 @@
...
+    throw new Error("Database is not correctly initialized");
+  }
+
+  await makeSessionDataChecks(
+    pluginSourceGateway,
+    pluginRecipientGateway,
+    sessionID,
+  );
+
+  expect(
```
petermetz, line 861: @AndreAugusto11 Could you please refactor these assertions to use one of the recommended ways from the Jest documentation?

https://jestjs.io/docs/asynchronous#asyncawait

These are my favorites, they are the least verbose and yet expressive IMO. Copying from the docs:

```typescript
test('the data is peanut butter', async () => {
  await expect(fetchData()).resolves.toBe('peanut butter');
});

test('the fetch fails with an error', async () => {
  await expect(fetchData()).rejects.toMatch('error');
});
```

packages/cactus-plugin-odap-hermes/README.md
```diff
@@ -123,6 +123,10 @@ A test with a rollback after a timeout (client crashed):
 
 https://github.com/hyperledger/cactus/blob/main/packages/cactus-plugin-odap-hermes/src/test/typescript/integration/odap/odap-rollback.test.ts
 
+A test with a backup gateway resuming the protocol after the client gateway crashed:
+
+https://github.com/hyperledger/cactus/blob/main/packages/cactus-plugin-odap-hermes/src/test/typescript/integration/odap/backup-gateway-after-client-crash.test.ts
```
petermetz, line 128: @AndreAugusto11 This link will die if the file gets renamed, moved or deleted on the main branch in the future. What I recommend instead is getting the direct link to the current latest commit on main (the head) and then link to the file through that. 
This way it doesn't matter what happens on the main branch, the link will be alive as long as GitHub is alive (which is also a risk we are taking, but an exponentially lower one compared to the first one I explained above - at least according to my own estimates)

packages/cactus-plugin-odap-hermes/src/test/typescript/integration/backup-gateway-after-client-crash.test.ts
```diff
@@ -0,0 +1,907 @@
...
+    const {
+      approveForMyOrgList,
+      installList,
+      queryInstalledList,
+      commit,
+      packaging,
+      queryCommitted,
+    } = lifecycle;
+
+    Checks.truthy(packageIds, `packageIds truthy OK`);
```
petermetz, line 414: @AndreAugusto11 Jest has truthiness assertions [1] built in and I would recommend using those instead of the `Checks` class which is meant for production code to validate its preconditions. They kind of produce the same end result here, technically, but with `Checks` the developer experience will be worse because if these assertions fail it will look like as if the test code itself had crashed due to a bug **in the test code itself** not the code that it is verifying to be working properly (very nuanced but important difference when you are trying to debug a failing test case and all you have to go on is some obscure exception message ;-))

[1] https://jestjs.io/docs/using-matchers#truthiness

petermetz (changes requested):

packages/cactus-plugin-odap-hermes/src/main/typescript/gateway/client/transfer-initialization.ts
```diff
@@ -35,7 +35,8 @@ export async function sendTransferInitializationRequest(
     sessionData.lastSequenceNumber == undefined ||
     sessionData.sourceGatewayDltSystem == undefined ||
     sessionData.recipientGatewayPubkey == undefined ||
-    sessionData.recipientGatewayDltSystem == undefined
+    sessionData.recipientGatewayDltSystem == undefined ||
```
petermetz, line 38: @AndreAugusto11 Adding these additional manadatory input parameters made me think: Is this a breaking API change? Let's say someone deployed the ODAP plugin as it was in `v1.0.0` and now has code running in production that sends requests to it (some business application). And now let's say next week we issue a new release that has this code, they upgrade, try to send their "old" requests without these mandatory parameters and start getting errors. 

Please let me know based on the above example if this is a breaking change or not and then we can take it from there.

AndreAugusto11: Hey @petermetz it won't break because this parameter is optional in the plugin options, and if not provided, the list will be initialized as empty.

In the constructor ([line 252 in the new plugin-odap-gateway.ts file](https://github.com/hyperledger/cactus/blob/83c96e727ac1ad15c38b243059667b3d7d2f9233/packages/cactus-plugin-odap-hermes/src/main/typescript/gateway/plugin-odap-gateway.ts#L252).

In the session data ([line 1140 in the new plugin-odap-gateway.ts file](https://github.com/hyperledger/cactus/blob/83c96e727ac1ad15c38b243059667b3d7d2f9233/packages/cactus-plugin-odap-hermes/src/main/typescript/gateway/plugin-odap-gateway.ts#L1140)).

packages/cactus-plugin-odap-hermes/src/test/typescript/integration/backup-gateway-after-client-crash.test.ts
AndreAugusto11: Fixed! I also changed all the other tests where I was using it incorrectly. Thank you

packages/cactus-plugin-odap-hermes/README.md
AndreAugusto11: It makes sense. The only one pointing to the main branch is the `odap-rollback.test` since I can only know the commit hash after the PR is merged

packages/cactus-plugin-odap-hermes/src/test/typescript/integration/backup-gateway-after-client-crash.test.ts
AndreAugusto11: I also changed this in the other tests where I was using`Checks`

AndreAugusto11:
@petermetz do you know why the `CodeQL` check is not passing?
Additionally, in the `Cactus_CI / build-node-14` some tests are causing Jest to exceed its timeout. What could be the reason for this?

RafaelAPB:
Looks good to me. Good job.

petermetz:
> @petermetz the `Cactus_CI / build-node-14` is failing because of an error: `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`. Do you know what could be the reason for this?

@AndreAugusto11 Yeah, more context here: https://github.com/hyperledger/cactus/pull/2009

packages/cactus-plugin-odap-hermes/src/main/typescript/gateway/client/transfer-initialization.ts
petermetz: @AndreAugusto11 Alright, makes sense.

petermetz (approved):
@AndreAugusto11 LGTM, thank you.

takeutak (approved):

takeutak:
Sorry to review late.  LGTM.

Issue #2008, opened 2022-05-09
test: fix jest out of memory errors on Node v16.14.2


petermetz:
## Summary

The Jest tests are failing with out of memorry errors after upgrading the CI to Node v16.14.2.
The NodeJS version upgrade seems to be the cause, but we cannot be entirely sure.
Upgrading Jest might fix the problem. If not we might be able to hack it together by giving a larger heap allocation to the text execution process spawned by Jest (there must be a parameter for it somewhere in the Jest CLI docs - I would be surprised if there wasn't one)

## Logs

Full log file: [2022-05-08-cactus-ci-fail-jest-heap-oom.log](https://github.com/hyperledger/cactus/files/8648082/2022-05-08-cactus-ci-fail-jest-heap-oom.log)

```sh
2022-04-29T18:38:18.4351978Z   ● Console
2022-04-29T18:38:18.4352471Z 
2022-04-29T18:38:18.4359461Z     console.info
2022-04-29T18:38:18.4360850Z       [2022-04-29T18:38:14.639Z] INFO (PluginKeychainMemory): Created PluginKeychainMemory. KeychainID=6784d25f-a7bb-4545-8632-c6bc43419422
2022-04-29T18:38:18.4361452Z 
2022-04-29T18:38:18.4363168Z       at Logger.info (node_modules/loglevel-plugin-prefix/lib/loglevel-plugin-prefix.js:106:24)
2022-04-29T18:38:18.4363839Z 
2022-04-29T18:38:18.4364310Z     console.warn
2022-04-29T18:38:18.4365155Z       [2022-04-29T18:38:14.642Z] WARN (PluginKeychainMemory): Never use PluginKeychainMemory in production. It does not support encryption. It stores everything in plain text.
2022-04-29T18:38:18.4365507Z 
2022-04-29T18:38:18.4365593Z       52 |
2022-04-29T18:38:18.4365884Z       53 |   public warn(...msg: unknown[]): void {
2022-04-29T18:38:18.4366197Z     > 54 |     this.backend.warn(...msg);
2022-04-29T18:38:18.4366450Z          |                  ^
2022-04-29T18:38:18.4366670Z       55 |   }
2022-04-29T18:38:18.4366942Z       56 |   public info(...msg: unknown[]): void {
2022-04-29T18:38:18.4367259Z       57 |     this.backend.info(...msg);
2022-04-29T18:38:18.4367423Z 
[...]

PR #2009, opened 2022-05-09
test: fix jest out of memory errors on Node v16.14.2

petermetz:
[test: fix jest out of memory errors on Node v16.14.2](https://github.com/hyperledger/cactus/pull/2009/commits/756d80f7a5249cda681e40dce407b19fb681d172)

Explicitly specify the heap size (RAM) for the Jest process
that executes the tests to be 3 GB instead of the default
2 GB that NodeJS allocates by default.

This is not a good solution, but a workaround instead.

The test runner itself using this much RAM is the acutal
problem that we need to fix longer term and I'll be opening
a follow-up task specifically to address that and once that
has been fixed we can lower back the test runner's
heap size to the default 2 GB (which is still a lot BTW)

Not sure why the out of memory errors starting occuring only
on NodeJS 16.14.2, probably something made it use
slightly more memory and it finally spilled over.
My assumption is that the memory usage of the test runner
was bad already regardless.

Fixes https://github.com/hyperledger/cactus/issues/2008

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "webpack": "lerna run webpack:dev",
     "webpack:dev:web": "lerna run webpack:dev:web",
     "webpack:dev:node": "lerna run webpack:dev:node",
-    "test:jest:all": "jest",
+    "test:jest:all": "NODE_OPTIONS=--max_old_space_size=3072 jest",
     "test:tap:all": "tap",
     "test:all": "yarn test:jest:all && yarn test:tap:all",
     "posttest:all": "tap --no-check-coverage --coverage-report=lcov && codecov",


jagpreetsinghsasan (approved):
LGTM, I see this option in another PR #2015 (I think its mentioned there to bypass the merge conflicts, is that right?)

petermetz:
> LGTM, I see this option in another PR #2015 (I think its mentioned there to bypass the merge conflicts, is that right?)

@jagpreetsinghsasan Yeah I have the same change in the other PR temporarily because without this there is no way to verify if what #2015 does actually works. When we merge this, then rebase #2015 that change will get disappeared so it won't be in the diff of #2015 anymore.

takeutak (approved):

PR #2011, opened 2022-05-10
docs(examples): explain supply chain live development

### LINKED ISSUE #2010: docs(examples): explain supply chain live development
We should have documentation for how to develop the supply chain app front-end with live-reloading enabled.
The ionic serve command has a built-in reverse-proxy feature that makes this possible quite easily with
the right configuration in place.

How to achieve this is explained in the README.md file of the supply chain app and the necessary updates
were made so that it actually works.

petermetz:
We should have documentation for how to develop the
supply chain app front-end with live-reloading enabled.
The ionic serve command has a built-in reverse-proxy
feature that makes this possible quite easily with
the right configuration in place.

How to achieve this is explained in the README.md
file of the supply chain app and the necessary updates
were made so that it actually works.

Fixes #2010

### DIFF
--- a/examples/cactus-example-supply-chain-backend/package.json
+++ b/examples/cactus-example-supply-chain-backend/package.json
@@ -65,6 +65,7 @@
     "@hyperledger/cactus-test-tooling": "1.0.0",
     "async-exit-hook": "2.0.1",
     "axios": "0.21.4",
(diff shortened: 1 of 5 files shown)


jagpreetsinghsasan (approved):
LGTM

takeutak (approved):

PR #2012, opened 2022-05-10
test: jestify jwt endpoint authz scope enforcement

Leeyoungone:
File Path:
packages/cactus-cmd-api-server/src/test/typescript/integration/jwt-endpoint-authz-scope-enforcement.test.ts

### DIFF
--- a/packages/cactus-cmd-api-server/src/test/typescript/integration/jwt-endpoint-authz-scope-enforcement.test.ts
+++ b/packages/cactus-cmd-api-server/src/test/typescript/integration/jwt-endpoint-authz-scope-enforcement.test.ts
@@ -1,11 +1,10 @@
-import test, { Test } from "tape-promise/tape";
+import "jest-extended";
 import { v4 as uuidv4 } from "uuid";
 import { generateKeyPair, exportSPKI, SignJWT } from "jose";
 import expressJwt from "express-jwt";
 import axios, { Method } from "axios";
-import { StatusCodes } from "http-status-codes";
 
-import { LoggerProvider, LogLevelDesc } from "@hyperledger/cactus-common";
+import { LogLevelDesc } from "@hyperledger/cactus-common";
 import { PluginRegistry } from "@hyperledger/cactus-core";
 
 import {
@@ -23,14 +22,18 @@ import { UnprotectedActionEndpoint } from "../fixtures/plugin-ledger-connector-s
 const testCase =
   "API server enforces scope requirements on top of generic authz";
 const logLevel: LogLevelDesc = "TRACE";
-const log = LoggerProvider.getOrCreate({
-  level: logLevel,
-  label: __filename,
-});
 
-test(testCase, async (t: Test) => {
-  try {
-    const jwtKeyPair = await generateKeyPair("RS256", { modulusLength: 4096 });
+describe(testCase, () => {
+  let apiServer: ApiServer;
+
+  afterAll(async () => {
+    await apiServer.shutdown();
+  });
+
+  test(testCase, async () => {
+    const jwtKeyPair = await generateKeyPair("RS256", {
+      modulusLength: 4096,
+    });
     const jwtPublicKey = await exportSPKI(jwtKeyPair.publicKey);
     const expressJwtOptions: expressJwt.Options = {
       algorithms: ["RS256"],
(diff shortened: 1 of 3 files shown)


packages/cactus-cmd-api-server/src/test/typescript/integration/jwt-endpoint-authz-scope-enforcement.test.ts
```diff
@@ -72,18 +76,14 @@ test(testCase, async (t: Test) => {
...
-    test.onFinish(async () => await apiServer.shutdown());
-
     const startResponse = apiServer.start();
-    await t.doesNotReject(
-      startResponse,
-      "failed to start API server with dynamic plugin imports configured for it...",
-    );
-    t.ok(startResponse, "startResponse truthy OK");
+
+    expect(startResponse).not.toReject();
```
petermetz, line 85: @Leeyoungone This needs to be awaited for.

packages/cactus-cmd-api-server/src/test/typescript/integration/jwt-endpoint-authz-scope-enforcement.test.ts
```diff
@@ -142,22 +139,12 @@ test(testCase, async (t: Test) => {
...
-      t.ok(out, "error thrown for forbidden endpoint truthy OK");
-      t.ok(out.response, "deploy contract response truthy OK");
-      t.equal(
-        out.response.status,
-        StatusCodes.FORBIDDEN,
-        "deploy contract response status === 403 OK",
-      );
-      t.notok(out.response.data.data, "out.response.data.data falsy OK");
-      t.notok(out.response.data.success, "out.response.data.success falsy OK");
+    } catch (out: any) {
```
petermetz, line 142: @Leeyoungone This can be made to work with `unkownn` as the type of the exception instead of `any`. Do you remember that pattern I showed earlier? Not sure when it was TBH.

petermetz (changes requested):

packages/cactus-cmd-api-server/src/test/typescript/integration/jwt-endpoint-authz-scope-enforcement.test.ts
Leeyoungone: got it! thank you

packages/cactus-cmd-api-server/src/test/typescript/integration/jwt-endpoint-authz-scope-enforcement.test.ts
```diff
@@ -142,22 +139,16 @@ test(testCase, async (t: Test) => {
...
-        StatusCodes.FORBIDDEN,
-        "deploy contract response status === 403 OK",
-      );
-      t.notok(out.response.data.data, "out.response.data.data falsy OK");
-      t.notok(out.response.data.success, "out.response.data.success falsy OK");
+    } catch (out: unknown) {
+      if (axios.isAxiosError(out)) {
+        expect(out).toBeTruthy();
+        expect(out.response).toBeTruthy();
+        expect((out as any).response.status).toBe(StatusCodes.FORBIDDEN);
```
petermetz, line 146: @Leeyoungone Please no casting to `any`. You can use the `toHaveProperty` assertions instead: https://stackoverflow.com/a/47904961

petermetz (changes requested):

packages/cactus-cmd-api-server/src/test/typescript/integration/jwt-endpoint-authz-scope-enforcement.test.ts
Leeyoungone: Got it! I've refactored so that the test for the status code uses the `toHaveProperty`, however there is a way to also use `toBe` and the `StatusCodes.FORBIDDEN` if we wanted to keep the code consistent with what we've had before. Below, I've listed the two ways to do this 


-- using the `toBe` and `StatusCodes.FORBIDDEN`

``` Typescript
expect(out.response?.status).toBe(StatusCodes.FORBIDDEN);
```


-- using the `toHaveProperty`
``` Typescript
expect(out).toHaveProperty(["response", "status"], 403);
```

Please let me know which one you'd prefer! For now I'm using the `toHaveProperty` because of your suggestion above.

petermetz (approved):
> Please let me know which one you'd prefer! For now I'm using the `toHaveProperty` because of your suggestion above.  

@Leeyoungone Yep, I prefer it this way, thank you! :-)

takeutak (approved):

PR #2015, opened 2022-05-10
test(jest): upgrade jest,jest-extended to latest

### LINKED ISSUE #2014: test(jest): upgrade jest,jest-extended to latest
`jest-extended` now has a `2.0.0` stable release instead of the beta that we were using earlier.
There also has been a major release of `jest` itself which contains some performance improvements
and therefore we should invest a little time into upgrading these now instead of having a higher
migration cost incurred later.

petermetz:
Bump the dependencies and update the yarn.lock file
accordingly.

Fixes #2014


jagpreetsinghsasan (approved):
LGTM

jagpreetsinghsasan:
I was checking, we have one more feature
`--optimize-for-size` (optimizes for the available memory instead of execution speed) 
So while looking for optimizations, shall we trade for execution speed for available memory? (given, it doesnt increase the execution by a significant amount)

takeutak (approved):

Issue #1995, opened 2022-04-27
test(connector-corda): fix flaky monitor transactions v4.8


petermetz:
**Describe the bug**

The `packages/cactus-plugin-ledger-connector-corda/src/test/typescript/integration/monitor-transactions-v4.8.test.ts` test case is flaky, meaning that it fails on occasion due to unknown problems (usually race conditions in the code, but it could anything really...)

**To Reproduce**

Run the test enough times so that it eventually fails despite having succeeded multiple times before and after the failed execution.

**Expected behavior**

The test is stable, either always fails or always succeeds.

**Logs/Stack traces**

Full Log file: [2022-04-27-failed-ci-monitor-transactions-v4.8.test.ts.log](https://github.com/hyperledger/cactus/files/8577968/2022-04-27-failed-ci-monitor-transactions-v4.8.test.ts.log)

Stack trace:

```sh
2022-04-27T18:32:43.7305546Z 
2022-04-27T18:32:43.7306859Z Summary of all failing tests
2022-04-27T18:32:43.7315078Z FAIL packages/cactus-plugin-ledger-connector-corda/src/test/typescript/integration/monitor-transactions-v4.8.test.ts (379.503 s)
2022-04-27T18:32:43.7318563Z   ● Monitor Tests › Multiple clients tests › State change can be read by all listening clients separately
2022-04-27T18:32:43.7319014Z 
2022-04-27T18:32:43.7321198Z     expect(received).toEqual(expected) // deep equality
2022-04-27T18:32:43.7322008Z 
2022-04-27T18:32:43.7322426Z     Expected: "net.corda.samples.example.states.IOUState"
2022-04-27T18:32:43.7322918Z     Received: null
2022-04-27T18:32:43.7323244Z 
2022-04-27T18:32:43.7323470Z       693 |       });
2022-04-27T18:32:43.7323945Z       694 |       expect(resGetTxSecondClient.status).toBe(200);
2022-04-27T18:32:43.7328293Z     > 695 |       expect(resGetTxSecondClient.data.stateFullClassName).toEqual(
2022-04-27T18:32:43.7331514Z           |                                                            ^
2022-04-27T18:32:43.7333233Z       696 |         stateToMonitor,
2022-04-27T18:32:43.7334816Z       697 |       );
2022-04-27T18:32:43.7338966Z       698 |       expect(resGetTxSecondClient.data.tx?.length).toBe(
2022-04-27T18:32:43.7339350Z 
[...]

PR #2021, opened 2022-05-12
test(connector-corda): fix flaky monitor transactions v4.8

outSH:
Fix race condition that caused user session removal during request processing,
that caused test to fail occasionaly.

Correlated fixes / improvements:
- Use paging when watching for vault changes.
This was necessary to perform long-running runs of a failing test.
  Without it, start monitor of a state with 200+ changes was failing instantly.
  I thought that paging was needed when there are 200+ changes in single observer push,
  but it seems that it breaks when total number of changes are that large,
  which makes it more critical to fix.
- To prevent deadlock of cleanup thread I introduced thread pool to corda kotlin server.
  Number of threads is configurable in environment variable.
  Updated readme to show this option.

Closes: #1995

### DIFF
--- a/packages/cactus-plugin-ledger-connector-corda/README.md
+++ b/packages/cactus-plugin-ledger-connector-corda/README.md
@@ -329,6 +329,8 @@ const res = await apiClient.invokeContractV1({
 ```json
 {
   "cactus": {
+    "threadCount": 3,
+    "sessionExpireMinutes": 10,
     "corda": {
       "node": {
         "host": "localhost"
--- a/packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/Application.kt
+++ b/packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/src/main/kotlin/org/hyperledger/cactus/plugin/ledger/connector/corda/server/Application.kt
@@ -5,18 +5,23 @@ import com.fasterxml.jackson.databind.SerializationFeature
 import net.corda.client.jackson.JacksonSupport
 import org.hyperledger.cactus.plugin.ledger.connector.corda.server.impl.NodeRPCConnection
 import org.springframework.beans.factory.annotation.Autowired
-import org.springframework.boot.runApplication
-import org.springframework.context.annotation.ComponentScan
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.runApplication
 import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.ComponentScan
 import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter
+import org.springframework.scheduling.TaskScheduler
 import org.springframework.scheduling.annotation.EnableScheduling
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler
 
+private const val ThreadPoolCount = "cactus.threadCount"
+private const val ThreadPoolCountDefault = "3"
 
 @SpringBootApplication
 @ComponentScan(basePackages = ["org.hyperledger.cactus.plugin.ledger.connector.corda.server", "org.hyperledger.cactus.plugin.ledger.connector.corda.server.api", "org.hyperledger.cactus.plugin.ledger.connector.corda.server.model"])
 @EnableScheduling
-class Application {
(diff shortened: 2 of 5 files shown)


outSH:
@petermetz @izuru0 Please review.

Also, please upload image `ghcr.io/outsh/cactus-connector-corda-server:corda-dev-4` to hyperledger ghcr so I can update the reference in the test, and remove my name from cspell exclusion.

petermetz:
> @petermetz @izuru0 Please review.

Also, please upload image `ghcr.io/outsh/cactus-connector-corda-server:corda-dev-4` to hyperledger ghcr so I can update the reference in the test, and remove my name from cspell exclusion.

@outSH Thank you for fixing the flake, much appreciated! Where is the diff for the `Dockerfile` changes?

petermetz (changes requested):
> > Thank you for fixing the flake, much appreciated! Where is the diff for the `Dockerfile` changes?

There are no Dockerfile changes, I've changed files packed in `cactus-connector-corda-server` image that is used in the test.
 
@outSH Could you please make the changes in the Dockerfile so that the image you created is reproducible via the regular build?

petermetz (changes requested):
> @petermetz Sorry, I don't understand. What changes in Dockerfile? Why the image wouldn't be reproducible via the regular build? 

@outSH No worries, I'm probably not explaining myself right (English isn't my first language so trouble sometimes ensues)

Let me try it again. 
Let's say we have image A (which is the image you get if you build the current dockerfile with the command from the readme).
Then let's say you have image B which is image A plus your changes ** in it.
The changes in image B are not done through the dockerfile and therefore if next week we change something in the dockerfile and publish image C, we would expect that image to be usable in place of image B and that everything would just work, but it wouldn't (from what I understand) because the changes you made in image B were done manually, not through the dockerfile and therefore when someone builds image C, they are not getting the changes that you made for B.

The key here is that technically image B is reproducible only if you do the regular build and then whatever manual steps you did** and what the expectation would be that image B (or any image that we have) is reproducible without manual extra steps apart from the build command itself.

** `I've changed files packed in cactus-connector-corda-server image that is used in the test.`

outSH:
> The changes in image B are not done through the dockerfile and therefore if next week we change something in the dockerfile and publish image C

@petermetz 
The changes are replicable in dockerfile, but current image in hyperledger ghcr is older than one from this PR. I'm afraid I might be missing something important here - otherwise I'd just ask Izuru-san to upload this ;) - so I'll explain it in more detail:
- I have a functional test [monitor-transactions-v4.8.test.ts](https://github.com/hyperledger/cactus/pull/2021/files#diff-8f6f872134f8bc99de3d43129a25c4921928a9307314e8605a172498284f1709)
- This test used to pull `cactus-connector-corda-server` image from hyperledger ghcr, which contains kotlin connector (`packages/cactus-plugin-ledger-connector-corda/src/main-server`). This image is based on sources from couple commits ago.
- I've done some changes in `cactus-plugin-ledger-connector-corda/src/main-server` source files.
- If I run the test now, it would still use upstream ghcr  kotlin-server from couple commits ago, i.e. not including my changes.
- For this reason, I've built a new image, using the same, unmodified dockerfile, to have an image that represents kotlin-server state from this commit (i.e. including my changes).
- I've uploaded them to my ghcr, and updated the test, so that now the functional test uses the kotlin-server from this commit.
- I need to have this image uploaded to hyperledger ghcr, which will contain the source version of kotlin-server from this commit.
- I think normally this image would be updated with this PR sources after merge, but I want to have this earlier so we don't have to commit a test with reference to my GHCR.

petermetz (changes requested):
@outSH Oh, okay, now I get it. Sorry, I thought you meant you were doing the changes with the `docker commit` command and that's why you wanted me to upload your custom image directly to ghcr.
If we are talking about me rebuilding the image myself and then uploading it, then it's all good of course.

Please update your image tag to this one and then it's LGTM: `ghcr.io/hyperledger/cactus-connector-corda-server:2022-05-26-0ff7407--pr-2021`

outSH:
> @outSH Oh, okay, now I get it. Sorry, I thought you meant you were doing the changes with the `docker commit` command and that's why you wanted me to upload your custom image directly to ghcr. If we are talking about me rebuilding the image myself and then uploading it, then it's all good of course.

Oh, OK, now I see I'v asked you to upload my image from my `ghcr.io/outsh`, that was misleading, my bad. Good that we came to an understanding, eventually ;)

I've rebased, updated the image using in the test and fixed cspell ignore file. Should be good to go now.

petermetz (approved):
@outSH Haha, yeah we got it all together in the end! :-)  LGTM

izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):
LGTM

PR #2023, opened 2022-05-12
docs(connector-besu): fix besu AIO image tag in README.md

### LINKED ISSUE #2022: docs(connector-besu): fix besu AIO image tag in README.md
This guide is broken at the moment:
https://github.com/hyperledger/cactus/tree/main/packages/cactus-plugin-ledger-connector-besu#testing-api-calls-with-the-container

Make sure it works.
The first mistake I noticed was that the container image tag is no longer
working because we moved from DockerHub to ghcr.io.

petermetz:
1. The besu connector container was using an outdated version
of the base image (which is the cmd-api-server image)
This was leading to the API server not recognizing the
besu plugin as a web service plugin and then not installing
the endpoints due to that.
Fixed this by specifying the current latest image tag of
the API server image:
ghcr.io/hyperledger/cactus-cmd-api-server:2022-05-12-2330a96

2. The environment variables for disabling authorization and gRPC TLS
were missing from the command that runs the besu connector
connector container. Adding those made it work and the cURL
requests from the example is now succeeding as expected.

Fixes #2022

### DIFF
--- a/packages/cactus-plugin-ledger-connector-besu/Dockerfile
+++ b/packages/cactus-plugin-ledger-connector-besu/Dockerfile
@@ -1,5 +1,4 @@
-FROM ghcr.io/hyperledger/cactus-cmd-api-server:2021-08-15--refactor-1222
-
+FROM ghcr.io/hyperledger/cactus-cmd-api-server:2022-05-12-2330a96
 RUN npm install -g yarn@1.22.17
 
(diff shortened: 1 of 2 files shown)


izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):
LGTM

PR #2026, opened 2022-05-15
test: jestify remote plugin imports 

Leeyoungone:
Migrated test from Tap to Jest

File Path:
packages/cactus-test-cmd-api-server/
src/test/typescript/integration/
remote-plugin-imports.test.ts

### DIFF
--- a/packages/cactus-test-cmd-api-server/src/test/typescript/integration/remote-plugin-imports.test.ts
+++ b/packages/cactus-test-cmd-api-server/src/test/typescript/integration/remote-plugin-imports.test.ts
@@ -1,4 +1,4 @@
-import test, { Test } from "tape-promise/tape";
+import "jest-extended";
 import { v4 as uuidv4 } from "uuid";
 
 import {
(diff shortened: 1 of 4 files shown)


takeutak (approved):

packages/cactus-test-cmd-api-server/src/test/typescript/integration/remote-plugin-imports.test.ts
```diff
@@ -27,86 +27,101 @@ import {
...
+  let pluginContainer: any;
 
-  test.onFinish(async () => {
-    await vaultTestContainer.stop();
-    await vaultTestContainer.destroy();
+  afterAll(async () => {
+    vaultTestContainer.stop();
+    vaultTestContainer.destroy();
+    apiServer.shutdown();
+    pluginContainer.stop();
```
petermetz, line 41: @Leeyoungone Please await for the stop to finish before calling destroy. Otherwise it might be a race condition to run the two concurrently.

packages/cactus-test-cmd-api-server/src/test/typescript/integration/remote-plugin-imports.test.ts
```diff
@@ -27,86 +27,101 @@ import {
...
+  let apiServer: ApiServer;
+  let pluginContainer: any;
 
-  test.onFinish(async () => {
-    await vaultTestContainer.stop();
-    await vaultTestContainer.destroy();
+  afterAll(async () => {
+    vaultTestContainer.stop();
+    vaultTestContainer.destroy();
+    apiServer.shutdown();
```
petermetz, line 40: @Leeyoungone I would probably just await all of these so that they are in sequence and therefore much easier to debug if one of them goes wrong.

packages/cactus-test-cmd-api-server/src/test/typescript/integration/remote-plugin-imports.test.ts
```diff
@@ -27,86 +27,101 @@ import {
...
-  const factory = new PluginFactoryKeychain({
-    pluginImportType: PluginImportType.Remote,
-  });
-
-  const plugin: IPluginKeychain = await factory.create({
-    keychainId: "_keychainId_",
-    instanceId: "_instanceId_",
-    remoteConfig: configuration,
+  test(testCase, async () => {
+    try {
```
petermetz, line 46: @Leeyoungone Why did the try/catch block needed to be added when in the original test code it wasn't there?

petermetz (changes requested):

packages/cactus-test-cmd-api-server/src/test/typescript/integration/remote-plugin-imports.test.ts
Leeyoungone: I think jest got mad about not catching a failed promise or something (even though all the test were passing then too). So that's why I added the `try-catch` however, it works now (probably because I `awaited` everything within the afterAll ? (I'm not too sure but if it works I'm not gonna poke it).

petermetz (changes requested):
@Leeyoungone LGTM once you squash the commits into one. ;-)

jagpreetsinghsasan (approved):

petermetz (approved):
@Leeyoungone LGTM, thank you!

Issue #2029, opened 2022-05-17
refactor(examples): include sample apps in monorepo build


outSH:
- Build electricity-trade and discounted-cartrade during main build. Include these apps in monorepo setup.
- This will solve recent build errors (single dependency management).
- Additional changes are required to solve this, they will be described in the PR itself.

PR #2030, opened 2022-05-17
refactor(examples): include sample apps in monorepo build

outSH:
- Include discounted-cartrade and electricity trade in monorepo setup,
- Update the devcontainer and include new environment setup instructions in BUILD.md,
- Fix sample apps build error,
- Rename of sample apps to cactus-example-*,
- Change any "car" references to "asset" (in code, comments, documentation, etc..),
- Rename asset-trade files to match the format of the rest of this repository.
- Remove asset-trade eth asset management class and endpoint
  (not used right now, and it would confuse with fabric asset).
- Renamed some uses of cartrade to asset-trade,
  (more complete changes are included in follow-up commit),
- Add indy build artifacts to .gitignore,
- Add cmake to devcontainer build (fixed warning during npm dependency build stage),
- Fix random whitespace and typos errors,

#### BREAKING CHANGE: building discounted-asset-trade app (or any future app that use indy validator) requires Indy SDK to be installed on the build machine.

Closes: #2029

Signed-off-by: Michal Bajer <michal.bajer@fujitsu.com


outSH:
@petermetz @izuru0 Please review.

outSH:
@petermetz I wrongly assumed that CI runs inside `.devcontainer/Dockerfile` cotainer, but it doesn't. Do you have an idea on how to add dependency (Indy SDK) to CI build environment? Install it in `ci.sh` (it set up npm and yarn already, not sure if OS packages fit here as well), or maybe use .devcontainer for CI build?

outSH:
> @petermetz I wrongly assumed that CI runs inside .devcontainer/Dockerfile cotainer, but it doesn't. Do you have an idea on how to add dependency (Indy SDK) to CI build environment? Install it in ci.sh (it set up npm and yarn already, not sure if OS packages fit here as well), or maybe use .devcontainer for CI build?

@jagpreetsinghsasan Do you have any ideas on this? I can add that Indy SDK is already required since initial indy support I suppose, but it did not affect the CI earlier since it had to be built manually.

outSH:
> @outSH is this what you are looking for? https://github.com/hyperledger/cactus/blob/main/.github/workflows/ci.yml

@jagpreetsinghsasan My question was: what is your preferred way to install additional dependency in CI environment. Do you mean that I should add additional step in build-node jobs that will install these dependencies?

petermetz (changes requested):
> > @outSH is this what you are looking for? https://github.com/hyperledger/cactus/blob/main/.github/workflows/ci.yml

> @jagpreetsinghsasan My question was: what is your preferred way to install additional dependency in CI environment. Do you mean that I should add additional step in build-node jobs that will install these dependencies?
 

@outSH Sorry for the slow response. The preferred way is to do it declaratively (same as we do for NodeJS).
I found an GH workflow action that is pre-built to do what you want (AFAICT) here: https://github.com/actions/setup-python

If you find it too troublesome to do it this way I would also be okay with hacking it in ci.sh as long as you open a follow-up issue in the tracker so that later on someone gets it done in the ci.yml file with the action (or some other action) declaratively. 
I can also offer to help with implementation of the workflow action, you can find me on the daily pair programming calls every morning that's on the HL Cactus calendar so if you stop by there we can collaborate live too.

outSH:
> Sorry for the slow response. The preferred way is to do it declaratively (same as we do for NodeJS).
- That was my assumpion as well, that's why I've proposed using `.devcontainer` image for CI build as well. This'd have an advantage of having environment setup logic in one place (in case we need to add yet another dependency, we have to edit both `ci.yml` and `.devcontainer` now), but after all I've followed your advice and edited `ci.yml` to speed up the process.

> I found an GH workflow action that is pre-built to do what you want (AFAICT) here: https://github.com/actions/setup-python
- I don't want to setup python, I want to install deb package. I've added plain setup scripts in `ci.yml`, please review the last commit if it's satisfactory and I'll squash it with the main commit.

takeutak:
@petermetz @jagpreetsinghsasan  (cc: @izuru0)
Could you review this PR?

petermetz (approved):
@outSH LGTM, thank you!

jagpreetsinghsasan (approved):

outSH:
@izuru0 @takeutak Finally adjusted the new CI to include Indy SDK, PR is ready to be merged now.

petermetz:
@outSH Thank you and sorry for the conflicts you (I guess) had with my PR.

> That was my assumpion as well, that's why I've proposed using .devcontainer image for CI build as well. This'd have an advantage of having environment setup logic in one place (in case we need to add yet another dependency, we have to edit both ci.yml and .devcontainer now), but after all I've followed your advice and edited ci.yml to speed up the process.

It would be cool if we migrated the CI to run on the dev container, I agree!

Issue #2017, opened 2022-05-11
feat(secret): remove python LP Validator/Verifier secret keys from repository


izuru0:
**Is your feature request related to a problem? Please describe.**

keys for authentication for python LPs are stored in these directories, but it is discouraged to have such files included in the repository:

https://github.com/hyperledger/cactus/tree/main/packages-python/cactus_validator_socketio_indy/sample-CA
https://github.com/hyperledger/cactus/tree/main/packages-python/cactus_validator_socketio_iroha/sample-CA

**Describe the solution you'd like**

Generate these files during build process and remove them from repository

Issue #2016, opened 2022-05-11
feat(secret): remove Validator/Verifier secret keys from repository


izuru0:
**Is your feature request related to a problem? Please describe.**

keys for authentication are stored in these directories, but it is discouraged to have such files included in the repository:

https://github.com/hyperledger/cactus/tree/main/packages/cactus-plugin-ledger-connector-fabric-socketio/sample-config/CA
https://github.com/hyperledger/cactus/tree/main/packages/cactus-plugin-ledger-connector-go-ethereum-socketio/sample-config/CA
https://github.com/hyperledger/cactus/tree/main/packages/cactus-plugin-ledger-connector-sawtooth-socketio/sample-config/CA

**Describe the solution you'd like**

Generate these files during build process and remove them from repository

PR #2033, opened 2022-05-20
feat(secret): remove Validator/Verifier secret keys from repository

outSH:
- Remove validator sample CA keys hardcoded inside the repository.
- Generate fresh ECDSA keys when starting up electricity-trade
  or discounted-asset-trade sample apps.
- Add support for RSA CA keys in fabric-socketio validator.
  I couldn't find any trivial way of generating ECDSA self-signed certificate
  (without calling openssl cmdline, which seems poor from functional test perspective),
  so I've added support for RSA keys to simplify the tests.
- Allow selection of jwt algorithm in  fabric-socketio validator.
  It must correspond to the key used.
- Update the READMEs, add short description of SSL config option of fabric-socketio validator.

Closes: #2016
Closes: #2017

Depends on #1977
Depends on #2030

--------------

I've included commits from the dependencies to simplify the review before they're merged to main. Please review only the last commit: [feat(secret): remove Validator/Verifier secret keys from repository](https://github.com/hyperledger/cactus/commit/037100e992328cb49213b4138d9324fe419c0630)


izuru0 (changes requested):
@outSH The CI script is complaining that pull request title " feat(secret): remove Validator/Verifier secret keys from repository" isn't following the convention. It appears there is a leading space character before 'feat'. Can you fix this?
@jagpreetsinghsasan
There is also a test that failed (GitGuardian Security Checks), but I suspect that this is a false positive. It is complaining that there are four secret keys in this PR, but this PR is removing those files. So I don't understand why it is complaining. Can you investigate this problem? thanks.

jagpreetsinghsasan:
> There is also a test that failed (GitGuardian Security Checks), but I suspect that this is a false positive. It is complaining that there are four secret keys in this PR, but this PR is removing those files. So I don't understand why it is complaining. Can you investigate this problem? thanks.

This is happening because the PR has 3 commits within, where the [most earliest commit](https://github.com/hyperledger/cactus/commit/4ec5b883cd0310055fd5681451cc210746548195) do contains crypto material. I am sure that a squash merge of the 3 commits shall resolve this.

izuru0:
> > There is also a test that failed (GitGuardian Security Checks), but I suspect that this is a false positive. It is complaining that there are four secret keys in this PR, but this PR is removing those files. So I don't understand why it is complaining. Can you investigate this problem? thanks.
> 
> This is happening because the PR has 3 commits within, where the [most earliest commit](https://github.com/hyperledger/cactus/commit/4ec5b883cd0310055fd5681451cc210746548195) do contains crypto material. I am sure that a squash merge of the 3 commits shall resolve this.

ah you are totally right. Thanks for correction.

packages/cactus-plugin-ledger-connector-fabric-socketio/package.json
```diff
@@ -30,6 +34,8 @@
     "socket.io": "4.4.1"
   },
   "devDependencies": {
+    "@hyperledger/cactus-api-client": "1.0.0",
+    "@hyperledger/cactus-cmd-api-server": "1.0.0",
```
petermetz, line 38: @outSH The API server does not explicitly depend on the plugins, but it is used to install them at runtime so I would recommend not creating a (hidden) circular dependency this way. 
When I made the self signed PKI generator I was lazy and didn't put it in a separate package but I should've so if you want to duplicate it in one of the test packages (test-tooling is my recommendation) then we can create an issue to remove the old one from the API server package in the future (breaking change, v2.0.0, etc.) and until then it can be just marked as deprecated.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-fabric-socketio/package.json
outSH: socketio-based validators has nothing to do with `cactus-cmd-api-server` plugin system, but I agree that it's good to keep up a convention, and that this logic belongs somewhere else. I've moved the PKI generation to test-tooling, and marked the old implementation as deprecated. Please review the last commit, I'll squash it if everything is OK.

izuru0 (approved):
LGTM

packages/cactus-plugin-ledger-connector-fabric-socketio/package.json
petermetz: @outSH Sounds good, thank you for adjusting and very sorry for the slow review. I'm trying to bring the CI back online.

petermetz (changes requested):
@outSH Please fix the merge conflicts + squash and then we should be good to go

jagpreetsinghsasan (approved):

outSH:
> @outSH Please fix the merge conflicts + squash and then we should be good to go

@petermetz Thanks for the review :) There's still one dependency that needs to me merged first - https://github.com/hyperledger/cactus/pull/2030. I will do the squash after that

outSH:
@petermetz @izuru0 @takeutak This could be merged but `GitGuardian Security Checks` complains about keys that were removed by this PR. Do you have any solution for this, or can this be ignored and merged with failing check anyway?

petermetz (approved):
@outSH LGTM, thank you!

> @petermetz @izuru0 @takeutak This could be merged but GitGuardian Security Checks complains about keys that were removed by this PR. Do you have any solution for this, or can this be ignored and merged with failing check anyway?

Yes, in my opinion. 
The Git Guardian checks are experimental for now because the ignore patterns are something that is not yet fully configurable and the examples/test code trips it up sometimes (like in this case).
So we are good to merge IMO, regardless of the GitGuardian check. Once we figured out all the details of how to make it work properly, we can set it up so that it becomes a hard stop for merging PRs, but that will only happen in the future.

petermetz:
FYI: I disabled the git guardian check being required for now.

PR #2035, opened 2022-05-22
test: jestify containers in test tooling

Leeyoungone:
File Path:
packages/cactus-test-
tooling/src/test/
typescript/integration/
common/
containers.test.ts


takeutak (approved):

packages/cactus-test-tooling/src/test/typescript/integration/common/containers.test.ts
```diff
@@ -12,114 +12,118 @@ import {
...
+            port: 9999,
+            socketPath: badSocketPath,
+          },
+        });
+        fail(Containers.getDiagnostics);
+      } catch (ex) {
+        expect(ex).toBeTruthy();
+        expect(ex.cause).toBeTruthy();
+        expect(ex.cause.message).toBeTruthy();
+        const causeMsgIsInformative = (ex as any).cause.message.includes(
```
petermetz, line 122: @Leeyoungone Could you refactor this to not use `any`? I'm thinking of something like what they are showing off here: https://github.com/facebook/jest/issues/7284#issuecomment-433426694

petermetz (changes requested):

packages/cactus-test-tooling/src/test/typescript/integration/common/containers.test.ts
Leeyoungone: Hello Peter! Thanks for the link, I went ahead and implemented this test to use `toHaveProperty`for this.

`connect ENOENT /some-non-existent-path/to-make-it-trip-up/`

However, to use the `toHaveProperty` I had to add two more lines of code (shown below) 

``` typescript
const connectEN = "connect ENOENT ";
const fullBadPath = connectEN.concat(badSocketPath);
expect(ex).toHaveProperty(["cause", "message"], fullBadPath);
```
and this is because the path `ex.cause.message` had `connect ENOENT /some-non-existent-path/to-make-it-trip-up/` instead of just `/some-non-existent-path/to-make-it-trip-up/`. Using the `toHaveProperty` does work, it's just a few more lines. 

Another way to do this with less lines by using `toMatch`, here it is below 

``` typescript 
expect(ex.cause.message).toMatch(badSocketPath);
```
Both of them work, I just wanted to ask which one you'd prefer before I refactor this test case and push again!!

petermetz: @Leeyoungone Sorry for the very slow response! The simpler one liner[1] should work fine in this case because as long as the message does contain the socket path, we are okay (so we don't need an exact match).

[1] `expect(ex.cause.message).toMatch(badSocketPath);`

petermetz: > @Leeyoungone Sorry for the very slow response! The simpler one liner[1] should work fine in this case because as long as the message does contain the socket path, we are okay (so we don't need an exact match).
> 
> [1] `expect(ex.cause.message).toMatch(badSocketPath);`

@Leeyoungone Sorry, nvm, I just checked it and it won't compile with the second option when the exception type is defined as `unknown` so better to go with the first option. I made the changes quickly just now so you don't have to worry about it, no action necessary. Thank you again!

petermetz (approved):
@Leeyoungone LGTM, thank you!

Issue #1941, opened 2022-03-22
feat(cactus-plugin-ledger-connector-iroha): Add monitoring and send sync/async functionality


StepniowskiP:
Add monitoring and send sync/async requests via SocketIO functionality

PR #1966, opened 2022-04-12
Add monitoring and send sync/async functionality

StepniowskiP:
Add monitoring and sending async/sync requests to ledger via SocketIO.

PR #2048, opened 2022-05-26
feat(iroha-connector): implement validator interface

outSH:
feat(iroha-connector): implement validator interface

Add monitoring and sending async/sync requests to ledger via SocketIO.

Closes: https://github.com/hyperledger/cactus/issues/1941

Author: stepniowskip <piotr.stepniowski@fujitsu.com>

Original PR: #1966

### DIFF
--- a/packages/cactus-core-api/src/main/typescript/plugin/ledger-connector/i-socket-api-client.ts
+++ b/packages/cactus-core-api/src/main/typescript/plugin/ledger-connector/i-socket-api-client.ts
@@ -8,17 +8,17 @@ import type { Observable } from "rxjs";
  */
 export interface ISocketApiClient<BlockType> {
   sendAsyncRequest?(
-    args: any,
-    method: Record<string, unknown>,
-    baseConfig?: any,
     contract?: Record<string, unknown>,
+    method?: Record<string, unknown>,
+    args?: any,
+    baseConfig?: any,
   ): void;
 
   sendSyncRequest?(
-    args: any,
-    method: Record<string, unknown>,
-    baseConfig?: any,
     contract?: Record<string, unknown>,
+    method?: Record<string, unknown>,
+    args?: any,
+    baseConfig?: any,
   ): Promise<any>;
 
   watchBlocksV1?(
--- a/packages/cactus-core-api/src/main/typescript/plugin/ledger-connector/i-socket-api-client.ts
+++ b/packages/cactus-core-api/src/main/typescript/plugin/ledger-connector/i-socket-api-client.ts
@@ -8,15 +8,17 @@ import type { Observable } from "rxjs";
  */
 export interface ISocketApiClient<BlockType> {
   sendAsyncRequest?(
-    contract: Record<string, unknown>,
-    method: Record<string, unknown>,
     args: any,
+    method: Record<string, unknown>,
+    baseConfig?: any,
+    contract?: Record<string, unknown>,
   ): void;
 
   sendSyncRequest?(
-    contract: Record<string, unknown>,
-    method: Record<string, unknown>,
     args: any,
+    method: Record<string, unknown>,
+    baseConfig?: any,
+    contract?: Record<string, unknown>,
   ): Promise<any>;
 
   watchBlocksV1?(
--- a/packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
+++ b/packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
@@ -917,6 +917,11 @@
             "type": "boolean",
             "nullable": false,
             "description": "Can only be set to false for an insecure grpc connection."
+          },
+          "monitorMode": {
+            "type": "boolean",
+            "nullable": true,
+            "description": "Flag used for monitoring. It changes default beahviour of transaction wrapper so it return error to caller instead of throwing RuntimeError straight away."
           }
         }
       },
@@ -976,6 +981,82 @@
             "type": "string",
             "nullable": false
           }
+        }
+      },
+      "IrohaSocketSessionEvent": {
+        "type": "string",
+        "enum": [
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Subscribe",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Next",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Unsubscribe",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Error",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Complete",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.SendAsyncRequest",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.SendSyncRequest"
+        ],
+        "x-enum-varnames": [
+            "Subscribe",
+            "Next",
+            "Unsubscribe",
+            "Error",
+            "Complete",
+            "SendAsyncRequest",
+            "SendSyncRequest"
+        ]
+      },
+      "IrohaBlockResponse": {
+        "type": "object",
+        "required": [
+           "payload",
+           "signaturesList"
+        ],
+        "properties": {
+          "payload": {
+            "type": "object",
+            "required": [
+              "txNumber",
+              "transactionsList",
+              "height",
+              "prevBlockHash",
+              "createdTime",
+              "rejectedTransactionsHashesList"
+            ],
+            "properties": {
+              "transactionsList": {
+                "type": "array",
+                "items": {}
+              },
+              "txNumber": {
+                "type": "number"
+              },
+              "height": {
+                "type": "number"
+              },
+              "prevBlockHash": {
+                "type": "string"
+              },
+              "createdTime": {
+                "type": "number"
+              },
+              "rejectedTransactionsHashesList": {
+                "type": "array",
+                "items": {}
+              }
+            }
+          },
+          "signaturesList":{
+            "type": "array",
+            "items": {}
+          }
+        }
+      },
+      "IrohaBlockProgress": {
+        "type": "object",
+        "required": ["transactionReceipt"],
+        "properties": {
+            "transactionReceipt": {
+                "$ref": "#/components/schemas/IrohaBlockResponse"
+            }
         }
       }
     }
--- a/packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
+++ b/packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
@@ -983,16 +983,14 @@
           }
         }
       },
-      "IrohaSocketSessionEvent": {
+      "WatchBlocksV1": {
         "type": "string",
         "enum": [
-            "org.hyperledger.cactus.api.async.iroha.SocketSession.Subscribe",
-            "org.hyperledger.cactus.api.async.iroha.SocketSession.Next",
-            "org.hyperledger.cactus.api.async.iroha.SocketSession.Unsubscribe",
-            "org.hyperledger.cactus.api.async.iroha.SocketSession.Error",
-            "org.hyperledger.cactus.api.async.iroha.SocketSession.Complete",
-            "org.hyperledger.cactus.api.async.iroha.SocketSession.SendAsyncRequest",
-            "org.hyperledger.cactus.api.async.iroha.SocketSession.SendSyncRequest"
+            "org.hyperledger.cactus.api.async.iroha.WatchBlocksV1.Subscribe",
+            "org.hyperledger.cactus.api.async.iroha.WatchBlocksV1.Next",
+            "org.hyperledger.cactus.api.async.iroha.WatchBlocksV1.Unsubscribe",
+            "org.hyperledger.cactus.api.async.iroha.WatchBlocksV1.Error",
+            "org.hyperledger.cactus.api.async.iroha.WatchBlocksV1.Complete"
         ],
         "x-enum-varnames": [
             "Subscribe",
@@ -1000,6 +998,17 @@
             "Unsubscribe",
             "Error",
             "Complete",
+            "SendAsyncRequest",
+            "SendSyncRequest"
+        ]
+      },
+      "IrohaSocketIOTransactV1": {
+        "type": "string",
+        "enum": [
+            "org.hyperledger.cactus.api.async.iroha.IrohaSocketIOTransactV1.SendAsyncRequest",
+            "org.hyperledger.cactus.api.async.iroha.IrohaSocketIOTransactV1.SendSyncRequest"
+        ],
+        "x-enum-varnames": [
             "SendAsyncRequest",
             "SendSyncRequest"
         ]
--- a/packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
+++ b/packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
@@ -6,17 +6,37 @@ import { LogLevelDesc, LoggerProvider } from "@hyperledger/cactus-common";
 import { Constants, ISocketApiClient } from "@hyperledger/cactus-core-api";
 import {
   DefaultApi,
-  IrohaSocketSessionEvent,
+  WatchBlocksV1,
+  IrohaSocketIOTransactV1,
   IrohaBlockProgress,
   IrohaBaseConfig,
 } from "../generated/openapi/typescript-axios";
-import { Configuration } from "../generated/openapi/typescript-axios/configuration";
+import {
+  Configuration,
+  ConfigurationParameters,
+} from "../generated/openapi/typescript-axios/configuration";
 import { RuntimeError } from "run-time-error";
+
+export interface IrohaApiClientParameters extends ConfigurationParameters {
+  logLevel?: LogLevelDesc;
+  wsApiHost?: string;
+  wsApiPath?: string;
+  timeoutLimit?: number;
+}
 
 export class IrohaApiClientOptions extends Configuration {
   readonly logLevel?: LogLevelDesc;
   readonly wsApiHost?: string;
   readonly wsApiPath?: string;
+  readonly timeoutLimit?: number;
+
+  constructor(param: IrohaApiClientParameters = {}) {
+    super(param);
+    this.logLevel = param.logLevel;
+    this.wsApiHost = param.wsApiHost;
+    this.wsApiPath = param.wsApiPath;
+    this.timeoutLimit = param.timeoutLimit;
+  }
 }
 
 export class IrohaApiClient
@@ -45,6 +65,7 @@ export class IrohaApiClient
     this.log.debug(`wsApiHost=${this.wsApiHost}`);
     this.log.debug(`wsApiPath=${this.wsApiPath}`);
     this.log.debug(`basePath=${this.options.basePath}`);
+    this.log.debug(`timeoutLimit=${this.options.timeoutLimit}`);
 
     Checks.nonBlankString(
       this.wsApiHost,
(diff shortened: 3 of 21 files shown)


outSH:
This is follow up PR on #1966

takeutak:
@petermetz @jagpreetsinghsasan  (cc: @izuru0)
Could you review this PR?

jagpreetsinghsasan (changes requested):
I just now checked HL Iroha, looks cool to me, will try experimenting and developing around this code base once it gets merged. Thanks for the contribution :)

packages/cactus-core-api/src/main/typescript/plugin/ledger-connector/i-socket-api-client.ts
```diff
@@ -8,15 +8,17 @@ import type { Observable } from "rxjs";
  */
 export interface ISocketApiClient<BlockType> {
   sendAsyncRequest?(
-    contract: Record<string, unknown>,
-    method: Record<string, unknown>,
-    args: any,
+    contract?: Record<string, unknown>,
+    method?: Record<string, unknown>,
+    args?: any,
+    baseConfig?: any,
```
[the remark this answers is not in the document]
jagpreetsinghsasan: I see, thanks for letting me know.

jagpreetsinghsasan (approved):

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
```diff
@@ -0,0 +1,247 @@
...
+    method: Record<string, unknown>,
+    args: any,
+    baseConfig?: IrohaBaseConfig,
+  ): void {
+    this.log.debug(`inside: sendAsyncRequest()`);
+    this.log.debug(`baseConfig=${baseConfig}`);
+    this.log.debug(`methodName=${method.methodName}`);
+    this.log.debug(`args=${args}`);
+
+    if (baseConfig === undefined || baseConfig === {}) {
```
petermetz, line 120: @outSH I recommend doing just a regular falsy-ness check. I think `baseConfig === {}` will always be false because you are asking something like: "Will this previously allocated memory address that I'm holding right now going to be the same memory address that I'm going to ask the allocator to give me in the future? (for the inline object literal creation)" - unless the OS memory allocator is broken, this should never be the case IMO. The other upside of the falsy-ness check is that it covers nulls as well not just undefined.

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
```diff
@@ -0,0 +1,247 @@
...
+      throw new RuntimeError("baseConfig object must exist and not be empty");
+    }
+
+    if (
```
petermetz, line 124: @outSH Is it okay if these fields are `null` instead of `undefined`? These current checks are not protecting against that.

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
```diff
@@ -0,0 +1,247 @@
...
+      baseConfig.privKey === undefined ||
+      baseConfig.creatorAccountId === undefined ||
+      baseConfig.irohaHost === undefined ||
+      baseConfig.irohaPort === undefined ||
+      baseConfig.quorum === undefined ||
+      baseConfig.timeoutLimit === undefined
+    ) {
+      throw new RuntimeError("Some fields in baseConfig are undefined");
```
petermetz, line 132: @outSH I know it's way more typing but I'm very confident that our users in the future will highly appreciate the exact list of fields that were missing instead of only being told that some are missing and some aren't.

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
```diff
@@ -0,0 +1,247 @@
...
+    method: Record<string, unknown>,
+    args: any,
+    baseConfig?: IrohaBaseConfig,
+  ): Promise<any> {
+    this.log.debug(`inside: sendSyncRequest()`);
+    this.log.debug(`baseConfig=${baseConfig}`);
+    this.log.debug(`method=${method}`);
+    this.log.debug(`args=${args}`);
+
+    if (baseConfig === undefined || baseConfig === {}) {
```
petermetz, line 173: @outSH Same question as above in the other method.

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
```diff
@@ -0,0 +1,247 @@
...
+      throw new RuntimeError("baseConfig object must exist and not be empty");
+    }
+
+    if (
```
petermetz, line 177: @outSH Same question as above in the other method.

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
```diff
@@ -0,0 +1,247 @@
...
+      baseConfig.privKey === undefined ||
+      baseConfig.creatorAccountId === undefined ||
+      baseConfig.irohaHost === undefined ||
+      baseConfig.irohaPort === undefined ||
+      baseConfig.quorum === undefined ||
+      baseConfig.timeoutLimit === undefined
+    ) {
+      throw new RuntimeError("Some fields in baseConfig are undefined");
```
petermetz, line 185: @outSH Same question as above in the other method.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/plugin-ledger-connector-iroha.ts
```diff
@@ -140,9 +140,63 @@ export class PluginLedgerConnectorIroha
...
+    wsApi.on("connection", (socket: SocketIoSocket) => {
+      let irohaSocketEndpoint: IrohaSocketIOEndpoint;
+
+      if (socket.id in this.socketSessionDictionary) {
+        this.log.debug(`Connected to existing socket session ID=${socket.id}`);
+        irohaSocketEndpoint = this.socketSessionDictionary[socket.id];
+      } else {
+        this.log.debug(`New Socket connected. ID=${socket.id}`);
+        irohaSocketEndpoint = new IrohaSocketIOEndpoint({ socket, logLevel });
+        this.socketSessionDictionary[socket.id] = irohaSocketEndpoint;
```
petermetz, line 160: @outSH Where in the code is this getting cleaned out of memory when the connection for the socket is closed?

packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
```diff
@@ -231,6 +236,82 @@
       "PrometheusExporterMetricsResponse": {
         "type": "string",
         "nullable": false
+      },
+      "IrohaSocketSessionEvent": {
```
petermetz, line 240: @outSH `IrohaSocketSessionEvent` is a thing not an action that an endpoint would undertake (e.g. it's a noun not a verb). Naming it like this crowds out everyone else from adding any other future endpoints that are SocketIO based because then we would have endpoints called `IrohaSocketSessionEvent` and `DoStuffV1` and `DoOtherStuffV1` but they would all be based on socket sessions and their events so people could end up getting pretty confused over their relationship to each other. I'd recommend naming it `WatchBlocksV1` instead, then it describes what it does instead of only hinting at the transport it uses and saying nothing about the actual operation that it undertakes.

packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
```diff
@@ -231,6 +236,82 @@
...
+        "type": "string",
+        "enum": [
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Subscribe",
```
petermetz, line 243: @outSH The same thing here as above regarding the name. Instead of `SocketSession` I'd go with `WatchBlocksV1` for the same reasons.

packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
```diff
@@ -231,6 +236,82 @@
...
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Next",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Unsubscribe",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Error",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.Complete",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.SendAsyncRequest",
+            "org.hyperledger.cactus.api.async.iroha.SocketSession.SendSyncRequest"
```
petermetz, line 249: @outSH The `SendAsyncRequest` and `SendSyncRequest` could be their own endpoints with their own enums defined here in the openapi.json file so that they both can also have their own event names for subscribing, erroring, completion, etc.
Mixing all of them together (watchblocks and send async/sync requests) like this would mean that they are all crammed in on the same logical channel and if a client starts monitoring the blocks but then also "sends a request" and finally gets an error they would not know if the "send request" thingy is having issues or the watchblocks functionality.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
outSH: Agree

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
outSH: @petermetz It seems they are marked as non nullable in openapi, so not OK

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
outSH: @petermetz Changed to separate checks

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
outSH: @petermetz done

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
outSH: @petermetz done

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
outSH: @petermetz I assume this is the same as comment at beginning of this if statement - done. Also could you please select exact code boundaries that you comment on, it's few clicks for you and huge help for me :)

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/plugin-ledger-connector-iroha.ts
outSH: @petermetz It's not. I think handling socketio reconnections is more complex problem so I've removed this whole session storage completely. It has similar logic as watchBlock in other modules.

packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
outSH: @petermetz Yeah I've changed them to match the other connectors.

packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
outSH: @petermetz Same as above

packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
outSH: @petermetz Agree. I've moved `SendAsyncRequest` and `SendSyncRequest` into new `IrohaSocketIOTransactV1`. The logic in code is still handled by common `IrohaSocketIOEndpoint`. Ideally I'd like to come up with solution to have more protocol-agnostic endpoints, but we have different priorities right now.

outSH:
@petermetz I've just pushed commit that contains fixes to your comments and some additional improvements, please review.

Also, there must be two separate commits (mine and Piotr's) for copyright reasons.

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
```diff
@@ -0,0 +1,261 @@
...
+    try {
+      socket.emit(IrohaSocketIOTransactV1.SendAsyncRequest, asyncRequestData);
+
+      // Connector should disconnect us after receiving this request.
+      // If he doesn't, disconnect after specified amount of time.
+      setTimeout(() => {
+        if (socket.connected) {
+          socket.disconnect();
+        }
+      }, baseConfig.timeoutLimit ?? 10 * 1000);
```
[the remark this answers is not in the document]
outSH: I think this wrongly connects `sendAsyncRequest` from other ledger in discounted-asset-trade sample with this ApiClient and reports error.

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
petermetz: @outSH Thank you!

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
petermetz: @outSH Thanks for the fix!

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
petermetz: @outSH Awesome, cheers!

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
petermetz: @outSH Got it, thank you!

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
petermetz: @outSH Nice, thanks!

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
petermetz: @outSH Yup, same thing. Sorry about not selecting the exact lines, I'll try to do better with that, it really is a very small effort thing on my part to do use that feature!

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/plugin-ledger-connector-iroha.ts
petermetz: @outSH Okay, that's fine by as well (end result is the same from the perspective of why I wrote the question). Thank you for addressing!

packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
petermetz: @outSH Thank you very much!

packages/cactus-plugin-ledger-connector-iroha/src/main/json/openapi.json
petermetz: @outSH 

> Agree. I've moved SendAsyncRequest and SendSyncRequest into new IrohaSocketIOTransactV1.

Awesome, thank you!

> The logic in code is still handled by common IrohaSocketIOEndpoint. Ideally I'd like to come up with solution to have more protocol-agnostic endpoints, but we have different priorities right now.

I can live with that for now because it's not getting exported (from what I could tell) to the public API surface of the package so if we postpone renaming/refactoring it to later then it won't have a breaking change attached to it.

I opened an issue for tracking it: https://github.com/hyperledger/cactus/issues/2122

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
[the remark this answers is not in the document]
petermetz: > outSH4 days ago
> I think this wrongly connects sendAsyncRequest from other ledger in discounted-asset-trade sample with this ApiClient and reports error.

@outSH From what I gather, the timeout is a request parameter that the caller can specify to be anything when coming in through `PluginLedgerConnectorIroha#transact()` (which ultimately ends up running this code if I read it right).
And it's saying that I could send requests with the timeout being set to a decade and then bring down the servers by sending a lot of those requests that allocate memory (technically) forever through the setTimeout.
A few things that you could do IMO:
1. Specify in the openapi.json that there is a maximum value for the timeout, something like 3 minutes or maybe 10 (hard to tell in the end because you don't want to run into issues with it being too short either...)
2. Validate here in the code that the timeout is not longer than what's allowed so that if someone tries to set it to a decade, they won't succeed.
3. Add a description to the timeout field that explains that it is to be specified in milliseconds and/or rename the field to `timeoutLimitMs`.
4. If after 1) and 2) and 3) CodeQL still complains then I think we are justified to mark it as "resolved" manually, but right now it looks like it is a legit concern, assuming that `baseConfig.timeoutLimit` is indeed something the requestors can specify themselves to be anything.

petermetz (changes requested):

petermetz:
> This is follow up PR on #1966, @izuru0 and @petermetz - please continue the review here.

@outSH Please next time don't duplicate the PRs. I just realized that I've been reviewing the same code... again.

petermetz:
> Also, there must be two separate commits (mine and Piotr's) for copyright reasons.

@outSH Could you break it up to two separate PRs in that case so that we know that both commits are stable (compiles and passes the tests on it's own)

packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/api-client/iroha-api-client.ts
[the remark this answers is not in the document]
outSH: @petermetz Not really, this is an `ApiClient` class used by the client and it sets the timeout on the client side. The only leak is socketio connection, but this can be caused as easily without timeout at all (sort of affects all the connectors at the moment).

The problem that CodeQL catches here is example app, which blindly passes the args to `sendSyncMethod`, so a person who uses BLP can cause BLP to leak timeouts - but it doesn't, since it uses different ledger apiClient, not this one, and in general should never be done in an actual app for various other security reasons.

Nevertheless I will change it anyway to match the other ApiClients, it seems to be a better practice to set this timeout in ApiClient constructor to better isolate it from any possible BLP-user input.

outSH:
> @outSH Please next time don't duplicate the PRs. I just realized that I've been reviewing the same code... again.

@petermetz As I described in the original PR, the author (P. Stepniowski) of this PR no longer works for us, we couldn't make changes in the old PR so I've opened the new one with his original commit and my changes to suit your review, fix CI etc...

> @outSH Could you break it up to two separate PRs in that case so that we know that both commits are stable (compiles and passes the tests on it's own).

Original commit doesn't pass the tests on it's own, so I'm afraid it's impossible. Initially I wanted to merge single coauthored commit but DCO CI was complaining about author and signoff mismatch. We can either push these two or silence DCO for this PR, don't know what are the other options here.

outSH:
@petermetz Rebased with the main, please merge when possible.

outSH:
@petermetz All issues solved, please merge when possible.

petermetz (changes requested):
> @petermetz All issues solved, please merge when possible.

@outSH Sorry, but there seem to be more merge conflicts at the moment. Please fix and pass it back for review!

outSH:
> @outSH Sorry, but there seem to be more merge conflicts at the moment. Please fix and pass it back for review!

Conflicts fixed

petermetz (approved):
@outSH LGTM, thank you!

PR #2050, opened 2022-05-31
build(openapi): phase 1- generate kotlin client

### LINKED ISSUE #2000: build(openapi): phase 1 - generate kotlin clients
This is phase one of the task.
All the details are in the phase 2 task's description which you can see here: https://github.com/hyperledger/cactus/issues/468

jagpreetsinghsasan:
# Commit to be reviewed

build(openapi): generate kotlin client

    Primary Changes
    ---------------
    1. Updated the root package.json and cactus-plugin-ledger-connector-corda
 	   package.json to incorporate generate-clients script
    2. Created a git action workflow to generate version-tagged client artifacts

Fixes #2000


takeutak (approved):

jagpreetsinghsasan:
Hi @petermetz , can you review this PR?

petermetz (approved):
Could you please create a task in the issue tracker about testing the published .jar files once we have a new release issued where this job got triggered? (This way the testing would be properly end to end and we would know for sure that the client works when used by developers)

.github/workflows/jars-publish.yaml
```diff
@@ -0,0 +1,48 @@
+name: Build kotlin artifacts and publish to github packages
```
petermetz, line 1: @jagpreetsinghsasan Sorry, some nit-pick here. We might have more kotlin artifacts for other connectors in the future so let's not crowd out everyone else by not specifying that this is the corda API client.

TLDR: Name the job Something like `Corda_API_Client` instead of `Build kotlin artifacts and publish to github packages`

petermetz: @jagpreetsinghsasan Please also rename the file in the same spirit so that it's less ambiguous once we have more jobs publishing more jars. 
Recommended filename: `corda-api-client-publish.yaml`

petermetz (changes requested):

.github/workflows/jars-publish.yaml
jagpreetsinghsasan: @petermetz I have also coded for the bigger task where the kotlin client is generated at 22 more places.
image

Are we expecting 22 different github workflows ?(I generalised it so that we can different steps for different openapi files instead of having different workflows altogether)

jagpreetsinghsasan: @petermetz any updates on this?

petermetz: @jagpreetsinghsasan Sorry for the slow response. I'm not sure TBH => If the tasks are long running (more than a few minutes) then I would break them up into separate workflow action *jobs* at the very least (e.g. they could stay in the same file but at least have separate jobs within).
So we don't necessarily need 22 workflows (and then we might not need to name the file so specifically either)

> @petermetz I have also coded for the bigger task where the kotlin client is generated at 22 more places.

Could you send a link to the version of the file that has the 22 more places and then I can provide a definitive answer!

jagpreetsinghsasan: I am not able to provide a link for the parent issue as there is an issue with the `openapi-generator-cli`
The latest version of the cli uses `openapi-generator` version 5.2.0 but in that there is a bug where the type `any` is rendered as `AnyType`. This bug got fixed in 5.2.1 version of `openapi-generator` which isnt currently supported by `openapi-generator-cli` (I have linked detailed information in the personal chat @petermetz )

jagpreetsinghsasan: The parent code is also ready, just stuck there because of the above blocker.

jagpreetsinghsasan: > @jagpreetsinghsasan Sorry for the slow response. I'm not sure TBH => If the tasks are long running (more than a few minutes) then I would break them up into separate workflow action _jobs_ at the very least (e.g. they could stay in the same file but at least have separate jobs within). So we don't necessarily need 22 workflows (and then we might not need to name the file so specifically either)
> 
> > @petermetz I have also coded for the bigger task where the kotlin client is generated at 22 more places.
> 
> Could you send a link to the version of the file that has the 22 more places and then I can provide a definitive answer!

Hi @petermetz . Here is a [link](https://github.com/jagpreetsinghsasan/cactus/actions/runs/3383545104/jobs/5619542626) to such workflow (please ignore the errors, need to see why some of them are still failing).

petermetz: > > @jagpreetsinghsasan Sorry for the slow response. I'm not sure TBH => If the tasks are long running (more than a few minutes) then I would break them up into separate workflow action _jobs_ at the very least (e.g. they could stay in the same file but at least have separate jobs within). So we don't necessarily need 22 workflows (and then we might not need to name the file so specifically either)
> > > @petermetz I have also coded for the bigger task where the kotlin client is generated at 22 more places.
> > 
> > 
> > Could you send a link to the version of the file that has the 22 more places and then I can provide a definitive answer!
> 
> Hi @petermetz . Here is a [link](https://github.com/jagpreetsinghsasan/cactus/actions/runs/3383545104/jobs/5619542626) to such workflow (please ignore the errors, need to see why some of them are still failing). And here's the [link](https://github.com/jagpreetsinghsasan/cactus/blob/feature-468/.github/workflows/jars-publish.yaml) to the corresponding code

@jagpreetsinghsasan The kotlin build runs for 30 seconds while the configure script takes 6 minutes so it makes sense to combine all the kotlin builds into one workflow so that they can share the configure script's results instead of having a bunch of only slightly smaller workflows where each of them runs the configure script for the majority of their execution time.

Also, as usual, sorry for the slow response!

jagpreetsinghsasan: Sure. This seems to be pretty optimised approach.

.github/workflows/kotlin-api-client-publish.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+  pull_request:
+    branches: [main, dev]
+
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
```
petermetz, line 15: @jagpreetsinghsasan Please lock it down to the same version of the checkout action that is being used by all the other actions as well. See https://github.com/hyperledger/cactus/pull/2188

petermetz:
Suggested edit:
```diff
diff --git a/.github/workflows/kotlin-api-client-publish.yaml b/.github/workflows/ghpkg-all-kotlin-api-clients-publish.yaml
similarity index 82%
rename from .github/workflows/kotlin-api-client-publish.yaml
rename to .github/workflows/ghpkg-all-kotlin-api-clients-publish.yaml
index d1b9c245..06f99561 100644
--- a/.github/workflows/kotlin-api-client-publish.yaml
+++ b/.github/workflows/ghpkg-all-kotlin-api-clients-publish.yaml
@@ -1,4 +1,4 @@
-name: Build kotlin artifacts and publish to github packages
+name: ghpkg-all-kotlin-api-clients-publish
 
 on:
   push:
@@ -33,7 +33,7 @@ jobs:
 
     - run: yarn configure
 
-    - name: Build the cactus-plugin-ledger-connector-corda kotlin package
+    - name: build-plugin-ledger-connector-corda-kotlin-client
       run: |
         WDIR=`pwd`
         cd ${WDIR}/packages/cactus-plugin-ledger-connector-corda/src/main/kotlin/generated/openapi/kotlin-client
@@ -41,7 +41,7 @@ jobs:
         ./gradlew clean
         ./gradlew build
 
-    - name: Upload the generated kotlin artifact for cactus-plugin-ledger-connector-corda to github packages
+    - name: publish-plugin-ledger-connector-corda-kotlin-client
       uses: actions/upload-artifact@v3.0.0
       with:
         name: cactus-plugin-ledger-connector-corda-client-kotlin-${{ env.GITVERSION }}.jar

```

petermetz (changes requested):
@jagpreetsinghsasan There are a few failing checks, please fix those as well and then pass it back for re-review. I also sent in a patch suggesting a rename to  the file and the workflow names, you can apply that patch directly from the VSCode GitHub extension if you want to keep it simple (there's a button for it when you open the PR page via the extension)

jagpreetsinghsasan:
> Suggested edit:
> 
> ```diff
> diff --git a/.github/workflows/kotlin-api-client-publish.yaml b/.github/workflows/ghpkg-all-kotlin-api-clients-publish.yaml
> similarity index 82%
> rename from .github/workflows/kotlin-api-client-publish.yaml
> rename to .github/workflows/ghpkg-all-kotlin-api-clients-publish.yaml
> index d1b9c245..06f99561 100644
> --- a/.github/workflows/kotlin-api-client-publish.yaml
> +++ b/.github/workflows/ghpkg-all-kotlin-api-clients-publish.yaml
> @@ -1,4 +1,4 @@
> -name: Build kotlin artifacts and publish to github packages
> +name: ghpkg-all-kotlin-api-clients-publish
>  
>  on:
>    push:
> @@ -33,7 +33,7 @@ jobs:
>  
>      - run: yarn configure
>  
> -    - name: Build the cactus-plugin-ledger-connector-corda kotlin package
> +    - name: build-plugin-ledger-connector-corda-kotlin-client
>        run: |
>          WDIR=`pwd`
>          cd ${WDIR}/packages/cactus-plugin-ledger-connector-corda/src/main/kotlin/generated/openapi/kotlin-client
> @@ -41,7 +41,7 @@ jobs:
>          ./gradlew clean
>          ./gradlew build
>  
> -    - name: Upload the generated kotlin artifact for cactus-plugin-ledger-connector-corda to github packages
> +    - name: publish-plugin-ledger-connector-corda-kotlin-client
>        uses: actions/upload-artifact@v3.0.0
>        with:
>          name: cactus-plugin-ledger-connector-corda-client-kotlin-${{ env.GITVERSION }}.jar
> ```

This is indeed a cool way to convey changes, if writing it in words is huge. I will also use this method.

jagpreetsinghsasan:
@petermetz it worked only by commenting the gitattributes rule for `.bat` files (fixed bat files crlf,lf issue) and adding lf to text files (fixed the `FILES` file being populated as generated in git diff, which originally was generated by openapi-generator)

petermetz:
> This is indeed a cool way to convey changes, if writing it in words is huge. I will also use this method.

@jagpreetsinghsasan I like it very much as well! You can make these suggests directly from VSCode by making local changes on the PR branch, stashing them and then there's a VSCode `suggest` command that submits it as a comment like this.

petermetz (changes requested):
> @petermetz it worked only by commenting the gitattributes rule for `.bat` files (fixed bat files crlf,lf issue) and adding lf to text files (fixed the `FILES` file being populated as generated in git diff, which originally was generated by openapi-generator)

@jagpreetsinghsasan I'm a bit worried about this because no one else has been reporting the issue and now we are changing it in a way that will change something for everyone, so we might end up in a situation where we have to revert this anyway if somehow the new config makes it go haywire for everyone but you. I have no idea if any of this will happen or not though so let's just fire from the hip on this one and see what happens. So, almost LGTM, but there's one thing though:
When I checkout this branch and then run the configure script, it makes changes compared to your branch.
I'm assuming it's because of the different generator version used but then you should commit those changes so that a regular build does not change the source code (reproducible builds principle).

Screenshot from 2022-11-26 10-03-49

jagpreetsinghsasan:
Hi @petermetz 
I am not sure why you are seeing changed files in your code upon running the `configure`. (If it were, then our ci build should have failed as it has a `git diff` check as well)
Because I ran the configure script just now (the terminal logs in screenshot) and then ran the git status (the terminal logs in screenshot), I didnt see any files changed.
Also if you see the version of openapi-generator (the opened file in the screenshot), its still `5.2.1`
image

petermetz:
> Hi @petermetz I am not sure why you are seeing changed files in your code upon running the `configure`. (If it were, then our ci build should have failed as it has a `git diff` check as well) Because I ran the configure script just now (the terminal logs in screenshot) and then ran the git status (the terminal logs in screenshot), I didnt see any files changed. Also if you see the version of openapi-generator (the opened file in the screenshot), its still `5.2.1` image

@jagpreetsinghsasan Alright, let's just give it a try then and see how it goes. Hopefully we don't start hearing reports of issues from other users either given that the CI seems to be fine with it.

petermetz (approved):
@jagpreetsinghsasan Thank you very much! Fingers crossed!

jagpreetsinghsasan:
> > Hi @petermetz I am not sure why you are seeing changed files in your code upon running the `configure`. (If it were, then our ci build should have failed as it has a `git diff` check as well) Because I ran the configure script just now (the terminal logs in screenshot) and then ran the git status (the terminal logs in screenshot), I didnt see any files changed. Also if you see the version of openapi-generator (the opened file in the screenshot), its still `5.2.1` image
> 
> @jagpreetsinghsasan Alright, let's just give it a try then and see how it goes. Hopefully we don't start hearing reports of issues from other users either given that the CI seems to be fine with it.

Yup  😅

jagpreetsinghsasan:
> @jagpreetsinghsasan Alright, let's just give it a try then and see how it goes. Hopefully we don't start hearing reports of issues from other users either given that the CI seems to be fine with it.

Yup  😅 😅

Issue #1671, opened 2021-12-21
fix(cmd-socketio-server): Enable tsconfig strict flag and fix all the warnings


outSH:
Depends on https://github.com/hyperledger/cactus/issues/1647

Some projects has temporarily disabled strict flag in their tsconfig. We must enable it and fix all the warnings that appears.

Affected projects:
- [x] ./packages/cactus-cmd-socketio-server
- [x] ./packages/cactus-plugin-ledger-connector-fabric-socketio
- [x] ./packages/cactus-plugin-ledger-connector-go-ethereum-socketio
- [x] ./packages/cactus-plugin-ledger-connector-sawtooth-socketio
- [x] ./examples/cactus-example-electricity-trade
- [x] ./examples/cactus-example-discounted-asset-trade

outSH:
All packages are fixed now

PR #2051, opened 2022-05-31
refactor(connector-go-ethereum-socketio): fix strict flag warnings

outSH:
cactus-plugin-ledger-connector-go-ethereum-socketio will compile with global strict flag.

Related issue: #1671

### DIFF
--- a/packages/cactus-plugin-ledger-connector-go-ethereum-socketio/src/main/typescript/common/core/bin/www.ts
+++ b/packages/cactus-plugin-ledger-connector-go-ethereum-socketio/src/main/typescript/common/core/bin/www.ts
@@ -69,7 +69,7 @@ server.on("listening", onListening);
  * Normalize a port into a number, string, or false.
  */
 
-function normalizePort(val) {
+function normalizePort(val: string) {
   const port = parseInt(val, 10);
 
   if (isNaN(port)) {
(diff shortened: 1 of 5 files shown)


izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):

packages/cactus-plugin-ledger-connector-go-ethereum-socketio/src/main/typescript/common/core/bin/www.ts
```diff
@@ -89,7 +89,7 @@ function normalizePort(val) {
  * Event listener for HTTPS server "error" event.
  */
 
-function onError(error) {
+function onError(error: any) {
```
petermetz, line 92: @outSH A recommendation for the future: change the `any` to `unknown` so that the compiler forces proper runtime type checking on the developers during development (it will force them to add checks to prevent this function crashing if `error` is null or undefined for example). Not a change request, just something to keep in mind for the future as a recommendation.
Also apologies for the slow reviews, I had a lot of trouble with reviving the CI but we are back in business now.

petermetz (approved):
@outSH LGTM, please take a quick look at this comment I left here for future reference: https://github.com/hyperledger/cactus/pull/2051/files#r920445350

Issue #2052, opened 2022-05-31
test(connector-go-ethereum-socketio): add functional test, bug fix


outSH:
Add a functional test to `connector-go-ethereum-socketio` so it's operations can be checked in CI. Refactor the connector to improve testability, fix any bugs that are revealed by the tests.

PR #2053, opened 2022-05-31
test(connector-go-ethereum-socketio): add functional test, bug fix

outSH:
- Add functional test of all functions from go-ethereum-socketio validator.
- Refactor go-ethereum validator to allow importing as a module, to simplify the functional test.
- Fix sendRawTransaction to work with Verifier protocol.
  It couldn't be reached by any client library until now, so I consider this as "private" interface.
- Add common web3 client object in openethereum test ledger helper class.
- Add few new functions to ethereum test ledger helper class:
  newEthPersonalAccount, transferAssetFromCoinbase, deployContract.

Depends on #2051
Depends on #2047

Closes: #2052

-------------

Please consider only the last commit - `test(connector-go-ethereum-socketio): add functional test, bug fix`. The rest are the dependencies that should be merged by separate PRs.

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/openethereum/openethereum-test-ledger.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/openethereum/openethereum-test-ledger.ts
@@ -2,7 +2,8 @@ import { EventEmitter } from "events";
 import Docker, { Container } from "dockerode";
 import { v4 as internalIpV4 } from "internal-ip";
 import Web3 from "web3";
-import { Account } from "web3-core";
+import { AbiItem } from "web3-utils";
+import { Account, TransactionReceipt } from "web3-core";
 import { v4 as uuidv4 } from "uuid";
 
 import {
@@ -14,6 +15,7 @@ import {
 } from "@hyperledger/cactus-common";
 
 import { Containers } from "../common/containers";
+import { RuntimeError } from "run-time-error";
 
 export interface IOpenEthereumTestLedgerOptions {
   envVars?: string[];
@@ -55,6 +57,7 @@ export class OpenEthereumTestLedger {
   private readonly httpPort: number;
   private _container: Container | undefined;
   private _containerId: string | undefined;
+  private _web3: Web3 | undefined;
 
   public get imageFqn(): string {
     return `${this.imageName}:${this.imageVersion}`;
@@ -69,6 +72,16 @@ export class OpenEthereumTestLedger {
       return this._container;
     } else {
       throw new Error(`Invalid state: _container is not set. Called start()?`);
+    }
+  }
+
+  private get web3(): Web3 {
+    if (this._web3) {
+      return this._web3;
+    } else {
+      throw new Error(
+        "Invalid state: web3 client is missing, start the ledger container first.",
+      );
     }
   }
 
@@ -123,7 +136,7 @@ export class OpenEthereumTestLedger {
       "--jsonrpc-cors=all",
       "--jsonrpc-interface=all",
       "--jsonrpc-hosts=all",
-      "--jsonrpc-apis=web3,eth,net,parity",
+      "--jsonrpc-apis=web3,eth,personal,net,parity",
       "--ws-interface=all",
       "--ws-apis=web3,eth,net,parity,pubsub",
       "--ws-origins=all",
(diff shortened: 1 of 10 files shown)


izuru0 (approved):
LGTM

packages/cactus-common/src/main/typescript/verifier/jwt-message-authentication.ts
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright 2020-2022 Hyperledger Cactus Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import crypto from "crypto";
+import jwt from "jsonwebtoken";
```
petermetz, line 7: @outSH Could you either make this cross-platform by any chance? (the common package is designed to be useable in both the browser and NodeJS environments). Alternatively it could go in another package too, I'm flexible on options, but want to avoid exporting functions to the browser that are known to not work.

I'm not sure if this is actually the diff of another PR that this one depends on or if this is the one I should be reviewing. If this is coming from another PR then this PR is LGTM (and I'll ask you to fix it in that other PR instead), just let me know which way it goes.

petermetz (changes requested):
Apologies for the slow reviews once again.

packages/cactus-common/src/main/typescript/verifier/jwt-message-authentication.ts
outSH: @petermetz Yeah you already made this comment here - https://github.com/hyperledger/cactus/pull/2047

outSH: This is fixed

outSH:
@petermetz Rebased with main and tested - ready to merge

packages/cactus-test-tooling/src/main/typescript/openethereum/openethereum-test-ledger.ts
```diff
@@ -178,20 +191,67 @@ export class OpenEthereumTestLedger {
...
-    const fnTag = `${this.className}#getEthTestAccount()`;
+    const ethTestAccount = this.web3.eth.accounts.create(uuidv4());
+
+    const receipt = await this.transferAssetFromCoinbase(
+      ethTestAccount.address,
+      seedMoney,
+    );
+
+    if (receipt instanceof Error) {
+      throw receipt;
```
petermetz, line 207: @outSH I think if you re-throw it like this then you chop off the previous call stack and then start a new one from this point. I'd recommend re-throwing a `RuntimeError` [1] that wraps the original error and also adds some contextual detail about what was being done and where when something went wrong.

[1]: https://www.npmjs.com/package/run-time-error

packages/cactus-test-tooling/src/main/typescript/openethereum/openethereum-test-ledger.ts
```diff
@@ -178,20 +191,67 @@ export class OpenEthereumTestLedger {
...
+    const account = await this.web3.eth.personal.newAccount(password);
 
-    const rpcApiHttpHost = await this.getRpcApiHttpHost();
-    const web3 = new Web3(rpcApiHttpHost);
-    const ethTestAccount = web3.eth.accounts.create(uuidv4());
+    const receipt = await this.transferAssetFromCoinbase(account, seedMoney);
 
-    const tx = await web3.eth.accounts.signTransaction(
+    if (receipt instanceof Error) {
+      throw receipt;
```
petermetz, line 231: @outSH I think if you re-throw it like this then you chop off the previous call stack and then start a new one from this point. I'd recommend re-throwing a `RuntimeError` [1] that wraps the original error and also adds some contextual detail about what was being done and where when something went wrong.

[1]: https://www.npmjs.com/package/run-time-error

petermetz (changes requested):

packages/cactus-common/src/main/typescript/verifier/jwt-message-authentication.ts
petermetz: @outSH Oh, that's why it seemed so familiar and redundant! /facepalm sorry!

packages/cactus-test-tooling/src/main/typescript/openethereum/openethereum-test-ledger.ts
outSH: @petermetz Done.

packages/cactus-test-tooling/src/main/typescript/openethereum/openethereum-test-ledger.ts
outSH: @petermetz Done.

petermetz (approved):
@outSH LGTM

Issue #1963, opened 2022-04-10
docs(build): add text build steps for Linux, Mac and Windows


AbhinavMir:
The current [build](https://github.com/hyperledger/cactus/blob/main/BUILD.md) instructions use YouTube videos to display how to build Cactus locally. While this is great in general, but people with accessibility issues might run into problems. The videos are divided and text is small, someone with slower internet or bad eyesight for example wouldn't be able to access this. Due to lack of sound, this is completely inaccessible to the visually impaired.

I propose an Iroha-like [documentation](https://iroha.readthedocs.io/en/main/build/index.html) in addition to current documentation.

If assigned, I'm happy to make these changes ASAP. Open to views.

petermetz:
@AbhinavMir Very much agreed and thank you so much for bringing this up! Assigning now and thank you in advance for all your contributions!

petermetz:
@AbhinavMir Please also note that I slightly refactored the issue title which now you can copy paste into your future commit message that you'll make the changes in (we have a commit message linter so that's why this is more important than it looks at first sight)

ashnashahgrover:
Please assign me

jagpreetsinghsasan:
@petermetz we have assigned @ashnashahgrover for covering specifics of Cacti setup for MacOS.
Once done, we will have someone to update it for Ubuntu 22.04

petermetz:
@jagpreetsinghsasan Nice, thank you! @ashnashahgrover Good luck and thank you as well!

PR #2072, opened 2022-06-10
docs(build): add text build steps for Linux, Mac and Windows

AbhinavMir:
Issue: #1963

Signed-off-by: Abhinav Srivastava (abhinavmir#1898 on Discord)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,91 +1,84 @@
-# Hyperledger Cactus Build Instructions
+### Cactus BUILD instruction
 
-This is the place to start if you want to give Cactus a spin on your local
-machine or if you are planning on contributing.
+#### Installing Git
+Git is a tool for version control that you will need to build and maintain Hyperledger Cactus from source. To install Git on a Debian-based Linux system, you should run the following command.
 
-> This is not a guide for `using` Cactus for your projects that have business logic
-> but rather a guide for people who want to make changes to the code of Cactus.
-> If you are just planning on using Cactus as an npm dependency for your project,
-> then you might not need this guide at all.
+```shell
+sudo apt install git-all
+```
 
-The project uses Typescript for both back-end and front-end components.
+Installation methods for other systems can be found [here](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git).
 
-## Developers guide
+#### Installing Docker
+ 
+Docker is a tool for containerization of softwares. It helps ship software and mitigates compatibility issues. To install Docker, please follow the installation guide provided [here](https://docs.docker.com/engine/install/).
 
-This is a video guide to setup Hyperledger Cactus on your local machine.
+To install Docker on Debian, one would follow the following steps.
 
-### Installing git
+```shell=
+### Uninstall old Docker versions
+sudo apt-get remove docker docker-engine docker.io containerd runc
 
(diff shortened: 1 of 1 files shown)


petermetz (changes requested):
@AbhinavMir 

1. apt is only there on Debian and it's downstream distros so I would somehow make it clear and then in separate sections you could detail the steps for Win and Mac so that people don't get confused.
2. Instead of npm, please use yarn where possible
3. Do not install the `nodejs` package via apt (not even on the Debian distros). We strongly encourage/prefer the use of the node version manager tool (nvm)
4. Let's push people towards JDK 11 instead of JDK 8

AbhinavMir:
have addressed 2,3,4 in latest commit. I think for 1, we can create a new issue to create separate build instructions on Windows and Mac.

BUILD.md
```diff
@@ -1,91 +1,91 @@
-# Hyperledger Cactus Build Instructions
+### Cactus BUILD instruction
```
jagpreetsinghsasan, line 1: I believe saying Hyperledger Cactus (instead of just Cactus) here is better.

BUILD.md
```diff
@@ -1,91 +1,91 @@
...
+curl -fsSL https://download.docker.com/linux/debian/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+echo \
+  "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/debian \ $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+  
+### Install the Docker Engine
+sudo apt-get update
+sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin
+
+### Check that the package has been correctly installed
+sudo docker run hello-world
```
jagpreetsinghsasan, line 47: After this line, reference can also be given to use docker without sudo, as mentioned here, https://docs.docker.com/engine/install/linux-postinstall/

BUILD.md
```diff
@@ -1,91 +1,91 @@
...
 
-We put a lot of thought and effort into making sure that fast developer iterations can be
-achieved (please file a bug if you feel otherwise) while working **on** the framework.
+#### Installing VS Code 
 
-If you find yourself waiting too much for builds to finish, most of the time
-that can be helped by using the `npm run watch` script which can automatically
-recompile packages as you modify them (and only the packages that you have
-modified, not everything).
+You can download and install VS Code from [here](https://code.visualstudio.com/download). It is suggested to install the git-graph and Docker plugins.
```
jagpreetsinghsasan, line 77: Here, we can also mention the `npm-scripts` plugin as well (https://marketplace.visualstudio.com/items?itemName=traBpUkciP.vscode-npm-scripts)

BUILD.md
```diff
@@ -1,91 +1,91 @@
...
+You can download and install VS Code from [here](https://code.visualstudio.com/download). It is suggested to install the git-graph and Docker plugins. 
 
-It also supports re-running the OpenAPI generator when you update any
-`openapi.json` spec files that we use to describe our endpoints.
+#### Clone and compile the Cactus repository 🌵
 
-The `npm run watch` script in action:
+```
+git clone https://github.com/hyperledger/cactus
+yarn add
```
jagpreetsinghsasan, line 83: We can specify the `npm run install-yarn` prior to`yarn add` (yarn-add isn't required)

jagpreetsinghsasan (changes requested):

PR #2095, opened 2022-06-22
fix: compile errors in the electricity-trade example

janvinsha:
This fix allows the cactus repository to configure without errors when `npm run configure` is run in the cactus root directory
It enables various examples like the electricity-trade to work
Signed-off-by: Jande Vincent [janvinsha@gmail.com](url)

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,6 +148,7 @@
     "lodash": ">=4.17.21",
     "minimist": ">=1.2.6",
     "node-forge": ">=1.3.0",
-    "underscore": "1.13.2"
+    "underscore": "1.13.2",
+    "@types/ssh2": "0.5.52"
   }
 }


takeutak (approved):

takeutak:
@janvinsha Thanks for your contribution!

jagpreetsinghsasan (approved):
LGTM. Thankyou for the contribution.

petermetz (changes requested):
@janvinsha Please squash the commits into a single one and then we are good to go AFAICT => LGTM

ryjones:
@petermetz I fixed this

petermetz (commented):
@janvinsha Is this needed though? I'm pretty sure the configure script works without it because otherwise the CI build+ tests would be broken.

petermetz:
Closing due to inactivity + I don't think we need this because the build is not broken on the main branch.

Issue #1356, opened 2021-09-20
style: 2021-09-20 linter warnings batch 7 / 26


petermetz:
style: 2021-09-20 linter warnings batch 7 / 26

### Description

`packages/cactus-test-plugin-htlc-eth-besu/src/test/typescript/integration/plugin-htlc-eth-besu/openapi/openapi-validation.test.ts`

### Acceptance Criteria
1. Each linter warning is either fixed or a specific new issue is created aimed at exactly that one linter warning where applicable: If fixing a linter warning requires a larger refactoring of the codebase, then it is justifiable to break it out into it's own GH task.
2. The pull request submitted links to all (if any) additional issues that were created for the linter warnings that did not fit in the scope of the issue.

### Full List of Linter Warnings

/workspaces/cactus/packages/cactus-test-plugin-htlc-eth-besu/src/test/typescript/integration/plugin-htlc-eth-besu/openapi/openapi-validation.test.ts
  204:45  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  211:50  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  234:45  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  241:50  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  313:46  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  320:50  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  349:46  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  356:50  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  412:41  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  419:50  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  437:41  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  444:50  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  518:43  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  525:50  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  546:43  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
[...]

PR #2098, opened 2022-06-24
fix: part 2 of lint warnings resolved for batch 7/26 #1356 : Commit to be reviewed

akankshadixit:
[style(common): fix batch 7/26 of linter warnings](https://github.com/hyperledger/cacti/pull/2098/commits/fda1910eac5289f56fa3ec30cba24e6a73d09afe)

Changes made:
- Fixed lint warnings associated with any-type for truthyness and nonBlankString checking
- Fixed lint warnings associated with type-aliases in logger
- Fixed lint warnings associated with  non-any type in getAllMethodNames and getAllFieldNames
- Fixed lint errors associated with unexpected any for private-public keypairs
- Added an isRecord user-defined type guard in the common package that
can narrow `unknown` types into `Record<string, unknown>`

Fixes https://github.com/hyperledger/cacti/issues/1356

Signed-off-by: akankshadixit <akanksha.dixit12@gmail.com>

### DIFF
--- a/packages/cactus-common/src/main/typescript/objects.ts
+++ b/packages/cactus-common/src/main/typescript/objects.ts
@@ -29,18 +29,19 @@ export class Objects {
    *
    * @param anObject
    */
-  public static getAllMethodNames(anObject: any): string[] {
+  public static getAllMethodNames(anObject: unknown): string[] {
+    let aRecord = anObject as Readonly<Record<string, unknown>>;
     let properties: string[] = [];
     do {
-      const symbols = Object.getOwnPropertySymbols(anObject);
+      const symbols = Object.getOwnPropertySymbols(aRecord);
       const symbolPropertyNames = symbols.map((aSymbol) => aSymbol.toString());
 
-      const propertyNamesCurrent = Object.getOwnPropertyNames(anObject)
+      const propertyNamesCurrent = Object.getOwnPropertyNames(aRecord)
         .concat(symbolPropertyNames)
         .sort()
         .filter((propertyName: string, index: number, arr) => {
           return (
-            typeof anObject[propertyName] === "function" &&
+            typeof aRecord[propertyName] === "function" &&
             propertyName !== "constructor" &&
             (index === 0 || propertyName !== arr[index - 1]) &&
             properties.indexOf(propertyName) === -1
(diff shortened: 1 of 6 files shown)


takeutak (approved):

jagpreetsinghsasan (changes requested):
The changes looks fine, just a squash merge of the commits required.

petermetz (approved):
@akankshadixit LGTM, thank you!

Some advice on the admin chores to speed things up next time: If a reviewer requested changes and then you addressed them (like you did here) then it helps if you then hit  the little circular arrows button next to their name (in this case Jagpreet) which is called the "re-request review" button which will send them a ping to come back and check on the changes you've made in response to their change request. Without this it doesn't actually pop up on their feed as the PR needing review again.

image

jagpreetsinghsasan (approved):

akankshadixit:
Hi @petermetz,

Thank you. I will ensure this in future.

petermetz:
> Hi @petermetz,
> 
> Thank you. I will ensure this in future.

@akankshadixit No worries at all, it's not a big deal either way, I'm just letting you know for future reference. :-)

petermetz (commented):
@akankshadixit The build seems to be broken now, please double check if it works locally and then let's go from there.

I'm not sure if you have access to the CI Logs or not but the `build-dev` check failed with this:

```sh
5:41:00 AM - Project 'packages/cactus-common/tsconfig.json' is out of date because output file 'packages/cactus-common/dist/lib/main/typescript/bools.js' does not exist
5:41:00 AM - Building project '/home/runner/work/cactus/cactus/packages/cactus-common/tsconfig.json'...
Error: packages/cactus-common/src/test/typescript/unit/objects/get-all-method-names.test.ts(8,49): error TS2345: Argument of type 'A' is not assignable to parameter of type 'Record<string, unknown>'.
  Index signature is missing in type 'A'.
5:41:04 AM - Project 'packages/cactus-core-api/tsconfig.json' can't be built because its dependency 'packages/cactus-common/tsconfig.json' has errors
```

petermetz (changes requested):
^^

packages/cactus-common/src/main/typescript/objects.ts
```diff
@@ -29,7 +29,7 @@ export class Objects {
    *
    * @param anObject
    */
-  public static getAllMethodNames(anObject: any): string[] {
+  public static getAllMethodNames(anObject: Record<string, unknown>): string[] {
```
petermetz, line 32: @akankshadixit Sorry for just noticing this, but specifically this change must be moved to a separate, brand new pull request because it is a breaking change that we cannot make prior to releasing version 2.0 (we can only make breaking changes while issuing so called major releases according to the rules of semantic versioning).

So what I ask here is for you to:
1. Open a new pull request that makes only this change (and the update to the test case that is needed to make it compile)
2. Remove this change from this particular pull request so that it contains everything else but not this change.
3. Then we can merge this pull request and will postpone the merging of the other one with the breaking change to the time when we are releasing 2.0

Sorry again for not catching this sooner.

petermetz (changes requested):

packages/cactus-common/src/main/typescript/objects.ts
petermetz: @akankshadixit I figured out a way to make the change needed but without having it become a breaking change! Resolving.

petermetz (approved):
@akankshadixit LGTM now with the additions I made just now it does not have to be a breaking change and we are good to go!

Issue #2107, opened 2022-06-30
test(connector-go-sawtooth-socketio): add Jest tests


petermetz:
### Description

Add Jest compatible unit and integration tests to the package under the path:
`packages/cactus-plugin-ledger-connector-go-sawtooth-socketio/src/test/typescript/...`

### Acceptance Criteria
1. ci.yml also updated to execute these tests as part of the CI

cc: @takeutak @outSH @izuru0

PR #2089, opened 2022-06-21
chore(connector-go-eth): upgrade web3.js to recent version

outSH:
- Upgrade web3 package used by go-ethereum-socketio connector from "0.20.7" to "1.8.1"
- Major cleanup and refactor of go-ethereum-socketio logic,
  without affecting the current functionalities.
- Add safety checking of method called in web3Eth and contract functions.
  Add tests to check if it works.
- Add websocket eth RPC support and some helper functions in openethereum-test-ledger.
- Expose WS RPC in geth-testnet, use this port in sample apps
  (required to monitor new blocks in updated web3.js)
- Change raw use of sendRawTransaction to sendSignedTransaction in sample apps
  in order to conform to new web3.js version.
- Add safeStringifyException to cactus-common that can be used by validators
  to sanitize response errors.
- Refactor socketio-based validators to use safeStringifyException.
- Adjust connector README files.
- Improve cleanup after running sample apps
- Fix discounted-cartrade sample ethereum transfer and checking

Closes #2088

Depends on #2053

Signed-off-by: Michal Bajer <michal.bajer@fujitsu.com>

------------

Please review only the last commit, the rest are dependencies that should be merged in separate PRs.

PR #2071, opened 2022-06-07
feat(fabric-socketio-connector): sending transactions signed on the client-side

outSH:
- Add functions to support entire process of sending transaction to the ledger without sharing
  any private key with the connector.
- Add test cases to check the new functionalities.
- Adjust asset-trade sample app to use new logic that doesn't send the private key to the connector.
- Move fabric signing utils to a separate file.
- Add fabric protobuf serializers and deserializers, so they can be send back and forth
  between connector and BLP without loosing type information.
- Add an option to FabricTestLedgerV1 to allow attaching to already running ledger container.
  This can speed up test development/troubleshoot process.
- Improve fabric-all-in-one ledger healthcheck, so that it waits for basic chaincode intialization.
  This is not strictly needed by current tests, but may be expected by future tests.
- Refactor duplicated logic in fabric-connector.
- Fix fabric-socketio module entry points (minor fix).

Closes: #2070

Depends on #2047

Signed-off-by: Michal Bajer <michal.bajer@fujitsu.com>

-------------

As always, please review only the last commit. Remaining ones are just dependencies, that will be squashed before an actual merge.

Issue #2069, opened 2022-06-07
chore(release): missing @hyperledger/cactus-cmd-socket-server npm package


outSH:
There's no `@hyperledger/cactus-cmd-socket-server:1.0.0` package in official NPM repository. It's required by socketio-based connector plugins (validators). Without this dependency, building of validator docker images fail. These images are required by electricty-trade and asset-trade sample apps, so without them, it's impossible to run them through the docker environment.

I'm pretty sure this package was available in NPM earlier on, please restore it if possible.

@izuru0 @jagpreetsinghsasan @petermetz

AbhinavMir:
Is it possible that cactus-cmd-socket-server has shifted to [cactus-cmd-api-server](https://www.npmjs.com/package/@hyperledger/cactus-cmd-api-server)?

petermetz:
@outSH Do you know for a fact that it used to be there?

outSH:
@petermetz No, I'm not 100% sure. Nevertheless, it would be nice to have it on npm like the other packages, so please upload it if possible.

outSH:
@petermetz Now that https://github.com/hyperledger/cactus/issues/2105 is merge and the name of package is consistent, can you publish this package to npm? I need to use it in some other Dockerfiles

petermetz:
@outSH Apologies for the delay! Yes, I'll publish a new release with this package included ASAP!

petermetz:
Re-opening until we can fix the release automation problems.

petermetz:
@outSH I think I already said it but just to be sure pinging here as well: v1.1.2 artifacts are up on npmjs.com now.

PR #2111, opened 2022-07-01
test(connector-sawtooth-socketio): add functional test, bug fix

outSH:
- Add functional test of all functions from sawtooth-socketio validator.
- Refactor sawtooth validator to allow importing as a module, to simplify the functional test.
- Add stopMonitor to terminate tests easily.
- Allow multiple clients to monitor for blocks.
- Fix parsing of URL from config so that it doesn't depend on trailing slash anymore.

Closes: https://github.com/hyperledger/cactus/issues/2107

Depends on: https://github.com/hyperledger/cactus/pull/2109
Depends on: https://github.com/hyperledger/cactus/pull/2110
Depends on: https://github.com/hyperledger/cactus/pull/2047

### DIFF
--- a/packages/cactus-plugin-ledger-connector-sawtooth-socketio/package.json
+++ b/packages/cactus-plugin-ledger-connector-sawtooth-socketio/package.json
@@ -2,6 +2,9 @@
   "name": "@hyperledger/cactus-plugin-ledger-connector-sawtooth-socketio",
   "version": "1.1.2",
   "license": "Apache-2.0",
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
     "start": "cd ./dist && node common/core/bin/www.js",
     "debug": "nodemon --inspect ./dist/common/core/bin/www.js",
@@ -28,6 +31,9 @@
     "xmlhttprequest": "1.8.0"
   },
   "devDependencies": {
+    "@hyperledger/cactus-test-tooling": "1.1.2",
+    "@hyperledger/cactus-common": "1.1.2",
+    "@hyperledger/cactus-api-client": "1.1.2",
     "@types/config": "0.0.41"
   }
 }
--- a/packages/cactus-plugin-ledger-connector-sawtooth-socketio/src/main/typescript/common/core/bin/www.ts
+++ b/packages/cactus-plugin-ledger-connector-sawtooth-socketio/src/main/typescript/common/core/bin/www.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 /*
- * Copyright 2021 Hyperledger Cactus Contributors
+ * Copyright 2022 Hyperledger Cactus Contributors
  * SPDX-License-Identifier: Apache-2.0
  *
  * www.js
@@ -16,7 +16,6 @@
  */
 
 import app from "../app";
-const debug = require("debug")("connector:server");
 import https = require("https");
 
 // Overwrite config read path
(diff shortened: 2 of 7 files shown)


jagpreetsinghsasan (approved):

izuru0 (approved):
LGTM

packages/cactus-plugin-ledger-connector-sawtooth-socketio/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@hyperledger/cactus-plugin-ledger-connector-sawtooth-socketio",
   "version": "1.0.0",
   "license": "Apache-2.0",
+  "main": "dist/common/core/bin/indes.js",
```
petermetz, line 5: @outSH is `index.js` a typo?

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-sawtooth-socketio/package.json
outSH: @petermetz Yes it was - fixed

outSH:
@petermetz Rebased and tested with current main - please review

packages/cactus-plugin-ledger-connector-sawtooth-socketio/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@hyperledger/cactus-plugin-ledger-connector-sawtooth-socketio",
   "version": "1.1.2",
   "license": "Apache-2.0",
+  "main": "dist/index.js",
```
petermetz, line 5: @outSH I recommend keeping the conventions of the other packages when it comes to declaring the paths & configuring the compiler (tsconfig.json file). That will avoid a big chunk of the issues like when some packages do not get published due to strange bugs: https://github.com/hyperledger/cactus/issues/2069 or other kinds of automation scripts failing if they make assumptions that some of these paths are uniform everywhere. If there is a specific reason for not going uniform with the package configuration properties then that's a different question of course but if it's all the same to you then I'd highly recommend changing it to match the rest of the packages.

The same goes for the tsconfig.json file, the include paths there ideally would match the rest of the tsconfig.json files which would lead to the compiler output being structured the same way as well.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-sawtooth-socketio/package.json
outSH: BTW the same applies to all `*-socketio-*` connectors, so maybe it is better to handle in separate PR

petermetz: > @petermetz Sorry, I don't understand, can you clarify the conventions you want to keep (would be even nice to have them documented somewhere)? Are the paths supposed to be like `dist/lib/main/typescript/index.js`? If so, it doesn't seem much work and I could do it in this PR I suppose.

@outSH Yeah, that's one of them, but in general every single property in the package.json file that does not have a specific reason (like contributor list or package name/description) to be different should be the same, uniform configuration (same for tsconfig.json files).

Agreed 100% that this should be documented. Best I could do earlier is to make the changes myself last year but then all of that got deleted by others from my own PR so in the end we got what we got.
I'll create a task in the tracker to have a custom check that enforces a set of rules for these files. That way it's not just a markdown file that will almost nobody ever read but instead people will get a failing CI check if anything is wrong with their packages. This will also make our review process more scalable as we continue to add more packages in the future.

---

> BTW the same applies to all *-socketio-* connectors, so maybe it is better to handle in separate PR

 Yeah I'm slowly but surely trying to get us to a state where all packages are as uniform as possible in terms of build configuration to make everybody's lives easier and this would be a part of that so thank you very much for offering. I'm totally fine with doing a separate and bigger PR with all of them, so I'll approve this one as-is and leave it up to you how you want to do it for this specific package. Thank you again!

outSH:
> Yeah I'm slowly but surely trying to get us to a state where all packages are as uniform as possible in terms of build configuration to make everybody's lives easier and this would be a part of that so thank you very much for offering. I'm totally fine with doing a separate and bigger PR with all of them, so I'll approve this one as-is and leave it up to you how you want to do it for this specific package. Thank you again!

@petermetz I've created a task for it here - https://github.com/hyperledger/cactus/issues/2216. I think I'll take care of it after following PR are merged (to limit the merging headaches):
- https://github.com/hyperledger/cactus/pull/2071
- https://github.com/hyperledger/cactus/pull/2089
- https://github.com/hyperledger/cactus/pull/2111

So please merge this when possible, thanks :)

petermetz:
> > Yeah I'm slowly but surely trying to get us to a state where all packages are as uniform as possible in terms of build configuration to make everybody's lives easier and this would be a part of that so thank you very much for offering. I'm totally fine with doing a separate and bigger PR with all of them, so I'll approve this one as-is and leave it up to you how you want to do it for this specific package. Thank you again!
> 
> @petermetz I've created a task for it here - #2216. I think I'll take care of it after following PR are merged (to limit the merging headaches):
> 
> * [feat(fabric-socketio-connector): sending transactions signed on the client-side #2071](https://github.com/hyperledger/cactus/pull/2071)
> * [chore(connector-go-eth): upgrade web3.js to recent version #2089](https://github.com/hyperledger/cactus/pull/2089)
> * [test(connector-sawtooth-socketio): add functional test, bug fix #2111](https://github.com/hyperledger/cactus/pull/2111)
> 
> So please merge this when possible, thanks :)

@outSH Got it, I'll merge this today then. Cheers!

Issue #2118, opened 2022-07-15
feat(connector-fabric): add WatchBlocks endpoint


outSH:
- Add new endpoint for monitoring new blocks.
- Add `FabricApiClient` which supports new socketio endpoint.
- Add functional test to check the new endpoint.
- Update readme with WatchBlocks usage.

This is part of decommissioning `fabric-socketio` in order to have single fabric connector.

PR #2119, opened 2022-07-15
feat(connector-fabric): add WatchBlocks endpoint

outSH:
- Add new endpoint for monitoring new blocks. Response type is determined from input argument.
  Monitor endpoints are stored in connector.runningWatchBlocksMonitors.
- Add `FabricApiClient` which supports new socketio endpoint.
- Add functional test to check the new endpoint.
- Upgrade fabric-sdk-node to solve test hand issue.
- Loose up `ISocketApiClient` to allow interface monitor options.
- Update readme with WatchBlocks usage.

Regarding fabric-sdk-node upgrade: with sdk2.3, the test used to hang occasionally.
I've ensured that all resources are being cleaned-up, and investigated the problem. The issue was
infinite recurse dns resolution (of fabric peer address) in grpc-js `ResolvingLoadBalancer`.
Could not find exact cause of this strange behaviour, the code there is pretty complex
and didn't want to dig to deep, but I've noticed that recently there were some fixes
in the functions I suspected so I've updated the fabric-sdk, which uses newer grpc-js,
and the problem went away, so I assume this was in fact bug in gprc-js. Run this test on repeat for
hours and didn't notice any error.

Closes: #2118


outSH:
OK, it seems that fabric SDK upgrade is not as trivial as I wanted it to be because it breaks all the custom identity providers feature tests:
- run-transaction-with-identities.test.ts
- run-transaction-with-ws-ids.test

When using 2.3.0-snapshot.69 or later (including 2.5 I've initially used) there's following error during runtime:
`TypeError: Channel credentials must be a ChannelCredentials object`

When using LTS 2.2 (which also solves my test), the fabric connector doesn't compile, because CryptoSuite ` in LTS is still sync instead of async. 

I couldn't make our identity `sign()` function sync in anyway, and I have no idea how to fix the error in recent fabric-sdk, so my proposal is to:
- Disable my test for now, but leave it in repo.
- Keep the current fabric-sdk version.
- Add tasks to upgrade the fabric-sdk, and to restore my test that is flaky at the moment.

@petermetz @jagpreetsinghsasan any thoughts?

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
```diff
@@ -0,0 +1,329 @@
...
+          chaincodeId,
+          transactionId,
+          functionName,
+          functionArgs: decodedArgs,
+        });
+      } catch (error) {
+        log.error(
+          "Could not retrieve transaction from received block. Error:",
+          safeStringifyException(error),
+        );
```
petermetz, line 134: @outSH If we only log here to the error logs of the back-end then how does the client application get notified that transactions will be missing from the response? 

If a sys-admin sees the logs and sees these errors, will they know what to do to fix it? Or if they passed those error logs to us, would we be able to do anything about it?

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
```diff
@@ -0,0 +1,329 @@
...
+        listenerType = "private";
+        break;
+      case WatchBlocksListenerTypeV1.CactusTransactions:
+        listener = async (blockEvent) =>
+          this.monitorCactusTransactionsCallback(blockEvent);
+        listenerType = "full";
+        break;
+      default:
+        // Will not compile if any type was not handled by above switch.
+        const unknownType: never = type;
```
petermetz, line 253: @outSH Why do you need to cast it to `never`? Couldn't you just log the value of `type` instead of this new `unknownType` variable?

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
```diff
@@ -0,0 +1,329 @@
...
+        throw new Error(
+          `Unknown block listen type - '${unknownType}'. Check name and connector version.`,
+        );
+    }
+
+    if (!listener || !listenerType) {
+      // Should never happen
+      throw new Error("Could not determine listener or listenerType.");
+    }
```
petermetz, line 262: @outSH I agree that this will never happen because it wouldn't compile if you didn't assign both of those variables, so IMO this can be removed. With that said if you'd still rather keep this in here (which I can also understand) then I'd make sure to include the `type` in this exception message as well so that people investigating it have more information.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
```diff
@@ -0,0 +1,329 @@
...
+
+    return { listener, listenerType };
+  }
+
+  /**
+   * Subscribe to new blocks on fabric ledger, push them to the client via socketio.
+   *
+   * @param options Block monitoring options.
+   */
+  public async subscribe(options: WatchBlocksOptionsV1) {
```
petermetz, line 272: @outSH Please specify the return type

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
```diff
@@ -0,0 +1,329 @@
...
+      const errorMessage = safeStringifyException(error);
+      log.error(errorMessage);
+      socket.emit(WatchBlocksV1.Error, {
+        code: 500,
+        errorMessage,
+      });
+    }
+  }
+
+  close() {
```
petermetz, line 323: @outSH Please specify the method return type

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
```diff
@@ -0,0 +1,329 @@
...
+        );
+        network.removeBlockListener(listener);
+        this.close();
+      });
+
+      socket.on(WatchBlocksV1.Unsubscribe, () => {
+        log.info(`${WatchBlocksV1.Unsubscribe} => clientId: ${clientId}`);
+        this.close();
+      });
+    } catch (error: any) {
```
petermetz, line 313: @outSH Could this work with the error being typed as `unknown`?

petermetz (changes requested):

petermetz:
> OK, it seems that fabric SDK upgrade is not as trivial as I wanted it to be because it breaks all the custom identity providers feature tests:
> 
> * run-transaction-with-identities.test.ts
> * run-transaction-with-ws-ids.test
> 
> When using 2.3.0-snapshot.69 or later (including 2.5 I've initially used) there's following error during runtime: `TypeError: Channel credentials must be a ChannelCredentials object`
> 
> When using LTS 2.2 (which also solves my test), the fabric connector doesn't compile, because CryptoSuite ` in LTS is still sync instead of async.
> 
> I couldn't make our identity `sign()` function sync in anyway, and I have no idea how to fix the error in recent fabric-sdk, so my proposal is to:
> 
> * Disable my test for now, but leave it in repo.
> * Keep the current fabric-sdk version.
> * Add tasks to upgrade the fabric-sdk, and to restore my test that is flaky at the moment.
> 
> @petermetz @jagpreetsinghsasan any thoughts?

@outSH `TypeError: Channel credentials must be a ChannelCredentials object` means that the grpc versions are different between your client and the server so you could fix this by upgrading the fabric dependency versions consistently across the packages. I'm also fine if you open an issue for doing that and for now just disable the flaky test (just make sure to leave a note in the issue that the test has to be re-enabled as part of the acceptance criteria)

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
outSH: @petermetz This means there's a bug in connector itself, or the connector we use is not compatible with fabric ledger we're connected to (i.e. fabric block data changed). Monitoring in fabric-socketio ignored block like that, maybe it should be more explicit in informing the client that there were some blocks it couldn't parse so that the client app could react accordingly? But should it stop entire monitoring? I will think about notifying the client

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
outSH: @petermetz There's a comment above - Will not compile if any type was not handled by above switch. It's not casting, type in default is already of `never` type if the switch check above is exhaustive. If, in the future, someone adds another type `Foo`, then type in default will be `Foo` (since it wasn't narrowed by any case above), and assigning this to never will break compilation - marking place where type handling has to be added. The log below is to catch type unexpected by compiler, and also make use of `unknownType` which would cause warning otherwise :)

So this is just a trick for additional compile time check.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
outSH: @petermetz Wow, I don't remember why it's here :D I think this is leftover from partial implementation where "unknown" type was popping out and I wanted a way to narrow it before return, but it is redundant here. Removed.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
outSH: @petermetz Sure thing, done :)

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
outSH: @petermetz Sure thing, done :)

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
outSH: @petermetz Sure

outSH:
> @outSH `TypeError: Channel credentials must be a ChannelCredentials object` means that the grpc versions are different between your client and the server so you could fix this by upgrading the fabric dependency versions consistently across the packages. I'm also fine if you open an issue for doing that and for now just disable the flaky test (just make sure to leave a note in the issue that the test has to be re-enabled as part of the acceptance criteria)

I read the same thing on the internet, but I've updated the test ledger to latest 2.4, to match fabric-sdk from this package and the problem remains (and only for identity related stuff). Or do you have different setup in mind??

izuru0 (approved):
LGTM

jagpreetsinghsasan:
@outSH can you re-request review from @petermetz  if the requested changes were addressed?

outSH:
> I read the same thing on the internet, but I've updated the test ledger to latest 2.4, to match fabric-sdk from this package and the problem remains (and only for identity related stuff). Or do you have different setup in mind??

@petermetz Funny thing, I've just needed to update fabric SDK on test ledger as well, it somehow interfered with these problematic tests. All works now, ready to merge :)

petermetz (changes requested):
> > I read the same thing on the internet, but I've updated the test ledger to latest 2.4, to match fabric-sdk from this package and the problem remains (and only for identity related stuff). Or do you have different setup in mind??
> 
> @petermetz Funny thing, I've just needed to update fabric SDK on test ledger as well, it somehow interfered with these problematic tests. All works now, ready to merge :)

@outSH Sorry for the slow response! Very glad to hear it figured itself out somehow! Please resolve the merge conflicts and then pass it back for review!

outSH:
> @outSH Sorry for the slow response! Very glad to hear it figured itself out somehow! Please resolve the merge conflicts and then pass it back for review!
Done

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
petermetz: @outSH Thank you!

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
petermetz: @outSH Thank you for the update and no worries, I forget stuff like this in all the time too. Better to be paranoid with validation haha

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
petermetz: @outSH Cheers!

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
petermetz: @outSH Cheers!

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
petermetz: @outSH Cheers!

petermetz:
Suggested edit:
```diff
diff --git a/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts b/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
index 82bac1f0..15f91a5b 100644
--- a/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
+++ b/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
@@ -19,6 +19,7 @@ import {
 
 import safeStringify from "fast-safe-stringify";
 import sanitizeHtml from "sanitize-html";
+import { RuntimeError } from "run-time-error";
 
 /**
  * WatchBlocksV1Endpoint configuration.
@@ -254,11 +255,9 @@ export class WatchBlocksV1Endpoint {
         listenerType = "full";
         break;
       default:
-        // Will not compile if any type was not handled by above switch.
-        const unknownType: never = type;
-        throw new Error(
-          `Unknown block listen type - '${unknownType}'. Check name and connector version.`,
-        );
+        const validTypes = Object.keys(WatchBlocksListenerTypeV1).join(";");
+        const errorMessage = `Unknown block listen type '${type}'. Check name and connector version. Accepted listener types for WatchBlocksListenerTypeV1 are: [${validTypes}]`;
+        throw new RuntimeError(errorMessage);
     }
 
     return { listener, listenerType };
diff --git a/packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/fabric-watch-blocks-v1-endpoint.test.ts b/packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/fabric-watch-blocks-v1-endpoint.test.ts
index bb713579..92941212 100644
--- a/packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/fabric-watch-blocks-v1-endpoint.test.ts
+++ b/packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/fabric-watch-blocks-v1-endpoint.test.ts
@@ -400,4 +400,31 @@ describe("watchBlocksV1 of fabric connector tests", () => {
 
     await monitorPromise;
   });
+
+  test("Invalid WatchBlocksListenerTypeV1 value gets knocked down", async () => {
+    const monitorPromise = testWatchBlock(
+      "CactusTransactionsTest",
+      "Some_INVALID_WatchBlocksListenerTypeV1" as WatchBlocksListenerTypeV1,
+      () => undefined, // will never reach this because it is meant to error out
+    );
+
+    try {
+      await monitorPromise;
+    } catch (ex: unknown) {
+      // Execution never reaches this point - I'm assuming because the
+      // testWatchBlock method somehow does not fulfil it's obligation of
+      // either succeeding or throwing (it seems to get stuck idling forever
+      // when I debug this in VSCode)
+      expect(ex).toBeTruthy();
+      expect(ex).toBe(Error);
+      expect(ex).toHaveProperty("message");
+      expect((ex as Error).message).toContainAllValues([
+        WatchBlocksListenerTypeV1.CactusTransactions,
+        WatchBlocksListenerTypeV1.Filtered,
+        WatchBlocksListenerTypeV1.Full,
+        WatchBlocksListenerTypeV1.Private,
+        ...Object.keys(WatchBlocksListenerTypeV1),
+      ]);
+    }
+  });
 });

```

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
petermetz: @outSH 

> There's a comment above - Will not compile if any type was not handled by above switch.

Yeah that's fine, I'm not saying we should remove the `default` block, I'm saying that this weird casting of the `type` variable doesn't do anything useful (as far as I can tell but could be totally wrong 'coz I do that a lot anyway)

> If, in the future, someone adds another type Foo, then type in default will be Foo (since it wasn't narrowed by any case above), and assigning this to never will break compilation - marking place where type handling has to be added.

> So this is just a trick for additional compile time check.

I agree with all of that, 100%.

What I'm saying is that something like this also works as the code for the default branch and it is less confusing because there is no weird casting (and I added this extra little thing that it dumps the valid enum values back into the error message so that people can know what their options were)

```typescript
        const validTypes = Object.keys(WatchBlocksListenerTypeV1).join(";");
        const errorMessage = `Unknown block listen type '${type}'. Check name and connector version. Accepted listener types for WatchBlocksListenerTypeV1 are: [${validTypes}]`;
        throw new RuntimeError(errorMessage);
```

which then produces error messages like this:

```sh
[2022-12-19T20:15:50.069Z] ERROR (WatchBlocksV1Endpoint): RuntimeError: Unknown block listen type 'Some_INVALID_WatchBlocksListenerTypeV1'. Check name and connector version. Accepted listener types for WatchBlocksListenerTypeV1 are: [Filtered;Full;Private;CactusTransactions]
        at WatchBlocksV1Endpoint.getBlockListener (/***/cactus/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts:260:15)
```

I tried to write a test for it as well (please see my suggested patch that you can apply with one click from VSCode via the GitHub extension) but the test case itself does not work because if I feed an invalid enum value to the watch endpoint the `testWatchBlock()` test utility method does not seem to complete the promise at all (e.g. neither reject nor resolve).

Please take a look at the patch and let me know what you think. If I'm just not understanding something then it's fine to merge as-is too because I don't want to put extra work/burden on you.

This is the patch comment I'm talking about: https://github.com/hyperledger/cactus/pull/2119#issuecomment-1358266362

petermetz (commented):
@outSH Please see https://github.com/hyperledger/cactus/pull/2119#discussion_r1052614429 and then pass it back for review after you decided how to proceed.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/watch-blocks/watch-blocks-v1-endpoint.ts
outSH: @petermetz 
Your proposal is extending the runtime error handling (i.e. we received some unknown value and want to handle it accordingly). My solution (with assignment to `never`) is to handle issue of not including specific option in switch during compilation, and throw an error during runtime.

[Have a look at this TS playground example I've prepared](https://www.typescriptlang.org/play?#code/KYOwrgtgBAyg9hYBRc0DeBYAUFXUCCUAvFAOQCGpANNnlAELFkBG1teAwk6QMZtYBfbNgBmYEDwAuASzggowAB4ALcmADOMgG7AOy4DwDWACkkBPAA7AAXLATJUASiiYcUdQHdpknstOXgZ1c6HnJ1YDtEFEgAOnxrdhC5dTgAG2AY1LgAc2MAIgBxOEkCPMdEvGYAJ2ByQwBuCtDwyIdY+gS3JJAU9Myc-KKS+jKK3GrahqawiPgo1BiOTrpcHmS0jKzcwuKoDlGuypq6xsOAE2ARNVTJZZXV5JLxQxA4DxAAFQDbEGAdKqY5isp3uuEkyiqbygvw8UCQVUhVWMYxWAAMAKogF5veRAiIAWjIABI0M9Xu8vlYBKRUTRDnRHCCoEJBMIsGIJDI5NDivAkCo1JppDo9AYTHjbHM2hAguxPN5fP4rLLDs1ZvZohA4ndur1NgMdiV8AcQjNWpqYh0UWseht+tshgwTXgLlcwDcdSsbZooGScZSbNC-sAASQ8fUoAB6SNQANkKWa0hQaTqHklMLqaTZEDkZjpKCSOAFgJkX7-UgxYwAJgAzFWq+V6XhwZDYTC4Qi4EiUXQMVjybiS4TSCS-RSAtTaT2oIz2CyWaJxFJZPIICnwmd6Mdxd9zU4XHKvD4-HiVaaWgmFvFres+ltBrtjY37hMTtMLxqFlam967ffDU6z4rK+Uyqmal6xEsN62neBqOvsQF0NGIFMrgrrXLc07ek8-b+ruZYhoCASoSsLZQu28KIsiTa9pi2LvMWVhQMOo64eOVI0nSoIzky85AA)

- `exhaustiveCheck` is working code, all cases are handled.
- `notSoExhaustiveCheck` is missing handling of one type (for example - we've added a new type and forgot about switch here). Assignment to `unknownType` fails because rhs is not `never` - there are cases it will have valid value we just don't support yet.
- `missedBreak` example show another kind of error - we didn't break in single case and it's possible to fall to default by accident. This is also catched.

Value not defined in our definitions is handled correctly in all cases.

outSH: @petermetz I've included `RuntimeError` and test you've proposed, but kept assignment to `never` because of reasons described above.

petermetz: @outSH Whoah, epic facepalm. I finally get it, thank you for the patience and the detailed explanation! 
Damn, I really thought the TS compiler was preventing unhandled enum values by default without this (seemingly) pointless extra `never` cast thingy. Sorry and thanks again!

petermetz (approved):
@outSH LGTM, thank you!

PR #2130, opened 2022-07-22
feat(cc-tx-visualization): initial commit

RafaelAPB:
The proposed plugin allows generating process models from arbitrary cross-chain use cases. Currently supports Fabric and Besu. More documentation to come soon.

This project is supported by Blockdaemon https://blockdaemon.com/

cc @petermetz


packages/cactus-plugin-cc-tx-visualization/src/main/csv/example-dummy-use-case.csv
```diff
@@ -0,0 +1,9 @@
+caseID;timestamp;blockchainID;invocationType;methodName;parameters;identity
+1;2022-04-28T16:14:23.922Z;TEST;send;registerEmission;"[""1,100""]";company_A
+2;2022-04-28T16:14:23.922Z;TEST;send;registerEmission;"[""2,100""]";company_B
+1;2022-04-29T20:01:03.922Z;TEST;send;registerEmission;"[""2,100""]";company_A
+1;2022-05-10T06:01:03.922Z;TEST;send;getEmissions;"[""company_A""]";auditor
+1;2022-05-11T09:47:43.922Z;TEST;send;mintEmissionToken;"[""uuidTokenEmissions""]";protocol_administrator
+2;2022-04-29T22:47:43.922Z;TEST;send;registerEmission;"[""2,100""]";company_B
+2;2022-05-11T09:47:43.922Z;TEST;send;getEmissions;"[""company_B""]";auditor
+2;2022-05-12T13:34:23.922Z;TEST;send;mintEmissionToken;"[""uuidTokenEmissions""]";protocol_administrator
```
petermetz, line 9: @RafaelAPB If this is test data, not something that would be used in production then I'd recommend moving it under `src/test` from `src/main`

packages/cactus-plugin-cc-tx-visualization/.gitignore
```diff
@@ -0,0 +1,4 @@
+cactus-openapi-spec-plugin-consortium-manual.json
+src/main/typescript/generated/openapi/typescript-axios/.npmignore
+src/main/csv
```
petermetz, line 3: @RafaelAPB If we are ignoring `src/main/csv` then why is there a file committed in that directory? 
This question of mine might become irrelevant if you move it to the test directory - unless if that move is done then you'd also plan on updating the ignore pattern here.

packages/cactus-plugin-cc-tx-visualization/src/main/typescript/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./public-api";
```
petermetz, line 1: @RafaelAPB Could you please also add an index.web.ts to match the common export structure of the other packages?

packages/cactus-plugin-cc-tx-visualization/src/main/typescript/plugin-cc-tx-visualization.ts
```diff
@@ -0,0 +1,453 @@
...
+      this.log.info(`No HTTP server found, skipping...`);
+    }
+  }
+
+  public async registerWebServices(
+    app: Express,
+  ): Promise<IWebServiceEndpoint[]> {
+    const webApp: Express = this.options.webAppOptions ? express() : app;
+
+    if (this.options.webAppOptions) {
```
petermetz, line 183: @RafaelAPB The plugins hosting their own web servers has been deprecated a while back IIRC. Could you please migrate to defining an endpoint class and then registering that with the API server (by returning it through `getOrCreateWebServices()`)

petermetz (changes requested):
@RafaelAPB Are the python notebooks deployed & used in production or are those just tooling for local testing purposes? If the latter I'd move them to somewhere in within `src/test/python/fixtures/` or something like it.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
```diff
@@ -1,6 +1,6 @@
+/* eslint-disable prettier/prettier */
```
petermetz, line 1: @RafaelAPB Please re-enable the linter

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
```diff
@@ -1063,6 +1114,70 @@ export class PluginLedgerConnectorFabric
...
+          const txReceipt = new amqp.Message(extendedReceipt);
+          this.amqpQueue?.send(txReceipt);
+          this.log.debug(`Sent simple transaction receipt to queue ${this.queueId}`);
+        }
+        const endTimeFabricReceipt = new Date();
+        this.log.debug(`EVAL-${this.className}-GENERATE-AND-CAPTURE-RECEIPT:${endTimeFabricReceipt.getTime()-startTimeFabricReceipt.getTime()}`);
+
+      }
+
       const outUtf8 = out.toString("utf-8");
```
petermetz, line 1181: @RafaelAPB I'd move all this logic to the cc viz plugin itself, have the connector plugin and the cc viz plugin decoupled from each other. If there must be some dependency from one to the other then I'd definitely go with the cc viz plugin depending on the connector plugin and not the other way around because the connector is a base plugin that provides something pretty much every application needs (connectivity to the ledger itself) while the cc viz plugin is a higher level feature.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1.test.ts
```diff
@@ -229,25 +227,6 @@ describe(testCase, () => {
...
-        K_CACTUS_FABRIC_TOTAL_TX_COUNT +
-        '"} 3';
-      expect(res).toBeTruthy();
-      expect(res.data).toBeTruthy();
-      expect(res.status).toEqual(200);
-      expect(res.data.includes(promMetricsOutput)).toBeTrue();
-    }
-
     {
       const req: RunTransactionRequest = {
```
petermetz, line 231: @RafaelAPB Was this deleted by mistake? Or if not, then why was it deleted? :-)

packages/cactus-test-tooling/src/main/typescript/socketio-test-setup-helpers/socketio-test-setup-helpers.ts
```diff
@@ -72,7 +72,7 @@ export function connectTestClient(socket: ClientSocket): Promise<ClientSocket> {
     socket.on("connect_timeout", errorHandlerFactory("connect_timeout"));
 
     socket.on("connect", () => {
-      socket.removeAllListeners();
+      //socket.removeAllListeners();
```
petermetz, line 75: @RafaelAPB Please bug me until I come up with an explanation to why this is not compiling.
BTW looking at the code in a little more detail, I'd suspect that maybe the problems with your package dependencies is the culprit. You are doing cross-package imports in the Typescript code, but not declaring the dependencies in the package.json files which is defintely a problem and it could lead to weird compilation errors like this one.

petermetz (changes requested):

packages/cactus-plugin-cc-tx-visualization/src/main/csv/example-dummy-use-case.csv
RafaelAPB: fixed.

packages/cactus-plugin-cc-tx-visualization/.gitignore
RafaelAPB: because that is a test file useful for people trying out the plugin. It was moved to the test directory now.

packages/cactus-plugin-cc-tx-visualization/src/main/typescript/index.ts
RafaelAPB: done!

packages/cactus-plugin-cc-tx-visualization/src/main/typescript/plugin-cc-tx-visualization.ts
RafaelAPB: Done

packages/cactus-plugin-ledger-connector-besu/src/main/json/openapi.json
```diff
@@ -828,6 +828,18 @@
                 ],
                 "additionalProperties": false,
                 "properties": {
+                    "caseID": {
```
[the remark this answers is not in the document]
RafaelAPB: this is tied to the proposed plugin. however, our plugin requires the involved connectors to emit events based on issued transactions; and those transactions have an optional caseID. In other words, I believe the case ID must be on the transaction schema of each connector (although optional)

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
RafaelAPB: This means that a cctxviz plugin needs to receive connectors as input. How can I generate a modified transaction receipt? Would I have to implement a modified connector class? @petermetz

packages/cactus-plugin-ledger-connector-besu/src/main/json/openapi.json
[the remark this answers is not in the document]
RafaelAPB: @petermetz please let me know what you think of this. If you request that no changes are done to the connectors and their data schema at all, then probably the best implementation choice is to create a modified connector plugin that extends the base connector. But in that case, we will keep this code separate from the main codebase due to refactoring effort.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -523,6 +567,47 @@ export class PluginLedgerConnectorBesu
       const out = await this.transact(txReq);
       const success = out.transactionReceipt.status;
       const data = { success, out };
+      const endTimeToTransaction = new Date();
```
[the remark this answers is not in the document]
RafaelAPB: @petermetz  in other words, would it be better for our plugin to instantiate a pair of modified connectors, like the odap-hermes plugin does?

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
RafaelAPB: @petermetz can we have input on this? I believe this is decisive for this PR to go onward (if not acceptable, we will just keep cctxviz as a fork from Cactus).

packages/cactus-plugin-cc-tx-visualization/.gitignore
petermetz: > because that is a test file useful for people trying out the plugin. It was moved to the test directory now.

@RafaelAPB Gotcha, thank you. It is fine to have it in the test directory then.

packages/cactus-plugin-ledger-connector-besu/package.json
```diff
@@ -57,6 +57,7 @@
     "@hyperledger/cactus-core": "1.0.0",
     "@hyperledger/cactus-core-api": "1.0.0",
     "axios": "0.21.4",
+    "amqp-ts": "1.8.0",
```
petermetz, line 60: @RafaelAPB I'd keep amqp specific things out of the connectors and instead use an RxJS subject/observable instead (same thing we were talking about a while back during one of the pair programming sessions).
I can show you more detailed code examples just let me know.

packages/cactus-plugin-ledger-connector-besu/src/main/json/openapi.json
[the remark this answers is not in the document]
petermetz: > 

@RafaelAPB Sorry for the slow response again! This goes back to the same thing I was talking about a little higher up in the comments talking about RxJS streams made with subjects/observables.
If the cc viz plugin is the one where these case IDs logically belong then you could create an interface there that contains it and consume instances of that interface through the RxJS piping. Again, I'm happy to show some code on how this is done (it's pretty simple once you see it written down IMO)

This way you avoid putting stuff in the connector's more generic data model but also avoid creating a fork of it (the modified version that you suggested) so that we still only have to maintain the single codebase for the connector but it just has one tiny extra feature which is mostly decoupled from the other features of it (so complexity does not increase substantially but neither does maintenance burden)

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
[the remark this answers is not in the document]
petermetz: > @petermetz in other words, would it be better for our plugin to instantiate a pair of modified connectors, like the odap-hermes plugin does?

@RafaelAPB Sorry again for the slow response. Yeah, that's roughly what I mean as long as `modified` just means that you are instantiating the connectors with whatever unique configuration you need, but not actual source code modifications that are only good for/used by the cc viz (see my other comments about a generic tx RxJS stream through replay subjects)

packages/cactus-plugin-ledger-connector-fabric/package.json
```diff
@@ -58,6 +58,7 @@
     "@hyperledger/cactus-core": "1.0.0",
     "@hyperledger/cactus-core-api": "1.0.0",
     "axios": "0.21.4",
+    "amqp-ts": "1.8.0",
```
petermetz, line 61: @RafaelAPB Same comment here as I made for the besu connector (I won't spam you twice with all the same comments, but the requests are the same)

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
```diff
@@ -112,8 +113,11 @@ import {
   getTransactionReceiptByTxID,
   IGetTransactionReceiptByTxIDOptions,
 } from "./common/get-transaction-receipt-by-tx-id";
+
 import { GetBlockEndpointV1 } from "./get-block/get-block-endpoint-v1";
 import { querySystemChainCode } from "./common/query-system-chain-code";
+import * as amqp from "amqp-ts";
+import {FabricV2TxReceipt, IsVisualizable} from "@hyperledger/cactus-plugin-cc-tx-visualization/src/main/typescript/models/transaction-receipt";
```
petermetz, line 120: @RafaelAPB Same thing here, the fabric connector should not depend on the cc viz plugin, it should be the other way around.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
petermetz: > @petermetz can we have input on this? I believe this is decisive for this PR to go onward (if not acceptable, we will just keep cctxviz as a fork from Cactus).

@RafaelAPB Sorry again for the slow response! You can add the generic RxJS subject to the connectors that I spoke of and those will have a generic type parameter which can be the interface type that contains what is actually in your custom tx receipts (caseID for example).
Happy to provide code examples just let me know if this is not enough information!
The PR is definitely acceptable just needs some refactor to make it generic enough so that it doesn't hinder future improvements by hardcoding too many specifics in the connectors (such as AMQP)

petermetz (changes requested):

petermetz:
@RafaelAPB  Just bumping the thread to make sure we don't forget: Are you still working on this one? I'm trying to clear out some of the backlog of the pull requests that we have.

petermetz (commented):
@RafaelAPB Are you still working on this one?

RafaelAPB:
@petermetz for now i don't have the bandwidth to work on this package to be production ready. Let us keep a fork with the R&D version of it and come back later if we have the time. I'll close it

PR #2153, opened 2022-08-31
feat(connector-iroha2): add support for Iroha V2

### LINKED ISSUE #2138: feat(connector-iroha2): add support for Iroha V2
We need to support Iroha V2 in cactus.

Iroha V2 has been completely rewrote, the client SDK and API are completely different from Iroha V1. For this reason I'd propose to introduce a new connector for this new version.

Extending current Iroha connector would cause a lot of mess both in npm dependencies and in code.

**Additional context**
I will try to introduce this new feature if parts: ledger container, TS helper class, connector and it functions.

outSH:
- Add new Iroha V2 cactus connector.
- Two OpenAPI endpoints are implemented: `transact` and `query`. Both endpoints support
  critical subset of instructions and queries supported by the upstream javascript iroha sdk.
- One SocketIO endpoint can be used to monitor new blocks from the ledger.
- New connector can be used through a verifier-client interface.
- All added functions are tested in functional test suites and documented.
- Added execution of Iroha2 tests to the CI.

Additional notes:
- Connector doesn't work well with cactus module system, the issue has been reported and described
  in README. PR is not merge-ready until this is fixed (the CI should fail now).
- Iroha V2 javascript packages are not available on official npm yet, had to include `.npmrc` with
  private npm address. I'm not sure if there's ETA of delivering these through NPM, so it might be
  necessary to commit it after all.

Closes #2138
Depends on #2140


outSH:
@petermetz @izuru0 This is almost ready, I'm only waiting for fixes in upstream javascript SDK. Please start the review in the meantime

outSH:
@petermetz @izuru0 After fixes from Iroha team this is ready for merge. Please review, I'll introduce future upgrades in separate PRs

izuru0 (approved):
LGTM

@petermetz can you review thanks

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/plugin-ledger-connector-iroha2.ts
```diff
@@ -0,0 +1,629 @@
...
+  }
+
+  /**
+   * Create Cactus IrohaV2 client using both defaultConfig (defined during class creation)
+   * and config specified in arg.
+   *
+   * @param baseConfig Iroha V2 base connection configuration.
+   * @returns `CactusIrohaV2Client`
+   */
+  public async getClient(
```
petermetz, line 321: @outSH I recommend calling a method that creates a new object X either as `newX()` or `createX()`.
`getX()` *usually* implies that `X` was already cached somewhere/created elsewhere and now we are just returning it.

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/plugin-ledger-connector-iroha2.ts
```diff
@@ -0,0 +1,629 @@
...
+    T extends (...args: any[]) => unknown
+  >(
+    client: CactusIrohaV2Client,
+    transactFunction: T,
+    params: unknown[] | undefined,
+    expectedCount: LengthOf<Parameters<T>>,
+  ): void {
+    transactFunction.apply(
+      client,
+      this.checkArgsCount(params, expectedCount, transactFunction.name),
```
petermetz, line 416: @outSH Please try to avoid nesting function calls inline where possible. It's easier to make the code read like English prose if you store the output of the checkArgsCount call in a variable that hints at what it is in there.

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/web-services/query-endpoint.ts
```diff
@@ -0,0 +1,106 @@
...
+import { safeStringifyException } from "../utils";
+
+import OAS from "../../json/openapi.json";
+
+export interface IQueryEndpointOptions {
+  logLevel?: LogLevelDesc;
+  connector: PluginLedgerConnectorIroha2;
+}
+
+export class QueryEndpoint implements IWebServiceEndpoint {
```
petermetz, line 31: @outSH Please make sure to include the Iroha version and the Cacti version in the name (it's not done consistently elsewhere in the code either so apologies for asking for this additional clarity).

I'd recommend `Iroha2QueryEndpointV1` but anything else will also do that makes it clear that this is a V1 endpoint of the framework that is used to communicate with V2 of Iroha ledgers.
I'll paste this same thing to a bunch of different places but the bottom line is the same for all of it: Over time it will get very confusing when you try to locate which file/class is where for which ledger/framework version. I know it might feel needlessly verbose now but the longer term it should pay dividends in reduced confusion and mistakes made when someone would try to use one in the place of the other.

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/web-services/transact-endpoint.ts
```diff
@@ -0,0 +1,106 @@
...
+import { safeStringifyException } from "../utils";
+
+import OAS from "../../json/openapi.json";
+
+export interface ITransactEndpointOptions {
+  logLevel?: LogLevelDesc;
+  connector: PluginLedgerConnectorIroha2;
+}
+
+export class TransactEndpoint implements IWebServiceEndpoint {
```
petermetz, line 31: @outSH Please make sure to include the Iroha version and the Cacti version in the name (it's not done consistently elsewhere in the code either so apologies for asking for this additional clarity).

packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts
```diff
@@ -0,0 +1,591 @@
...
+  });
+
+  describe("Account tests", () => {
+    let newAccountName: string;
+    let newAccountDomainName: string;
+    let newAccountCredentials: Iroha2KeyPair;
+
+    // Register new account (RegisterAccount)
+    beforeAll(async () => {
+      newAccountName = addRandomSuffix("fooAcc");
```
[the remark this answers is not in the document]
petermetz: @outSH Can a UUID be used here instead?

packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts
```diff
@@ -0,0 +1,591 @@
+/**
+ * Tests for executing Iroha instructions and queries through the cactus connector.
```
petermetz, line 2: @outSH Please make sure to include the Iroha version and the in the comments and the file name (it's not done consistently elsewhere in the code either so apologies for asking for this additional clarity).

packages/cactus-plugin-ledger-connector-iroha2/package.json
```diff
@@ -0,0 +1,80 @@
...
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "codegen": "run-p 'codegen:*'",
+    "codegen:openapi": "npm run generate-sdk",
+    "generate-sdk": "openapi-generator-cli generate -i ./src/main/json/openapi.json -g typescript-axios -o ./src/main/typescript/generated/openapi/typescript-axios/ --reserved-words-mappings protected=protected"
+  },
+  "dependencies": {
+    "@hyperledger/cactus-common": "1.0.0",
```
petermetz, line 48: @outSH FYI: Once the 1.1.0 release goes out (which is pending ATM) you'll need to manually update these versions to 1.1.0 as well (otherwise the build will start breaking in really strange ways in some places). There's no automated way to do this yet for PRs that get caught up in a release, apologies for the complication!

packages/cactus-plugin-ledger-connector-iroha2/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "hada": "0.0.8",
+    "rxjs": "7.3.0",
+    "express": "4.17.1",
+    "socket.io": "4.4.1",
+    "fast-safe-stringify": "2.1.1",
+    "sanitize-html": "2.7.0",
+    "undici": "5.10.0"
+  },
+  "devDependencies": {
+    "@hyperledger/cactus-test-tooling": "1.0.0",
```
petermetz, line 64: @outSH Same thing about manual version bumps here for the dev dependencies as above.

packages/cactus-plugin-ledger-connector-iroha2/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "@hyperledger/cactus-common": "1.0.0",
+    "@hyperledger/cactus-core": "1.0.0",
+    "@hyperledger/cactus-core-api": "1.0.0",
+    "@iroha2/client": "3.0.0",
+    "@iroha2/crypto-core": "0.1.1",
+    "@iroha2/crypto-target-node": "0.4.0",
+    "@iroha2/data-model": "3.0.0",
+    "hada": "0.0.8",
+    "rxjs": "7.3.0",
+    "express": "4.17.1",
```
petermetz, line 57: @outSH I only saw type imports of express (if this is correct, then) please move express to the `devDependecies` section instead because it's not used at runtime and therefore we don't need to be using resources to install it separately.

packages/cactus-plugin-ledger-connector-iroha2/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "socket.io": "4.4.1",
```
petermetz, line 58: @outSH Same exact request about socket.io as I did with `express` one line above.
And also goes to any other dependencies that you ONLY `import type {...}` => they can be declared as dev dependencies because we only use them to compile the TS code but not at runtime.

packages/cactus-plugin-ledger-connector-iroha2/package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "@hyperledger/cactus-plugin-ledger-connector-iroha2",
+  "version": "1.0.0",
```
petermetz, line 3: @outSH Same thing here about the version number needing a manual bump after the 1.1.0 release, apologies.

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts
```diff
@@ -0,0 +1,591 @@
...
+  describe("Account tests", () => {
+    let newAccountName: string;
+    let newAccountDomainName: string;
+    let newAccountCredentials: Iroha2KeyPair;
+
+    // Register new account (RegisterAccount)
+    beforeAll(async () => {
+      newAccountName = addRandomSuffix("fooAcc");
+      expect(newAccountName).toBeTruthy();
+      newAccountDomainName = addRandomSuffix("newAccDomain");
```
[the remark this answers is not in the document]
outSH: @petermetz These has been resolved. I kept distinctive names with random suffix for easier debugging of tests.

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/plugin-ledger-connector-iroha2.ts
outSH: @petermetz Good idea. Normally I don't think the user should be concerned on how the response is created but here the client must be freed so an information that this creates new object fits well.

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/plugin-ledger-connector-iroha2.ts
outSH: @petermetz OK, the response of `checkArgsCount` can't be infered from the name so separate var will help.

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/web-services/query-endpoint.ts
outSH: @petermetz Agree, that's why majority of names in this module have verbose names. I've simply overlooked this place - fixed :)

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/web-services/transact-endpoint.ts
outSH: @petermetz Done.

packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts
outSH: @petermetz Which file name? The test file name?

packages/cactus-plugin-ledger-connector-iroha2/package.json
outSH: @petermetz Done

packages/cactus-plugin-ledger-connector-iroha2/package.json
outSH: Done

petermetz:
Suggested edit:
```diff
diff --git a/packages/cactus-plugin-ledger-connector-iroha2/openapitools.json b/packages/cactus-plugin-ledger-connector-iroha2/openapitools.json
index 29f5d069..601ac1d6 100644
--- a/packages/cactus-plugin-ledger-connector-iroha2/openapitools.json
+++ b/packages/cactus-plugin-ledger-connector-iroha2/openapitools.json
@@ -2,6 +2,6 @@
   "$schema": "node_modules/@openapitools/openapi-generator-cli/config.schema.json",
   "spaces": 2,
   "generator-cli": {
-    "version": "5.2.0"
+    "version": "5.2.1"
   }
 }

```

petermetz:
Suggested edit:
```diff
diff --git a/packages/cactus-plugin-ledger-connector-iroha2/package.json b/packages/cactus-plugin-ledger-connector-iroha2/package.json
index aab130a5..83e9a8d8 100644
--- a/packages/cactus-plugin-ledger-connector-iroha2/package.json
+++ b/packages/cactus-plugin-ledger-connector-iroha2/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hyperledger/cactus-plugin-ledger-connector-iroha2",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "description": "Allows Cactus nodes to connect to an Iroha V2 ledger.",
   "keywords": [
     "Hyperledger",

```

packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts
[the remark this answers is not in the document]
petermetz: @outSH Got it, thanks. I'm dismissing the alert because Math.random is bad, but it's fine for test cases IMO since that Math.random will never run in production code.

packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts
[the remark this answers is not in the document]
petermetz: @outSH Got it, dismissing the alert.

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/plugin-ledger-connector-iroha2.ts
petermetz: @outSH Thank you for updating!

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/plugin-ledger-connector-iroha2.ts
petermetz: @outSH Thank you!

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/web-services/query-endpoint.ts
petermetz: @outSH Thank you very much! This helps a lot!

packages/cactus-plugin-ledger-connector-iroha2/src/main/typescript/web-services/transact-endpoint.ts
petermetz: @outSH Thank you!

petermetz:
Suggested edit:
```diff
diff --git a/packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts b/packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha2-instructions-and-queries.test.ts
similarity index 99%
rename from packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts
rename to packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha2-instructions-and-queries.test.ts
index befa9a8e..cbe51a48 100644
--- a/packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts
+++ b/packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha2-instructions-and-queries.test.ts
@@ -1,5 +1,5 @@
 /**
- * Tests for executing Iroha instructions and queries through the cactus connector.
+ * Tests for executing Iroha 2 instructions and queries through the cactus connector.
  */
 
 //////////////////////////////////

```

packages/cactus-plugin-ledger-connector-iroha2/package.json
petermetz: @outSH Thank you!

packages/cactus-plugin-ledger-connector-iroha2/package.json
petermetz: @outSH Thank you!

packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha-instructions-and-queries.test.ts
petermetz: @outSH Yeah. I posted a suggestion edit patch as a comment here: https://github.com/hyperledger/cactus/pull/2153#issuecomment-1299342415 to show what I mean (rename + also in the comment I said Iroha 2 instead of just Iroha

petermetz:
@outSH Update about the 1.1.0 release packages being issued: we no longer have permission to directly publish to npm ourselves so it has to go through the github actions which I have a PR open for, but we'll have to test it to see if it even works and because we don't have the npm token of our own, this can only be tested "in production" as in the PR of mine that is adding the auto-publish action has to be merged and then we need to re-tag the release and see if it triggers and if the publishing process succeed or not (and if it does not then the whole PR+review+merge all over again until it gets fixed).

Kinda off-topic but I wanted to keep you in the loop about what's taking so long.

petermetz (changes requested):
@outSH Please rebase onto upstream/main to disappear the parent PR (the one that this one used to depend on) from the commit log of this PR

petermetz:
@outSH FYI the v1.1.2 package artifacts are now published onto npm and can be used.

outSH:
@petermetz 

- I've rebased with current main
- Changed package version
- Done changes you've recommended in review (mostly to make it more distinct that this is for Iroha V2)

I've also merged all the follow-up PR for this connector to avoid juggling back and forth fixes and make it more manageable in single PR. All PR I've merged were already approved so I hope it won't be a problem. The PR list:
- https://github.com/hyperledger/cactus/pull/2168
- https://github.com/hyperledger/cactus/pull/2171
- https://github.com/hyperledger/cactus/pull/2172
- https://github.com/hyperledger/cactus/pull/2178

I'll close them if you agree to merge it as one.
Please do a final review and let me know if there's anything more you'd like to improve.

.gitignore
```diff
@@ -44,7 +44,6 @@ bin/
 lerna-debug.log
 cactus-openapi-spec.json
 cactus-openapi-spec-*.json
-.npmrc
```
petermetz, line 47: @outSH The .npmrc file can contain tokens and therefore is strongly recommended to not keep it under version control. 
Even if there are no tokens in somebody's own .npmrc file, the fact that they have one will give them all sorts of grief if we now start committing one and leave them to manually deal with that every single time they make a contribution.

My recommendation here is to:
1. Add a new npm script in the root package.json called `"init-registries"` that will do the same thing that the config file does now (`npm config set @iroha2:registry=https://nexus.iroha.tech/repository/npm-group/`)
2. Change the scripts (probably the configure script the right place for it but let me know if you have any other place you think it would be best) so that they invoke this other utility script from 1) to set the npm config up with the registry informaiton.
3. Delete the .npmrc file you committed
4. Revert the change that makes git not ignore the .npmrc files

This way I'm hoping that everything will still work transparently to contributors, as-in, there will be no extra manual configuration step involved when setting up the project build, but without the problem of us committing the .npmrc file and then condemning everybody to having to deal with that.

petermetz (changes requested):
@outSH

> I'll close them if you agree to merge it as one.

Works for me. Ideally each PR would be as small as possible as a self contained change by itself, but in this case I think it's fine to combine them in the interest of moving things along a bit faster (I know it doesn't look like it but I try to be cognizant of productivity and velocity with my change requests haha)

> Please do a final review and let me know if there's anything more you'd like to improve.

I left a little essay about the .npmrc thing in the comments above, please take a look at that, but other than that it's LGTM

.gitignore
outSH: @petermetz Yeah, I agree, I was kinda hoping they'll release these packages sooner and was aiming for simplest solution now, didn't think about secret leaking though. 

I've followed your suggestion (and additionally added `init-registries`  to `reset:yarn-lock` since it uses yarn install). The two problems I see:
- `npm config` adds entry into user `~/.npmrc`, not project one. I think silently modifying user environment is not a very good practice, but the harm and possible drawback is small so I think it's OK (we also have script for installing yarn globally, but this requires explicit user action).
- Calling `yarn install` directly, without running configure / reset:yarn-lock first will fail with fairly cryptic error. This is very rare I think, but we can keep and eye if someone doesn't report it.

If these are acceptable for you then we can merge I think :) The alternative would be to put .npmrc in iroha2 package only - that way the risk of leaking secrets is minimized and the process is transparent, but the current state seem OK as well (ideally, we'll remove it once Iroha V2 releases it's official packages so no use in dwelling over this for too long).

petermetz: @outSH Thank you very much for the updates! Agreed on both of these also being issues that you mentioned so I'm keeping my fingers crossed for a quick iroha v2 client library release!

petermetz (approved):
@outSH LGTM, thank you!

petermetz (changes requested):
@outSH Sorry, last one: the tests need to pass too. (I retried them just now and they failed twice in a row so I'm thinking it's probably not a flake either).

outSH:
> Sorry, last one: the tests need to pass too. (I retried them just now and they failed twice in a row so I'm thinking it's probably not a flake either).

@petermetz Yeah, there was a bug in Iroha2 test ledger healthcheck, sorry for that. Should be all good now. Please upload `tools/docker/iroha2-all-in-one` image from this PR to hyperledger ghcr, I'll replace the links in the test and we should be good to go.

petermetz (changes requested):
@outSH Got it, thanks for the fix! I built and pushed the new image just now as: `ghcr.io/hyperledger/cactus-iroha2-all-in-one:2022-12-22-6c715bdb` please re-test with the new image and then pass it back for review!

outSH:
Done

petermetz (approved):
@outSH Cheers, LGTM!

PR #2154, opened 2022-09-01
feat(cbdc-bridging-app): refactor ODAP plugin implementation

### LINKED ISSUE #2132: refactor(odap-plugin): make implementation ledger agnostic
1. The existing implementation of the ODAP plugin is hard-coded for Hyperledger Fabric and Hyperledger Besu, which makes it difficult to develop new gateways for other ledgers.
2. Additionally, the assets involved in the transfer are specified at the Gateway level, and not at the Request level. At the moment, in order to perform N transfers using ODAP, it is required to create N gateways, which is not desirable.

1. Have a clear separation of what is the gateway functionality and the specific implementation for different ledgers.
Example: instead of having the current definition (Fabric and Besu parameters hard-coded), we could have one entity that takes care of the gateway functionality, which is extended by the specific implementations (BesuOdapGateway and FabricOdapGateway). Among others, these gateways would need to implement the methods regarding the create/delete/lock/unlock functionality specific to that ledger.
2. It is crucial that, within the same gateway object, we can perform multiple asset transfers without having the need of creating new ones. We would remove the asset IDs from the Gateway attributes and add them to SessionData, which would be populated when receiving a ClientV1Request.

AndreAugusto11:
* From now there is one PluginOdapGateway that must be extended by any implementation of a gateway
* This commit intends to decouple the gateway implementation from any ledger
* There is still a Gateway implementation for Fabric and Besu
* The asset being transferred is now in the scope of the request and not the gateway, which makes possible the transfer of multiple assets without having to create new gateways

closes #2132 


petermetz (approved):
@AndreAugusto11 Thank you, LGTM!

PR #2171, opened 2022-10-14
feat(connector-iroha2): wait for transaction commit

outSH:
- Add a new argument to transact endpoint to wait for transaction commit. When this flag is true, request will return with final transaction status (commit / reject).
- Add new transaction hash to all the responses from the transact endpoint.
- Alter tests to wait for status instead of waiting for X seconds to make them more reliable.
- Add new tests to check this PR features.

Relates to https://github.com/hyperledger/cactus/issues/2077

Depends on https://github.com/hyperledger/cactus/pull/2168

Signed-off-by: Michal Bajer <michal.bajer@fujitsu.com>

PR #2172, opened 2022-10-17
feat(connector-iroha2): sending queries signed on the client-side

outSH:
- Add new helper method for signing query payload on the client side.
- Change `query` and `generateTransaction` to support both query and transaction payload generation and
  sending.
- Refactor `CactusIrohaV2QueryClient` to return query context instead of single request method.
- Update tests to fit the new API.
- Add tests for signing queries on the client side.

Depends on https://github.com/hyperledger/cactus/pull/2171


petermetz (changes requested):
@outSH Please resolve the merge conflicts! (also squash+rebase)

PR #2173, opened 2022-10-17
chore(release): publish v1.1.0

### LINKED ISSUE #2069: chore(release): missing @hyperledger/cactus-cmd-socket-server npm package

petermetz:
Also manually changing the "publishConfig" property of the
packages/cactus-cmd-socketio-server/package.json
file because it was not set to public which excluded it from the
release automation altogether in previous releases, causing this bug
to be reported:
https://github.com/hyperledger/cactus/issues/2069

Fixes #2069


jagpreetsinghsasan (approved):

jagpreetsinghsasan:
@petermetz  should we also mention how we can skip the Indy SDK blocker (if someone have) by simply removing the `indy-sdk` dependency under `examples/cactus-example-discounted-asset-trade/package.json` (so that we can build the project in case of this blocker)?

petermetz:
> @petermetz should we also mention how we can skip the Indy SDK blocker (if someone have) by simply removing the `indy-sdk` dependency under `examples/cactus-example-discounted-asset-trade/package.json` (so that we can build the project in case of this blocker)?

@jagpreetsinghsasan Yes. But in my opinion, that should be explained in the relevant parts of the documentation in general (BUILD.md/CONTRIBUTING.md) not just specifically for this release. So I agree that we need to have that information available. It should also be clarified under what circumstances it does not work (I kept myself on Ubuntu 20.04 and still have not seen the issue at all so there's some nuance to it)

PR #2179, opened 2022-10-25
chore(typescript): bump tsc globally and switch to nodenext for iroha2

### LINKED ISSUE #2158: chore(typescript): upgrade and switch to nodenext
@petermetz
When implementing Iroha V2 support in cactus I encountered some problems with @iroha2/client package. Their team provided some solution, but they require the use of `moduleResolution` of `nodenext`. This also requires upgrade of TS to 4.7. We will opt for a package that supports our current `moduleResolution`/typescript but, just in case and for the future, what is your opinion regarding such change?

Here's info on nodenext modules: https://www.typescriptlang.org/docs/handbook/esm-node.html

aldousalvarez:
Here are all the changes implemented for
moduleResolution to nodeNext and Typescript upgrade
-------------------------------------------
1. Upgraded TypeScript to 4.7.4 globally and switched
moduleResolution nodeNext in cactus-plugin-ledger-connector-iroha2
2. Upgraded @typescript-eslint/eslint-plugin &
@typescript-eslint/eslint/parser to 5.27.0
so that it supports the newer version of Typescript.
3. Added useUnknownInCatchVariables: false in tsconfig
and tsconfig.base.json to fix the errors related to
'unknown' type in catch variables that occurred after
upgrading the typescript.
4. Updated rxjs to v7.8.1 globally in all package.json files
to match the version in IrohaV2. This is better than adding a resolution
in the root package.json because it makes it visible and explicit that
we've upgraded the dependency everywhere. The global resolutions
override is a last resort option because it is pretty confusing to
beginners when they specify a dependency version in their own package.json
but it doesn't ever get installed because of the override coming from the
top level package.json.
The reason the upgrade was needed to begin with: Typescript compiler
errors were being caused by the mismatching version of rxjs that the
iroha2 client packages depend on vs. the versions that our codebase
uses. Now with the upgrade they are in sync and the build passes.

---
Secondary change:
Peter: I'm sneaking in another hot-fix for the codegen task which makes
the .jar download (that's part of the warmup script) SEQUENTIAL instead
of having it PARALLEL which appeared to be working fine earlier but it
no longer seems to be the case, e.g., I've seen some race condition
looking errors pop up after having merged the PR with the earlier hotfix.
The root cause of the issue likely the openapi-generator-cli package
itself not being safe for concurrent use in general, but a full audit of
that code would have to be performed to validate this theory. For now
it is just easier to set it to SEQUENTIAL and be done with it (hopefully)

Fixes https://github.com/hyperledger/cacti/issues/2158


packages/cactus-api-client/src/main/typescript/socketio-api-client.ts
```diff
@@ -375,7 +375,7 @@ export class SocketIOApiClient implements ISocketApiClient<SocketLedgerEvent> {
 
       return this.monitorSubject.pipe(
         finalize(() => {
-          if (this.monitorSubject && !this.monitorSubject.observed) {
+          if (this.monitorSubject && !this.monitorSubject.observers) {
```
outSH, line 378: @aldousalvarez @petermetz `this.monitorSubject.observers` is always truthy (it's an array), `stopMonitor` is never emitted, causing #2191. What is the reason for this change?

aldousalvarez: @outSH  Apologies, while doing the fix on upgrade migration of typescript to nodeNext the path of the imports changed. Based on the suggested fix observers must be used thus the issue. Upon checking, the observed method returns the observers.length and I've already implemented it to <**packages/cactus-api-client/src/main/typescript/socketio-api-client.ts**> which is now passing see the latest [commit](https://github.com/hyperledger/cactus/pull/2179/commits/7f6493e6a866b98a51b72664b5658e7fa867a42d).

jagpreetsinghsasan (changes requested):
LGTM, just fix the PR commit message.
You can refer to this: [`57eb605` (#2207)](https://github.com/hyperledger/cacti/pull/2207/commits/57eb605abbf005db0714145ef94d568e96de222d)

petermetz (commented):
@aldousalvarez This is a great work, but I'd strongly prefer breaking it down to multiple smaller pull requests which are all focused on changing a single thing (for example one pull request that does nothing but upgrading to Typescript 4.7.x)

But, before we even get to that, here's another question for me to get more context: It looks to me like we could do the change to `moduleResolution` on a per-package basis (once the compiler is up on >=v4.7.x) therefore we could limit the scope of this PR to a much smaller scale that does nothing but exactly satisfying the requirement from @outSH which is that in a specific package he'd like to use a dependency that requires the new module resolution mechanism.

So, the question: Could we break this whole work down in this way:
1. The compiler upgrade to >v4.7
2. PR that switches only the specific iroha connector package to moduleResolution nodenext
3. N number of PRs later on that can change more packages to the new module resolution over time - thus breaking this giant diff down to smaller, digestible chunks?


Also note (to myself mostly) that we want the flag turned on in the future that forces developers to deal with the runtime type of the thrown exceptions (which is something that Typescript 4.4 introduced as a feature that we - me at least - very much want enabled in our codebase)

aldousalvarez:
Hello @petermetz here is the updated [commit ](https://github.com/hyperledger/cacti/pull/2179/commits/5f91646b7bacd4ca485e3d25d2c0e71bc06a6163)with moduleResolution to nodenext and typescript upgrade to 4.7.4 is only on iroha2 connector

petermetz:
> Hello @petermetz here is the updated [commit ](https://github.com/hyperledger/cacti/pull/2179/commits/5f91646b7bacd4ca485e3d25d2c0e71bc06a6163)with moduleResolution to nodenext and typescript upgrade to 4.7.4 is only on iroha2 connector

@aldousalvarez You upgrade the compiler to v4.7 project wide not just for the package but that's okay because this is the only way to do it anyway. I meant to send that in a separate PR, but the diff is not that large now that you cut down on the rest of the packages so I'm just gonna let it go because the CI is way too slow right now so I'm trying to be a little less nit-picky about mixing different commits into a single PR.

petermetz (changes requested):
@aldousalvarez Please explain in the commit message the reason behind each change.
Your point 4 explains why it was done but 1,2,3 just says "this is what I did" but not "why".

If someone comes back to this later because we broke something else that we didn't realize at the time, they have to be able to figure out why we made the changes we made so that they can make better/more educated decisions about what the fix might be (and ideally they need to be able to do that without having to have a discussion about it with you or me)

Other than that, the PR diff itself is LGTM, thank you!

aldousalvarez:
Hello @petermetz already updated the commit message and I have explained there why I did those changes. Thank you

petermetz (approved):
@aldousalvarez LGTM now, but I've made some changes: 
1. Instead of a global resolution I've upgrade `rxjs` explicitly via the dependency declarations. Same end result but less confusion. I further explained this in the commit message. The diff is much larger this way but much easier to understand/reason about it and so it's less technical debt (IMO).
2. I snuck in yet another hot-fix for the `codegen` script which is still suffering from concurrency issues.


@jagpreetsinghsasan @takeutak @izuru0 @VRamakrishna @sandeepnRES Please re-review with the above change in mind.

jagpreetsinghsasan (approved):

izuru0 (approved):
LGTM

PR #2204, opened 2022-11-14
docs(examples): add workshop 2022-11-14 examples

RafaelAPB:
Simple examples using Cacti tailored for the workshop.

Thanks to @AndreAugusto11  and Monica Gomez <monicasrgomez@gmail.com> @msrgomez for the amazing work.  

Works with version 1.0.0 of cactus-tooling.  

Depends on #2213 
Depends on #2223

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2481,6 +2481,14 @@
     crc-32 "^1.2.0"
     ethereumjs-util "^7.1.4"
 
+"@ethereumjs/common@^2.6.4":
+  version "2.6.5"
+  resolved "https://registry.yarnpkg.com/@ethereumjs/common/-/common-2.6.5.tgz#0a75a22a046272579d91919cb12d84f2756e8d30"
+  integrity sha512-lRyVQOeCDaIVtgfbowla32pzeDv2Obr8oR8Put5RdUBNRGr1VGPGQNGP6elWIpgK3YdpzqTOh4GyUGOureVeeA==
+  dependencies:
+    crc-32 "^1.2.0"
+    ethereumjs-util "^7.1.5"
+
 "@ethereumjs/ethash@^1.1.0":
   version "1.1.0"
   resolved "https://registry.yarnpkg.com/@ethereumjs/ethash/-/ethash-1.1.0.tgz#7c5918ffcaa9cb9c1dc7d12f77ef038c11fb83fb"
@@ -2507,6 +2515,14 @@
   dependencies:
     "@ethereumjs/common" "^2.6.1"
     ethereumjs-util "^7.1.4"
+
+"@ethereumjs/tx@^3.3.2":
+  version "3.5.2"
+  resolved "https://registry.yarnpkg.com/@ethereumjs/tx/-/tx-3.5.2.tgz#197b9b6299582ad84f9527ca961466fce2296c1c"
+  integrity sha512-gQDNJWKrSDGu2w7w0PzVXVBNMzb7wwdDOmOqczmhNjqFxFuIbhVJDwiGEnxFNC2/b8ifcZzY7MLcluizohRzNw==
+  dependencies:
+    "@ethereumjs/common" "^2.6.4"
+    ethereumjs-util "^7.1.5"
 
 "@ethereumjs/vm@^5.5.2":
   version "5.7.1"
@@ -5415,7 +5431,6 @@
 
 "@ubiquity/ubiquity-ts-client-modified@https://github.com/RafaelAPB/ubiquity-ts-client-mirror.git":
   version "1.0.0"
-  uid "1971019809f74889e150c3eaec061e0af841f953"
   resolved "https://github.com/RafaelAPB/ubiquity-ts-client-mirror.git#1971019809f74889e150c3eaec061e0af841f953"
   dependencies:
     axios "0.21.1"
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -2814,6 +2814,47 @@
     secp256k1 "4.0.2"
     sha3 "2.1.4"
 
+"@hyperledger/cactus-common@1.0.0":
```
petermetz, line 2817: @RafaelAPB Let's wait for the fix to the substrate test ledger (hopefully in v1.1.3) and that way we can avoid adding an old dependency to the lock file like the v1.0.0 test tooling package (could lead to interesting build failures and confusion).

Please:
1. Create a bug in the issue tracker for the substrate test ledger problem we discovered the morning of the workshop
2. Mark this PR dependent on that.
3. That way when that issue gets fixed we'll know that this one is ready for prime time (at which time I'll ask you to bump the test-tooling package dependency version to the one with the fix)

examples/cactus-workshop-examples/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@hyperledger/cactus-workshop-examples",
+  "version": "1.0.0",
```
petermetz, line 3: @RafaelAPB Please when the time comes bump this version to whatever the next release will be at the time.

petermetz (changes requested):

takeutak:
@RafaelAPB Thanks for your contribution and hard work to develop a workshop example!
As Peter suggested, the directory name (`cactus-workshop-example`) should be changed to some identified names such as `cacti-workshop-examples-2022-11-14` because there will be another workshop example in the future.

yarn.lock
RafaelAPB: Done. Could we instead accept the PR with test-tooling version 1.0.0 (one that passes), so that we can use the code on the PT workshop, next week? @petermetz

examples/cactus-workshop-examples/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "engines": {
+    "node": ">=10",
+    "npm": ">=6"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "browserMinified": "dist/cactus-workshop-examples.web.umd.min.js",
+  "mainMinified": "dist/cactus-workshop-examples.node.umd.min.js",
+  "watch": {}
```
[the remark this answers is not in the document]
RafaelAPB: Can you provide an example?

RafaelAPB:
@petermetz and @takeutak, the updates you recommended have been incorporated

examples/cactus-workshop-examples/package.json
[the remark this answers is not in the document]
petermetz: > Can you provide an example? The watch script is already included

@RafaelAPB Nah, my bad. I mixed it up with the open API watch. Sorry!

petermetz (changes requested):
> @RafaelAPB Both the description and the package name (and directory name) should be pinned to the actual workshop instance (2022-11-14) so that we can add more of these packages later

1. Please fix the above request from earlier and
2. Squash into a single commit
3. Update the tsconfig.json to match the conventions of the other packages if possible (if not possible for whatever reason, then let's discuss the details)
4. Mark the issue dependent with the correct syntax both in the PR description and the commit message.
5. Set the PR title to match the final remaining (one) commit's subject line.
6. Make sure that the commit subject line is also identifying the workshop instance.
7. Make sure that the README title is also identifying the workshop instance.

takeutak:
@RafaelAPB 
> @petermetz and @takeutak, the updates you recommended have been incorporated
Thanks.   It is no problem for me, but could you revise this pull-request according to Peter's request?  After that I will approve it.

RafaelAPB:
@

> > @RafaelAPB Both the description and the package name (and directory name) should be pinned to the actual workshop instance (2022-11-14) so that we can add more of these packages later
> 
> 1. Please fix the above request from earlier and
> 2. Squash into a single commit
> 3. Update the tsconfig.json to match the conventions of the other packages if possible (if not possible for whatever reason, then let's discuss the details)
> 4. Mark the issue dependent with the correct syntax both in the PR description and the commit message.
> 5. Set the PR title to match the final remaining (one) commit's subject line.
> 6. Make sure that the commit subject line is also identifying the workshop instance.
> 7. Make sure that the README title is also identifying the workshop instance.

@petermetz updated. Not sure about the changes required for steps 3 and 4.

petermetz (changes requested):
@RafaelAPB If you compare your tsconfig.json (packages/cactus-plugin-ledger-connector-besu/tsconfig.json) to for example the besu connector one, you'll catch a couple of differences:
1. The include paths are universal (more or less) whereas yours are package specific
2. The build cache you specified puts the file in a sub-folder while the others aren't (by convention to have a flat file structure for the build-cache files)

For point 4 see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

And also note that you can mark dependencies the same way but instead of saying "fixes #????" you can say "depends on #????" and then the robots will notice that and block the PR from getting merged accidentally before the parent PR (the one this depends on) has been merged.

The commit message is still not consistent with the PR description, it's fine to add *more* information to the commit message compared to the PR description but not the other way around.

RafaelAPB:
@petermetz updated!

jagpreetsinghsasan (approved):
LGTM, its nice to have such simple examples as well, for getting started.

takeutak (approved):

takeutak:
LGTM, but this pull-request cannot be merged without resolving this change-request from Peter due to GitHub system.

@petermetz If you have some improvement comments, please feel free to share them.

petermetz (changes requested):
LGTM but will mark it with a change request so that we know that it has to be updated after v1.1.3 comes out (which is dropping soon)

petermetz (changes requested):
@RafaelAPB Please update to v1.1.3 and then if you run `yarn` it should disappear the unnecessary diff from the lock file.

petermetz (changes requested):
@RafaelAPB You don't appear to have ran `yarn` to make the lock file changes necessary.

RafaelAPB:
> @RafaelAPB You don't appear to have ran `yarn` to make the lock file changes necessary.

should be good to go now! Thanks for the patience

petermetz (approved):
@RafaelAPB No worries and thanks for the updates, LGTM.

PR #2212, opened 2022-11-22
build: migrate to Yarn v3

### LINKED ISSUE #1142: build(yarn): migrate from v1.2.17 to v3.6.0
### Description
 
Depends on https://github.com/hyperledger/cactus/pull/1141

**This is an expert level good first issue, meaning that you don't need to know much about Cactus specifically, but you do need to be an expert in configuring build systems/package managers/etc.**

Yarn v3 has some performance improvements (among other things) compared to Yarn v1 so we should migrate over as soon as possible (when it hits GA).

### Acceptance Criteria
1. Developer flow has not changed, commands are the same (so that we don't have to re-write half the documentation)
2. If the dev flow must change then the PR must include the new documentation for this and also a migration guide for people who have the old project build working (it's okay to say that they need to delete everything and do a fresh clone - as long as that actually works)
3. No plug and play for now, let's take it one step at a time.
4. Existing documentation example commands work (such as the readme of the supply chain app which has a step by step guide)
5. Test case execution works as it is shown in the contributing.md and build.md files (`yarn jest some/path/to/test.case.test.ts` and `yarn tap --ts --timeout=9999 some/path/to/other/test-case.test.ts`)
6. `yarn configure` works as expected - does not crash due to missing or unlinked binaries from the `./node_modules/.bin/` folder

zondervancalvez:
Signed-off-by: zondervancalvez <zondervan.v.calvez@accenture.com>
Fixes #1142


packages/cactus-cmd-api-server/package.json
```diff
@@ -43,14 +43,14 @@
     "dist/*"
   ],
   "scripts": {
-    "codegen": "run-p 'codegen:*'",
+    "codegen": "yarn run codegen:openapi && yarn run codegen:proto",
```
petermetz, line 46: @zondervancalvez Could you please explain why was this specific change necessary?

extensions/cactus-plugin-htlc-coordinator-besu/package.json
```diff
@@ -43,9 +43,9 @@
     "dist/*"
   ],
   "scripts": {
-    "codegen": "run-p 'codegen:*'",
-    "codegen:openapi": "run-p generate-sdk",
-    "generate-sdk": "run-p generate-sdk:*",
+    "codegen": "yarn run codegen:openapi",
```
petermetz, line 46: @zondervancalvez Same question here as in my other comment. Why run only a specific script instead of the pattern?

tools/ci.sh
```diff
@@ -116,6 +116,7 @@ function mainTask()
   if [ "${DEV_BUILD_DISABLED:-false}" = "true" ]; then
     echo "$(date +%FT%T%z) [CI] Dev build disabled. Skipping..."
   else
+    yarn install
```
petermetz, line 119: @zondervancalvez The configure script already runs `yarn install`, must we run it twice, e.g. during and right before the configure script as well?

BUILD.md
```diff
@@ -135,6 +133,12 @@ cd cactus
 npm run enable-corepack
 ```
 
+* Run this command to migrate lockfile (Also, run this everytime you have changes before running `yarn run cofugure`)
```
petermetz, line 136: @zondervancalvez I'm guessing that's a typo in the script name, please fix!

petermetz, line 136: @zondervancalvez Please go into a little more detail about the lockfile migration: explain that it is needed when someone was using yarn v1 but once it was migrated then there is no need. Also explain in a few words how to recognize which version it is (the lockfile has a version declaration in it somewhere IIRC)

> Also, run this everytime you have changes before running `yarn run cofugure`

Is this true though? If I change some .ts code and then want to run the configure script, what difference does the extra `yarn install` will make in that case in general if I already migrated my lockfile earlier?

petermetz (changes requested):
@zondervancalvez Thank you!
I added a few comments/questions and change requests, please see above!

petermetz (changes requested):
@zondervancalvez A couple more asks:
1. Please write a legible commit message body explaining in detail the what/why/how/etc. of the changes.
2. Rebase onto upstream/main and fix the merge conflicts

packages/cactus-cmd-api-server/package.json
zondervancalvez: After few tests, we revert this to original and added minor changes.

extensions/cactus-plugin-htlc-coordinator-besu/package.json
zondervancalvez: After few tests, we revert this to original and added minor changes. We added single quote to `generate-sdk:*`

tools/ci.sh
zondervancalvez: We updated the `yarn run configure` script so that it will run first the `yarn install` command.

BUILD.md
zondervancalvez: After some changes and re-testing, we already removed this.

petermetz: @zondervancalvez It is still in the diff.

Issue #2236, opened 2022-12-19
ServerMonitorPlugin in connector-go-ethereum-socketio should have option to report empty blocks


rwat17:
startMonitor in ServerMonitorPlugin should have option to report empty blocks.

Now monitor only reports blocks with at least one transaction. It will be useful in the future that it could report empty blocks too. 
- startMonitor should take parameter for monitorOptions that will allow it to report also empty blocks 

cactus/packages/cactus-plugin-ledger-connector-go-ethereum-socketio/src/main/typescript/connector/ServerMonitorPlugin.ts

PR #2237, opened 2022-12-19
feat: connector-go-ethereum now can report empty blocks

rwat17:
- added parameter in monitorOptions what triggers monitor to report empty blocks

Closes: #2236

Signed-off-by: Tomasz Awramski <tomasz.awramski@fujitsu.com>


petermetz (approved):
@rwat17 LGTM as-is but I do have a couple of recommendations for the future:
1. For the plugin to be more useful/reusable you should export things other than a single function that starts a web server
2. Once you did that though, a change like this would be a breaking change (right now it's not because what you changed is not exported to begin with but for future reference)

izuru0 (approved):
LGTM

PR #2244, opened 2022-12-27
test(connector-go-ethereum-socketio): enable nullish coalescing checks

rwat17:
Test file was only testing monitor. Added nullish coalescing in monitor.

Signed-off-by: tomasz awramski <tomasz.awramski@fujitsu.com>


petermetz (changes requested):
@rwat17 Is this a bug-fix in the code itself that a user reading the release notes would care about? If not (e.g. it's just a fix in the test code itself and nothing has been fixed nor changed in the connector) then could you change the commit message (and PR title) to something like

`test(connector-go-ethereum-socketio): enable nullish coalescing checks`

That way it won't show up in the auto-generated release notes when we issue v1.1.4.

izuru0 (approved):
LGTM

rwat17:
@petermetz commit title and content changed

PR #2256, opened 2023-01-09
feat(connector-go-ethereum): add getBlock and getTransactionReceipt methods to connector

rwat17:
- getBlock and getTransactionReceipt added in go-ethereum-socketio-connector
- Added nullish coalescing in monitor options

Closes: #2255
Signed-off-by: tomasz awramski <tomasz.awramski@fujitsu.com>

PR #2254, opened 2023-01-09
feat(supabase-all-in-one): add docker image for test supabase instance

outSH:
- Add a new docker image `supabase-all-in-one` that will setup supabase instance for tests.
- Supabase is used as a backend for Cactus GUI.

Closes: #2253

Signed-off-by: Michal Bajer <michal.bajer@fujitsu.com>

PR #2259, opened 2023-01-10
feat(cactus-plugin-persistence-ethereum): add new persistence plugin

outSH:
- Add a new plugin for storing ledger data into a database (or any other storage in the future).
- Add functional tests for plugin and data access layer operations.
- Increase gas limit on openethereum contract to solve occasional issues in automatic tests.
- Tests assume any postgres database, but for final deployment supabase is assumed.
- Data fed by this plugin can later by visualized by a GUI application or analyzed directly.

Depends on https://github.com/hyperledger/cactus/pull/2254
Depends on https://github.com/hyperledger/cactus/pull/2256


packages/cactus-plugin-persistence-ethereum/src/main/typescript/plugin-persistence-ethereum.ts
```diff
@@ -0,0 +1,960 @@
...
+    await this.refreshMonitoredTokens();
+  }
+
+  /**
+   * Parse entire block data, detect possible token transfer operations and store the new block data to the database.
+   * Note: token balances are not updated.
+   *
+   * @param block `web3.js` block object.
+   */
+  public async parseBlockData(block: BlockTransactionObject): Promise<void> {
```
izuru0, line 866: this function not only parse a block data, but it also inserts the parsed data into a database.
can you rename this function name to e.g. parseAndStoreBlockData to reflect that?

jagpreetsinghsasan (approved):
LGTM

outSH:
Depends on: https://github.com/hyperledger/cacti/pull/2254

packages/cactus-plugin-persistence-ethereum/src/main/typescript/generated/openapi/typescript-axios/.openapi-generator-ignore
petermetz, line 1: @outSH Sorry for the slow review!
Please 
1. use the `.openapi-generator-ignore` file that all the other packages is are (hopefully) also using (it ignores some of the files that are currently in your diff like the .gitignore and the .npmignore if I remember correctly. 
2. Remove the files from the diff that are usually ignored by the other packages. To give an exmaple, the API server has this file which you could copy paste from: `packages/cactus-cmd-api-server/src/main/typescript/generated/openapi/typescript-axios/.openapi-generator-ignore`

As usual if there's some specific reason for not to do it the "usual" way just let me know and we can talk about it.

packages/cactus-plugin-persistence-ethereum/README.md
```diff
@@ -0,0 +1,237 @@
...
+const persistencePlugin = new PluginPersistenceEthereum({
+  instanceId: uuidv4(),
+  apiClient: new SocketIOApiClient(apiConfigOptions),
+  logLevel: "info",
+  connectionString:
+    "postgresql://postgres:your-super-secret-and-long-postgres-password@localhost:5432/postgres",
+});
+
+// Initialize the connection to the DB
+persistencePlugin.onPluginInit();
```
petermetz, line 52: @outSH `onPluginInit()` is an async call so I'd make sure to `await` on it in the examples otherwise people using this code as inspiration might get race conditions in their applications where they start initializing the plugin but then immediately try to do something else that depends on the plugin having finished initialization.

packages/cactus-plugin-persistence-ethereum/README.md
```diff
@@ -0,0 +1,237 @@
...
+const persistencePlugin = await factory.create({
+  instanceId: uuidv4(),
+  apiClient: new SocketIOApiClient(apiConfigOptions),
+  logLevel: "info",
+  connectionString:
+    "postgresql://postgres:your-super-secret-and-long-postgres-password@localhost:5432/postgres",
+});
+
+// Initialize the connection to the DB
+persistencePlugin.onPluginInit();
```
petermetz, line 75: @outSH `onPluginInit()` is an async call so I'd make sure to `await` on it in the examples otherwise people using this code as inspiration might get race conditions in their applications where they start initializing the plugin but then immediately try to do something else that depends on the plugin having finished initialization.

packages/cactus-plugin-persistence-ethereum/README.md
```diff
@@ -0,0 +1,237 @@
+# `@hyperledger/cactus-plugin-persistence-ethereum`
+
+This plugin allows `Cactus` to persist Ethereum data into some storage (currently to a `PostgreSQL` database, but this concept can be extended further).
+Data in the database can later be analyzed and viewed in a GUI tool.
```
petermetz, line 4: @outSH If someone comes in and submits a complicated pull request with a giant refactor that makes it so that this plugin supports SQLite as well, would you rather `a)` accept that with the additional complexity or `b)` prefer to ask that contributor to compartmentalize their change into a different plugin altogether?

The reason I ask is because if in that case `b)` is preferred then I'd name this plugin `cactus-plugin-persistence-ethereum-pg` so that in the imaginary scenario the other person can call their implementation `cactus-plugin-persistence-ethereum-sqlite` without any sort of confusion about which plugin does what.

This is just a question, not a change request (unless you'd prefer `b)` in which case then I'll ask to rename the plugin to `cactus-plugin-persistence-ethereum-pg`.

petermetz (changes requested):

packages/cactus-plugin-persistence-ethereum/README.md
outSH: @petermetz I want to support / be able to support other data storages in the future, and for this reason data access layer is separated from the monitoring functions of the main plugin, so I'd say option `a)`. Ideally such PR would introduce another db-client and some sort of dynamic selector that would choose data storage based on input args to the plugin.

packages/cactus-plugin-persistence-ethereum/README.md
outSH: @petermetz Fixed.

packages/cactus-plugin-persistence-ethereum/README.md
outSH: @petermetz Fixed.

packages/cactus-plugin-persistence-ethereum/src/main/typescript/generated/openapi/typescript-axios/.openapi-generator-ignore
outSH: @petermetz Sure, did not know that - fixed :)

packages/cactus-plugin-persistence-ethereum/src/main/typescript/plugin-persistence-ethereum.ts
outSH: @izuru0 Fixed

izuru0 (approved):
LGTM

packages/cactus-plugin-persistence-ethereum/README.md
petermetz: @outSH Kk!

packages/cactus-plugin-persistence-ethereum/src/main/typescript/generated/openapi/typescript-axios/.openapi-generator-ignore
petermetz: @outSH Thank you very much!

packages/cactus-plugin-persistence-ethereum/package.json
```diff
@@ -0,0 +1,91 @@
+{
+  "name": "@hyperledger/cactus-plugin-persistence-ethereum",
+  "version": "1.1.3",
```
petermetz, line 3: @outSH `1.2.0` went out yesterday, please update this to match that version as well.

packages/cactus-plugin-persistence-ethereum/package.json
```diff
@@ -0,0 +1,91 @@
...
+    "fast-safe-stringify": "2.1.1",
+    "sanitize-html": "2.7.0",
+    "web3-utils": "1.5.2",
+    "async-mutex": "0.4.0",
+    "uuid": "8.3.2"
+  },
+  "devDependencies": {
+    "@hyperledger/cactus-core-api": "1.1.3",
+    "@hyperledger/cactus-api-client": "1.1.3",
+    "@hyperledger/cactus-test-tooling": "1.1.3",
```
petermetz, line 74: @outSH `1.2.0` went out yesterday, please update these as well to match that version also.

yarn.lock
```diff
@@ -3022,6 +3022,17 @@
   resolved "https://registry.yarnpkg.com/@hutson/parse-repository-url/-/parse-repository-url-3.0.2.tgz#98c23c950a3d9b6c8f0daed06da6c3af06981340"
   integrity sha512-H9XAx3hc0BQHY6l+IFSWHDySypcXsvsuLhgYLUGywmJ5pswRVQJUHpOsobnLYp2ZUaUlKiKDrgWWhosOwAEM8Q==
 
+"@hyperledger/cactus-api-client@1.1.2":
```
petermetz, line 3025: @outSH This looks like it might be a leftover change, but I'm not sure. Please double check if setting up all the cross-package dependencies (one of our package depending on another one of our packages) to use `1.2.0` and then running `yarn install` makes this go away from the diff (or at least bumps up up to `1.2.0`

petermetz (changes requested):
@outSH Thanks for the changes! I added a few more asks because of the release that went out just yesterday made the dependency versions outdated.

Last thing: There are 2 commits now, one of them from the parent PR. I'm guessing if you do a rebase + conflict resolution it will go back down to 1 commit with that being just the one we want but in case it doesn't, please look into that as well.

yarn.lock
outSH: @petermetz This entry was removed after `yarn install`

petermetz: > @petermetz This entry was removed after `yarn install`

@outSH Yaay, that's what I was hoping it would do in a best case scenario, thank you for the updates!

packages/cactus-plugin-persistence-ethereum/package.json
petermetz: Cheers!

packages/cactus-plugin-persistence-ethereum/package.json
petermetz: @outSH Cheers!

petermetz (approved):
@outSH Thank you for the updates, LGTM!

PR #2293, opened 2023-02-16
feat(quorum): private transaction support

### LINKED ISSUE #951: feat(quorum): Add private transaction support

Feature request: I need to be able to issue private transactions to a Quorum ledger via the Cactus API client.

1. Either a separate endpoint next to the current POST /run-transaction one or as an extension of it. Not yet sure which one is more feasible/makes more sense.
2. Test coverage for the new feature
3. Private transactions must not be logged by Cactus for obvious reasons.
4. Documentation to explain specifically that this is not a privacy preserving feature, it only allows access to the pre-existing Quorum APIs as-is

Nothing. There is no doubt in my mind that we need this feature.

Additional context

Here's the Quorum documentation regarding private transactions:
https://docs.goquorum.consensys.net/en/stable/Concepts/Privacy/PrivateAndPublic/

cc: @takeutak @sfuji822 @jonathan-m-hamilton @petermetz @kikoncuo @travis-payne

aldousalvarez:
Fixes #951

Co-authored-by: Travis Payne <travis.payne@accenture.com>
Co-authored-by: johnhomantaring <john.h.o.mantaring@accenture.com>
Co-authored-by: jagpreetsinghsasan <jagpreet.singh.sasan@accenture.com>
Co-authored-by: aldousalvarez aldousss.alvarez@gmail.com

### DIFF
--- a/packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json-private.test.ts
+++ b/packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json-private.test.ts
@@ -1,0 +1,325 @@
+import "jest-extended";
+import Web3 from "web3";
+import { v4 as uuidV4 } from "uuid";
+
+import { AbiItem } from "web3-utils";
+
+import { LogLevelDesc } from "@hyperledger/cactus-common";
+
+import HelloWorldContractJson from "../../../../solidity/hello-world-contract/HelloWorld.json";
+
+import Web3JsQuorum, { IWeb3Quorum } from "web3js-quorum";
+
+const keyStatic = {
+  tessera: {
+    member1: {
+      publicKey: "BULeR8JyUWhiuuCMU/HLA0Q5pzkYT+cHII3ZKBey3Bo=",
+    },
+    member2: {
+      publicKey: "QfeDAys9MPDs2XHExtc84jKGHxZg/aj52DTh0vtA3Xc=",
+    },
+    member3: {
+      publicKey: "1iTZde/ndBHvzhcl7V68x44Vx7pl8nwx9LqnM/AfJUg=",
+    },
+  },
+  quorum: {
+    member1: {
+      name: "member1",
+      url: "http://127.0.0.1:20000",
+      wsUrl: "ws://127.0.0.1:20001",
+      privateUrl: "http://127.0.0.1:9081",
+      privateKey:
+        "b9a4bd1539c15bcc83fa9078fe89200b6e9e802ae992f13cd83c853f16e8bed4",
+      accountAddress: "f0e2db6c8dc6c681bb5d6ad121a107f300e9b2b5",
+    },
+    member2: {
+      name: "member2",
+      url: `http://127.0.0.1:20002`,
+      wsUrl: `http://127.0.0.1:20003`,
+      privateUrl: "http://127.0.0.1:9082",
+      privateKey:
+        "f18166704e19b895c1e2698ebc82b4e007e6d2933f4b31be23662dd0ec602570",
+      accountAddress: "ca843569e3427144cead5e4d5999a3d0ccf92b8e",
+    },
+    member3: {
+      name: "member3",
+      url: `http://127.0.0.1:20004`,
+      wsUrl: `http://127.0.0.1:20005`,
+      privateUrl: "http://127.0.0.1:9083",
+      privateKey:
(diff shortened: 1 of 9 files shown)


takeutak (approved):

jagpreetsinghsasan (changes requested):
LGTM, just write a detailed commit message for this and create follow-up issues to address the issue of `privateUrl` being truely optional. Right now we are passing privateUrl in all the instances of QuorumConnector even though we don't need to do private transactions everywhere.
The follow up issue can also include 2 test cases, one showcasing private tx (the newly created file in this PR) and a seperate test file showing public tx (a similar file as `packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json-private.test.ts`, but just doing the public tx). Also that issue can include renaming of this privateUrl to something more specific, maybe a `rpcApiHttpHost` and another field being `privacyEnabled` set as true/false.

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json-private.test.ts
```diff
@@ -0,0 +1,242 @@
...
+    contractAddress,
+  );
+  const mem2Response = await member2Contract.methods.getName().call();
+  t.strictEqual(mem2Response, "CaptainCactus");
+
+  const member3Contract = new web3QuorumMember3.eth.Contract(
+    HelloWorldContractJson.abi as AbiItem[],
+    contractAddress,
+  );
+  await t.doesNotReject(member3Contract.methods.getName().call);
```
petermetz, line 241: @aldousalvarez @jagpreetsinghsasan @johnhomantaring This test looks like it just deploys a public contract and then does a public `CALL` on it. No actual transaction is executed that would modify private state (`SEND` op).

Could you please make this test case a little more thorough so that:
1. It deploys a private contract that only 2 members out of 3 can access
2. It verifies that when the third member is trying to access the contract from point 1): it fails (assert the fact of the rejection of the promise)
3. It verifies that member 1 and 2 can indeed run `SEND` transactions privately to the contract. (You can update the name in the contract instance variable to some UUID and then verify that member 1 and 2 can read that and member 3 cannot (either with asserting complete rejection or by observing that the name variable for member 3 is still the old/original/default name not the UUID that was set by the other members)
4. Make the tests cases use the connector itself not the `web3js-quorum` library because otherwise we are just testing that the `web3js-quorum` library itself works correctly not that the connector plugin that uses it under its own hood.

petermetz (changes requested):

petermetz:
> @petermetz are the expectations from the follow-up issues fine? (Or anything more required, given this PR being the current state of enabling private tx)

@jagpreetsinghsasan Sorry for the slow response! We do need a good test case verifying that the private transactions that the connector can now allegedly run actually are private (otherwise the implementation in the connector is broken).
Please see my comment for details: https://github.com/hyperledger/cacti/pull/2293#discussion_r1150972067

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json-private.test.ts
jagpreetsinghsasan: Yes, the test case should be changed to incorporate the testing of the connector class itself via connector class methods related to private transactions.

jagpreetsinghsasan:
@aldousalvarez I see some commits being pushed into the PR after the last review comment, if you think its ready, please `re-request review`

aldousalvarez:
@jagpreetsinghsasan I tried to update my branch with the latest commit at that time. my apologies I am still working on it as of now. Thank you

packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v2.3.0-deploy-contract-from-json-private.test.ts
petermetz: The new test case looks good, thank you all!

petermetz (approved):
@aldousalvarez @jagpreetsinghsasan I pushed some minor changes to greatly reduce the diff and also did a rebase onto upstream/main. LGTM as-is now. Thank you all!

jagpreetsinghsasan (approved):
LGTM

PR #2303, opened 2023-02-27
fix(git): broken line endings corrupt git index after weaver merge

### LINKED ISSUE #2302: fix(git): broken line endings corrupt git index after weaver merge
**Describe the bug**

There 34 files in the git index (with no actual changes) after a rebase onto upstream/main.
This only happens on Linux based operating systems (Ubuntu in this case).
The issue is invisible on the Windows operating system.

**To Reproduce**

Clone the project on Ubuntu 20.04
Observe that your git index is now corrupted as shown on the screenshot below.

**Expected behavior**

Cloning the project leaves you with an empty index instead of having it corrupted.

**Logs/Stack traces**

https://stackoverflow.com/a/4683783

**Screenshots**

Screenshot from 2023-02-27 11-37-49

Dev machine

Ubuntu 20.04

`main`

 N/A

**Additional context**

More information on the issue and the proposal to fix it: https://stackoverflow.com/a/15646791

petermetz:
Not yet sure if this will fix the issue, but based on my research this
is the easiest and most probable solution.

Leaving a "paper trail" of the magical git commands that were executed
in order to have an idea in the future on how to fix potential negative
side-effects that might come out of this later (not likely but if it
does happen this should be immensely helpful information)

Without further ado, the list of commands executed to achieve this diff:

git rm --cached -r .

git config core.autocrlf input

git diff --cached --name-only -z | xargs -0 git add

Fixes #2302


petermetz:
Source and credits for the git commands go to: https://stackoverflow.com/a/1511273

takeutak (approved):

jagpreetsinghsasan (approved):
LGTM

Issue #2309, opened 2023-03-03
feat(quorum-connector): add script for checking connection status


outSH:
We need a simple script for checking if quorum-connector is connected to the ledger. The script will try to read latest block for now, but in the future a separate healthcheck endpoint can be implemented and queried by this script

PR #2310, opened 2023-03-03
feat(quorum-connector): add script for checking connection status

outSH:
- Simple script will check ledger connection by getting block through the connector.

Closes: #2309

### DIFF
--- a/packages/cactus-plugin-ledger-connector-quorum/package.json
+++ b/packages/cactus-plugin-ledger-connector-quorum/package.json
@@ -39,6 +39,9 @@
   "module": "dist/lib/main/typescript/index.js",
   "browser": "dist/cactus-plugin-ledger-connector-quorum.web.umd.js",
   "types": "dist/lib/main/typescript/index.d.ts",
+  "bin": {
+    "cacti-quorum-connector-status": "dist/lib/scripts/get-quorum-connector-status.js"
+  },
   "files": [
     "dist/*"
   ],
@@ -59,6 +62,7 @@
     "@hyperledger/cactus-core-api": "2.0.0-alpha.1",
     "axios": "0.21.4",
     "express": "4.17.3",
+    "minimist": "1.2.8",
     "prom-client": "13.2.0",
     "run-time-error": "1.4.0",
     "rxjs": "7.3.0",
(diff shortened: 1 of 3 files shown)


izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):
LGTM

packages/cactus-plugin-ledger-connector-quorum/package.json
```diff
@@ -64,14 +67,17 @@
...
   },
   "devDependencies": {
     "@hyperledger/cactus-plugin-keychain-memory": "1.1.3",
     "@hyperledger/cactus-test-tooling": "1.1.3",
     "@types/express": "4.17.13",
     "@types/sanitize-html": "2.6.2",
-    "web3-eth": "1.5.2"
+    "@types/minimist": "1.2.2",
+    "web3-eth": "1.5.2",
+    "chalk": "4.1.2"
```
petermetz, line 80: @outSH Please make sure that the dependencies are ordered alphabetically! (silly nit-pick I know, apologies)

petermetz (changes requested):
@outSH LGTM just please 
1. order the dependencies in the package.json alphabetically and 
2. resolve the merge conflict

packages/cactus-plugin-ledger-connector-quorum/package.json
outSH: @petermetz Sure thing :D I've used https://www.npmjs.com/package/sort-package-json on entire project and there are more files out of order, I can create an issue to fix them all at once and maybe add this check to CI && run it on configure

petermetz: @outSH Dejavu haha. There's this blast from the past ( https://github.com/hyperledger/cacti/issues/1341 ) that we did to do exactly what you just described but I see now that it's been broken this whole time because it does the package.json sorting but then it doesn't actually report an error if the package.json was unsorted, it just sorts it but that means that when it runs on the CI the changes it made to sort things are never actually committed.

We could make it part of the configure script, but with all the checks that it runs it would add an extra 20 seconds to the (already too slow) build so I'll suggest that we just fix it so that the CI fails if there are unsorted package.json files AND advise people to run the check manually before they open their PRs (so that they don't wait hours for the CI to pass but then it fails because of the sorting check).

See `tools/custom-checks/check-package-json-sort.ts` for the implementation (that is currently broken because it's supposed to error out if there were unsorted package.json files)

petermetz: > Done btw.

Cheers!

petermetz (approved):
@outSH Thank you, LGTM!

petermetz (changes requested):
@outSH Please fix the merge conflicts and then we are good to go!

petermetz (approved):
@outSH LGTM

FYI: I resolved the merge conflicts and rebased onto upstream/main so this is ready to merge now.

petermetz (commented):
@outSH FYI: I resolved the merge conflicts again (another PR introduced new ones)

PR #2318, opened 2023-03-14
build(yarn): migrate to CorePack

### LINKED ISSUE #2238: build(yarn): migrate to CorePack
### Description
As a contributor I want to have the easiest build tooling available so that I'm less bogged down with dev tooling and more focused on valuable contributions of mine.

https://yarnpkg.com/getting-started/install

Install Corepack

The preferred way to manage Yarn is through [Corepack](https://nodejs.org/dist/latest/docs/api/corepack.html), a new binary shipped with all Node.js releases starting from 16.10. It acts as an intermediary between you and Yarn, and lets you use different package manager versions across multiple projects without having to check-in the Yarn binary anymore.
Node.js >=16.10

Corepack is included by default with all Node.js installs, but is currently opt-in. To enable it, run the following command:

```
$ corepack enable
```

### Acceptance Criteria
1. Our documentation and scripts are updated so that CorePack is used

charellesandig:
Fixes #2238

Signed-off-by: charellesandig [charelle.wrk@gmail.com](mailto:charelle.wrk@gmail.com)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -129,6 +129,12 @@ git config --system core.longpaths true
 cd cactus
 ```
 
+* Run this command to enable corepack (Corepack is included by default with all Node.js installs, but is currently opt-in.)
+
+```sh
+npm run enable-corepack
+```
+
 * Run the initial configuration script (can take a long time, 10+ minutes on a low-spec laptop)
 
 ```sh
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "reset": "run-s reset:git reset:node-modules reset:yarn-lock configure",
     "configure": "yarn run init-registries && yarn install --frozen-lockfile --non-interactive && yarn run build:dev:backend",
     "install-yarn": "npm install --global yarn@1.22.17",
+    "enable-corepack": "npm i -g corepack && corepack enable && corepack prepare yarn@1.22.17 --activate",
     "custom-checks": "TS_NODE_PROJECT=./tools/tsconfig.json node --trace-deprecation --experimental-modules --abort-on-uncaught-exception --loader ts-node/esm --experimental-specifier-resolution=node ./tools/custom-checks/run-custom-checks.ts",
     "tools:validate-bundle-names": "TS_NODE_PROJECT=./tools/tsconfig.json node --trace-deprecation --experimental-modules --abort-on-uncaught-exception --loader ts-node/esm --experimental-specifier-resolution=node ./tools/validate-bundle-names.js",
     "generate-api-server-config": "node ./tools/generate-api-server-config.js",


takeutak (approved):

package.json
```diff
@@ -21,6 +21,7 @@
     "reset": "run-s reset:git reset:node-modules reset:yarn-lock configure",
     "configure": "yarn run init-registries && yarn install --frozen-lockfile --non-interactive && yarn run build:dev:backend",
     "install-yarn": "npm install --global yarn@1.22.17",
+    "corepack": "npm i -g corepack && corepack enable && corepack prepare yarn@stable --activate",
```
jagpreetsinghsasan, line 24: I have 2 questions here.
1. Should we not install a specific version of yarn (currently: 1.22.17)?
2. If we make yarn available via corepack, do we still need `install-yarn` script? (@petermetz what's your view on this?)

jagpreetsinghsasan (changes requested):

package.json
petermetz, line 24: @charellesandig Please call the npm script `enable-corepack` instead of just `corepack` so that the name hints at what it's doing. Having a noun for the name for something that modifies operating system level dependencies on the machine where the script runs is a bit of a misnomer.

petermetz: @jagpreetsinghsasan 

> Should we not install a specific version of yarn (currently: 1.22.17)?

100% agreed. @charellesandig Please fix this as well, gotta pin it to an exact version like we do with everything else in general.

> If we make yarn available via corepack, do we still need install-yarn script? (@petermetz what's your view on this?)

I would still leave it in because the enablement of corepack is optional at the end of the day. If someone prefers to just manually installing yarn and doing it the old way, they should be able to.

With that said, once all the LTS NodeJS versions that don't have yarn enabled by default have been end-of-life'd, THEN we could remove the install-yarn script because no one will ever need it again (unless they are purposefully running on some extremely outdated version of NodeJS but I don't want to support that in any way TBH). This will be years from now though because the CorePack thing is relatively new and LTS goes 5 or 6 years I think.

cc: @charellesandig

petermetz (changes requested):
@charellesandig Please see my change requests in the comments.

BUILD.md
```diff
@@ -135,6 +135,12 @@ cd cactus
 yarn run configure
 ```
 
+* Run to enable corepack which helps with managing versions of your package managers.
+
+```sh
+npm run corepack
+```
+
```
petermetz, line 143: @charellesandig I would move this up to be just above the `configure` script so that the corepack enablement happens first, which pins the yarn version to the preferred one and THEN the configure script is executed with that desired yarn version as defined by the corepack enablement script.

petermetz (changes requested):

petermetz (approved):
@charellesandig LGTM, thank you for the updates!

ryjones:
@jagpreetsinghsasan ?

jagpreetsinghsasan (approved):
LGTM, thankyou for the PR

PR #2340, opened 2023-03-24
chore(tools): script to bump openapi spec dependency versions

### LINKED ISSUE #2206: chore(tools): script to bump openapi spec dependency versions
### Description
As a release manager I want to have a simple and automated way of upgrading open API spec dependencies when I push a release so that everything is kept more up to date as we go along which leads to issues being detected faster/earlier.

Currently we have URL references to tagged versions of the spec files such as this one:
https://raw.githubusercontent.com/hyperledger/cactus/v0.8.0/packages/cactus-core-api/src/main/json/openapi.json#/components/schemas/JWSGeneral

The issue is that we are already on v1.1.2 at the time of this writing but there are still references to relatively old (especially pre-v1.0.0) spec files which runs the risk of us having bugs or breaking API changes added later by accident.

Manually looking and updating these references are tedious but since the spec files are valid JSON documents, we can certainly have a script that goes looking for `**/openapi.json` and replaces all the references with the latest available git tag.

Note: Tagging a release on the upstream is the very last step of issuing a release so we cannot change these references atomically as part of the release process (yet) but having the script doing the upgrade (even if that has to happen post-release in a separate PR) is a step in the right direction!

### Acceptance Criteria
1. Create a script in the tools directory written in Typescript that when executed: 1) determines the latest `v*` git tag on `main/upstream` and then finds and lists all the open API spec references. Finally it updates them.
2. It is important to print to stdout a complete and easily readable list of all the files that are being found and all the references for each file that as going to be updated.
3. Also make sure to include in the final log output the total number of open API spec files found AND also separately count the total count of references (since each file can have any number of references in it)

aldousalvarez:
1. Adds a new script that can be executed in order to bump all release
versions of the open API specs. These are one of two main categories:
1.1. The top level .info.version property of the OpenAPI spec JSON object
1.2. Any schema component references embedded anywhere deep inside the
JSON object that we define our specs as, for example when a connector
plugin's endpoint returns a response object whose schema is defined in
another package's OpenAPI spec file such as the core-api package's.
2. Fixed a bug in the "prettier" npm script which was pointing to a non-
existent configuration file. This was necessary to be fixed in this commit
because tests had to be carried out whether the auto-formatting applied
by the version bumping script is consistent with the auto-formatting that
is applied when we run prettier via the CLI (there was some struggle here
to figure out that we have to pre-format the JSON string held in-memory
via JSON.stringify() first otherwise the formatted version would diverge
from what the CLI does to the .json files on-disk)
3. We had to exclude the tag called v2.0.0-alpha-prerelease because it
was incorrectly named unfortunately: it is considered newer than
v2.0.0-alpha.1 by the semver-parser package but it is actually older,
e.g. we've issued v2.0.0-alpha-prerelease before we issued v2.0.0-alpha.1
(which is our mistake and this is the workaround to rectify that after
the fact without having to force delete the tag that we've already
pushed to the upstream main branch).
This exclusion is applied by default which might turn out to be a mistake
later but right now it does not look like there would be an edge case
where we need to handle this in any different way.
4. The script is re-entrant and will save a thorough report of what it did
to temporary .json files whose names are date and timestamped within the
project root's ./build/ subfolder.
5. The target version to bump to can be specified explicitly via the
--target-version CLI parameter but if it not provided then will default
to whatever is the latest sem-ver compatibly named tag in the git log.

Example to specify target version explicitly:
`yarn tools:bump-openapi-spec-dep-versions --target-version=v3.2.1`

Example to let the script figure out what the target version should be:
`yarn tools:bump-openapi-spec-dep-versions`

Fixes https://github.com/hyperledger/cacti/issues/2206


takeutak (approved):

tools/custom-checks/bump-openapi-spec-dep-version.ts
```diff
@@ -0,0 +1,161 @@
+import fs from "fs-extra";
+import path from "path";
+import { fileURLToPath } from "url";
+import { globby, Options as GlobbyOptions } from "globby";
+import { RuntimeError } from "run-time-error";
+import { isStdLibRecord } from "./is-std-lib-record";
+import { exec } from "child_process";
+import util from "util";
+
+interface paths {
```
petermetz, line 10: @aldousalvarez Please follow naming conventions and have the interface start with `I` and also be CapitalCased in general.

tools/custom-checks/bump-openapi-spec-dep-version.ts
```diff
@@ -0,0 +1,161 @@
...
+  post: {
+    requestBody: unknown;
+  };
+}
+interface packageJson {
```
petermetz, line 15: @aldousalvarez Please follow naming conventions and have the interface start with I and also be CapitalCased in general.

tools/custom-checks/bump-openapi-spec-dep-version.ts
```diff
@@ -0,0 +1,161 @@
...
+  const DEFAULT_GLOB = "**/src/main/json/openapi.json";
+  /** cannot be empty */
+  const pkgJsonPaths = await globby(DEFAULT_GLOB, globbyOpts);
+
+  const execPromise = util.promisify(exec);
+  const repoUrl = "https://github.com/hyperledger/cacti";
+  const branchName = "main";
+  const latestVersion = execPromise(
+    `git describe --tags --abbrev=0 ${branchName}`,
+  ).then(({ stdout }) => {
```
petermetz, line 73: @aldousalvarez Please use `await` unless there's a specific reason not to. It makes the code slighly more readable and avoids the first steps down onto callback hell.

tools/custom-checks/bump-openapi-spec-dep-version.ts
```diff
@@ -0,0 +1,161 @@
...
+    const latestTag = stdout.trim().substring(1);
+    console.log(
+      `${TAG} Hyperledger Cacti current version of ${branchName} branch  ${repoUrl} is version ${latestTag}`,
+    );
+    return latestTag;
+  });
+
+  const currentLatestVersion = await latestVersion;
```
petermetz, line 81: @aldousalvarez It's probably the best to extract the entire logic of obtaining the latest git release tag into a function of it's own so that it could potentially be re-used elsewhere in the tooling code as well.

tools/custom-checks/bump-openapi-spec-dep-version.ts
```diff
@@ -0,0 +1,161 @@
...
+        `${TAG} VERSION CHECK FAILED --------: ${pathRel} the package itself has no "version" field. Expected version: ${currentLatestVersion} \n`,
+      );
+    }
+    console.log("");
+    const spec = await fs.readJSON(filePathAbs);
+    console.log("");
+    console.log(
+      `${TAG} = = = = CHECKING TOTAL NUMBER OF REFERENCES = = = = \n`,
+    );
+    function countReferences(obj: any): number {
```
petermetz, line 138: @aldousalvarez Pleae don't define nested functions like this and instead aim to have a flat structure for them. This make make them have to have more parameters but this is a good thing in the sense that it will make the functions be closer to pure functions.

tools/custom-checks/bump-openapi-spec-dep-version.ts
```diff
@@ -0,0 +1,161 @@
...
+        `${TAG} VERSION CHECK FAILED --------: ${pathRel} the package itself has no "version" field. \n`,
+      );
+    } else if (packageVersion == currentLatestVersion) {
+      console.log(`${TAG} ---- CHECKING OPENAPI.JSON VERSION ---- \n`);
+      console.log(
+        `${TAG} VERSION CHECK PASSED --------: ${pathRel} the package itself correct has version { ${packageVersion} }. \n`,
+      );
+    } else if (packageVersion != currentLatestVersion) {
+      pkgJson.info.version = currentLatestVersion;
+      await fs.writeFile(filePathAbs, JSON.stringify(pkgJson, null, 2));
```
petermetz, line 118: @aldousalvarez When I ran the script, it produced a diff that changed the entire file due to whitespace changes. Let's figure out a way to have all the openapi.json's auto-formatted consistently with the same tool (prettier is my recommendation) and then you can apply that formatting after you've saved the JSON files and so the diff will only contain the lines that you actually changed, e.g. the vX.Y.Z declarations.

tools/custom-checks/bump-openapi-spec-dep-version.ts
```diff
@@ -0,0 +1,161 @@
...
+      console.log(`${TAG} ---- CHECKING OPENAPI.JSON VERSION ---- \n`);
+      console.log(
+        `${TAG} VERSION CHECK FAILED --------: ${pathRel} the package itself has no "version" field. Expected version: ${currentLatestVersion} \n`,
+      );
+    }
+    console.log("");
+    const spec = await fs.readJSON(filePathAbs);
+    console.log("");
+    console.log(
+      `${TAG} = = = = CHECKING TOTAL NUMBER OF REFERENCES = = = = \n`,
```
petermetz, line 136: @aldousalvarez Please clean up the log statements to just be regular sentences announcing either a) what's about to happen or b) what just happened. I get that making the log messages appear as headings of a document is helpful when you are debugging the code yourself during development, but then they lose their meaning when blended with all the other log messages from all the other components.

petermetz (changes requested):
@aldousalvarez Mostly LGTM but I have a few nits as usual. Please see the comments above for the details!

petermetz (changes requested):
@aldousalvarez FYI: I've pushed some changes that were easier to make then explain. Please make the following changes on top of what I had just pushed:
1. There are still nested functions, that point was never addressed.
2. Don't use regular expression search & replace to perform the version bumps. The files need to  be parsed and then the versions need to be updated based on the parsed data not via regex so that the replacements are deterministic and immune to bugs that will later stem from the content of the files changing in a way that will break the regex.

aldousalvarez:
@petermetz Thank you for the update, I have now addressed the changes that you have requested and I already fixed the nested functions and also I have implemented prettier as the one that will format the JSONs.

petermetz (approved):
LGTM

PR #2346, opened 2023-03-25
fix(security): upgrade express-jwt to v8.4.1

### LINKED ISSUE #2231: fix(security): upgrade express-jwt to v8.4.1
### Description

Depends on https://github.com/auth0/express-jwt/issues/308

The upgrade is needed due to cascading dependencies of which the root is vulnerabilities in other dependencies of other packages.
More info: https://github.com/hyperledger/cactus/issues/2228

petermetz:
Fixes/changes that needed to be done in order to make the upgrade work:
1. The HTTP verbs for exempted endpoints are now specified both as lowercase and uppercase meaning that if a specific endpoint is configured to be exempt from JWT authorization then it's method will be specified twice, once as 'POST' and once as 'post' because the underlying library (which is called express-unless) does not have the ability to handle verbs in a case insensitive way.

2. In the registerWebServiceEndpoint function, the configuration of the express-jwt-authz library had to be changed because the scope enforcement was broken due to express-jwt changing the default request property where it places the decoded JWT payload from `"user"` to `"auth"` and this made it incompatible by default with the behavior of express-jwt-authz Luckily there is a parameter to set the request property name and that is now being specified explicitly as `"auth"` so that they are playing nice with each other once again and the authorization's scope based access control works just fine.

Fixes #2231


takeutak (approved):

jagpreetsinghsasan (approved):
LGTM

Issue #2373, opened 2023-04-06
test(connector-iroha2): fix flaky iroha2-setup-and-basic-operations


petermetz:
**Describe the bug**

The flaky test case file in question:
`packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha2-setup-and-basic-operations.test.ts`

**To Reproduce**

Run  the CI enough times that it fails again. Unfortunately I could not reproduce it locally at all.

**Expected behavior**

Test succeeds consistently across different environments.

**Logs/Stack traces**

[2023-04-06T01-10-50-connector-iroha2-ci-test-crash.log](https://github.com/hyperledger/cacti/files/11165829/2023-04-06T01-10-50-connector-iroha2-ci-test-crash.log)

**Cloud provider or hardware configuration:**

GitHub/BuildJet runners

**Operating system name, version, build:**

Ubuntu-20.04

**Hyperledger Cactus release version or commit (git rev-parse --short HEAD):**

`main`

**Hyperledger Cactus Plugins/Connectors Used**

`iroha2`

cc: @outSH

Issue #2370, opened 2023-04-04
test(connector-iroha2): fix flaky generate-and-send-signed-transaction


petermetz:
**Describe the bug**

`yarn jest packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha2-generate-and-send-signed-transaction.test.ts`
succeeds on my local host, but occasionally fails on the CI.

**To Reproduce**

Run it enough times that the flake magically reappears.

**Expected behavior**

Test case should be  consistently passing in different environments.

**Logs/Stack traces**

[2023-04-04T19-23-07-connector-iroha2-ci-test-crash.log](https://github.com/hyperledger/cacti/files/11153277/2023-04-04T19-23-07-connector-iroha2-ci-test-crash.log)

**Cloud provider or hardware configuration:**

BuildJet runners

**Operating system name, version, build:**

Ubuntu-20.04

**Hyperledger Cactus release version or commit (git rev-parse --short HEAD):**

`main`

**Hyperledger Cactus Plugins/Connectors Used**

Iroha 2

PR #2381, opened 2023-04-07
fix(iroha2-connector): fix flaky tests to solve #2370 and #2373

outSH:
- Use waitForCommit for all transactions used in test suites to prevent spurious CI test failures.

Closes: #2370
Closes: #2373

### DIFF
--- a/packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha2-generate-and-send-signed-transaction.test.ts
+++ b/packages/cactus-plugin-ledger-connector-iroha2/src/test/typescript/integration/iroha2-generate-and-send-signed-transaction.test.ts
@@ -25,7 +25,6 @@ import {
 import {
   generateTestIrohaCredentials,
   IrohaV2TestEnv,
-  waitForCommit,
 } from "../test-helpers/iroha2-env-setup";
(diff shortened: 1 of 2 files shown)


petermetz (approved):
@outSH LGTM just please try to make the commit subject a little more unique. If neither of the file paths of the test cases fit in the subject because they are too long for the linter then I would at least include the two issue numbers in the subject as well just so that it's easy to distinguish this commit from the other commits in the future that will do the same thing (flaky tests are something that always comes up even if we disappear all of them at once new race conditions get added all the time)

jagpreetsinghsasan (approved):

outSH:
> @outSH LGTM just please try to make the commit subject a little more unique. If neither of the file paths of the test cases fit in the subject because they are too long for the linter then I would at least include the two issue numbers in the subject as well just so that it's easy to distinguish this commit from the other commits in the future that will do the same thing (flaky tests are something that always comes up even if we disappear all of them at once new race conditions get added all the time)

@petermetz Sure thing, done :)

petermetz (changes requested):
Dummy change request so that we don't accidentally merge this before the other orgs approval.

sandeepnRES (approved):
LGTM

PR #2384, opened 2023-04-10
feat(relay): configurable db_open retry mechanism added and in driver

sandeepnRES:
To manage concurrent accesses to a LevelDB instance so that protocols
don't fail because of temporary resource contention, added two env
variables in driver and config variables in relay to configure retry
mechanism while opening database:

- Max number of retries
- Backoff time in milli seconds.

Additionally upgraded go versions to v1.20 in publish workflows.


jagpreetsinghsasan (approved):
LGTM

VRamakrishna:
@sandeepnRES Add a bit more description to your initial comment: Go version upgrades, need for the retry logic (to manage concurrent accesses to a LevelDB instance so that protocols don't fail because of temporary resource contention.)

VRamakrishna (approved):

sandeepnRES:
Hi @petermetz I've modified the the PR description as per Rama's comments, and also added same in commit message.

petermetz (approved):
@sandeepnRES Thank you for making those changes! LGTM

PR #2391, opened 2023-04-13
chore(ci): try new builers

ryjones:
rename yaml->yml

update actions versions

Signed-off-by: Ry Jones <ry@linux.com>

### DIFF
--- a/.github/workflows/.dast-nuclei-cmd-api-server.yaml
+++ b/.github/workflows/.dast-nuclei-cmd-api-server.yaml
@@ -24,7 +24,7 @@ jobs:
             && sudo rm -f /etc/apt/sources.list.d/sovrin.list*
 
       - name: Set up NodeJS v16.9.1
-        uses: actions/setup-node@v2.1.2
+        uses: actions/setup-node@v3.6.0
         with:
           node-version: v16.9.1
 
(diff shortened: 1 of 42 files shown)


petermetz (approved):
@ryjones Thank you, LGTM!

takeutak (approved):

petermetz (changes requested):
@ryjones There are still some leftover yml -> yaml renames that need to be removed from the diff. I just checked out why the CodeQL is failing and it's also because of that (it only looks for it's own config file with .yml but not with .yaml)

Screenshot from 2023-04-14 00-33-18

Screenshot from 2023-04-14 00-33-04

ryjones:
@petermetz I renamed that one file back to `.yml`

jagpreetsinghsasan (approved):
LGTM

sandeepnRES:
Hi @ryjones I think you missed updating runner for `.github/workflows/test_weaver-data-sharing.yaml`, because the runner that used there was `buildjet-4vcpu-ubuntu-2204`. I don't know if this was intentional, hence asking you here.

petermetz (approved):
@ryjones LGTM, thank you!

PR #2402, opened 2023-04-22
chore(ci): revert to free runners

ryjones:


petermetz (approved):
@ryjones LGTM, thank you!

@jagpreetsinghsasan @takeutak @izuru0 @VRamakrishna @sandeepnRES I'll quickly merge this out of bounds because:
1. It doesn't do anything to the code
2. It is urgent because the foundation is incurring a lot of cost right now due to our massive test coverage
3. Hoping that you'll agree after the fact (but if not I'll revert of course)

PR #2420, opened 2023-05-09
chore(ci): update actions

ryjones:
many actions were using old versions.

update rtd so docs build.


petermetz (approved):
@ryjones LGTM, thank you!

PR #2422, opened 2023-05-11
chore(release): publish v2.0.0-alpha-prerelease

sandeepnRES:
Testing release with `alpha-prerelease`.

Have to change `gomodule` names due to golang mandatory requirement that any version with v2 or above, must have major version suffixed to the `gomodule` name. Also had to modify `go-gen-checksum` script according to this convention.


petermetz (approved):
@sandeepnRES LGTM - there'll most likely be issues with this one but we are on the right track to hammer out the details. Thank you for all the work you are putting in!

izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):
LGTM, thankyou for the PR !

PR #2430, opened 2023-05-22
docs(weaver/samples): pin solc to v0.8.8 and turn off IR for Besu asset exchange

### LINKED ISSUE #2423: fix(ci): Asset Exchange Besu workflow failing after PR #2420
**Describe the bug**

After the PR #2420 , the asset exchange besu workflow fails at `truffle migrate` step, i.e. when deploying contracts to the network.

**To Reproduce**

Run the workflow: `test_weaver-asset-exchange-besu.yaml`

**Expected behavior**

Should pass

**Logs/Stack traces**

Logs: [asset-exchange-besu.log](https://github.com/hyperledger/cacti/files/11449228/asset-exchange-besu.log)

petermetz:
1. Without having it pinned to v0.8.8, some breaking changes that solc
has snuck in [1] [2] around v0.8.15 (for IR) are breaking the contract
deployment in a way that opcodes end up in the migration contract's
constructor that Besu does not recognize ("opcode INVALID" in the besu
logs if you set the log level to "ALL") in the Besu logs and then sends
back an "internal error" message via the JSON-RPC response, which is a
bug IMO it should state that the user input was invalid (user input
being the contract)
2. Disabled IR in the truffle (solc) config which is a step backwards
because it's a new feature of the solidity compiler that has certain
benefits to it compared to the legacy compilation mode, but enabling it
breaks the build so it just had to be done. In the future if someone
has time to do a deep dive on why exactly it's failing, then it should
be re-enabled because having the legacy compilation mode being our default
is technical debt that should be paid off rather sooner than later because
we never know how it will come back to cause different issues later on.

When IR is enabled, the following error occurs (even on the pinned v0.8.8 solc):
> Compiling ./contracts/transferInterface.sol
> YulException: Variable var_amount_3290 is 9 slot(s) too deep inside the stack.

[1] https://docs.soliditylang.org/en/v0.8.15/ir-breaking-changes.html#semantic-only-changes
[2] https://github.com/ethereum/solidity/issues/13311#issuecomment-1199274310

Fixes #2423

### DIFF
--- a/weaver/samples/besu/simpleasset/truffle-config.js
+++ b/weaver/samples/besu/simpleasset/truffle-config.js
@@ -5,20 +5,34 @@ module.exports = {
 			port: 9544, // 7545 - default for Ganache
 			network_id: "1338", // 4447 - default for Ganache
 			//type: "fabric-evm",
-			from: "0x3c7F5262b873F637F8AdA2B09C34624B7F5fbCBF",
+			from: "0x35D99e41ce0Aa49d1B2B62f286b9F4189df7B7ae",
 			gasPrice: 0,
-			gas: "0x1ffffffffffffe"
+			gas: "0x1ffffffffffffe",
+			verbose: true, // helps with debugging contract deployment issues
 		}
 	},
 	compilers: {
 		solc: {
-			version: "^0.8.8",
+			// Unless this is pinned to 0.8.8, some breaking changes that solc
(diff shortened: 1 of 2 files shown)


jagpreetsinghsasan (approved):

VRamakrishna (approved):

izuru0 (approved):
LGTM

PR #2448, opened 2023-06-12
feat(quorum): add WebsocketProvider options to quorum LP

rwat17:
- add WebsocketProvider options to quorum LP
- add test for calling getBlock with verifier

### DIFF
--- a/packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
+++ b/packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
@@ -12,6 +12,10 @@ import Web3JsQuorum, {
   ISendRawTransaction,
   IPrivateTransactionReceipt,
 } from "web3js-quorum";
+import {
+  WebsocketProviderOptions,
+  HttpProviderOptions,
+} from "web3-core-helpers";
 import { AbiItem } from "web3-utils";
 import { Contract } from "web3-eth-contract";
 import { ContractSendMethod } from "web3-eth-contract";
@@ -84,6 +88,8 @@ export interface IPluginLedgerConnectorQuorumOptions
   prometheusExporter?: PrometheusExporter;
   pluginRegistry: PluginRegistry;
   privateUrl?: string;
+  wsProviderOptions?: WebsocketProviderOptions;
+  httpProviderOptions?: HttpProviderOptions;
 }
 
 export class PluginLedgerConnectorQuorum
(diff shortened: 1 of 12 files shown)


packages/cactus-plugin-ledger-connector-quorum/src/main/typescript/plugin-ledger-connector-quorum.ts
```diff
@@ -12,6 +12,10 @@ import Web3JsQuorum, {
   ISendRawTransaction,
   IPrivateTransactionReceipt,
 } from "web3js-quorum";
+import {
+  WebsocketProviderOptions,
+  HttpProviderOptions,
+} from "web3-core-helpers";
```
petermetz, line 18: @rwat17 Please declare this dependency in the package.json of the connector as well. While working on the dev build it is more or less guaranteed that this package will be available in the project root's `node_modules` but there are other situations in which it might not be present unless we explicitly specify the dependency.

packages/cactus-test-tooling/src/main/typescript/quorum/quorum-test-ledger.ts
```diff
@@ -105,6 +105,10 @@ export class QuorumTestLedger implements ITestLedger {
     const hostPort = await this.getRpcApiPublicPort();
     return `http://${ipAddress}:${hostPort}`;
   }
+  public async getRpcApiWsHost(): Promise<string> {
+    const ipAddress = "127.0.0.1";
+    return `ws://${ipAddress}:8546`;
```
petermetz, line 110: @rwat17 If it's not too much trouble please store the default port in a single place within this codefile and also make it queryable by other code via a method the same way `getRpcApiPublicPort()`  makes it possible for the HTTP port.

Making it this way will make the code less britle (resistant to changes around it) because we don't have to replace the port in multiple places if for whatever reason it has to be changed and so the probability of mistakes will be reduced.

petermetz (changes requested):

jagpreetsinghsasan (commented):
LGTM

petermetz (commented):
@rwat17 Did you see my comments from earlier?

rwat17:
@petermetz Yup, did what u asked for :)

izuru0 (approved):
LGTM

petermetz (changes requested):
> @petermetz Yup, did what u asked for :)

@rwat17 Nice! Thank you for that! Please rebase onto upstream main and resolve the merge conflicts and then pass it back for review!

petermetz (approved):
I know that @rwat17 got reassigned so I made the rebase myself, LGTM.
Also sneaking in an urgent  fix to the build (which was completely broken by me earlier - sorry everyone)

PR #2454, opened 2023-06-14
CI Scan

adrianbatuto:


petermetz:
@adrianbatuto This looks like it has gone stale - closing it for now. As always, feel free to re-open it as needed of course.
If you just opened it to run tests on the CI then I recommend using your personal fork for that prupose to reduce the noise on the upstream repo (and also the CI resource usage on the Hyperledger Foudnation's account)

PR #2453, opened 2023-06-13
feat(weaver): added multiple participants support for data sharing in…

sandeepnRES:
… corda

    - refactor(weaver): add pledge verification in interop contract
    - fix(corda-at): typo in ResponderRole for weaver asset transfer
    - fix(ci): rename asset-transfer-fabric job to asset-transfer
    - feat(weaver): add Consume command for external state
    
Closses #2447

PR #2459, opened 2023-06-15
fix: wrong session.receive type in responder flow of AssetTransferFlows.kt

philipbeaucamp:
issuerSession.send(AssetTransferResponderRole.ISSUER) and obsSession.send(AssetTransferResponderRole.OBSERVER) are sending objects of type AssetTransferResponderRole, but we were receiving
`val role = session.receive<ResponderRole>().unwrap { it }`
which results in a deserialization error.


sandeepnRES:
Hi @philipbeaucamp thanks for the PR, as I said on Weaver's issue, here in Cacti, I already opened PR #2453 for this issue (along with some changes). So am closing this as it will create conflicts since both are doing same changes.

PR #2460, opened 2023-06-15
docs ( update ) update FAQ.md file 

barnapa:
adding Alternative cacti install guide for developers using local machines in FAQ.md

Signed-off-by: Barnaba Pawelczak <barnaba.pawelczak@fujitsu.com>

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -186,3 +186,149 @@ this.quorum = new QuorumTestLedger({
   ...
 });
 ```
+## Setting up the machine for cacti as developing environment and modyfications.
+I ) install curl
+`sudo apt-get install curl`
+
+
+II) install docker 👍 and docker compose you can follow oficial instructions 👍
+https://docs.docker.com/engine/install/ubuntu/
+
+
+setup repository 👍
+`sudo apt-get update`
+
+
+`sudo apt-get install ca-certificates curl gnupg`
+
+
+add keys :
+`sudo install -m 0755 -d /etc/apt/keyrings`
+
+
+`curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg`
+
+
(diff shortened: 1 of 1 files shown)


petermetz (changes requested):
@barnapa Thank you! Could you please do these: 

1. put this in `./BUILD.md` instead? (and remove the overlaps - I think do
cker and docker compose, NVM being dependencies might already be in there)
2. `splint`: I'm all for using it, but could you please document what it is needed for specifically? I read the docs and it says it's for C code security vulnerability linting which is awesome (we need as much of that as we can possibly have) but would also be good to know where that code is because AFAIK there is no `.c` code files in the repo as of right now.
3. Please make sure to put it under the section that talks about Ubuntu/Debian Linux because not all of this would work on macOS (and none of it on vanilla Windows) - this way beginners don't get confused about whether they should try and run these commands on their machines or not (or at least will be a little less confused).
4. Change the commit subject line to express what the change is in a way that is not redundant with 99% of every other commit ever made to the same file. 
Reasoning: Literally every commit that does anything with file X could be described as "update X file" so it has zero actual information if you have that as the commit subject line and is only distinguishable from other similar commits if you go and inspect the diff and/or the commit metadata like timestamps which is a huge time drain when you are tracking down specific changes that were made in the past.
5. Change the PR title to match whatever the commit subject line is based on point 4)

jagpreetsinghsasan (changes requested):
There are some styling related issues
image
(we can use similar code blocks as the one used by `Fabric` and `Quorum` sections just above it, for better readability)

petermetz:
@barnapa Are you still working on this? If you don't have time to finish it right now, we can also just close it down in the meantime and pick it back up later. (I'm trying to clearo out the massive backlog of PRs that we have open right now)

petermetz:
Closing for now due to inactivity, please feel free to reopen it later as needed.

PR #2462, opened 2023-06-16
build(codegen): fix openapi concurrent jar download

### LINKED ISSUE #2461: build(codegen): fix openapi concurrent jar download issue
**Describe the bug**

With the newly added codegen tasks as part of the build process (kotlin API clients) we now have a large enough number of openapi-generator invocations happening concurrently that it causes problems for every single CI run (this used to be a once in a while problem).

The bug is that the .jar file for the openapi generator gets downloaded if not already, but this logic is probably not concurrency safe and so as we run multiple openapi CLI calls (the npm package) it gets twisted up and crashes with a file not found exception when it's trying to launch the jar.

**To Reproduce**

Run the CI for any PR, it reproduces consistently but always on a different job.

**Expected behavior**

Build is stable and it passes without needing 3 or 4 retries...

**Logs/Stack traces**

```sh                 
An unexpected error occurred while trying to open file 
cacti/node_modules/@openapitools/openapi-generator-cli/versions/6.3.0.jar
```

[2023-06-13-cacti-build-crash-openapi-jar-download-concurrency-problem.log](https://github.com/hyperledger/cacti/files/11764032/2023-06-13-cacti-build-crash-openapi-jar-download-concurrency-problem.log)

**Screenshots**

Are you running the software on a dev machine or somewhere in the cloud?

Ubuntu 20.04 and 22.04

`main`

N/A

**Additional context**

I'm hoping that this can be resolved by doing a cache warmup sort of operation as part of the build, meaning that it'll download the .jar file via a single(ton) execution of the openapi generator cli package and then when it is time to run the parallel codegen there won't be any chance for race conditions.
If that doesn't work, we need to look deeper into the code of the npm package that manages the downloads and maybe fix the issue there, contribute it back, wait for a release, finally upgrade.

petermetz:
Added a script to  warm up the file-system cache where the
.jar files are downloaded (which contain the OpenAPI generator)

This way the .jar download does not happen concurrently for 26
packages but instead just the one time right before the codegen
script is executed.

IMPORTANT NOTE: If you change the OpenAPI generator version
that is being used by the project from the current v5.2.1 and v6.3.0
then these warm-up scripts must also be updated at the same time
otherwise the same issue that we are fixing here now will creep back
in immediately.

 Fixes #2461


sandeepnRES (approved):

sandeepnRES:
Looks good but `build-dev` is failing.

petermetz:
> Looks good but `build-dev` is failing.

@sandeepnRES My bad, not sure why that extra file in the diff was there but I fixed it now, thanks!

jagpreetsinghsasan (approved):

takeutak (approved):

petermetz:
Oh no there was another race condition in there that I managed to add by doing the file deletion twice. /facepalm
@sandeepnRES  Thanks for merging! I'll have to submit another quick one soon though, apologies for the PR spam :-(

PR #2466, opened 2023-06-17
ci(github): evaluate cache viability

### LINKED ISSUE #2117: ci(github): evaluate cache viability
### Description

Currently we use the GitHub Action Cache to store/restore the yarn installed dependencies for the 40-something jobs that need it.
I caught the cache restore process taking hours instead of the usual 60 seconds because the bandwidth of it was 0.1 MB per second (you read that right...).
So it's worth asking the question at this point if it does more good than bad if it's this slow occasionally... I don't have a definitive answer because I'd need to know how often does this actually happen. It might still be worth sticking with it assuming that they'll improve on the stability of the performance and not make it worse over time (though in general software gets worse over time a lot of times)

image

### Acceptance Criteria
1. Make a decision on whether to stop using the cache or not.
2. Explain the decision. What are you basing it on? Why do you think it will be better this or that way?
3. Based on the decision the rest of the task is either a no-op or to implement the removal of the cache from the CI.

petermetz:
WORK IN PROGRESS

Fixes #2117

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,27 +20,86 @@ jobs:
       TAPE_TEST_RUNNER_DISABLED: true
     runs-on: ubuntu-20.04
     steps:
+      - uses: actions/checkout@v3.5.2
+
       - name: Use Node.js v16.14.2
         uses: actions/setup-node@v3.6.0
         with:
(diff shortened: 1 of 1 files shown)


petermetz:
Note to self: we don't need to do codegen on the CI.

PR #2533, opened 2023-07-03
fix(weaver-corda): throw error correctly in responder flows

sandeepnRES:
Bug fix in corda interop app (weaver), where flows would get stuck if responder flow verification fails.


petermetz (approved):
@sandeepnRES LGTM, thank you!

takeutak (approved):

Issue #2531, opened 2023-07-02
refactor(plugin-htlc-eth-besu): stop using the deprecated api.encodePacked


petermetz:
### Description

Stop using the `abi.encodePacked()` call in the `packages/cactus-plugin-htlc-eth-besu/src/main/solidity/contracts/HashTimeLock.sol` contract because that method has been deprecated according to [1]

[1] https://github.com/ethereum/solidity/issues/11593

### Acceptance Criteria
1. Tests are still passing
2. The new encoding call does not make this a breaking change, e.g. the outputs are consistent so that the test cases do not need to be changed to keep passing after the contract have been re-deployed.
3. Make sure to update the compiled `.json` artifact of the `.sol` source code file so that the tests do end up using the new revision of the `.sol` code.

RafaelAPB:
I'll do this task

RafaelAPB:
Done, PR #2538

PR #2538, opened 2023-07-06
refactor(plugin-htlc-eth-besu): stop using the deprecated api.encodePacked

RafaelAPB:
1. Add tests to vanilla HashTimeLock
2. Replace abi.encodePacked by abi.encode
3. Compile and replace generated ABIs (for HashTimeLock and PrivateHashTimeLock)

Addresses #2531 
cc @petermetz


petermetz (changes requested):
@RafaelAPB LGTM but could you please recompile the contracts with https://marketplace.visualstudio.com/items?itemName=JuanBlanco.solidity so that the .json artifacts structure is not changed at the same time?

petermetz (approved):

jagpreetsinghsasan:
LGTM (Peter has already approved to not approving it, as we need 3 approvals from different orgs)

RafaelAPB:
@petermetz thank you for updating the contracts!

Issue #2411, opened 2023-05-02
fix(tessera): updated tessera version error in quorum-all-in-one


ruzell22:
**Describe the bug**

Tessera version is updated in #2240 in the dockerfile of quorum-all-in-one to fix the vulnerabilities. However,  after the update, when running the docker image of quorum-all-in-one, tessera is having an issue. Versions lower than 21.1.7 works with the older version that is being used which is 21.1.1. The tessera was updated to 22.1.7 which has drastic changes which will be needed to be updated on our end as well in order for tessera to work for quorum-all-in-one.

**To Reproduce**

[process]
DOCKER_BUILDKIT=1 docker build ./tools/docker/quorum-all-in-one/ -t qaio
docker run --rm qaio

[error]
No such file or directory.

**Expected behavior**

Tessera will run when quorum-all-in-one docker image is ran. Quorum-all-in-one will run successfully as well.

**Additional context**

Related to #2240 (merged) and #2274 (fixed but on hold for tessera update)

ruzell22:
Hello @petermetz , can you assign to this me? Thank you.

petermetz:
> Hello @petermetz , can you assign to this me? Thank you.

@ruzell22 Done, thank you!

Issue #2274, opened 2023-01-26
bug(quorum-all-in-one): error "Failed to enumerate USB devices"


takeutak:
**Describe the bug**
In the process of the container building of quorum-all-in-one, there are the following error and the container cannot be launched.

```
[process]
DOCKER_BUILDKIT=1 docker build ./tools/docker/quorum-all-in-one/ -t qaio
docker run --rm qaio

[error]
"Failed to enumerate USB devices"
```

**To Reproduce**

When I edited [start-quorum.sh](https://github.com/hyperledger/cactus/blob/main/tools/docker/quorum-all-in-one/start-quorum.sh) and add "--nousb", the error does not happen.

```
geth \
  --identity node${NODE_ID}-${QUORUM_CONSENSUS:-raft} \
  --datadir ${DDIR} \
  --permissioned \
  --nodiscover \
  --verbosity 2 \
  --networkid ${NETWORK_ID} \
  --rpc \
  --rpccorsdomain "*" \
  --rpcvhosts "*" \
  --rpcaddr 0.0.0.0 \
  --rpcport 8545 \
  --rpcapi admin,db,eth,debug,miner,net,shh,txpool,personal,web3,quorum,${QUORUM_CONSENSUS:-raft} \
  --port 21000 \
  --allow-insecure-unlock \
  --unlock 0 \
  --password /passwords.txt \
  --nousb
  ${QUORUM_GETH_ARGS:-} ${GETH_ARGS_raft}
```
c.f.: https://github.com/ethereum/go-ethereum/issues/20568

aldousalvarez:
Hello @petermetz I would like to work on this one, Thank you.

ruzell22:
Hello @petermetz , can you assign this to me? Thank you.

petermetz:
@aldousalvarez Sorry for the slow response!
@ruzell22 You got it.

ruzell22:
Hi @takeutak , I have conducted multiple tests regarding the ticket including the one in the provided link.

Initial test without changes
![image](https://user-images.githubusercontent.com/89818529/229980360-355d18e2-1560-43ca-9e50-6dfbf07e40a3.png)

Added --nousb to replicate
![image](https://user-images.githubusercontent.com/89818529/229980430-a0018b29-93a5-420f-b0bc-87ebd0fad791.png)

changed --rpc flags to --http as per this log wherein it says it will be removed soon
[...]

PR #2240, opened 2022-12-23
fix(tools/quorum-all-in-one): address CVE-2021-36159 and CVE-2022-28391

aldousalvarez:
Other, lower severity vulnerabilities are also being addressed by this
change but the two big ones are the critical severity ones mentioned
in the commit subject.

Most of the vulnerabilities are now fixed in quorum-all-in-one but
there are still some that are not because most of the remaining
vulnerabilities are still new and is still waiting for the new changes
to be pulled in and released on their respective package versions.

And we tried to ask on the quorum discussions on github as you can
see here. (ConsenSys/quorum#1513).

Here are the remaining vulnerabilities for quorum-all-in-one:
CVE-2022-3602
CVE-2022-3786
CVE-2022-3602
CVE-2022-3786
CVE-2022-42003
CVE-2022-42004
CVE-2022-45868
CVE-2022-1471
CVE-2022-21698
CVE-2022-27664
CVE-2022-32149
CVE-2022-21698
CVE-2022-27664
CVE-2022-32149

Fixes #2059

Signed-off-by: aldousalvarez <aldousss.alvarez@gmail.com>
Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

PR #2549, opened 2023-07-12
fix(tessera): updated tessera version error in quorum-all-in-one #2411

ruzell22:
fixes: #2411

related to: #2240 and #2274

Tessera version was updated and some flags were deprecated. It has been
fixed in this commit. Also, the quorum and tessera compatibility with
quorum-all-in-one. The container and image has been built and ran
successfully.

### DIFF
--- a/tools/docker/quorum-all-in-one/Dockerfile
+++ b/tools/docker/quorum-all-in-one/Dockerfile
@@ -4,11 +4,15 @@ ARG TESSERA_VERSION=22.1.7
 FROM quorumengineering/quorum:$QUORUM_VERSION AS quorum
 FROM quorumengineering/tessera:$TESSERA_VERSION AS tessera
 
-COPY --from=quorum /usr/local/bin/geth /usr/local/bin/
 COPY --from=quorum /usr/local/bin/bootnode /usr/local/bin/
+
+#LIBSODIUM
+RUN apt-get update && apt-get install -y libsodium23
 
 # BASH
 RUN apt update && apt-get -y install bash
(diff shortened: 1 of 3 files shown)


ruzell22:
docker run --priveleged --rm qaio successfully ran and built

image

petermetz (commented):
@ruzell22 Please also run one of the tests against the locally built image (by modifying the test case to use it) and then post the exact path of that test case here. Once we have that, we know for sure that it doesn't just work as in the image does boot without crashing, but also that the test cases are still passing with the new image changes.

Other than that, it looks good to me already. Thank you!

ruzell22:
@petermetz  I have tested the following with launch.json JEST test:

• deploy-contract-via-web-service.test.ts (cactus/packages/cactus-test-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/deploy-contract-via-web-service.test.ts)
image

• v21.4.1-deploy-contract-from-json.test.ts (cactus/packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v21.4.1-deploy-contract-from-json.test.ts)
image

• v21.4.1-invoke-contract.test.ts (cactus/packages/cactus-plugin-ledger-connector-quorum/src/test/typescript/integration/plugin-ledger-connector-quorum/deploy-contract/v21.4.1-invoke-contract.test.ts)
image

petermetz (approved):
Thank you very much for confirming @ruzell22 - LGTM!

PR #2556, opened 2023-07-17
ci(tools/docker): stop compiling Iroha as part of the all-in-one image build

### LINKED ISSUE #2524: ci(tools/docker): stop compiling Iroha as part of the all-in-one image build
We should try and use the base-image that the Iroha maintainers publish.
This way our image could be built much faster and this would in turn make the
performance of the CI much better.

petermetz:
1. Upgraded the Iroha AIO image to use v1.4.0-patch-3
2. The new Iroha version now supports configuraton via ENV variables
3. Because of 2) we don't need to apply config file changes in the
entrypoint script of the container anymore, yay.
4. The new image uses the iroha release image as the base and  it
does NOT recompile the entire project which makes the build
orders of magnitude faster, wasting much less CI time/resources.
5. Exposed the healthcheck port for through the IrohaTestLedger class
so the other improvement we'll be able to do after this is to phase out
the hacky python script that does the healthcheck at the moment if
the new built-in healthcheck endpoint does actually work fine.
6. Migrated all the test cases to the new image version.

Fixes #2524

### DIFF
--- a/packages/cactus-plugin-ledger-connector-iroha/src/test/typescript/integration/generate-and-send-signed-transaction.test.ts
+++ b/packages/cactus-plugin-ledger-connector-iroha/src/test/typescript/integration/generate-and-send-signed-transaction.test.ts
@@ -8,9 +8,11 @@
 // Constants
 //////////////////////////////////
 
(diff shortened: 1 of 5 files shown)


jagpreetsinghsasan (approved):

izuru0 (approved):
LGTM

PR #2563, opened 2023-07-22
build(deps): fix npm (grpc) build on NodeJS v20.4.0

petermetz:
1. Eliminates all uses of the old `grpc` dependency from the codebase.
1.1. Upgraded all outdated fabirc-network, fabric-client and fabric-ca-client
dependencies to the latest stable 2.2.x version which is 2.2.18 at the time.
1.2. Also outrighted swapped `grpc` declarations with `@grpc/grpc-js`.
The rest of the diff is due to the incompatibility of the two mentioned.

Fixes https://github.com/hyperledger/cacti/issues/2562
Fixes https://github.com/hyperledger/cacti/issues/1507


sandeepnRES:
Hi @petermetz, I've corrected `fabric-ca-client` (along with other fabric packges) version in weaver packages, also upgraded `grpc-js` version in other places

petermetz:
> Hi @petermetz, I've corrected `fabric-ca-client` (along with other fabric packges) version in weaver packages, also upgraded `grpc-js` version in other places

@sandeepnRES The protos-js folder's package.json still had a `    "grpc": "1.24.11",` in it. I'm removing it.

jagpreetsinghsasan:
@petermetz the below test case fails
`packages/cactus-plugin-ledger-connector-fabric-socketio/src/test/typescript/integration/fabric-socketio-connector.test.ts` (as startFabricSocketIOConnector function is cleaned in the PR and returns void)
stating
image

takeutak (approved):

petermetz:
> packages/cactus-plugin-ledger-connector-fabric-socketio/src/test/typescript/integration/fabric-socketio-connector.test.ts

@jagpreetsinghsasan Thanks for catching that! I've removed the failing code for now just to get it moving along.

@takeutak Apologies but I had to make some further modifications so please take a look at the new diff if you have the chance!

packages/cactus-plugin-ledger-connector-fabric-socketio/src/main/typescript/connector/sign-utils.ts
```diff
@@ -71,19 +70,8 @@ function sign(
...
-  const ecdsa = new elliptic.ec(ecdsaCurve);
-  const key = jsrsa.KEYUTIL.getKey(privateKey) as any;
-
-  const signKey = ecdsa.keyFromPrivate(key.prvKeyHex, "hex");
-  const digest = hashFunction(proposalBytes);
-
-  let sig = ecdsa.sign(Buffer.from(digest, "hex"), signKey);
-  sig = preventMalleability(sig, key.ecparams);
-
-  return Buffer.from(sig.toDER());
```
VRamakrishna, line 86: It looks like the only dependency to the old `fabric-client` library is a hash function mapping. There's some code in the Weaver SDK that does something similar to what the deleted code is doing: https://github.com/hyperledger/cacti/blob/main/weaver/sdks/fabric/interoperation-node-sdk/src/eciesCrypto.js#L186. Perhaps you can use this as a dependency (or just copy the code for now) instead of deleting the functionality altogether?

VRamakrishna (changes requested):
LGTM. Just had a suggestion below for a refactoring instead of a deletion. See if that works. Otherwise, if you don't want to spend the time to refactor code right now, I'll approve.

petermetz:
> LGTM. Just had a suggestion below for a refactoring instead of a deletion. See if that works. Otherwise, if you don't want to spend the time to refactor code right now, I'll approve.

@VRamakrishna Thank you, this helps a lot! What you suggest makes a lot of sense to me and I agree it should be done. With that said, this pull request has been pending for a while and due to it being a fix for critical severity vulnerabilities I would suggest to merge it now and send a separate pull request later with the refactor as you described (I'd go with re-using the code not copying it unless it somehow introduces a circular dependency among the packages that we cannot get out of).
In terms of time: I will have some in the second half of next month (e.g. 4 weeks from now) but until then I have to ration my time unfortunately.

VRamakrishna (approved):

PR #2564, opened 2023-07-22
feat(weaver/common): add data view protocol buffer spec & RFCs for Besu

VattiPraveen:
feat(weaver/common): add data view protocol buffer spec & RFCs for Besu

### DIFF
--- a/weaver/common/protos/besu/view_data.proto
+++ b/weaver/common/protos/besu/view_data.proto
@@ -1,0 +1,41 @@
+syntax = "proto3";
+
+package besu;
+
+option java_package = "org.hyperledger.cacti.weaver.protos.besu";
+option go_package = "github.com/hyperledger/cacti/weaver/common/protos-go/v2/besu";
+
+
+// link below will take you to besu view rfc 
+// https://github.com/hyperledger/cacti/blob/main/weaver/rfcs/formats/views/besu.md
+//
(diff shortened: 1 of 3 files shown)


petermetz (changes requested):
@VattiPraveen Thank you! Please do the following:
1. Squash into a single commit
2. Make sure the message uniquely describes what is being done - the current commit messages of "modified X" are not unique because literally every commit ever does exactly `that` so it's hard to read the commit log that way if too many of these slip into the log.
3. Once you've come up with a good commit message please sync that up to the PR title and the PR description as well.
4. Please read the contributing.md guide for further details

Other than my usual administrative nit-picks: LGTM

petermetz (changes requested):
@VattiPraveen I modified the PR title with a commit subject suggestion. If you agree then you can just make that your commit subject when you squash.

VattiPraveen:
> feat(weaver/common): add data view protocol buffer spec & RFCs for Besu

Hi @petermetz.

I have made the changes as you suggested. Please review the PR.

sandeepnRES (approved):

petermetz (changes requested):
@VattiPraveen Nice! Now you just need to fix the DCO from what I can tell and then we should be good to go.

petermetz (approved):
@VattiPraveen LGTM, thank you!

PR #2566, opened 2023-07-24
chore(deps): clean up typescript compiler dependency declarations

### LINKED ISSUE #2532: chore(deps): clean up typescript compiler dependency declarations
We have declared 27 times so far the typescript compiler dependency in the project
throughout all the package.json files, but we only need it once, in the root, 
to compile the typescript code of the monorepo.

Acceptance criteria:

1. Get rid of all but one instance of it.
2. Make sure that every org's maintainers approve not just the minimum 2 because this affects the entire codebase and could cause annoying breakages in the build and/or test cases, examples, tutorials, etc.
3. For cases where we do have a justification for keeping around another dependency declration of it, list each and every case and why the decision was made to keep it around
3.1. If the reason for keeping it around is because something breaks then include the exact error messages generated and how to reproduce them exactly from scratch as well instead of just saying "build is breaking without it" or other not-so-helpful messaging.

petermetz:
Pinned all of the versions to 4.9.5 in all of the packages.

Fixes #2532

### DIFF
--- a/examples/cactus-example-carbon-accounting-backend/src/utility-emissions-channel/typescript/package.json
+++ b/examples/cactus-example-carbon-accounting-backend/src/utility-emissions-channel/typescript/package.json
@@ -22,8 +22,7 @@
     "mocha": "^8.3.2",
     "sinon": "^10.0.0",
     "ts-node": "^9.1.1",
-    "tslint": "^6.1.3",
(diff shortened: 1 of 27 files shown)


izuru0 (approved):
LGTM

PR #2574, opened 2023-07-24
build(deps): pin ALL dependency versions in package.json files 2023-07-24

### LINKED ISSUE #2571: build(deps): pin ALL dependency versions in package.json files 2023-07-24
Background information on why doing this is crucial:
https://reproducible-builds.org/

petermetz:
1. Replaced all tilde and caret characters in package.json
files with nothing so that all versions are pinned down
for safety and stability of the build/publishing process.
2. On top of just removing the tildes and carets, I also upgraded the
declared versions to whatever their caret upgrades would've been (e.g.
performed minor upgrades on all the versions). This is important because
this is how we achieve parity with the actual dependencies that were used
by the code prior to this PR - this also highlights why it is necessary to
have reproducible builds and pinned versions: because without it one has
no actual answer to the question of "So what dependencies does your code
need to function exactly?"
4. Removed the jasmine types from the root package because they
were causing build problems with the weaver tsc compilation.
The proper, longer term solution here is to make the weaver packages'
tsc configuration on-boarded to the monorepo structure, but I did not want
to blow up this commit to be 10x the current size just because of this.
I opened a couple of issues regarding this.
5. I added protobufjs as a dev dependency to the API server package because
artillery suddenly stopped working when grpcjs was upgraded and it was
complaining that one of its dependencies (protobufjs) cannot be imported.
6. I added the protocol buffer compiler to the dev container.
7. Added a new VSCode extension to the list of recommended ones because
I found it extremely useful while performing the minor upgrades in the
dozens of package.json files (100+ dependencies in total). Why? Because
all the other tooling that exists in the NodeJS ecosystem assumes that
doing major upgrades on dependencies in batch is the only thing needed.
None of the tools are able to do a minor upgrades only mass upgrade
operation, even the ones that claim this as a feature do not have it.
8. I synced the list of recommended VSCode extensions to the dev container
definition file so that the same extensions get installed automatically
in the container as well.

Fixes #2571


jagpreetsinghsasan (approved):
LGTM

izuru0 (approved):
LGTM

PR #2575, opened 2023-07-24
refactor(cacti-cmd-gui-app): rename to cacti-ledger-browser

### LINKED ISSUE #2572: refactor(cacti-cmd-gui-app): rename to cacti-ledger-browser
Rename the package to something more specific so that it doesn't squat on the entire
namespace of all GUI applications we ever wrote/will write.

This would be a breaking change but luckily we haven't released a new version of
it to npm yet so renaming it now is easy.

petermetz:
Fixes #2572


jagpreetsinghsasan (approved):
LGTM.
Just one quick question @petermetz , the folder itself has the old name, `package/cacti-cmd-gui-app`. So shouldn't we change that as well to have everything uniform?

izuru0:
I'm ok with changing the name but my preference is to use `cacti-ledger-viewer` or `cacti-ledger-browser` 
see the issue #2572 page

sandeepnRES (approved):

petermetz:
@takeutak @izuru0 @outSH Thank you all for the feedback and the wait (and sorry for the slow responses as well)! I've renamed it according to the guidance to `cacti-ledger-browser`!

@takeutak Very sorry to hear about your sickness, I hope you've gotten over it since then or if not, then that you are very close it being done!

Issue #2579, opened 2023-07-25
test(geth-all-in-one): improve geth helper class tests coverage


outSH:
- https://github.com/hyperledger/cacti/issues/2577 introduced geth all in one and helper class with limited test coverage. 
- Add new tests to improve package reliability.
 
Features to be tested:
 - `fullContainerImageName` test
 - `className` test
 - `container`/`web3` throws when not started, works after
 - `getContainerStatus` (check if healthy after test start)
 - dont test `useRunningLedger` or `omitPull` options
 - `createEthTestAccount` test
 - `newEthPersonalAccount` test
   - Refactor to use `Clef` instead of `eth.personal`
- `transferAssetFromCoinbase` test
- `deployContract` test
- `getRpcApiHttpHost` / `getRpcApiWebSocketHost` - valid URL returned

@rwat17

PR #2588, opened 2023-07-28
test(test-geth-ledger): add test for test-geth-ledger

rwat17:
- add tests for geth-test-ledger
- fix small errors in GethTestLedger class
    
Closes: #2579 
Depends on: #2577


packages/cactus-test-geth-ledger/src/test/typescript/solidity/simple-token-contract/SimpleToken.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity ^0.5.1;
```
petermetz, line 1: @rwat17 Please use a newer compiler if possible something like this `pragma solidity 0.8.19;` that we have in the other contracts. Also important to pin the version down so that the hard to trace bugs are circumvented where the ABI gets broken with minor updates of the solidity compiler (I spent days debugging one such issue a few weeks back)

petermetz (changes requested):
@rwat17 In addition to my above comment, could you please
1. Rename the .json file so that it is an exact match to the `.sol` file but with a different extenssion.
2. Obtain the JSON file by compiling it with the solidity VSCode extension which adds the `contractName` attribute (among other things) automatically as useful metadata that we sometimes use programmatically in test cases for reducing the number of magic strings the tests are depending on to function.

As an example of such .json file please take a look at this one: `packages/cactus-plugin-htlc-eth-besu-erc20/src/main/solidity/contracts/HashedTimeLockContract.json`

packages/cactus-test-geth-ledger/src/main/typescript/geth-test-ledger.ts
```diff
@@ -0,0 +1,414 @@
...
+  // For test development, attach to ledger that is already running, don't spin up new one
+  readonly useRunningLedger?: boolean;
+}
+
+/**
+ * Default values used by GethTestLedger constructor.
+ */
+export const GETH_TEST_LEDGER_DEFAULT_OPTIONS = Object.freeze({
+  containerImageName: "cactus_geth_all_in_one",
+  containerImageVersion: "latest",
```
petermetz, line 31: @rwat17 Sorry, I just noticed this one: Could you please pin the image version to a specific tag that will not change in the future? It will avoid the problem of the tests breaking down later due to the floating tags.

packages/cactus-test-geth-ledger/src/test/typescript/integration/contractData.json
```diff
@@ -0,0 +1,52 @@
+{
```
petermetz, line 1: @rwat17 Do you still need this and with this file name or was it left in the diff by mistake?

packages/cactus-test-geth-ledger/package.json
```diff
@@ -0,0 +1,71 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/hyperledger/cactus.git"
+  },
+  "license": "Apache-2.0",
+  "author": {
+    "name": "Hyperledger Cactus Contributors",
+    "email": "cactus@lists.hyperledger.org",
+    "url": "https://www.hyperledger.org/use/cactus"
+  },
```
petermetz, line 25: @rwat17 Could you please update these to refer to the project as cacti not cactus. I also just noticed this, sorry for missing it in the previous review.

petermetz (changes requested):

packages/cactus-test-geth-ledger/package.json
rwat17: changed

packages/cactus-test-geth-ledger/src/test/typescript/integration/contractData.json
rwat17: yup that was a mistake, forgot to delete this file :)

packages/cactus-test-geth-ledger/src/main/typescript/geth-test-ledger.ts
rwat17: done

packages/cactus-test-geth-ledger/src/test/typescript/solidity/simple-token-contract/SimpleToken.sol
rwat17: i decided to reuse already compiled contract from other package :)

packages/cactus-test-geth-ledger/src/main/typescript/geth-test-ledger.ts
```diff
@@ -0,0 +1,414 @@
...
+  readonly envVars?: string[];
+  // For test development, attach to ledger that is already running, don't spin up new one
+  readonly useRunningLedger?: boolean;
+}
+
+/**
+ * Default values used by GethTestLedger constructor.
+ */
+export const GETH_TEST_LEDGER_DEFAULT_OPTIONS = Object.freeze({
+  containerImageName: "cacti-geth-all-in-one",
```
petermetz, line 30: @rwat17 The correct image name should be fully qualified with the registry and the org IDs (GHCR and Hyperledger)

petermetz: @rwat17 the image name itself is also not what I published (cacti vs cactus)

packages/cactus-test-geth-ledger/src/test/typescript/integration/geth-test-ledger.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    log.info("Start GethTestLedger...");
+    ledger = new GethTestLedger({
+      emitContainerLogs: true,
+      logLevel: testLogLevel,
+    });
+    expect(ledger).toBeTruthy();
+    log.debug("Geth image:", ledger.fullContainerImageName);
+
+    await ledger.start(true);
```
petermetz, line 63: @rwat17 Omitting the image pull by default will make the test case permanently broken

packages/cactus-test-geth-ledger/src/test/typescript/solidity/simple-token-contract/SimpleToken.sol
petermetz: @rwat17 Even better! Thank you!

petermetz (approved):
LGTM

izuru0 (approved):
LGTM

petermetz (changes requested):
@rwat17 Please
1. Fix the DCO check
2. Fix the dependency declaration syntax (the robot is pretty simple minded unfortunately and won't understand `Depends on: #2577` only `Depends on #2577` (notice the colon difference) - this is important so that we don't accidentally merge the PR before the parent have been merged
3. Fix the `build-dev` check
4. When everything else is ready and the parent PR(s) have been merged please do the usual rebase + squash onto upstream/main

Marking it as a change request so we don't accidentally merge while the dependencies are not met. (once done with the chores above just pass it back for review)

petermetz:
I'm guessing this one has gone stale too since @rwat17 moved on, closing for now. We can re-open it anytime as usual.

PR #2589, opened 2023-08-01
build(openapi): license presence check

### LINKED ISSUE #490: build(openapi): license presence check
### Description

The build/CI should bail if any of our openapi.json files is missing the `"license"` property.

petermetz:
1. Adds a custom check that ensures that all the properties of the
license object are specified (name, identifier, url).
2. Optimized the custom-check that verifies the opan API json specs so
that it ignores **/node_modules/ not just node_modules (meaning that it
was only excluding the root dir not all sub-dirs of the packages as well)

Fixes #490


sandeepnRES (approved):

jagpreetsinghsasan (approved):
LGTM

PR #2597, opened 2023-08-05
test(jest): reduce Jest log verbosity - make logs more compact

### LINKED ISSUE #2595: test(jest): reduce Jest log verbosity - make logs more compact
Currently the Jest test logs are very spammy because it 3 or 4 lines for every
single line that the actual logger produces.

It is meant to be useful by way of including the stack trace of where the log
is originating from but in our case (since we use a logger library) it is always
pointing at the source code of said library which is not useful at all but at
the same time very distracting when one is reading/parsing the logs mentally.

petermetz:
Configure jest to override it's internal console logger with the
regular console object provided by the JS runtime.
This way it does not have the extra information printed that is
not useful for 99% of our use-cases (100% of the known ones).

Fixes #2595

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -4,7 +4,7 @@ module.exports = {
   testEnvironment: "node",
   maxWorkers: 1,
   maxConcurrency: 1,
-  setupFilesAfterEnv: ["jest-extended/all"],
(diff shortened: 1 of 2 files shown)


izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):
LGTM

PR #2596, opened 2023-08-04
build(docker-compose): upgrade to docker compose V2

outSH:
- `quorum-multi-party-all-in-one`: use newest `quorum-quorum-dev-quickstart`, update quorum versions to most recent available. Run ledger as `quorum` user (required by newer versions). Use docker compose V2 from alpine package registry instead of V1 from pip.
- `besu-multi-party-all-in-one`: similar changes as for quorum-multi-party-all-in-on. Fix broken besu private transaction tests.
- `fabric-all-in-one`: Use docker compose V2 from alpine package registry instead of V1 from pip.
- `sawtooth-all-in-one`: Use docker compose V2 from alpine package registry instead of V1 from pip.

Closes: #2593
Closes: #2557

PR #2598, opened 2023-08-07
fix(indy-validator): fix package dependencies

outSH:
- Update indy validator python dependencies.
- Add README chapter on updating python dependencies to simplify this process in the future.
- Use pinend ubuntu base image in indy-sdk-cli dockerfile.
- Do some minor README improvements and cleanups.

Tested with `discounted-asset-trade` (should work without an issue now, at least dockerless one)

Depends on: #2596


izuru0 (approved):
LGTM

petermetz (changes requested):
@outSH Thank you and LGTM, but please rebase onto upstream/main and resolve the merge conflicts!

outSH:
> @outSH Thank you and LGTM, but please rebase onto upstream/main and resolve the merge conflicts!

Done

tools/docker/fabric-all-in-one/Dockerfile_v1.4.x
```diff
@@ -41,7 +41,7 @@ RUN apk add --no-cache libc6-compat
 ENV CACTUS_CFG_PATH=/etc/hyperledger/cactus
 RUN mkdir -p $CACTUS_CFG_PATH
 # OpenSSH - need to have it so we can shell in and install/instantiate contracts
-RUN apk add --no-cache openssh augeas
+RUN apk add --no-cache augeas
```
petermetz, line 44: @outSH I think we need the OpenSSH server installed for contract deployments. Is this somehow part of the indy-validator fix or just conflict resolution gone wrong?

LGTM otherwise, but I want to make sure about this one.

petermetz (changes requested):
@outSH I just spotted that the parent PR's commit is still in here with a few lines of diff, but the parent PR has already been merged onto main so please squash or remove the second commit otherwise when we merge this there'll be two commits with the same message but one of them will have the actual diff while the other just a couple of these lines of changes.

tools/docker/fabric-all-in-one/Dockerfile_v1.4.x
outSH: @petermetz Yes, sorry, this is leftover from temporary change from dependent commit when I was testing this change. This has been reverted, openssh remains.

outSH:
> @outSH I just spotted that the parent PR's commit is still in here with a few lines of diff, but the parent PR has already been merged onto main so please squash or remove the second commit otherwise when we merge this there'll be two commits with the same message but one of them will have the actual diff while the other just a couple of these lines of changes.

@petermetz You're right, sorry for that. Fixed, rebased and ready for merge

petermetz (approved):
> @petermetz You're right, sorry for that. Fixed, rebased and ready for merge

@outSH No worries, thank you very much for the updates! LGTM

PR #2602, opened 2023-08-11
fix: use common conventions: tsconfig.json, package.json

rwat17:
- change package.json and tsconfig.json files to follow common
convention

Peter's additions to fix the broken build:
- Included shelljs typings as a dev dependency for the package called
cactus-plugin-ledger-connector-go-ethereum-socketio
- Also included "socket.io-client" as another missing dev dependency for
the package above.

Closes: https://github.com/hyperledger/cacti/issues/2216

Signed-off-by: Tomasz Awramski <tomasz.awramski@fujitsu.com>

### DIFF
--- a/packages/cactus-plugin-ledger-connector-go-ethereum-socketio/src/test/typescript/unit-test/validatorDriver_getNonceHex.ts
+++ b/packages/cactus-plugin-ledger-connector-go-ethereum-socketio/src/test/typescript/unit-test/validatorDriver_getNonceHex.ts
@@ -60,7 +60,7 @@ import { io } from "socket.io-client";
     reqID: reqID,
   };
 
-  const json2str = (jsonObj) => {
+  const json2str = (jsonObj: object) => {
     try {
       return JSON.stringify(jsonObj);
     } catch (error) {
@@ -68,25 +68,25 @@ import { io } from "socket.io-client";
     }
   };
 
-  socket.on("connect_error", (err) => {
+  socket.on("connect_error", (err: object) => {
     console.log("####connect_error:", err);
     // end communication
     socket.disconnect();
     process.exit(0);
   });
 
-  socket.on("connect_timeout", (err) => {
+  socket.on("connect_timeout", (err: object) => {
     console.log("####Error:", err);
     // end communication
     socket.disconnect();
     process.exit(0);
   });
 
-  socket.on("error", (err) => {
+  socket.on("error", (err: object) => {
     console.log("####Error:", err);
   });
 
-  socket.on("eventReceived", function (res) {
+  socket.on("eventReceived", function (res: object) {
     // output the data received from the client
     console.log("#[recv]eventReceived, res: " + json2str(res));
   });
--- a/examples/cactus-example-carbon-accounting-backend/package.json
+++ b/examples/cactus-example-carbon-accounting-backend/package.json
@@ -21,7 +21,7 @@
   "author": {
     "name": "Hyperledger Cactus Contributors",
     "email": "cactus@lists.hyperledger.org",
-    "url": "https://www.hyperledger.org/use/cactus"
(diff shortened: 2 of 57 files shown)


rwat17:
Also sorted them with $npm run custom-checks from other PR :) https://github.com/hyperledger/cacti/pull/2434

tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
+{"program":{"fileNames":[],"fileInfos":[],"root":[],"options":{"declaration":true,"esModuleInterop":true,"inlineSourceMap":true,"module":1,"removeComments":false,"strict":true,"target":4,"useUnknownInCatchVariables":false},"referencedMap":[],"exportedModulesMap":[]},"version":"5.1.6"}
```
petermetz, line 1: @rwat17 This looks like a build artifact that somehow escaped git ignore. If I'm wrong, could you please explain why we need this in the diff?

packages/cactus-plugin-ledger-connector-go-ethereum-socketio/src/test/typescript/unit-test/validatorDriver_getNonceHex.ts
```diff
@@ -60,33 +60,33 @@ import { io } from "socket.io-client";
     reqID: reqID,
   };
 
-  const json2str = (jsonObj) => {
+  const json2str = (jsonObj: object) => {
```
petermetz, line 63: @rwat17 Are these unrelated linter fixes? (which are great by the way) I'm fine with it either way but I do wnat to make sure it is clarified what this actually is in terms of purpose.

petermetz (changes requested):
@rwat17 Mostly LGTM, please see my comments above!

> Also sorted them with $npm run custom-checks from other PR :) #2434

**For future reference** 
(it's OK for this one because I'm trying to be a little less annoying with my nit-picking)
When you combine two separate changes when one is relatively small and the other one is relatively big, it makes it hard to interpret the changes later. E.g., next year if you come back tyring to figure out something specific about the way you changed the compiler root directory, you'll have to dig through the hundreds of lines of diff from the sorting which could've been in a separate commit not shadowing the meaningful changes that have actual impact on the build system (the sorting is important for develoer experience too but it does not actually change anything so it's just noise when someone is looking at the diff)

sandeepnRES (approved):
LGTM, but please address peter's comment.
Also all the `package.json` files in `examples` that are in the diff are just re-arranging of lines, and nothing new in that, would recommend to revert them, but not strongly against it.

packages/cactus-plugin-ledger-connector-go-ethereum-socketio/src/test/typescript/unit-test/validatorDriver_getNonceHex.ts
rwat17: TypeScript required to type these arguments while building at some point.

tsconfig.tsbuildinfo
rwat17: sorry I forgot to delete this :)

rwat17:
@petermetz I agree about not sorting packages in this commit, it is not very relevant and not purpose of this one. It should be separated. I think it will be better to sort it after I fix my other PR about not working sorting packages functionality. Please review this changes and tell me if I have to revert what Sandeepn mentioned too :)

petermetz (approved):
@rwat17 Awesome, thank you very much for going the extra mile to decouple the sorting changes from the rest of it! LGTM

petermetz (commented):
@rwat17 FYI: I fixed the build just now - there were missing dependencies. I also made an attempt at making the commit subject a little more specific

rwat17:
@petermetz Thanks and sorry for my oversights

PR #2607, opened 2023-08-12
build(deps): upgrade Jest to v29.6.2 project-wide

### LINKED ISSUE #2606: build(deps): upgrade Jest to v29.6.2 project-wide
Also upgrade the `@types/jest` pacakge at the same time so that the typings
are consistent with the new version of the code being used by the tests.

The reason why we need to upgrade jest is because older versions have vulnerable [1]
versions of `ansi-regex` declared as transitive dependencies.

[1] https://github.com/hyperledger/cacti/security/dependabot/122

[why-ansi-regex.log](https://github.com/hyperledger/cacti/files/12326242/why-ansi-regex.log)

petermetz:
Also upgraded these packages to make sure that
they all work correctly together:
1. @types/jest
2. jest-extended
3. ts-jest

Also disabled build scripts for nodemon which was
crashing during yarn install.

Fixes #2606


izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):

PR #2617, opened 2023-08-15
fix(security): crash in HeaderParser in dicer

petermetz:
Addresses the advisory:
https://github.com/hyperledger/cacti/security/dependabot/176

The dicer dependency is included via the express-openapi-validator
package which needed to be upgraded project-wide for a fix.

The test cases for verifying that the OpenAPI validation still works had
to be updated because the way invalid field names are represented by
the new version of the validator have been changed. It used to say if
in the request body there was a problematic field ".body.SOME_FIELD"
but now it says instead "/body/SOME_FIELD" so a minor change was needed
in the test cases assertions to make sure they are not ending up with
false negative results.

URL https://github.com/advisories/GHSA-wm7h-9275-46v2
CVE ID CVE-2022-24434
GHSA ID GHSA-wm7h-9275-46v2

Fixes https://github.com/hyperledger/cacti/issues/2616

### DIFF
--- a/packages/cactus-cmd-api-server/package.json
+++ b/packages/cactus-cmd-api-server/package.json
@@ -75,7 +75,7 @@
     "express": "4.17.3",
     "express-http-proxy": "1.6.2",
     "express-jwt": "8.4.1",
-    "express-openapi-validator": "4.12.12",
+    "express-openapi-validator": "5.0.4",
     "express-rate-limit": "6.7.0",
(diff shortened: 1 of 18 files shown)


izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):
LGTM

VRamakrishna (approved):

PR #2632, opened 2023-08-23
build(tools): fix tooling globs where cactus is hardcoded

### LINKED ISSUE #2576: build(tools): fix tooling globs where cactus is hardcoded
## Summary

A few of the checks we have that are linting manifest files for example are
hardcoding glob patterns with `cactus` in them to find the files they need to 
parse but this won't work because now we started addnig new packages with 
`cacti` in them as well.

Ultimately the best solution is to keep the lerna/yarn workspace globs as the
single source of truth on what paths we must check when doing these tasks.

`tools/custom-checks/run-custom-checks.ts`

`tools/custom-checks/check-sibling-dep-version-consistency.ts`

`tools/custom-checks/check-package-json-sort.ts`

adrianbatuto:
Fixes #2576

### DIFF
--- a/tools/custom-checks/check-package-json-sort.ts
+++ b/tools/custom-checks/check-package-json-sort.ts
@@ -28,7 +28,9 @@ export async function checkPackageJsonSort(
   const __filename = fileURLToPath(import.meta.url);
   const __dirname = path.dirname(__filename);
   const SCRIPT_DIR = __dirname;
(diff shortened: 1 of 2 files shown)


jagpreetsinghsasan (approved):
LGTM

sandeepnRES (approved):

petermetz (approved):
@adrianbatuto LGTM

Issue #2299, opened 2023-02-23
feat(openapi): upgrade to 6.3.0 - phase 2


jagpreetsinghsasan:
### Description
As a developer, I want to upgrade the below mentioned packages to use the latest 6.3.0
Upon initial spike, I found out that some packages need some in-depth code changes to possibly have an upgrade for 6.3.0
Phase 2 is separated out from Phase 1, due to estimated high complexity of fixing the below mentioned broken packages with 6.3.0 upgrade

Broken packages | Reasons
-- | --
packages/cactus-plugin-ledger-connector-fabric | the generated code has quoted fields, which are messing up with the current state of the fabric connector class
packages/cactus-plugin-ledger-connector-iroha | packages/cactus-plugin-ledger-connector-iroha/src/main/typescript/generated/openapi/typescript-axios/api.ts file has variables changed. RunTransactionSignedRequestV1 has signedTransaction which was earlier of type any, but now it is File
packages/cactus-plugin-ledger-connector-iroha2 | Same as packages/cactus-plugin-ledger-connector-iroha
packages/cactus-plugin-ledger-connector-corda | the generated code fails to run this: `docker build ./packages/cactus-plugin-ledger-connector-corda/src/main-server/ -f ./packages/cactus-plugin-ledger-connector-corda/src/main-server/Dockerfile`

### Acceptance Criteria
1. The above mentioned packages should use 6.3.0 version of openapi-generator

PR #2633, opened 2023-08-23
ci(codegen)!: fix the yarn codegen job in ci.yaml

### LINKED ISSUE #2618: ci(codegen): fix  the yarn codegen job in ci.yaml
### CI Logs Attachment

[2023-08-14T07-19-09-cacti-ci-yarn-codegen-crash.log](https://github.com/hyperledger/cacti/files/12353706/2023-08-14T07-19-09-cacti-ci-yarn-codegen-crash.log)

### Likely cause

The older version of the generator that we use in parallel (v.5.2.1) has a different schema defined for the spec files by default and our changes to the license object are not compatible with that (only the 6.x version)

Potential solution is to upgrade all the generators to 6.x but I have some vague memory of this not being feasible right now because the way some of our specs are written that breaks compilation of the generated code.

petermetz:
BREAKING CHANGES:
1. Fabric connector endpoints are using base64 encoding instead of JSON
when needing to transfer otherwise binary data (UInt8Array)
2. Same as the above for the Iroha 1 and Iroha 2 connectors.

Primary changes:
-----------------
1. Removed the `"identifier": "Apache-2.0",` section from all OpenAPI
specification documents (./**/openapi.json) project-wide because the
OpenAPI generator is not yet compatible with it and because of that the
new property just makes it crash with a false-negative validation error.
2. Downgraded all OpenAPI specification documents version from `v3.1.0` to
`v3.0.3` (which is how it used to be) because the generator is not yet
compatible with it and it had caused types to be mangled.
3. Eliminated incorrect usage of the "format": "binary" declarations in
the OpenAPI specifications where the format was declared as binary but
the data was still traveling as JSON strings despite it (divergence
between our spec .json files and the implementation under the hood)
4. Updated the tests and their related endpoints to use base64 encoding
instead of JSON serializing UInt8Array instances: It is more efficient
this way and also has the benefit of honoring the specification. The
reason why it's more efficient is because JSON serializing a typed UInt8Array
results in an object literal that maps every single byte to a property
of said object, roughly tripling the size of the network transfer. It
probably also is slower to serialize/deserialize but I haven't actually
measured this.
5. All of the packages are now using the OpenAPI generator v6.6.0 instead
of some being stuck on v5.2.1. This might cause build issues in kotlin
API clients but because we don't have a freeze on breaking changes right
now (leading up to v2.0.0) it is the right time to make drastic changes
like this.

Fixes #2618
Fixes https://github.com/hyperledger/cacti/issues/2299


petermetz:
All: Apologies for the huge diff, changing the generator configuration unfortunately has this bad habit of generating huge diffs.
I also had to break it up to two separate commits so that the changes that are not (strictly) related to the generator update are in a separate commit for easier reading.

jagpreetsinghsasan (approved):
LGTM
This also fixes #2299

jagpreetsinghsasan (approved):

izuru0 (approved):
LGTM

PR #2185, opened 2022-11-02
feat(cbdc-bridging-app): implementation of CBDC bridging example

AndreAugusto11:
This PR presents a new example to demonstrate the capabilities of Cactus. This application demonstrates a CBDC bridging between a Fabric and a Besu network using the ODAP business logic plugin, a protocol under standardization for cross-chain operations in the IETF.

The goal is to demonstrate the use of npm packages of Cactus so that anyone can build a solution such as this. (Some dependencies are not using npm packages and the Dockerfile is not functional yet since some npm packages are not available yet – an example is the ODAP plugin).

Summary of the contributions:
* CBDC application backend (setting up and interacting with the underlying ledgers through smart contracts)
* Add support for multiple organizations in some operations in the FabricTestLedger
* Add support for minting some Ether to a given address in the BesuTestLedger

Future Work:
* CBDC application frontend (UI demonstrating the use of the app). It is already built, but there are problems with the dependencies clashing with the ones of the whole project. For now, I will leave a note for the docker image (`aaugusto11/cactus-example-cbdc-bridging-frontend`) so that one can run the entire app and play with it.
* A Dockerfile for the backend will be created as soon as this package is published to npm
* More documentation will be added in the next weeks.

Depends on #2205
Depends on #2251 
Depends on #2249

Signed-off-by: André Augusto <andre.augusto@tecnico.ulisboa.pt>

PR #2642, opened 2023-08-28
feat(cbdc-bridging): add frontend code for the CBDC example

AndreAugusto11:
* UI to interact with the CBDC example backend

This PR closes one action pointed out in PR #2185


takeutak (approved):

petermetz (changes requested):
@AndreAugusto11 Please integrate the package into the monorepo, right now it seems completely independent (build scripts, name, version, lock file, etc.).
The easiest way (to me at least) to get it done is to 
1. copy paste an existing package as a whole and then 
2. throw out the source code of it and 
3. put in the one you are contributing
4. ensure everything was renamed as needed (free text search, check both lower case, upper case, snake, camel case etc.)

This usually guarantees that I don't forget things that the CI don't explicitly check for.

AndreAugusto11:
> @AndreAugusto11 Please integrate the package into the monorepo, right now it seems completely independent (build scripts, name, version, lock file, etc.). The easiest way (to me at least) to get it done is to
> 
> 1. copy paste an existing package as a whole and then
> 2. throw out the source code of it and
> 3. put in the one you are contributing
> 4. ensure everything was renamed as needed (free text search, check both lower case, upper case, snake, camel case etc.)
> 
> This usually guarantees that I don't forget things that the CI don't explicitly check for.

@petermetz thanks for the reminder. The code is all in js (not ts) is there a problem with that? Is there already any js package that I can base myself on (I'm expecting some configs to be different)?

petermetz (approved):
@AndreAugusto11 Sorry for the slow response!
I made the monorepo integration at least on the level that the yarn install in the root will now install the dependencies of this one as well.
I also fixed the build (it would fail due to linter issues)
and updated the package.json metadata to have cacti instead of cactus, etc.

Verified that it works with the back-end that we already had in the codebase.

JS only code: Well, it would be nice to migrate it to TS but I'm also very short on time and trying to balance good enough vs. perfection as usual, so let's just go ahead and merge this in now that it does work and is technically part of the monorepo (even though some of the conventions are not yet followed)

Huge thanks for the contribution this helps a lot!

AndreAugusto11:
> @AndreAugusto11 Sorry for the slow response! I made the monorepo integration at least on the level that the yarn install in the root will now install the dependencies of this one as well. I also fixed the build (it would fail due to linter issues) and updated the package.json metadata to have cacti instead of cactus, etc.
> 
> Verified that it works with the back-end that we already had in the codebase.
> 
> JS only code: Well, it would be nice to migrate it to TS but I'm also very short on time and trying to balance good enough vs. perfection as usual, so let's just go ahead and merge this in now that it does work and is technically part of the monorepo (even though some of the conventions are not yet followed)
> 
> Huge thanks for the contribution this helps a lot!

@petermetz thank you very much, as usual!! 💯💯

I believe it would not be hard to migrate to TS, so I will try to work on that once I have some spare time.

PR #2653, opened 2023-09-02
feat: Add VSCode extension

RANJITHROSAN17:


docs/source/conf.py
```diff
@@ -37,7 +37,7 @@
 extensions = [
     'sphinx.ext.todo',
     'sphinx_markdown_tables',
- #   'recommonmark',
+    'recommonmark',  # Uncomment this line
```
izuru0, line 40: is it possible to achieve your objective, which isn't clear because there are no descriptions in this PR, without using the recommonmark?
https://github.com/readthedocs/recommonmark says it is now deprecated

petermetz (changes requested):
@RANJITHROSAN17 +1 to what @izuru0 said above. Please
1. Use the GitHub syntax indicating which issue this is fixing and make sure that the automatic linking is done.
2. Write a legible commit message and PR description and make sure that they match each other.
3. Remove white-space formatting from the diff entirely.
4. Remove unnecessary changes like altering python dependencies in unrelated packages.

petermetz (commented):
@RANJITHROSAN17 In addition to the above, if you need any help with the git mechanics or otherwise in general, just let me know (and invite me as a collaborator to your fork so that I can step in as needed)

PR #2654, opened 2023-09-06
chore(tools): add script to strip out non-production sources

### LINKED ISSUE #2652: chore(tools): add script to strip out non-production sources
### Description
As a maintainer I want to have a convenient way to have a script do a fresh git clone and then strip out non-production files/source code so that it is easy for me to submit the codebase for static code analysis.

`./tools/create-production-only-archive.ts`

### Acceptance Criteria
1. The script does a git clone to a temporary folder
2. After finishing with the operation it deletes the temporary folder to clean up after itself
3. The output is a .zip file that is produced into the current working directory date and timestamped uniquely
4. The script can be ran with a one-liner terminal command

petermetz:
To run the script you can execute this on a shell:
    yarn tools:create-production-only-archive

The above will create a temporary clone of the project sources, then
delete a lot of the files that are documentation/test code (e.g. things
that do not go into production). Finally it zips the remaining files
together into something that includes a date & timestamp and the git
hash that the archive was generated from.

There were plans to make the generated zip file also encrypted, but this
is not supported by the library that we are using for compression. The
other library was not necessarily purely Javascript implemented and so
there was some reluctance to use it because of the huge overhead on the
the dependency installation process that we already have from packages
that use native code to satisfy their stated purpose.
So, with all that said, encrypting the .zip file is left as a to-do for
later. The issues tracking this feature are here:
1. https://github.com/cthackers/adm-zip/issues/259
2. https://github.com/cthackers/adm-zip/issues/398

Fixes #2652


jagpreetsinghsasan (approved):
This is a good feature to have. Will suggest a similar feature for HL Bevel as well

sandeepnRES (approved):

PR #2660, opened 2023-09-07
build(weaver-go): Update code and dependencies in Go packages for Fabric

VRamakrishna:
- Updated Fabric Go SDK code in `weaver/sdks/fabric/go-sdk` to work with the latest protos and release a new version of the library.
- Updated dependencies in `go.mod` for Fabric Interop Chaincode, Fabric Go CLI, and sample chaincodes.
- Updated dependencies in relay `Cargo.lock`.
- Updated Fabric testnet launch script to use the right Fabric Interop Chaincode package.
- Set correct release versions for Fabric driver and IIN Agent local builds.


jagpreetsinghsasan (approved):
LGTM

petermetz (commented):
@VRamakrishna Please fix the commit lint check [1]. On top of that, some of my recommendations:
1. I'd also recommend specifying the component/scope as something else than `update` in the commit messages
2. I would also consider using `build` as the type of the commits instead of `fix` because users will see these in the release notes and not be able to assign any meaning to it because it's not something that was broken in the previous release and now fixed but instead it's just that the build was broken for a period of time during development in-between releases (to which the users would most likely just be noise rather than signal).
Sorry for the nit-picking, these are just recommendations, not an actual change request because I also consider these highly subjective.

[1]
image

weaver/core/relay/Cargo.lock
```diff
@@ -1810,9 +1810,9 @@ dependencies = [
 
 [[package]]
 name = "webpki"
-version = "0.22.1"
+version = "0.22.0"
```
sandeepnRES, line 1813: This is reducing the version, is it intended?

weaver/tests/network-setups/fabric/dev/scripts/setupCC.sh
```diff
@@ -1,11 +1,12 @@
 directory=$(dirname $0)
 
+CACTI_VERSION=v2@v2.0.0-alpha.1
```
sandeepnRES, line 3: if you are putting `v2` to variable, then may be do it same for `line 9` as well?
Also instead of `v2@v2.0.0-alpha.1`, may be just store the `v2.0.0-alpha.1` as the version, and extract `v2` (first substring till the first `.`) for the first part, because that will be the convention for version like vA@A.B.C

petermetz: @VRamakrishna @sandeepnRES Just in case it is useful here: In other places in the code there's also examples of us using CloudFlare's bash semantic version parser in case you need to write logic based on major/minor/patch versions:

Usage:
https://github.com/hyperledger/cacti/blob/main/tools/docker/fabric-all-in-one/run-fabric-network.sh#L16

Where and how to get the .sh file to include: 
https://github.com/hyperledger/cacti/blob/main/tools/docker/fabric-all-in-one/Dockerfile_v2.x#L125

weaver/core/relay/Cargo.lock
VRamakrishna: I'll revert the version change. Not sure how it crept in.

weaver/tests/network-setups/fabric/dev/scripts/setupCC.sh
VRamakrishna: @sandeepnRES I fixed the script as per your suggestion. Please take a look again.

VRamakrishna:
@petermetz @sandeepnRES Fixed commits and code as per your suggestions. Please review again.

@sandeepnRES It'll need your approval as Jagpreet has added his.

weaver/tests/network-setups/fabric/dev/scripts/setupCC.sh
```diff
@@ -1,11 +1,12 @@
 directory=$(dirname $0)
 
+CACTI_VERSION=v2.0.0-alpha.1
 TMP_PATH=$PWD/../shared/tmp
 CHAINCODE_PATH=$PWD/../shared/chaincode
 rm -rf $CHAINCODE_PATH/interop
 
 # interop cc module
-INTEROPCC_MOD=github.com/hyperledger-labs/weaver-dlt-interoperability/core/network/fabric-interop-cc/contracts/interop
+INTEROPCC_MOD=github.com/hyperledger/cacti/weaver/core/network/fabric-interop-cc/contracts/interop/v2
```
sandeepnRES, line 9: instead of hardcoding `v2` take it from first number of `CACTI_VERSION` variable.

VRamakrishna: Fixed

sandeepnRES (approved):

VRamakrishna:
@petermetz `cactus-cmd-api-server` test is failing because of an NPM dependency failure. This has nothing to do with my code changes. Can you check?

petermetz:
> @petermetz `cactus-cmd-api-server` test is failing because of an NPM dependency failure. This has nothing to do with my code changes. Can you check?

@VRamakrishna Yeah, I traced it back to the new npm publish of `v2.0.0-alpha.1` and it can't get fixed until we publish alpha.2 on npm so for now I just disabled the CI check for cmd-api-server.

PR #2631, opened 2023-08-22
feat(cactus-plugin-ledger-connector-ethereum): refactor connector API

outSH:
- Refactor to single `invokeContract` endpoint that can accept multiple methods of 
    supplying contract definition (directly, from keychain, etc..).
- Same for `deployContract` method.
- Update cactus-common `safeStringifyException` function to better handle axios errors
    (and other custom exceptions that support `toJSON()` method.).
- Use common error handling in all connector endpoints (except prometheus).
- Fix the tests.
- Switch to `Cacti` work in entire connector (except for package name)

Depends on #2630

Signed-off-by: Michal Bajer <michal.bajer@fujitsu.com>

PR #2707, opened 2023-09-18
feat(cactus-plugin-ledger-connector-ethereum): add stress test

outSH:
- Return JSON error before checking for Error instance in safeStringifyException
    (for cases when custom errror extends `Error` with `toJSON()` method)
- Move artillery dependency up to the root (since its common tool depdenecy)
- Add web3js http/ws provider options to connector.
- Make http provider optional (can run on ws web3js provider only)
- Don't register async endpoints when WS provider is not available.
- Return 400 for invalid responses from ethereum ledger (to distinguish from connector errors)
- In docker `geth-all-in-one` allow overwriting of default options when starting the ledger.
- Add artillery stress test scripts
    - Template config and artillery functions file
    - Common environment setup file
    - CLI for running the test environment on a separate machine.
    - Jest test to run quickly stress test on same machine.
    - 
Depends on #2631

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ site/
 tools/docker/geth-testnet/data-geth1/
 
 .history/
+.manual-geth-artillery-config.yaml
--- a/packages/cactus-common/src/main/typescript/error-utils.ts
+++ b/packages/cactus-common/src/main/typescript/error-utils.ts
@@ -24,10 +24,6 @@ export class CodedError extends Error {
  * @returns Safe string representation of an error.
  */
 export function safeStringifyException(error: unknown): string {
-  if (error instanceof Error) {
-    return sanitizeHtml(error.stack || error.message);
-  }
-
   // Axios and possibly other lib errors produce nicer output with toJSON() method.
   // Use it if available
   if (
@@ -36,7 +32,11 @@ export function safeStringifyException(error: unknown): string {
     "toJSON" in error &&
     typeof error.toJSON === "function"
   ) {
-    return sanitizeHtml(error.toJSON());
+    return sanitizeHtml(safeStringify(error.toJSON()));
+  }
+
+  if (error instanceof Error) {
+    return sanitizeHtml(error.stack || error.message);
   }
 
   return sanitizeHtml(safeStringify(error));
--- a/packages/cactus-plugin-ledger-connector-ethereum/README.md
+++ b/packages/cactus-plugin-ledger-connector-ethereum/README.md
@@ -157,6 +157,32 @@ To check that all has been installed correctly and that the pugin has no errors 
 npx jest cactus-plugin-ledger-connector-ethereum
 ```
 
+### Stess test
+- Use CLI for manual setup of test environment and geneartion of artillery config.
+- `artillery` must be installed separately (we do not recommend running it if they are any known open vulnerabilities)
+
+#### Setup
+
+``` sh
+# Start the test environment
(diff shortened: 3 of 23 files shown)


petermetz (commented):
@outSH 

1. Artillery had to be removed earlier because of critical severity dependencies that they did not have a fix.
Do you feel strongly about adding it back or could we run the stress tests in some other way? I was thinking of using JMeter because we already need JDK installed for the code generator anyway so it's not a big change to start using JMeter (though I would prefer something JS/TS/NodeJS too...).
Kind of an open question because all the options have trade-offs.
2. Please rebase to upstream/main and resolve the merge conflicts and pass it back for review!

outSH:
> 1. Artillery had to be removed earlier because of critical severity dependencies that they did not have a fix.

@petermetz Isn't artillery a dev dependency? Do we have similar requirements for both dev and prod packages or is this issue with artillery really that critical? In the end I can remove jest script and leave the rest (which just generates artillery config), that way we can remove artillery completely.

As for alternative stress test framework to use from now on I'd go with https://github.com/grafana/k6

petermetz:
> > 1. Artillery had to be removed earlier because of critical severity dependencies that they did not have a fix.
> 
> @petermetz Isn't artillery a dev dependency? Do we have similar requirements for both dev and prod packages or is this issue with artillery really that critical? In the end I can remove jest script and leave the rest (which just generates artillery config), that way we can remove artillery package completely.
> 
> As for alternative stress test framework to use from now on I'd go with https://github.com/grafana/k6

@outSH Yup, dev dependency indeed, but I was worried that if I mark it false positive in the tool then we'll just forget that it's there and could later add some other package to prod that has the same issue which will not get masked by the earlier false positive. I'm not a 100% sure if this would happen, maybe the scanners are sophisticated enough to catch this kind of problem, but I didn't want to risk it with critical and high severity vulnerabilities (so my policy was so far that high and critical severity needs to be addressed even if they are dev dependencies).

My other fear is that contributors & maintainers still have to install the vulnerable dependencies so them not causing any trouble in production on the servers doesn't mean that you or me couldn't accidentally get compromised by some vulnerability in the dev stack (probably low probability but...)

petermetz:
> In the end I can remove jest script and leave the rest (which just generates artillery config), that way we can remove artillery package completely.

Works for me, I'm trying to be as flexible as possible.

> As for alternative stress test framework to use from now on I'd go with https://github.com/grafana/k6

This also could work. I'll leave it up to you with the ask that if you go for k6 (which looks like a high quality tool at a first glance) please update the relevant docs and CI scripts , dev container image, etc. so that people won't get hit by the issue that the native code (golang) parts of k6 are not installed and they have no idea what's going wrong because it wasn't listed anywhere in the docs that they should be)
The other thing is to cover at least ubuntu and macOS for the install steps (windows is not working right now anyway)

petermetz (changes requested):
@outSH LGTM but please pass it back for review once the parent PRs were done and the merge conflicts have been resolved (e.g. the diff has settled close to it's final state)

izuru0 (approved):
LGTM

outSH:
@petermetz OK, I've: 
- Removed artillery from depdendency
- Updated README on how to run the stress test manually
- Added logic for closing watchBlock web3js connection too (similar to work done in previous commits)
- Rebased

petermetz (approved):
@outSH All understood, thank you very much for the updates! LGTM

PR #2709, opened 2023-09-20
build(plugin-keychain-x): fix tsc compiler errors - missing V1 suffix

petermetz:
This was caused by the bumping of the references in the openapi.json
files which had been pointing to older versions of the openapi.json
files through git direct links but now we have upgraded to them.

Previously the keychain request/response entities were missing the
V1 suffix and the cross-references were still using these older versions
prior to us making the bumps.

[skip ci]


petermetz:
I'll sneak this change in as part of https://github.com/hyperledger/cacti/pull/2448 to speed things up.

PR #2724, opened 2023-09-28
feat(cactus-core-api): add ISendRequestResultV1<T> for Fujitsu verifier

petermetz:
An interface representing the response object of the send sync and send async
request methods on the connection-chain verifier API.

Also great when trying to help bridge the gap between the API returning
results as `unknown` and the tests needing auto-completion for the properties
to assert for type of data at runtime.

@see {ISocketApiClient<BlockType>}
@see {Verifier<LedgerApiType extends ISocketApiClient<unknown>>}
@see {IVerifier}

refactor(core-api,verifier-client): use unknown instead of any type

1. Changes the ISocketApiClient and the Verifier type definitions so that
instead of `any` they are heavily relying on `unknown` which helps spotting
bugs in the code where type checks should be done at runtime but currently
aren't.
2. It also helps with spotting breaking changes in the future because if
the test cases stop compiling it means you've done a breaking API change,
but they'll always compile if the parameters and the return values are
using `any` for their own types.

3. I've updated a big chunk of the test cases to have explicit type casts
onto the shapes that are needed for them to compile (without changing
any of the runtime behavior).
4. Also made it so that the json2str function of the DriverCommon
code file now accepts unknown instead of object type which makes one of
the linter warnings go away as well.

### DIFF
--- a/examples/cactus-example-discounted-asset-trade/balance-management.ts
+++ b/examples/cactus-example-discounted-asset-trade/balance-management.ts
@@ -7,12 +7,13 @@
 
 import { LPInfoHolder } from "@hyperledger/cactus-cmd-socketio-server";
 import { ConfigUtil } from "@hyperledger/cactus-cmd-socketio-server";
+import { ISendRequestResultV1 } from "@hyperledger/cactus-core-api";
 import {
   VerifierFactory,
   VerifierFactoryConfig,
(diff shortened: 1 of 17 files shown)


jagpreetsinghsasan (approved):
LGTM

izuru0 (approved):

PR #2729, opened 2023-10-02
chore(release): forgot to run codegen before v2.0.0-alpha.2 publish

### LINKED ISSUE #2728: chore(release): forgot to run codegen before v2.0.0-alpha.2 publish
### Description

This means that the generated code comments/declarations still have `alpha.1` in them instead of `alpha.2`.

I've already updated the release checklist document so that we don't forget this next time, but we still need a commit to 
sync it up after the fact. This issue is about making that change happen.

petermetz:
1. Ran the `yarn codegen` command so that all the generated docs/code comments/etc
get updated with the correct version (which changed from alpha.1 to alpha.2)

[skip ci]

Fixes #2728


jagpreetsinghsasan (approved):
LGTM

izuru0 (approved):
LGTM

Issue #2676, opened 2023-09-10
style(cmd-api-server): fix Unexpected any linter W in getOrCreateWebServices()


petermetz:
### Description

![image](https://github.com/hyperledger/cacti/assets/1583231/123bbfa5-c028-4e9e-bfae-c7c775b96f9a)

Yogesh01000100:
Hi! I'm interested in working on this issue. Could you please assign it to me? I'll start working on a solution and create a PR once it's ready. Thanks!

Yogesh01000100:
I got this error while installing the dependency can you specify the exact version so that it resolves the dependency error? 
![image](https://github.com/hyperledger/cacti/assets/90953665/00497ced-52e2-4451-af3e-bc90a0bb5c8f)

Yogesh01000100:
I got another ESPERM error, while configuring command=> npm run configure : 
![image](https://github.com/hyperledger/cacti/assets/90953665/9fe8f231-0deb-40b7-a140-a08202093bdf)

petermetz:
> Is this issue sorted? Because I don't get any 'Unexpected any' linter errors in the latest pull.

@Yogesh01000100 It should still be there. The `lint` script won't catch it by default because of the `--quiet` flag (so that we focus on errors not warnings) so you have to run it like this `yarn eslint '*/*/src/**/*.{js,ts}' --fix`

petermetz:
Yes, assigning now! Thank you!

RafaelAPB:
Until there is an explicit need for this feature, this issue is closed.

PR #2751, opened 2023-10-06
style(cmd-api-server): fix any linter in getOrCreateWebServices()

Yogesh01000100:
## Description:
This pull request addresses issue #2676  by fixing an Unexpected any linter warning in the getOrCreateWebServices() method of the ApiServer class. The warning occurred due to a typecast using (app as any). 

### Changes:
Replaced (app as any) with (app as express.Express)[httpVerbPrometheus as keyof express.Express](httpPathPrometheus, prometheusExporterHandler);

### Importance of keyof:
The use of keyof here is crucial because it ensures that httpVerbPrometheus and httpVerb are treated as the valid key of the express.Express type. This is necessary because HTTP verbs, like GET, POST, PUT etc... should be treated as keys within the express.Express type. By casting httpVerbPrometheus as keyof express.Express, we enforce type safety and prevent potential runtime errors that could occur if an invalid HTTP verb is used. In essence, it ensures that only valid HTTP verbs are accepted, improving the robustness of the code. 

### Benefits:
This change enhances code readability and maintainability using appropriate types and expressions.

### DIFF
--- a/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
+++ b/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
@@ -593,7 +593,7 @@ export class ApiServer {
     const { "/api/v1/api-server/healthcheck": oasPath } = OAS.paths;
     const { http } = oasPath.get["x-hyperledger-cactus"];
     const { path: httpPath, verbLowerCase: httpVerb } = http;
-    (app as any)[httpVerb](httpPath, healthcheckHandler);
+    (app as express.Express)[httpVerb as keyof express.Express](httpPath, healthcheckHandler);
 
     this.wsApi.on("connection", (socket: SocketIoSocket) => {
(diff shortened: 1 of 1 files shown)


jagpreetsinghsasan (approved):
Thankyou for the PR, @Yogesh01000100 
This looks good to me !

petermetz (commented):
@Yogesh01000100 This is fine to fix the linter warnings themselves so LGTM (just commenting instead of explicit approval to avoid accidental pre-quorum merge).

For fixing it with runtime type validation baked in, please look at the 2 PRs I just opened We'll merge your first because the linter fixes here are legit and then my 2 PRs coming after it will extend the fix with the runtime type validation. This is just an FYI in case you have the time & willingness to look into the extended solution.

izuru0 (approved):
LGTM

PR #2818, opened 2023-10-20
build: codegen to call clean on a per pkg basis

### LINKED ISSUE #2736: build: codegen to call clean and do so on a per package basis
### Description

This is a follow-up with a longer term solution to the short term fix that we applied in #2730 for the same problem that the referenced issue is about (which could and will prop up again in the future in a different form for sure)

For the future: The current solution is kind-of lackluster, we have this giant npm script cleaning out a long list of directories and it is separate from the codegen script so the contributors are responsible for running both. A better long term solution would be to make sure that the codegen script calls the clean automatically before it starts generating the new code so that this situation we were in just now is impossible to happen (otherwise it will keep happening for sure)

Historically I've shied away from doing the above because we have some packages that have code generation that is not fully automated (the  rust crates need a lot of customization  that is manual IIRC) but if we do it right, it could be done right for most of the packages.

### Acceptance Criteria
1. The `codegen` script works and it does NOT create any changes in the diff after repeated runs on the same code (assuming you've committed the changes after the first run).
2. The `clean` script is refactored so that it calls the package's own `codegen:clean` script which is responsible for the package's generated code to be cleaned out and also for excluding files that do not need to be cleaned out such as the `openapi-generator-ignore` files
3. `openapi-generator-ignore` files are not deleted

cc: @jagpreetsinghsasan

jagpreetsinghsasan:
Primary Changes
    ---------------
    1. Updated codegen scripts of all affected packages to include a global
       openapi-generator-ignore file
    2. Created script to clean openapi folders generated by codegen for
       typescript-axios and kotlin

    Secondary Changes
    -----------------
    1. Removed duplicated code generated previously when codegen clear wasn't
       enabled
    2. Created openapi-generator-ignore file at project root for global ignore
       rules
    3. Updated package.json of all affected packages, extensions and examples

Fixes #2736

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


openapi-generator-ignore
```diff
@@ -0,0 +1,5 @@
+**/src/main/typescript/generated/openapi/typescript-axios/git_push.sh
+**/src/main/typescript/generated/openapi/typescript-axios/.npmignore
+**/src/main/typescript/generated/openapi/typescript-axios/.gitignore
+**/src/main/kotlin/generated/openapi/kotlin-client/docs/**
+**/src/main/kotlin/generated/openapi/kotlin-client/gradle/**
```
petermetz, line 5: @jagpreetsinghsasan Question: Do we have to specify these paths here? E.g., couldn't it just work if we had the contents of this file defined with the same content as we used to have in the per-package ignore files (the ones that this one will now supersede)?

petermetz (changes requested):
@jagpreetsinghsasan Huge diff, I might need to take multiple passes on reviewing this, apologies in advance. Quick findings on the first pass that I'd like to ask if you could maybe change up:
1. The `tools/clear-openapi-codegen-folders.js` script is not formatted (Prettier/ESLint)
2. Withing the logic of `tools/clear-openapi-codegen-folders.js`:
2.1. There is a brittleness in it in terms of having the directory depth hard-coded in the globs. Instead of the patterns you have right now, I would pull the list of package directories straight from lerna.json and then do something like `src/main/**/generated/` or similar (I'm making this as I go so it most likely need tweaks). My reasoning is that there were already multiple proposals for us to completely revamp the monorepo package directory structure from flat to hierarchical (tree structure) which would mean that the packages would no longer be at a fixed 2 deep from the project root. So if/when that change takes place, the current globs will break because it doesn't start from the package dir globs defined in lerna.json
3. If we no longer use the localized openapi-generator-ignore files (see all of them were reverted back to the default spat out by the generator) then we should just 1) delete them AND 2) ignore them in the centralized openapi-generator-ignore file that we have in the root, so that the package specific ignore files won't even get generated. This (in my opinion) would make it less confusing because there is no ignore file that isn't being used (meaning that the package local ignore files are being ignored but their existence would imply otherwise to begin with)
4. There are merge conflicts and for some reason not just in the yarn.lock file like the GitHub UI is suggesting; when I tried to rebase this to upstream/main it ran into hundreds of files being conflicted for some reason.

jagpreetsinghsasan:
@petermetz all good suggestions!
I will fix/update the code and re-request for review asap

openapi-generator-ignore
jagpreetsinghsasan: @petermetz It works if we have this file content changed to
```
**/git_push.sh
**/.npmignore
**/.gitignore
**/generated/openapi/kotlin-client/docs/**
**/generated/openapi/kotlin-client/gradle/**
```
If we have just `.npmignore` then the generator will not try to generate the root `.npmignore` which is not what this is intended for (a tradeoff for using a global generator-ignore file)
So shall I change it to the above one, or do you have any suggestions?

jagpreetsinghsasan:
> @jagpreetsinghsasan Huge diff, I might need to take multiple passes on reviewing this, apologies in advance. Quick findings on the first pass that I'd like to ask if you could maybe change up:
> 
> 1. The `tools/clear-openapi-codegen-folders.js` script is not formatted (Prettier/ESLint)
> 2. Withing the logic of `tools/clear-openapi-codegen-folders.js`:
>    2.1. There is a brittleness in it in terms of having the directory depth hard-coded in the globs. Instead of the patterns you have right now, I would pull the list of package directories straight from lerna.json and then do something like `src/main/**/generated/` or similar (I'm making this as I go so it most likely need tweaks). My reasoning is that there were already multiple proposals for us to completely revamp the monorepo package directory structure from flat to hierarchical (tree structure) which would mean that the packages would no longer be at a fixed 2 deep from the project root. So if/when that change takes place, the current globs will break because it doesn't start from the package dir globs defined in lerna.json
> 3. If we no longer use the localized openapi-generator-ignore files (see all of them were reverted back to the default spat out by the generator) then we should just 1) delete them AND 2) ignore them in the centralized openapi-generator-ignore file that we have in the root, so that the package specific ignore files won't even get generated. This (in my opinion) would make it less confusing because there is no ignore file that isn't being used (meaning that the package local ignore files are being ignored but their existence would imply otherwise to begin with)
> 4. There are merge conflicts and for some reason not just in the yarn.lock file like the GitHub UI is suggesting; when I tried to rebase this to upstream/main it ran into hundreds of files being conflicted for some reason.

@petermetz 
1. Done
2. I have updated the globs as how other scripts deal with similar paths
3. There is no way we can prevent the generation of per package openapi-generator-ignore files. The latest code update in this PR still includes them for the same reason. If having a post-cleanup script is fine (or any suggestions from your side), I can add that, to manually get rid of these per package generator ignore files (and can use the same script for doing pre and post cleanup just to not have lots of script files)
4. I have updated it, for some reason, something very weird was happening related to merge conflicts in a lot of files during rebase, but I have fixed them

petermetz:
> > @jagpreetsinghsasan Huge diff, I might need to take multiple passes on reviewing this, apologies in advance. Quick findings on the first pass that I'd like to ask if you could maybe change up:
> > 
> > 1. The `tools/clear-openapi-codegen-folders.js` script is not formatted (Prettier/ESLint)
> > 2. Withing the logic of `tools/clear-openapi-codegen-folders.js`:
> >    2.1. There is a brittleness in it in terms of having the directory depth hard-coded in the globs. Instead of the patterns you have right now, I would pull the list of package directories straight from lerna.json and then do something like `src/main/**/generated/` or similar (I'm making this as I go so it most likely need tweaks). My reasoning is that there were already multiple proposals for us to completely revamp the monorepo package directory structure from flat to hierarchical (tree structure) which would mean that the packages would no longer be at a fixed 2 deep from the project root. So if/when that change takes place, the current globs will break because it doesn't start from the package dir globs defined in lerna.json
> > 3. If we no longer use the localized openapi-generator-ignore files (see all of them were reverted back to the default spat out by the generator) then we should just 1) delete them AND 2) ignore them in the centralized openapi-generator-ignore file that we have in the root, so that the package specific ignore files won't even get generated. This (in my opinion) would make it less confusing because there is no ignore file that isn't being used (meaning that the package local ignore files are being ignored but their existence would imply otherwise to begin with)
> > 4. There are merge conflicts and for some reason not just in the yarn.lock file like the GitHub UI is suggesting; when I tried to rebase this to upstream/main it ran into hundreds of files being conflicted for some reason.
> 
> @petermetz
> 
>     1. Done
> 
>     2. I have updated the globs as how other scripts deal with similar paths
> 
>     3. There is no way we can prevent the generation of per package openapi-generator-ignore files. The latest code update in this PR still includes them for the same reason. If having a post-cleanup script is fine (or any suggestions from your side), I can add that, to manually get rid of these per package generator ignore files (and can use the same script for doing pre and post cleanup just to not have lots of script files)
> 
>     4. I have updated it, for some reason, something very weird was happening related to merge conflicts in a lot of files during rebase, but I have fixed them

@jagpreetsinghsasan Sorry for the slow response! Please re-request review next time!

1) and 2): Thank you very much!

> There is no way we can prevent the generation of per package openapi-generator-ignore files. 

What's wrong with ignoring the ignore files in the root ignore file like I just did on my local branch of your PR (see below)?

image

> I have updated it, for some reason, something very weird was happening related to merge conflicts in a lot of files during rebase, but I have fixed them

Thank you for that as well!

petermetz (commented):
@jagpreetsinghsasan Please see my comment above^^

jagpreetsinghsasan:
> @jagpreetsinghsasan Please see my comment above^^

@petermetz , yes I was able to do it this way (It would have worked earlier too when I was trying, its just that the default globby options were not removing any files or folder with dot as the starting character ex `.openapi-generator-ignore`, so even after adding it to the global ignore file, it did nothing. I fixed it with `onlyFiles: false` in the globby options)
Now the local .openapi-generator-files are ignored and code works as intended

petermetz (approved):
@jagpreetsinghsasan Nice, thank you very much for the updates! LGTM

Issue #2807, opened 2023-10-18
fix(ci): tests are failing after recent package bumps


outSH:
@petermetz @izuru0 @jagpreetsinghsasan 

**Describe the bug**

After recent dependency bumps many tests are failing in CI - see https://github.com/hyperledger/cacti/pull/2805. I confirmed some of these errors on main branch (run locally).

TLDR I'd propose to rollback the following commits and investigate how to bump these packages without breaking the CI:
- https://github.com/hyperledger/cacti/commit/71734b053379ff089ad5a2f03598aab14afdfecc
- https://github.com/hyperledger/cacti/commit/61fc7001b1dd0849ab1d9bcab08e2475c695adae

As for now I've identified the following issues:

## iroha-helpers

``` sh
Cannot find module 'grpc' from 'packages/cactus-plugin-ledger-connector-iroha/node_modules/iroha-helpers/lib/proto/endpoint_grpc_pb.js'

    Require stack:
      packages/cactus-plugin-ledger-connector-iroha/node_modules/iroha-helpers/lib/proto/endpoint_grpc_pb.js
      packages/cactus-plugin-ledger-connector-iroha/dist/lib/main/typescript/iroha-transaction-wrapper.js
      packages/cactus-plugin-ledger-connector-iroha/dist/lib/main/typescript/plugin-ledger-connector-iroha.js
      packages/cactus-plugin-ledger-connector-iroha/dist/lib/main/typescript/public-api.js
      packages/cactus-plugin-ledger-connector-iroha/dist/lib/main/typescript/index.js
      packages/cactus-verifier-client/dist/lib/main/typescript/get-validator-api-client.js
      packages/cactus-verifier-client/dist/lib/main/typescript/public-api.js
      packages/cactus-verifier-client/dist/lib/main/typescript/index.js
      packages/cactus-test-plugin-ledger-connector-ethereum/src/test/typescript/integration/api-client/verifier-integration-with-ethereum-connector.test.ts

      21 | } from "iroha-helpers/lib/proto/primitive_pb";
      22 |
    > 23 | import { CommandService_v1Client as CommandService } from "iroha-helpers/lib/proto/endpoint_grpc_pb";
         | ^
      24 | import { QueryService_v1Client as QueryService } from "iroha-helpers/lib/proto/endpoint_grpc_pb";
      25 |
      26 | import commands from "iroha-helpers/lib/commands/index";

[...]

PR #2830, opened 2023-10-24
build(deps): bulk fix CVEs via dependency resolution overrides 2023-10-23

### LINKED ISSUE #2544: build(plugin-ledger-connector-ubiquity): stop using GitHub URL as client version
We need to get rid of this dependency declaration that depends on a GitHub URL:
`"@ubiquity/ubiquity-ts-client-modified": "https://github.com/RafaelAPB/ubiquity-ts-client-mirror.git",`
because it is technical debt and a security concern at the same time.

The package in question: `packages/cactus-plugin-ledger-connector-ubiquity/package.json``
Gitlab repo: https://gitlab.com/Blockdaemon/ubiquity/ubiquity-ts-client/-/releases

- Plan A: Have the maintainers of this package publish an up to date & working
version to npmjs.com so that we can use it from there.
- Plan B: If they publish it to an alternative registry that's fine too we just
need to create the mapping for the scope (as long as they do have an org scope)
- Plan C: Their website appears to have a much newer version of the OpenAPI specification
for their APIs (v3.0.0). As a plan C we could even just generate our own client
and use that as part of the connector. This is not desired because then we would
have to maintain code that we don't really control (we don't write the specs
nor we develop the back-end APIs that they must comply with) but as a last 
resort it is an acceptable solutin if all else had failed.

OpenAPI specs: https://ubiquity.docs.blockdaemon.com/

cc: @RafaelAPB

### LINKED ISSUE #2828: build(deps): bulk fix CVEs via dependency resolution overrides 2023-10-23
### Description

Upgrading vulnerable dependencies have become a game of whack-a-mole: Vulnerabilities that I've submitted pull requests for in the past few weeks are reappearing because new pull requests keep adding back the vulnerable versions of these dependencies to the project build. 
We could have a CI step scanning for these and block the PR's from getting merged but that's extra resources used on the already stretched-thin CI so the simpler route is to start using the central (root package.json file's) resolution overrides more aggressively. Previously I was advocating for this to be a last resort, but seeing in recent weeks how the accidental re-adding of vulnerabilities happen very frequently nowadays (due to project growth which is a good thing in itself) I changed my mind on it and decided that my new recommendation is to lock ourselves away from vulnerable versions of dependencies entirely via the centralized overrides. 

The scope of this task: Submit one PR with resolution overrides.

### Acceptance Criteria
1. All the resolution overrides are defined with a "great or equal" qualifier to ensure that we are not blocking people from upgrading further (e.g., if v0.9.0 is vulnerable, we should say `>=1.0.0` instead of `=1.0.0` because the former is not forcing us to not use v1.1.0 in the future as it comes out)
2. Each resolution override is detailed in the commit message body, links to the github security advisory and the CVE IDs are included
3. Build is still working
4. There is no significant slowdown in the build
5. The tests are passing as well

petermetz:
1. Couldn't get rid of vulnerable versions in a couple of dependencies
because the underlying dependencies have gone ESM only which is a blocker
for us at the moment unfortunately.
2. Swapped out the ubiquity TS client to a version of it that I self
published onto npm after a full renovation of all of its dependencies.

Depends on #2807 (because that one also has a couple of dependency bumps
that are needed to eliminate the vulnerabilities)

Fixes #2828
Fixes #2544

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,16 +78,27 @@
     "init-registries": "npm config set @iroha2:registry=https://nexus.iroha.tech/repository/npm-group/"
   },
   "resolutions": {
-    "ansi-html": ">0.0.8",
+    "@babel/traverse": ">=7.23.2",
+    "ansi-html": ">=0.0.8",
     "axios": ">=0.27.2",
-    "glob-parent": "5.1.2",
+    "engine.io": ">=6.4.2",
+    "get-func-name": ">=2.0.1",
+    "glob-parent": ">=5.1.2",
     "http-cache-semantics": ">=4.1.1",
+    "jsonwebtoken": ">=9.0.0",
     "lodash": ">=4.17.21",
     "minimist": ">=1.2.6",
     "nano": ">=10.0.0",
     "node-forge": ">=1.3.0",
+    "postcss": ">=8.4.31",
     "protobufjs": ">=7.2.5",
-    "underscore": "1.13.2"
+    "semver": ">=7.5.2",
+    "socket.io-parser": ">=4.2.3",
+    "tough-cookie": ">=4.1.3",
+    "xml2js": ">=0.5.0",
+    "yargs-parser": ">=18.1.1",
+    "underscore": ">=1.13.2",
+    "zod": ">=3.22.3"
   },
   "devDependencies": {
     "@commitlint/cli": "17.7.1",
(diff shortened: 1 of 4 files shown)


petermetz:
cc: @RafaelAPB (please see point 2)

jagpreetsinghsasan (approved):
LGTM

jagpreetsinghsasan (approved):
LGTM

izuru0 (approved):
LGTM

PR #2859, opened 2023-11-02
fix(deps): bulk add missing dependencies - 2023-11-02

### LINKED ISSUE #2857: fix(deps): bulk add missing dependencies - 2023-11-02
### Description

On top of the obvious solution of adding the missing dependency to the package.json, we also need to make sure that this can be checked against somehow with tooling and automation.
Up until now I thought that disabling hoisting in yarn completely fixed this issue but it turns out that it's only a partial fix.
The edge case where it still fails is when the root package.json declares a dependency that we then import from a sub-package and the compiler walks up the directory tree to the root, finds the package and is happy. This can happen because the package was in the root node_modules with or without hoisting to begin with.

https://yarnpkg.com/package?name=depcheck

### Acceptance Criteria
1. A one liner or script (preferably a script) that finds these issues
2. The solution to 1) should also be part of the custom-checks
3. It should clearly list which package json is missing which dependencies so that taking action is trivial without further investigation having to be performed
4. In addition to fixing the particular missing dependency in the subject, all other missing dependencies should be fixed as well due to the urgent nature of this problem (packages are being released as broken when this happens)

petermetz:
1. Added missing dependencies everywhere to all the packages
2. The exception to the above is test runners and related dependencies.
You can see the detailed exclusion list in the source code of the script
itself [1]. We should make these configurable later on as well. For
reference, this is what the exclusions were declared as when I ran the
tool in order to then proceed to update the package.json files:

```typescript
ignorePatterns: [
    // files matching these patterns will be ignored
    "sandbox",
    "dist",
    "bower_components",
    "node_modules"
],
ignoreMatches: [
    // ignore dependencies that matches these globs
    "grunt-*",
    "jest-extended",
    "tape-promise",
    "tape",
    "tap",
    "@ionic-native/*"
],
```
3. There were instances of missing dependency usages where we did NOT
have to add the dependencies to package.json files because what we could
do instead is just import types at development time by using the
`import type { .. } from "...";`  syntax of Typescript which means that
the import is disappeared during transpilation completely. So this is why
some source code files were also modified and not strictly just package.json
files.
4. Added a script in the ./tools/custom-check directory to audit the
entire mono-repo for missing NodeJS dependencies.

We've had at least a dozen packages that were missing production
dependency declarations from their package.json files. The usual suspects
here are packages that are contained by the root node_modules folder
which masks the problem at development time (e.g. the compiler won't
let you know about the missing dependencies).

For a future-proof solution we should add a commit hook or other validation
that runs the custom check that I added [2] here to verify that there are no
missing dependencies in the project

[1] `./tools/custom-checks/check-missing-node-deps.ts`
[2] `$ yarn tools:check-missing-node-deps`

Fixes #2857

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


examples/cactus-example-discounted-asset-trade/package.json
```diff
@@ -36,20 +36,25 @@
     "fabric-ca-client": "2.2.19",
     "fabric-network": "2.2.19",
     "http-errors": "1.6.3",
+    "indy-sdk": "1.16.0-dev-1655",
```
outSH, line 39: indy-sdk was removed from package.json on purpose, it is described in this sample readme to install it manually. It was removed because it requries indy-sdk library to be installed on the system, which causes some troubles for CI and users.

petermetz: @outSH Ahhh, I remember now! I knew there was something special about that dependency! I think at that time we were still on Yarn v1 and it did not have this feature that solves this problem very easily in the root package.json by declaring that the native code of indy-sdk should not be built (that's how I got the CI to work in this PR).

This is not a perfect solution either though because people will still need to manually install the native dependencies themselves so we probably just went from confusing people in one way to confusing them in another. I'm not even sure which one is worse. 
If you think we should rather just not have the indy-sdk dependency declared in the package.json at all then I can update the PR with that, no problem, just let me know what you think in light of this new option we have in Yarn v3.

outSH: @petermetz Please choose whichever you like, I'm removing indy SDK in a separate commit anyway :)

petermetz: @outSH Oh, fair enough haha! I'll just leave it as-is for now for the sake of simplicity (trying to save time, things are very busy!)

outSH:
@petermetz Can't give approval on this (or any other PR) because I'm missing permissions to this repo, I think Sato-san mentioned it on discord. Could you look into that in spare moment and give me write access so I can perform the reviews?

petermetz:
> @petermetz Can't give approval on this (or any other PR) because I'm missing permissions to this repo, I think Sato-san mentioned it on discord. Could you look into that in spare moment and give me write access so I can perform the reviews?

@outSH Oops, sorry, we need to add you to the Hyperledger GitHub org and then I can grant you permissions. I can't add you to the GH org myself but I sent in a request to the HL staff to get it done here: https://github.com/hyperledger/governance/issues/177
You should see an email in your inbox soon with an invitation to join the Hyperledger GitHub org and once you accept through the link in there then I can fix the permissions!

petermetz:
@outSH You should have full permissions now on the repo (admin rights)

outSH (approved):
@petermetz Thank you!

LGTM

jagpreetsinghsasan (approved):

PR #2893, opened 2023-11-16
feat(plugin-keychain-memory): add REST API endpoint implementations

petermetz:
**BREAKING CHANGE**: Some of the schema elements that were redundant
got removed from the openapi.json spec of the memory plugin.
The same schemas are now being pulled from the core-api package.
In this sense there is no breaking change because these two
schemas were already equivalent anyway, but the fact is
the openapi.json had schema elements removed from it anyway
so this is something that people need to be aware of as they
upgrade to v2.0.0 in the near future.

1. Added the REST API endpoints to the plugin so that it is accessible over
HTTP for testing/demo purposes as necessary.
2. Also refactored the openapi.json spec file so that it pulls the schema
definitions from the core-api package the same way as the
keychain-memory-wasm plugin is doing already.
3. Migrated the test API client test case to Jest from tape
4. Refactored the `handleRequest` methods of the endpoint classes so that
they perform the error handling with respect to the HTTP status codes being
set correctly (sanitization is pending because I have to send a separate
PR for that to the cactus-core package)

Special thanks to @outSH for the suggestions to improve things!

---

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


jagpreetsinghsasan (approved):

packages/cactus-plugin-keychain-memory/src/main/json/openapi.json
```diff
@@ -11,148 +11,136 @@
...
-            "description": "The key that was used to retrieve the value from the keychain.",
-            "minLength": 1,
-            "maxLength": 1024,
-            "nullable": false
+        },
+        "operationId": "getKeychainEntryV1",
+        "summary": "Retrieves the contents of a keychain entry from the backend.",
+        "parameters": [],
+        "requestBody": {
+          "$ref": "https://raw.githubusercontent.com/hyperledger/cactus/v2.0.0-alpha.2/packages/cactus-core-api/src/main/json/openapi.json#/components/requestBodies/keychain_get_entry_request_body"
```
outSH, line 33: I'm not sure about using URL references here. If we want to introduce and import any new component from `cactus-core-api` (which is possible), we must update URL to a local path of an updated spec, test it, push it, and replace the reference to the new URL (which is a lot of work). Same should be done if we introduce any breaking change (before the next major release), and do this procedure with all packages that link it like this (which will be forgotten probably). This will be more problematic if we introduce more specs like this.

On the other hand I see that using relative path here would render openapi.json useless on it's own (since user would need whole repo to make any use of it), and hardcoding project structure is something we want to avoid.

I was thinking about solution to this some time ago and my rough idea was to add a symlink of the dependent openapi.json in the location of parent openapi.json, reference it as local file (like `$ref: 'cactus-core-api.json#/components/requestBodies/keychain_get_entry_request_body'`), and then copy all openapi.json files to `dist` during build. The downside is that symlink is still bound to project structure, but on the other hand detecting broken links is easier than detecting broken paths in files.

In the end I was just copy-pasting relevant structures to keep things simple (KISS over DRY I suppose). Please consider other options, none of them are perfect but some may fit better than the other :)

packages/cactus-plugin-keychain-memory/src/test/typescript/unit/plugin-keychain-memory-api-client.test.ts
```diff
@@ -0,0 +1,138 @@
+import test, { Test } from "tape-promise/tape";
```
outSH, line 1: Can this test be written in `jest` instead of `tape`? Also tape tests must be listed in `jest.config` (and probably in `.taprc` as well)

outSH (changes requested):
Added some comments

outSH (changes requested):

packages/cactus-plugin-keychain-memory/src/main/json/openapi.json
petermetz: @outSH Agreed on all of those concerns! 
1. I'll request that we leave it as URLs in this PR and start a brand new PR that refactors the whole project to the new way.
2. Regarding the new way: Agree with KISS > DRY! A slight adjustment idea: I'm paranoid about symlinks and it still doesn't completely address the self-containment problem (when a package is installed alone in another project) so I would suggest to just take it all the way and vendor the openapi.json spec files across packages as needed but with explicit git revisions in their filenames so that we know exactly which ones they are. 
3. An example of my suggestion in `2)` would be in the context of this PR is to have `src/main/json/openapi.json` as usual but also have `src/main/json/vendor/openapi-v2.0.0-alpha.2-cactus-core-api.json` (the exact naming convention is not that important as long as the file path and the name immediately makes it obvious that this is a vendored copy not something that should be locally edited)
4. With the example in `3)` we would avoid the problems with symlinks (macOS and Windows barf on them every now and then) but also satisfy all the requirements that you mentioned above.
5. When we upgrade to a newer version of the core-api specs, we just vendor the new file as well and this will of course feel very repetitive but what I just realized (inspired by your idea, so thank you very much for that) is that we are monorepo and so we need to think as if each package had it's own git repository and in a package architecture like that it would be pretty reasonable to just vendor the pinned openapi.json spec files. It just looks a bit silly to copy the files across directories in the same project because we are a monorepo, but the self-containment principle makes it a rational choice IMO.

I'm including a list of reasons why we originally went with the URLs (this is old informatio that predates the enlightment that I had just reached as I was reading your symlink proposal). The reason for me putting these here is so that future generations can delve into the full decision process and (hopefully) avoid some of the pitfalls. In a way this is just to establish the institutional memory of some sort. 

---

The old list of concerns with the other way of doing it (relative local paths):
1. Breaking changes get masked implicitly (I change spec A which uses spec B, I have compilation errors, so I change spec B as well to accomodate this and then to the reviewers it's not immediately obvious that I'm making a breaking change in spec A)
2. If a package is installed by itself in a project as a dependency and someone tries to use the spec file of it for generating their own API clients (because let's say they want a C# client or one of other dozens of languages that is supported by the generator) then they'll fail because the openapi.json is no longer valid on it's own (violates the self-containment principle) so they have to come back to the upstream repo, file a bug, then realize the problem, etc. (or hopefully find the FAQ answering their question).
3. Local relative paths are a flavor of auto-upgrades (which I'm always a on a crusade against) because if spec A is using spec B and then I change spec B now I implicitly upgraded spec B onto the new version which might be completely harmless but it might also cause a bunch of other issues. In this regard I also have to state my usual disclosure: I know that it's much less convenient to pin the versions (which we can achieve with the URLs).

packages/cactus-plugin-keychain-memory/src/main/typescript/web-services/delete-keychain-entry-endpoint-v1.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.log.debug(reqTag);
+    try {
+      const { key } = req.body as DeleteKeychainEntryRequestV1;
+      const resBody = await this.options.plugin.delete(key);
+      res.json(resBody);
+    } catch (ex) {
+      this.log.error(`Crash while serving ${reqTag}`, ex);
+      res.status(500).json({
+        message: "Internal Server Error",
+        error: ex?.stack || ex?.message,
```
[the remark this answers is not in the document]
outSH: @petermetz Of course sounds great!

packages/cactus-plugin-keychain-memory/src/main/json/openapi.json
outSH: @petermetz Sure, copying of files works for me as well :)

outSH (approved):

PR #2907, opened 2023-11-25
fix(weaver): usage of weak PRNG issue

### LINKED ISSUE #2765: fix(weaver): usage of weak PRNG
## Description
Static source code assessment has picked up a potential vulnerability regarding use of weak pseudo RNG. Use of secure pseudo random number generators rather than using basic random methods can fix the above vulnerability

## The report from which the above information was summarized
Risk Rating: Low
Category: Security Misconfiguration

### Description
The application uses a weak method of generating pseudo-random values, such that other numbers could be determined from a relatively small sample size. Since the pseudo-random number generator used is designed for statistically uniform distribution of values, it is approximately deterministic. Thus, after collecting a few generated values, it would be possible for an attacker to calculate past or future values. These values are then used as parameters in a cryptographic function. If an attacker were to obtain the original value from the weak PRNG, they may be able to predict or crack values produced by the cryptographic function.

### Impact
Depending on what this random value is used for, an attacker with knowledge of the PRNG used would be able to predict the next values generated, or previously generated values, based on sources often used to derive certain randomness; however, while they may seem random, large statistical samples would demonstrate that they are insufficiently random, producing a much smaller space of possible "random" values than a truly random sample would. In the context of cryptography, this could enable an attacker to derive or guess the value, and thus crack the cryptographic mechanism or its outputs.

### Remediation Recommendation
• Always use a cryptographically secure pseudo-random number generator, instead of basic random methods, particularly when dealing with a security context such as cryptographic functions
• Use the cryptorandom generator that is built-in to your language or platform, and ensure it is securely seeded. Do not seed the generator with a weak, non-random seed. (In most cases, the default is securely random). 
• Ensure you use a long enough random value, thus making brute-force attacks unfeasible.

### Affected files (path - line number)

weaver/core/network/fabric-interop-cc/contracts/interop/certificate_utils.go - 304
weaver/sdks/corda/src/main/kotlin/org/hyperledger/cacti/weaver/sdk/corda/HashFunctions.kt - 46
weaver/sdks/fabric/interoperation-node-sdk/src/eciesCrypto.js - 177

## Snapshot of the sourcecode at the time of scan

image
image
image

---

Source: `APP PE Hyperledger Cacti v2.0.0 - Static Application Assessment Report.odt`

cc: @takeutak @izuru0 @outSH @petermetz

KaganCanSit:
The Logic Behind the Problem
When RNG (Random Number Generator) values are not received through a hardware TRNG, seed values apply a certain pattern. (It takes a seed value such as a mathematical formula or time.) In response to this situation, there are various secure random classes to increase security.

Solution
Changes have been made to get random values using safe randomness instead of mathematical randomness. This increases the complexity of the pattern, making it difficult to discover even if data is listened to for long periods of time.

The changes that have been made;
- In the "certificate_utils.go" file, the random value was taken from the math class (mrand "math/rand") and used. By taking this random value from the secure random class, we obtain a more reliable random value. I added HmacGenerate and generateSecureRandomKey functions for readability and ease of use. If you want to generate a key again, the generateSecureRandomKey function, which uses secure random, can be used.

- In "HashFunctions.kt", kotlin.random.Random class has been replaced with the more reliable java.security.SecureRandom class.

- The reason for the change in "eciesCrypto.js" is that the length of aes-128-ctr is not considered reliable by various standards. For this reason, I preferred the more reliable 256 length.

Fixes #2765
------- 
**Pull Request Requirements**
- [☑] Rebased onto `upstream/main` branch and squashed into single commit to help maintainers review it more efficient and to avoid spaghetti git commit graphs that obfuscate which commit did exactly what change, when and, why.

- [☑] Have git sign off at the end of commit message to avoid being marked red. You can add `-s` flag when using `git commit` command. You may refer to this [link](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) for more information.
- [  ] Follow the Commit Linting specification. You may refer to this [link](https://www.conventionalcommits.org/en/v1.0.0-beta.4/#specification) for more information. 

**Character Limit**
- [☑] Pull Request Title and Commit Subject must not exceed 72 characters (including spaces and special characters).

**A Must Read for Beginners**

### DIFF
--- a/weaver/core/network/fabric-interop-cc/contracts/interop/certificate_utils.go
+++ b/weaver/core/network/fabric-interop-cc/contracts/interop/certificate_utils.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"hash"
 	"math/big"
-	mrand "math/rand"
 	"time"
 
 	"golang.org/x/crypto/ed25519"
@@ -207,7 +206,7 @@ func ecdsaVerify(verKey *ecdsa.PublicKey, msgHash, signature []byte) error {
 	return nil
 }
 
-//Validate Ed25519 signature
+// Validate Ed25519 signature
 func verifyEd25519Signature(pubKey []byte, hashedMessage []byte, signature []byte) error {
 
 	result := ed25519.Verify(pubKey, hashedMessage, signature)
@@ -297,12 +296,31 @@ func encryptWithEd25519PublicKey(message []byte, pubKey []byte) ([]byte, error) 
 	return []byte(""), nil
 }
 
+func generateSecureRandomKey(length int) ([]byte, error) {
+	key := make([]byte, length)
+	_, err := rand.Read(key)
+	if err != nil {
+		return nil, err
+	}
+	return key, nil
+}
+
+func generateHMAC(data, key []byte) ([]byte, error) {
+	hmacHash := hmac.New(sha256.New, key)
+	_, err := hmacHash.Write(data)
+	if err != nil {
+		return nil, err
+	}
+	return hmacHash.Sum(nil), nil
+}
+
 func generateConfidentialInteropPayloadAndHash(message []byte, cert string) ([]byte, error) {
 	// Generate a 16-byte random key for the HMAC
-	hashKey := make([]byte, 16)
-	for i := 0; i < 16 ; i++ {
-		hashKey[i] = byte(mrand.Intn(255))
+	hashKey, err := generateSecureRandomKey(16)
+	if err != nil {
+		return []byte(""), err
 	}
+
 	confidentialPayloadContents := common.ConfidentialPayloadContents{
 		Payload: message,
 		Random: hashKey,
@@ -311,22 +329,26 @@ func generateConfidentialInteropPayloadAndHash(message []byte, cert string) ([]b
 	if err != nil {
 		return []byte(""), err
 	}
+
 	x509Cert, err := parseCert(cert)
 	if err != nil {
 		return []byte(""), err
 	}
+
 	encryptedPayload, err := encryptWithCert(confidentialPayloadContentsBytes, x509Cert)
 	if err != nil {
 		return []byte(""), err
 	}
 
-	payloadHMAC := hmac.New(sha256.New, hashKey)
-	payloadHMAC.Write(message)
-	payloadHMACBytes := payloadHMAC.Sum(nil)
+	payloadHMAC, err := generateHMAC(message, hashKey)
+	if err != nil {
+		return []byte(""), err
+	}
+
 	confidentialPayload := common.ConfidentialPayload{
 		EncryptedPayload: encryptedPayload,
 		HashType: common.ConfidentialPayload_HMAC,
-		Hash: payloadHMACBytes,
+		Hash: payloadHMAC,
 	}
 	confidentialPayloadBytes, err := proto.Marshal(&confidentialPayload)
 	if err != nil {
--- a/weaver/sdks/corda/src/main/kotlin/org/hyperledger/cacti/weaver/sdk/corda/HashFunctions.kt
+++ b/weaver/sdks/corda/src/main/kotlin/org/hyperledger/cacti/weaver/sdk/corda/HashFunctions.kt
@@ -9,7 +9,7 @@ package org.hyperledger.cacti.weaver.sdk.corda;
 import java.util.Base64
 import net.corda.core.utilities.OpaqueBytes
 import net.corda.core.crypto.sha256
-import kotlin.random.Random
+import java.security.SecureRandom
 import org.hyperledger.cacti.weaver.protos.common.asset_locks.AssetLocks.HashMechanism
 import org.hyperledger.cacti.weaver.imodule.corda.states.sha512
 
@@ -42,8 +42,9 @@ class HashFunctions {
         
         override fun generateRandomPreimage(length: Int)
         {
-            val bytes = ByteArray(length)
-            Random.nextBytes(bytes)
+            val secureRandom = SecureRandom.getInstanceStrong();
+            val bytes = ByteArray(length);
+            secureRandom.nextBytes(bytes);
             this.setPreimage(Base64.getEncoder().encodeToString(bytes));
         }
         override fun setPreimage(preImage: String) {
--- a/weaver/sdks/fabric/interoperation-node-sdk/src/eciesCrypto.js
+++ b/weaver/sdks/fabric/interoperation-node-sdk/src/eciesCrypto.js
@@ -161,7 +161,7 @@ function eciesEncryptMessage(recipientPublicKey, msg, options) {
     const hKm = bitsToBytes(hmacKeyHash.finalize());
 
     const iv = crypto.randomBytes(IVLength);
-    const cipher = crypto.createCipheriv("aes-128-ctr", Buffer.from(aesKey), iv);
+    const cipher = crypto.createCipheriv("aes-256-ctr", Buffer.from(aesKey), iv);
     const encryptedBytes = cipher.update(msg);
     const EM = Buffer.concat([iv, encryptedBytes]);
     const D = hmac(hKm, EM, options);


KaganCanSit:
Hello,
The current problem and locations are well defined.(#2765) I'm a C++ developer myself. But I wanted to try myself to solve this problem. Since I am not an expert in Go, I made changes using various sources. However, I do not have time to create and execute the project as a whole. I examined the files one by one, taking into account their compilation status.

If there is an easy way I can test it or if we can create the environment in a short time, I can make additional arrangements. Please check the current status when you are available.

I'm just getting into the Open Source world. I can do my best as long as I have the opportunity.
See you later.

petermetz (approved):
@KaganCanSit Thank you very much for the detailed explanation and the context! Welcome on-board to our list of contributors!

If you ever have any more questions about contributing, we have daily pair programming call [1] that runs at 10 AM Pacific Time on most weekdays on our discord channel.

To answer your specific question as well and also give some tips, here's what I just did to figure out how to test the changes you made to `certificate_utils.go`

1. I did a full text search in VSCode to `go build` which gave me a long list of hits
2. I found the first file that was a match to my search that also had the right path [2] as in, the go build runs in the directory where you changed the code.
3. From then on, I saw that the `.github/workflows/test_weaver-go.yaml` contains what I need in terms of reproducing the test cases as they are executed on the CI
4. So what I did then is checked out your PR's branch in git and then ran these commands (basically just replaying what the github action in the .yml does)
5. `cd weaver/core/network/fabric-interop-cc/contracts/interop`
6. `go build -v ./...`
7. `go test -v ./...`
8. The output of the test indicated that all the tests passed.
9. Based on all the above, I think this is looking good.

[1] https://wiki.hyperledger.org/display/cactus/Daily+Pair+Programming+Calls
[2] image

petermetz (changes requested):
@KaganCanSit Sorry, a few more things in terms of administrative details:
1. Please make sure to explain in the commit message thoroughly the what/why/how (weak encryption as the root cause is kinda self explanatory but the specific solution could be explained a little further so that everyone is on the same page about it for sure)
2. Please use the `Fixes #???` syntax in your commit message to indicate the issue number that you are fixing. It has to be exactly this way because then the bots pick it up and auto-close the issue that you are fixing the moment the pull request is merged (more automation makes it easier for the maintainers to review a higher number of pull requests and keeps everyone a little happier in general)

petermetz (approved):
@KaganCanSit No worries. LGTM!

VRamakrishna:
@KaganCanSit One of the changes made in this PR was incorrect, causing the code to fail. The unit test for that module would have caught the error but was disabled.

Can you please work on a fix ASAP? I've added appropriate comments against the offending change.

@petermetz @sandeepnRES FYI.

VRamakrishna:
@KaganCanSit 

> The reason for the change in "eciesCrypto.js" is that the length of aes-128-ctr is not considered reliable by various standards

Can you point to a source for this? I need to understand the implications. Per my understanding, 128 bits for AES is still highly secure. (But my knowledge could be outdated.)

VRamakrishna:
I checked with a cryptography expert and he informed me that 128-bit AES is secure enough for classical computers but is quantum-unsafe. To be quantum-safe, we do need to upgrade to 256 bits.

KaganCanSit:
Hello @VRamakrishna,

First of all, I would like to help you and learn something new as soon as possible. Since this is my first contribution to this project, I may have missed some parts.

I would like to start by explaining as best I remember why I changed the AES-128 length to AES-256. When I examined the content of the opened issue (https://github.com/hyperledger/cacti/issues/2765), I saw that mostly the random seed values were not from reliable sources. Basically, the first two parts mentioned were like this. However, the "crypto" class was used in the "weaver/sdks/fabric/interoperation-node-sdk/src/eciesCrypto.js" file and was specified as weak. When I examined various sources, I thought that the AES-128-CTR key was mentioned as the part that needed to be replaced. This was the first idea that led me to increasing the key length.

I took a look at various contents. Basically, in many sources, the length of 128 is considered sufficient, as you said. Including in banking. (https://www.pcisecuritystandards.org/glossary/strong-cryptography/)

In other cases, there is information that it was increased as a precaution against Quantom computers and military content. They think it would be better to use 256 to preserve data over time.(https://crypto.stackexchange.com/questions/76738/has-aes-128-been-complete-broken/76746#76746,
https://web.archive.org/web/20160306104007/http://research.microsoft.com/en-us/projects/cryptanaliz/aesbc.pdf) It is important how long the data will remain useful. Essentially, the situation is that the key takes longer to break. Here it may be better to determine how long the key will serve and its security level. Of course, I might have misunderstood the error at first glance.

Since I'm a C++ developer, I was a little hesitant about this pr. For testing I relied on the build states and tests on the path @petermetz outlined for me.


> @KaganCanSit Thank you very much for the detailed explanation and the context! Welcome on-board to our list of contributors!
> 
> If you ever have any more questions about contributing, we have daily pair programming call [1] that runs at 10 AM Pacific Time on most weekdays on our discord channel.
> 
> To answer your specific question as well and also give some tips, here's what I just did to figure out how to test the changes you made to `certificate_utils.go`
> 
> 1. I did a full text search in VSCode to `go build` which gave me a long list of hits
> 2. I found the first file that was a match to my search that also had the right path [2] as in, the go build runs in the directory where you changed the code.
> 3. From then on, I saw that the `.github/workflows/test_weaver-go.yaml` contains what I need in terms of reproducing the test cases as they are executed on the CI
> 4. So what I did then is checked out your PR's branch in git and then ran these commands (basically just replaying what the github action in the .yml does)
> 5. `cd weaver/core/network/fabric-interop-cc/contracts/interop`
> 6. `go build -v ./...`
> 7. `go test -v ./...`
> 8. The output of the test indicated that all the tests passed.
> 9. Based on all the above, I think this is looking good.
> 
> [1] https://wiki.hyperledger.org/display/cactus/Daily+Pair+Programming+Calls [2] image


It's been a busy week, but I can use my weekend to try to improve. I would be very happy if you leave various instructions and little tips. Please share your knowledge with me.

Have a nice day.

petermetz:
> @KaganCanSit One of the changes made in this PR was incorrect, causing the code to fail. The unit test for that module would have caught the error but was disabled.
> 
> Can you please work on a fix ASAP? I've added appropriate comments against the offending change.
> 
> @petermetz @sandeepnRES FYI.

@VRamakrishna Sorry, this is totally my fault! I still haven't re-enabled requiring the tests because the optimizations on the CI that would bring down the resource usage to acceptable levels are still pending. I'll work on tightening back up the CI because this is definitely a wake-up call in that sense.

VRamakrishna:
@KaganCanSit Thanks for your explanation. Really appreciate your enterprise in finding potential vulnerabilities in the code base and trying to fix them. My comments were not meant as criticism of your effort and initiative but rather the completion of the task. The main problem was that the relevant test was disabled. If that wasn't the case, you'd have been alerted to the fact that more changes were needed in the code after your 128-->256 bit modification.

The right steps to test the module you touched are as follows:
- `cd weaver/common/protos-js`
- 'make build` (_Note_: this builds data structures the SDK library depends on)
- `cd weaver/sdks/fabric/interoperation-node-sdk`
- 'make build-local`
- 'npm run test' (_Note_: All tests should pass. Code coverage results can be ignored for now.)

The above steps are coded in the GitHub Action `unit_test_weaver_node_sdk_local` (see https://github.com/hyperledger/cacti/blob/main/.github/workflows/test_weaver-node-pkgs.yaml#L25).

I think the change to AES-256-CTR does need to be done (based on long-term security concerns), but you can take your time with it. For now, to avoid inconsistencies, I'm going to submit a PR reverting just that change (the rest of your changes in this PR looked good to me.)

VRamakrishna:
> > @KaganCanSit One of the changes made in this PR was incorrect, causing the code to fail. The unit test for that module would have caught the error but was disabled.
> > Can you please work on a fix ASAP? I've added appropriate comments against the offending change.
> > @petermetz @sandeepnRES FYI.
> 
> @VRamakrishna Sorry, this is totally my fault! I still haven't re-enabled requiring the tests because the optimizations on the CI that would bring down the resource usage to acceptable levels are still pending. I'll work on tightening back up the CI because this is definitely a wake-up call in that sense.

@petermetz Don't worry about it. I understood the rationale for disabling many of the tests earlier, but perhaps it's time to re-enable the mandatory unit tests now so problems like these don't slip through the cracks.

VRamakrishna:
@KaganCanSit @petermetz I opened a PR which should fix this issue by making the key length configurable: https://github.com/hyperledger/cacti/pull/2953.

KaganCanSit:
> [VRamakrishna](/VRamakrishna)

@VRamakrishna thank you for sharing your knowledge with me. I know you don't criticize me and you guide me. As an engineer candidate at the beginning of my journey, I think I will learn a lot from the open source world that will broaden my horizons. I respect your knowledge and effort. I will try to research and make a correction for my mistake here as soon as possible.

PR #2915, opened 2023-12-05
fix(cactus-common): coerceUnknownToError() now uses HTML sanitize

petermetz:
1. This is a security fix so that the exception serialization does not
accidentally XSS anybody who is looking at crash logs through some
admin GUI that is designed to show logs that are considered trusted.

Related discussion about `1)` can be seen at this other pull request:
https://github.com/hyperledger/cacti/pull/2893

---

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-common/src/main/typescript/exception/coerce-unknown-to-error.ts
+++ b/packages/cactus-common/src/main/typescript/exception/coerce-unknown-to-error.ts
@@ -1,4 +1,5 @@
 import stringify from "fast-safe-stringify";
+import sanitizeHtml from "sanitize-html";
 import { ErrorFromUnknownThrowable } from "./error-from-unknown-throwable";
 import { ErrorFromSymbol } from "./error-from-symbol";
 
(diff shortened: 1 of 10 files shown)


packages/cactus-common/src/main/typescript/exception/coerce-unknown-to-error.ts
```diff
@@ -26,10 +27,11 @@ export function coerceUnknownToError(x: unknown): Error {
   } else if (x instanceof Error) {
     return x;
```
outSH, line 28: @petermetz I think we should also sanitize regular exceptions as well (in case dependent library doesn't do so and throws errors with user input in it). 

And a side note - would it be possible for `handleRestEndpointException` to use `safeStringifyException` on the message being returned, so that axios errors are formatted correctly and stack is returned for exceptions? These were very useful in debugging process and would love to have them when I switch to using `handleRestEndpointException`

outSH (changes requested):

packages/cactus-common/src/main/typescript/exception/coerce-unknown-to-error.ts
petermetz: @outSH Agreed! Sorry for the slow response, I went down multiple rabbit-holes with this. I was trying to refactor it all in a way that the double quotes in the JSON serialized exceptions doesn't have to be escaped by the sanitizer, but ultimately gave up on it and now it just does what you asked and I have added a test case too to verify that malicious payloads in thrown `Error`s or `string`s are sanitized correctly.

The other refactor I did is that now the `handleRestEndpointException` is async which I needed because of the ESM/CJS drama we have going on (a lib I'm using to manage HTTP codes/status messages is forcing my head with a dynamic import).
I figured this was okay because we are still pre-2.0.0 plus the method hasn't been widely adapted yet (I've been dog-fooding it in a few endpoints to discover issues like this)

outSH (approved):

jagpreetsinghsasan (approved):
LGTM

PR #2952, opened 2023-12-18
feat(connector): remove keychain dependency

### LINKED ISSUE #963: feat(connector): remove hard dependency on keychain
### Description
As a user of Cactus I want to be able to omit using the keychain if I want to so that I can pick and chose what complexity I take on when I decide to use the framework for my project.

https://chat.hyperledger.org/channel/cactus?msg=WgTabjQho3oFGskSa

### Acceptance Criteria
1. Test coverage for both with and without keychain scenarios
2. CI suite passing
3. Avoid depending on shared mutable state, the API server must be horizontally scalable. No file system usage either.

cc: @takeutak @sfuji822 @hartm @jonathan-m-hamilton @AzaharaC @jordigiam @kikoncuo @petermetz @arnab-roy @jagpreetsinghsasan @TonyRowntree @travis-payne

### LINKED ISSUE #1163: feat(connector-besu): remove hard dependency on keychain

jagpreetsinghsasan:
## Commit to be reviewed
------------------------------------

feat(connector): remove keychain dependency   

    Primary Changes
    ---------------
    1. Updated besu, ethereum, quorum and xdai connectors
    to remove hard dependencies on keychain

    2. Updated openapi.json file of the above mentioned connectors
       to include the new no-keychain endpoints
    3. Generated code and updated related web-services for the same

Fixes #963
Fixes #1163

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -912,6 +913,77 @@ export class PluginLedgerConnectorBesu
...
+        timeoutMs: req.timeoutMs || 60000,
+      },
+      web3SigningCredential,
+      privateTransactionConfig: req.privateTransactionConfig,
+    });
+
+    const { transactionReceipt: receipt } = runTxResponse;
+    const { status, contractAddress } = receipt;
+
+    if (status && contractAddress) {
```
petermetz, line 960: @jagpreetsinghsasan We could check for these being falsy and if they are, then we just throw within the if-condition. This way the nested if conditions that you have here can be flattened into just a depth of 1 instead of 2 which makes the code easier to read, debug, refactor, etc.

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/plugin-ledger-connector-besu.ts
```diff
@@ -912,6 +913,77 @@ export class PluginLedgerConnectorBesu
...
+        this.log.debug("Contract JSON: \n%o", JSON.stringify(contractJSON));
+        const contract = new this.web3.eth.Contract(
+          contractJSON.abi,
+          contractAddress,
+        );
+        this.contracts[contractName] = contract;
+        const network = { [networkId]: networkInfo };
+        contractJSON.networks = network;
+      } else {
+        throw new Error(
```
petermetz, line 973: @jagpreetsinghsasan I've got a new requirement here in terms of error handling, sorry for not saying this earlier. I wrote a little essay about it in another PR review and it's quite long and I'm tryin to save time so I'll just copy paste the relevant parts of that one here but if it's not immediately obvious how it applies then please let me know and I'll just go the whole way and re-write the text more clearly.

---

Exception throws and re-throws are done with the appropriate utilities which are either 
* `createRuntimeErrorWithCause()` from the file at `packages/cactus-common/src/main/typescript/exception/create-runtime-error-with-cause.ts` or
* The `http-errors` library's function collection accessed via `createHttpError` .  For HTTP REST API endpoint related error handling where your exceptions communicate with the web layer, have a look at the patterns established in this diff 
https://github.com/hyperledger/cacti/pull/2869
<br>In short, instead of this:
  ```typescript
  // This is wrong - the HTTP response's status code will be
  // set to 500 but it should be 400 because it was user error
  // not a bug in the server's production code.
  throw new Error(
  `${fnTag} Cannot deploy contract without keychainId and the contractName`
	);
	```
	Do this:
	```typescript
	if (!keychainHasContract) {
	// This is better, because the HTTP response will 
	// contain a status code of 400 instead of 500
	const errorMessage =
		`${fnTag} Cannot create an instance of the contract instance because` +
		`the contractName in the request does not exist on the keychain`;
	throw new createHttpError[400](errorMessage);
	}
	```

petermetz (changes requested):
@jagpreetsinghsasan Mostly LGTM, I Just left a few comments in the besu connector's implementation and I have one more question: Is there an endpoint implementation for the new method included or is it just the connector method for now and you are planning on adding the REST endpoint later in another PR (which is also reasonable because the PR is already pretty big)

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
```diff
@@ -953,24 +961,25 @@ export class PluginLedgerConnectorEthereum
...
-      });
-    } else if (isContractKeychainDefinition(req.contract)) {
-      return this.deployContractFromKeychain(req, req.contract);
-    } else {
-      // Exhaustive check
-      const unknownContract: never = req.contract;
-      throw new Error(
-        `Unknown contract definition provided: ${unknownContract}`,
-      );
+  public async deployContractNoKeychain(
```
outSH, line 967: I think I'd prefer to keep a single deploy endpoint for both scenarios with / without a keychain.
- This is roughly the same for the user (no keychain - don't use keychain option in endpoints, if you do the error will be thrown on runtime. This is similar to error thrown when you use regular deploy instead of deployContractNoKeychain)
- This will spread to other methods that may use keychain in the future, and will require both method flawors in order to work (invokeNoKeychain etc..). Also similar strategy must be assumed in the future, if we support other optional plugins in methods, like DB.
- We could use default in-memory keychain plugin in case user doesn't provide one, this would simplify implementations in the future but I'm not sure if this is valid default security-wise.
- These comments are mostly for ethereum connector.

@petermetz You are the creator of this issue, any thoughts on it?

outSH (changes requested):

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
petermetz: @outSH TLDR: I recommend re-using code in the form of smaller functions under the hood but keeping the endpoints separate so that the validation logic is straight forward. The guiding principle being that we want to reduce if conditions because they are complexity. (I was talking to @jagpreetsinghsasan about the same thing the other day).

The obvious downside is the strong potential for code repetition and some of it will be unavoidable but I believe that with the right structuring of the code in smaller functions we can avoid 95% of that and the remaining 5% is just worth it as a trade-off.

Of course if we find out while writing the actual code that the 5/95 ratio is more like 50/50 then we should reevaluate our choices but I'm hopeful that we can collaborate and get to the (currently just imagined) 5/95 ratio.

## Code Re-use Example

This is what I mean roughly speaking as a vision (it is very similar to my other message on the connector inheritance thread):

```typescript
class DeployEndpointX {
   handleRequest() {
    // ...
   }
}


class DeployEndpointY {
   handleRequest() {
    // ...
   }
}

class Connector {

   deployX() {
      this.smallerReusableFunction1();
      this.smallerReusableFunction2();
   }

   // Mostly the same as deployX() but
   // it has some more logic.
   // No need for if conditions though
   // which is nice
   deployY() {
      this.smallerReusableFunction1();
      this.smallerReusableFunction2();
      this.smallerReusableFunction3();
   }

   smallerReusableFunction1() {

   }

   smallerReusableFunction2() {
      
   }

   smallerReusableFunction3() {
      
   }

   smallerReusableFunction4() {
      
   }
}
```


## Specifically on validation

The other big benefit of separate endpoints is that the validation can be locked down much tighter.
If our request definitions in the OpenAPI specs are designed to handle multiple different use-cases (such as keychain & no-keychain) then we are either
1. Forced to mark some of the properties non-mandatory just because we don't know at design time whether it's mandatory for that code path or not (results in more if conditions within the implementation just to check this as well)
2. Use inheritance (and still handle that with switch cases or if conditions too) which is poorly supported by a lot of the code templates that the generator has (we currently have probably 80% of the code generation related issues due to class inheritance not being fully supported either by the generator itself or by the template that we'd like to use)

outSH: @petermetz

For me, endpoints are supposed to define single business action without technical details on how it's implemented. I'm still afraid this will burst exponentially into more specialty methods like that if we add more features (another auth method, another storage for contracts, etc..). As for openapi, right now multiple options can be supported with `oneOf` which works in TS and is informative for the user, I don't think we should mess our API just because the tooling requires us to do so.

jagpreetsinghsasan:
> @jagpreetsinghsasan Mostly LGTM, I Just left a few comments in the besu connector's implementation and I have one more question: Is there an endpoint implementation for the new method included or is it just the connector method for now and you are planning on adding the REST endpoint later in another PR (which is also reasonable because the PR is already pretty big)

@petermetz I have added the endpoint for deployContractNoKeychain (do we need to do it for all the endpoints or just showcasing it for one endpoint works?)

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
petermetz: > @petermetz
> 
> For me, endpoints are supposed to define single business action without technical details on how it's implemented. I'm still afraid this will burst exponentially into more specialty methods like that if we add more features (another auth method, another storage for contracts, etc..). As for openapi, right now multiple options can be supported with `oneOf` which works in TS and is informative for the user, I don't think we should mess our API just because the tooling requires us to do so.

@outSH Sorry for the very slow response. I had think long and deep about this one. Based on the comments so far I think we are both pretty set on our viewpoints which happen to be the exact opposite of each other. How about the following plan:
1. You'll be the principal maintainer of `cactus-plugin-ledger-connector-ethereum` and consequently we leave that out of the keychain dependency refactor because that's how you prefer it.
2. I'll be the principal maintainer of `cactus-plugin-ledger-connector-besu` and we'll make the keychain optional there because that's just how I'd prefer it (not exactly with the current diff of this PR because I see some things I would change still but me and Jagpreet we'll push through that together)
3. The xdai and quorum seem pretty dead so TBH we can just remove them IMO and then it's that much less code to maintain / have this conversation about.

Does this seem like a reasonable compromise?

petermetz:
> > @jagpreetsinghsasan Mostly LGTM, I Just left a few comments in the besu connector's implementation and I have one more question: Is there an endpoint implementation for the new method included or is it just the connector method for now and you are planning on adding the REST endpoint later in another PR (which is also reasonable because the PR is already pretty big)
> 
> @petermetz I have added the endpoint for deployContractNoKeychain (do we need to do it for all the endpoints or just showcasing it for one endpoint works?)

@jagpreetsinghsasan Sorry for the slow response, this one is a toughie. Let's keep it as small as possible initially. It's hard to review the bigger diffs. So just a single endpoint in a single package. Let's not do it all together. Also, if @outSH agrees to my idea presented in https://github.com/hyperledger/cacti/pull/2952#discussion_r1589824232 then we can cut down the scope of this PR to just the besu connector and just the single endpoint. After all that is done, I'll do a deeper dive with a much more detailed review where there'll be a lot of nit-picking (apologies in advance) because this is a very important PR/change in the sense that I'll want to use this as THE example for everyone else in the future for them to use as a template/inspiration for similar changes where we decouple things.

petermetz (commented):
^^ @jagpreetsinghsasan @outSH Please see both of my comments above.

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
outSH: @petermetz I'd prefer to have these connectors merged some day (i.e. use common base code in the future, as described in connector inheritance discussion some time ago) so I'd prefer to stay in line as much as possible. If you're sure we want to go with separate endpoinds then we can do that :) We'll just have to keep an eye on how this progresses in the future as you've mentioned before.

outSH: Oh, and in general +1 to make this PR smaller and easier to review and reason about :)

jagpreetsinghsasan:
> > > @jagpreetsinghsasan Mostly LGTM, I Just left a few comments in the besu connector's implementation and I have one more question: Is there an endpoint implementation for the new method included or is it just the connector method for now and you are planning on adding the REST endpoint later in another PR (which is also reasonable because the PR is already pretty big)
> > 
> > 
> > @petermetz I have added the endpoint for deployContractNoKeychain (do we need to do it for all the endpoints or just showcasing it for one endpoint works?)
> 
> @jagpreetsinghsasan Sorry for the slow response, this one is a toughie. Let's keep it as small as possible initially. It's hard to review the bigger diffs. So just a single endpoint in a single package. Let's not do it all together. Also, if @outSH agrees to my idea presented in [#2952 (comment)](https://github.com/hyperledger/cacti/pull/2952#discussion_r1589824232) then we can cut down the scope of this PR to just the besu connector and just the single endpoint. After all that is done, I'll do a deeper dive with a much more detailed review where there'll be a lot of nit-picking (apologies in advance) because this is a very important PR/change in the sense that I'll want to use this as THE example for everyone else in the future for them to use as a template/inspiration for similar changes where we decouple things.

Sure @petermetz . I will then reduce the scope of this to just the Besu connector. Will close this PR once the conversation is concluded from all the involved maintainers.

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
petermetz: > @petermetz I'd prefer to have these connectors merged some day (i.e. use common base code in the future, as described in connector inheritance discussion some time ago) so I'd prefer to stay in line as much as possible. If you're sure we want to go with separate endpoinds then we can do that :) We'll just have to keep an eye on how this progresses in the future as you've mentioned before.

@outSH Got it, thank you for the compromise! Alright, that's our plan then. We can plan on doing the merging together for Cacti v3.0.0 (because we'll be able to do breaking changes that time again) and apply all that we've learned from the current refactors (could be I'll become a convert and start advocating for your endpoint design too, it happens often that once I write the code for something my opinion changes).

petermetz:
> > > > @jagpreetsinghsasan Mostly LGTM, I Just left a few comments in the besu connector's implementation and I have one more question: Is there an endpoint implementation for the new method included or is it just the connector method for now and you are planning on adding the REST endpoint later in another PR (which is also reasonable because the PR is already pretty big)
> > > 
> > > 
> > > @petermetz I have added the endpoint for deployContractNoKeychain (do we need to do it for all the endpoints or just showcasing it for one endpoint works?)
> > 
> > 
> > @jagpreetsinghsasan Sorry for the slow response, this one is a toughie. Let's keep it as small as possible initially. It's hard to review the bigger diffs. So just a single endpoint in a single package. Let's not do it all together. Also, if @outSH agrees to my idea presented in [#2952 (comment)](https://github.com/hyperledger/cacti/pull/2952#discussion_r1589824232) then we can cut down the scope of this PR to just the besu connector and just the single endpoint. After all that is done, I'll do a deeper dive with a much more detailed review where there'll be a lot of nit-picking (apologies in advance) because this is a very important PR/change in the sense that I'll want to use this as THE example for everyone else in the future for them to use as a template/inspiration for similar changes where we decouple things.
> 
> Sure @petermetz . I will then reduce the scope of this to just the Besu connector. Will close this PR once the conversation is concluded from all the involved maintainers.

@jagpreetsinghsasan OK, just please make sure to make a comment on that new PR with the relevant comment threads directly linked (so not just a link to the PR but also the direct links to the comment threads - I'm asking for this because the way the GitHub UI collapses the comment threads sometimes it's not easy to find later on where the conversation was actually happening)

jagpreetsinghsasan:
Sure @petermetz . I will have the right references to the conversations happened here in the new PR.
Closing this PR

petermetz:
> Sure @petermetz . I will have the right references to the conversations happened here in the new PR. Closing this PR

@jagpreetsinghsasan Got it, thank you!

PR #2963, opened 2023-12-22
feat(fabric-driver): added weaver fabric driver as cacti package

sandeepnRES:
* This commit is a step towards of our roadmap of creating an integrated API, where fabric driver from weaver will be merged with ledger connector from cactus. 
* Currently this is only creating a package for the fabric driver with a symbolic link, trying to experiment how to move the weaver packages as cacti packages.
* Following this we want to convert all the weaver modules to cacti packages.
* And then deeper integration will be the future steps.

**Pull Request Requirements**

**Character Limit**

### DIFF
--- a/packages/cacti-plugin-weaver-driver-fabric/package.json
+++ b/packages/cacti-plugin-weaver-driver-fabric/package.json
@@ -1,0 +1,76 @@
+{
+    "name": "@hyperledger/cacti-plugin-weaver-driver-fabric",
+    "version": "2.0.0-alpha.2",
+    "description": "Driver Server for communication with a Fabric Network as part of weaver data sharing protocol",
+    "keywords": [
+        "Hyperledger",
+        "cacti",
+        "Integration",
+        "Blockchain",
+        "Distributed Ledger Technology",
+        "Weaver"
+    ],
+    "homepage": "https://github.com/hyperledger/cacti#readme",
+    "bugs": {
+        "url": "https://github.com/hyperledger/cacti/issues"
+    },
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/hyperledger/cacti.git"
+    },
+    "license": "Apache-2.0",
+    "author": {
+        "name": "Hyperledger cacti Contributors",
+        "email": "cacti@lists.hyperledger.org",
+        "url": "https://www.hyperledger.org/use/cacti"
+    },
+    "contributors": [
+        {
+            "name": "Please add yourself to the list of contributors",
+            "email": "your.name@example.com",
+            "url": "https://example.com"
+        },
+        {
+            "name": "Venkatraman Ramakrishna",
+            "email": "vramakr2@in.ibm.com",
+            "url": "https://researcher.watson.ibm.com/researcher/view.php?person=in-vramakr2"
+        },
+        {
+            "name": "Sandeep Nishad",
+            "email": "sandeep.nishad1@ibm.com",
+            "url": "https://github.com/sandeepnRES"
+        },
+        {
+            "name": "Krishnasuri Narayanam",
+            "email": "knaraya3@in.ibm.com",
+            "url": "https://research.ibm.com/people/krishnasuri-narayanam"
+        }
+    ],
+    "main": "dist/lib/main/typescript/index.js",
+    "module": "dist/lib/main/typescript/index.js",
+    "types": "dist/lib/main/typescript/index.d.ts",
+    "files": [
+        "dist/*"
+    ],
+    "scripts": {
+        "build": "cd src/main/typescript && make build",
+        "build-image": "cd src/main/typescript && make build-image",
+        "build-image-local": "cd src/main/typescript && make build-image-local",
+        "build-local": "cd src/main/typescript && make build-local",
+        "clean": "cd src/main/typescript && make clean",
+        "clean-local": "cd src/main/typescript && make clean-local",
+        "publish": "cd src/main/typescript && make push-image && make push-image-latest",
+        "postpublish": "cd src/main/typescript && make push-image-latest",
+        "watch": "npm-watch"
+    },
+    "engines": {
+        "node": ">=18",
+        "npm": ">=8"
+    },
+    "publishConfig": {
+        "access": "public"
+    },
+    "browserMinified": "dist/cacti-weaver-driver-fabric.web.umd.min.js",
+    "mainMinified": "dist/cacti-weaver-driver-fabric.node.umd.min.js",
+    "watch": {}
+}
--- a/packages/cacti-plugin-weaver-driver-fabric/readme.md
+++ b/packages/cacti-plugin-weaver-driver-fabric/readme.md
@@ -1,0 +1,15 @@
+<!--
+ Copyright IBM Corp. All Rights Reserved.
+
+ SPDX-License-Identifier: CC-BY-4.0
+ -->
+# Cacti Fabric-Driver
+
+The term "driver" has been used in Weaver parlance, and is synonymous with "connector" as used in Cactus (and not in Cacti). Both terms refer to a module with an interface and functions to "connect" to a ledger of a given DLT type and "drive" transactions through that ledger for querying and state update purposes whenever required in the context of a cross-network transaction.
+There are some distinctive features of the Weaver Fabric driver that are not covered by the Cactus Fabric connector package, which is why the two continue to co-exist at this time. Our goal is to eventually merge them into a single connector/driver package that offers both the distinctive and overlapping features of both the existing packages.
+
+For detailed information about fabric driver visit [here](src/main/typescript/readme.md).
+
+To use fabric-driver in your application please refer [documentation](https://hyperledger.github.io/cacti/weaver/getting-started/guide/). 
+
+
--- a/packages/cacti-plugin-weaver-driver-fabric/src/main/typescript
+++ b/packages/cacti-plugin-weaver-driver-fabric/src/main/typescript
@@ -1,0 +1,1 @@
+../../../../weaver/core/drivers/fabric-driver
--- a/weaver/core/drivers/fabric-driver/server/events.ts
+++ b/weaver/core/drivers/fabric-driver/server/events.ts
@@ -66,7 +66,7 @@ async function subscribeEventHelper(
     if (newRequestId == requestId) {
       // event being subscribed for the first time
       // Start an appropriate type of event listener for this event subscription if one is not already active
-      const [listenerHandle, error] = await handlePromise(
+      const [, error] = await handlePromise(
         registerListenerForEventSubscription(
           call_request.getEventMatcher()!,
           network_name,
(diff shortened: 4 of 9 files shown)


VRamakrishna:
@sandeepnRES Can you edit the opening comment and add some description (in bullet points) about what this feature is and how it is part of the Cacti integration roadmap?

packages/cacti-weaver-driver-fabric/readme.md
```diff
@@ -0,0 +1,14 @@
+<!--
+ Copyright IBM Corp. All Rights Reserved.
+
+ SPDX-License-Identifier: CC-BY-4.0
+ -->
+# Cacti Fabric-Driver
+
+This driver is for communication with a Fabric Network.
```
VRamakrishna, line 8: Add here that the term "driver" has been used in Weaver parlance, and is synonymous with "connector" as used in Cactus (and not in Cacti). Both terms refer to a module with an interface and functions to "connect" to a ledger of a given DLT type and "drive" transactions through that ledger for querying and state update purposes whenever required in the context of a cross-network transaction.

Add also that there are some distinctive features of the Weaver Fabric driver that are not covered by the Cactus Fabric connector package, which is why the two continue to co-exist at this time. Our goal is to eventually merge them into a single connector/driver package that offers both the distinctive and overlapping features of both the existing packages.

VRamakrishna (approved):

VRamakrishna:
Added package LGTM. Please add some descriptions as per my comments before merging though.

packages/cacti-weaver-driver-fabric/package.json
```diff
@@ -0,0 +1,81 @@
+{
+    "name": "@hyperledger/cacti-weaver-driver-fabric",
```
outSH, line 2: If this package is meant to serve similar purpose as fabric connector, I'd propose to name it according to cacti standards (i.e. something like `cactus-plugin-ledger-connector-weaver-fabric` or `cactus-plugin-ledger-connector-fabric-weaver`)

packages/cacti-weaver-driver-fabric/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "version": "1.0.0",
```
outSH, line 3: Version should be `2.0.0-alpha.2`

packages/cacti-weaver-driver-fabric/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "description": "",
```
outSH, line 4: Description should be set

packages/cacti-weaver-driver-fabric/package.json
```diff
@@ -0,0 +1,81 @@
...
+      "postpublish": "cd src/main/typescript && make push-image-latest",
+      "clean": "cd src/main/typescript && make clean",
+      "clean-local": "cd src/main/typescript && make clean-local"
+    },
+    "dependencies": {
+    },
+    "devDependencies": {
+    },
+    "engines": {
+      "node": ">=10",
```
outSH, line 71: node should be `>=18`, npm `>=8`

outSH (changes requested):

packages/cacti-weaver-driver-fabric/package.json
sandeepnRES: Done.

packages/cacti-weaver-driver-fabric/package.json
sandeepnRES: Done.

packages/cacti-weaver-driver-fabric/package.json
sandeepnRES: Added.

packages/cacti-weaver-driver-fabric/package.json
VRamakrishna: @outSH The names for several packages contain the `cactus-` prefix because they were carried over from the Cactus project into Cacti when the merger with Weaver occurred. Our intention, going forward, was to use `cacti` instead of `cactus` (or `weaver`). The reason we are still using `cactus` and `weaver` in several package names is because the packages of both the contributing repos are still distinct within Cacti.

We are aiming for a gradual integration whereby we eliminate redundant packages. This PR is simply a stepping stone to that whereby we expose the Weaver packages through the standard Cacti monorepo. After this, we will expose all other Weaver packages the same way. Later we will combine, merge, and eliminate various packages within the `packages` folder. For example, `cacti-weaver-driver-fabric` and `cactus-plugin-ledger-connector-fabric` are meant to serve basically the same purpose, so we ought to merge their functions and APIs into a single module called, let's say, `cacti-plugin-ledger-connector-fabric`. But this kind of refactoring (actually, it's more than that) will take time and effort.

So, to cut to the chase, any new package we create within this repository I believe ought to have the `cacti-` prefix. The suffix is something we can debate.

For your reference, see the roadmap and target architecture sketches in https://github.com/hyperledger/cacti/blob/main/ROADMAP.md#cacti-v2. (I know we've long passed the dates indicated there, so we will need to update that page with a more realistic estimate.)

@petermetz Any thoughts or suggestions on this?

outSH: > We are aiming for a gradual integration whereby we eliminate redundant packages. This PR is simply a stepping stone to that whereby we expose the Weaver packages through the standard Cacti monorepo. After this, we will expose all other Weaver packages the same way. Later we will combine, merge, and eliminate various packages within the `packages` folder. For example, `cacti-weaver-driver-fabric` and `cactus-plugin-ledger-connector-fabric` are meant to serve basically the same purpose, so we ought to merge their functions and APIs into a single module called, let's say, `cacti-plugin-ledger-connector-fabric`. But this kind of refactoring (actually, it's more than that) will take time and effort.

@VRamakrishna That was clear to me, my intention was to keep standardized naming convention for temporary weaver packages as well, so that they can be easily identified and managed. I'm don't have strong opinions on the exact naming. 
As for `cacti` prefix, some scripts still have (or at least had a while ago) hard-coded `cactus` for identifying packages, that's why I keep using `cactus`. I kinda assumed we will do the project-wide switch to `cacti` in a single commit at some point, but if it works already then that's even better.

packages/cacti-weaver-driver-fabric/src/main/typescript
```diff
@@ -0,0 +1 @@
+../../../../weaver/core/drivers/fabric-driver
```
petermetz, line 1: @sandeepnRES Was this meant to be a directory instead of a file by any chance?

packages/cacti-weaver-driver-fabric/package.json
petermetz: @outSH @VRamakrishna Sorry for being late to the party! I could argue for and against all of these and also agree that it's not a big deal either way if the pacakge is temporary. So if it is temporary, I'm good with the pkg name as is. If it will stick around I'll also suggest to call it something like `cacti-plugin-ledger-connector-fabric-weaver` but it's really hard to pin down which one makes most sense because it's (for now) not an implementation of the connector plugin interface. So we are trying to name something that we already know will change over time and so it could stand to reason either way, as in, the current state of it could inform the name to be less confusing now and more confusing later, or the future state of it could inform the name to be more confusing now and less later... :-)

1. I agree that for the new packages `cacti` is a better prefix since the end goal is to have all of them end up that way.
2. I also agree that the hierarchical & flattened naming structure we have in place helps a lot (at least to me) to find the code I'm looking for and should be kept where possible (even if it feels annoyingly verbose at times)
3. Also agree that we need to do the "big rename" rather sooner than later and definitely before the 2.0.0 release goes out because we get to make breaking changes prior to that date (this is my usual reminder that now is the "easy" time when we can do things like that - next time will be right before 3.0.0)

With all that said, I think I'm OK with the name for now as-is and we should rename it later, potentially in v3.0.0.

What I'll nit-pick on is much worse... :D I'd love it if we called the `readme.md` file `README.md` just keep with the conventions.

petermetz (changes requested):

outSH (approved):

packages/cacti-weaver-driver-fabric/src/main/typescript
sandeepnRES: It is a symbolic link file to weaver directory

petermetz: @sandeepnRES Oh, OK, got it! Sorry for not coming back to this earlier, please hit the re-request review button next time otherwise it doesn't show up in my review queue!

petermetz (approved):

petermetz:
> Added package LGTM. Please add some descriptions as per my comments before merging though.

@sandeepnRES Big +1 on that. However trivial/straightforward it seems to us, it is always good to explain for future contributors/maintainers what the change is in a little more detail with some historical context, etc.

outSH:
@sandeepnRES CI is failing, could you investigate? I think the package name is duplicated in both places and that's causing the scripts to fail

petermetz:
@sandeepnRES Do you need any help investigating this one?

sandeepnRES:
Hi @petermetz What do you think will be better, to rename this package. Or to remove the one inside `weaver` directory from `packages` declared in root `package.json`

sandeepnRES:
Hi @petermetz @VRamakrishna @outSH 
Sorry for such a long delay, I've now renamed the package to `cacti-plugin-weaver-driver-fabric`
Haven't called it ledger connector yet, just to differentiate the original ledger connector and driver, once we combine both, then probably we can use that name.

petermetz:
@sandeepnRES OK, works for me. Please run yarn install to update the lock file.

```sh
> configure
> npm run init-registries && yarn install && yarn build:dev:backend


> init-registries
> npm config set @iroha2:registry=https://nexus.iroha.tech/repository/npm-group/

➤ YN0000: Yarn detected that the current workflow is executed from a public pull request. For safety the hardened mode has been enabled.
➤ YN0000: It will prevent malicious lockfile manipulations, in exchange for a slower install time. You can opt-out if necessary; check our documentation for more details.

➤ YN0000: · Yarn 4.1.0
➤ YN0000: ┌ Resolution step
Resolution step
➤ YN0000: └ Completed in 23s 737ms
➤ YN0000: ┌ Post-resolution validation
Post-resolution validation
➤ YN0028: +"@hyperledger/cacti-plugin-weaver-driver-fabric@workspace:packages/cacti-plugin-weaver-driver-fabric":
➤ YN0028: +  version: 0.0.0-use.local
➤ YN0028: +  resolution: "@hyperledger/cacti-plugin-weaver-driver-fabric@workspace:packages/cacti-plugin-weaver-driver-fabric"
➤ YN0028: +  languageName: unknown
➤ YN0028: +  linkType: soft
➤ YN0028: +
➤ YN0028: The lockfile would have been modified by this install, which is explicitly forbidden.
➤ YN0000: └ Completed in 0s 243ms
➤ YN0000: · Failed with errors in 24s 2ms
Error: Process completed with exit code 1.
```

PR #2973, opened 2024-01-11
build(api-client): generate java, go clients

### LINKED ISSUE #393: build(api-client): generate java, go clients
### Description

As a developer I want to not be forced to use the NodeJS/Javascript API client if my primary programming language is something else so that I can achieve high productivity through Cactus no matter the language I'm using.

- This is different from a similar looking issue which discusses the possibility of writing plugins in different languages as well.
- We do not need write much Go/Java code because 99% of it is auto-generated. See the Acceptance criteria 4)

Here is a complete list of programming languages that are supported by the OpenAPI generator (which we can take full advantage of since we are using the OpenAPI specs)
https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/languages

### Acceptance Criteria
1. Clients for Java and Go are generated as part of the full build
2. Dev builds only build the Typescript API client to remain as fast as possible
3. Publishing of the Java and Go API clients is included in the release process => if I run the release scripts the full build runs and the Java, Go clients are published/made available for public consumption the same way we do with the Typescript API client (it gets published to npm)
4. Client side routing is supported by the Go/Java API clients as well.

jagpreetsinghsasan:
# Commit to be reviewed

build(api-client): generate java, go clients    
    
    Primary Changes
    ---------------
    1. Updated package.json of packages to include the new
       codegen scripts
    2. Added a new workflow to create and upload the jars from
       the newly added code

    3. Added replace package to assist the codegen scripts written in package.json

Fixes #393

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


petermetz (approved):
@jagpreetsinghsasan Thank you for this! In general, LGTM. 

Any reason why we can't use the kotlin based templates (I can imagine many, but we should documetn the decision and the reasons for it in the commit message)

P.S.: The custom-checks were broken because the package.json files became unsorted, I remedied the situation and did a force push to your branch. Please make sure not to accidentally erase those changes!

jagpreetsinghsasan:
> @jagpreetsinghsasan Thank you for this! In general, LGTM.
> 
> Any reason why we can't use the kotlin based templates (I can imagine many, but we should documetn the decision and the reasons for it in the commit message)
> 
> P.S.: The custom-checks were broken because the package.json files became unsorted, I remedied the situation and did a force push to your branch. Please make sure not to accidentally erase those changes!

Here we are generating the java client code and their corresponding artifacts, so 
1. For the java client code, we can transpile the code from kotlin to java (but then thats more like using a transpiler and not the raw openapi specs directly to generate the java code). Shall I look into generating the java code in this way then, if this reasoning is not right?
2. The failure in generating jars from generated java code (in certain plugins), also documents that straightaway `openapi specs` -> `java code` -> `jar` doesnt work as intented and if a developer wants to extensively use the straightaway java codegen, then our work done around this (and figuring out what actually is wrong from openapi codegen side will be helpful while investing time and will give a right direction).

Everything seems fine after your force push, compiling correctly, so no issues there.

outSH (changes requested):
Sorry, I forgot about this PR and introduced more connector packages in the meantime :( Could you please generate the new client for them as well?
- https://github.com/hyperledger/cacti/tree/main/packages/cactus-plugin-ledger-connector-aries
- https://github.com/hyperledger/cacti/tree/main/packages/cactus-plugin-ledger-connector-cdl

Everything else looks great, thank you very much!

petermetz:
> Here we are generating the java client code and their corresponding artifacts, so
> 
>     1. For the java client code, we can transpile the code from kotlin to java (but then thats more like using a transpiler and not the raw openapi specs directly to generate the java code). Shall I look into generating the java code in this way then, if this reasoning is not right?
> 

@jagpreetsinghsasan I'd say we transpile from the JSON spec to the target language (of our choice) directly in each case (be that java or kotlin). 
Right now you are saying `-g java` in the package.json scripts and I'm suggesting to keep with the pre-existing convention [1] and say `-g kotlin` unless there is a good reason to not to do that.


[1] =>  image

petermetz:
>     2. The failure in generating jars from generated java code (in certain plugins), also documents that, straightaway `openapi specs` -> `java code` -> `jar` doesnt work as intented and if a developer wants to extensively use the straightaway java codegen, then our work done around this (and figuring out what actually is wrong from openapi codegen side will be helpful while investing time and will give a right direction).


@jagpreetsinghsasan This works for me on the main branch, so maybe these issues are only happening for the `java` template and not the `kotlin` one?

```sh
yarn install
yarn configure
yarn codegen
cd packages/cactus-cmd-api-server/src/main/kotlin/generated/openapi/kotlin-client
```

```sh
$ gradle build
Starting a Gradle Daemon, 1 incompatible Daemon could not be reused, use --status for details

BUILD SUCCESSFUL in 21s
3 actionable tasks: 3 executed
```

```sh
$ gradle jar

BUILD SUCCESSFUL in 496ms
3 actionable tasks: 3 up-to-date
```

```sh
$ ls -alt build/libs/kotlin-client-1.0.0.jar 
-rw-r--r-- 1 peter peter 74501 Jan 31 17:33 build/libs/kotlin-client-1.0.0.jar
```

petermetz (commented):
@jagpreetsinghsasan ^^ Please see my comments above and pass it back for review as applicable!

jagpreetsinghsasan:
> > Here we are generating the java client code and their corresponding artifacts, so
> > 
> > 1. For the java client code, we can transpile the code from kotlin to java (but then thats more like using a transpiler and not the raw openapi specs directly to generate the java code). Shall I look into generating the java code in this way then, if this reasoning is not right?
> 
> @jagpreetsinghsasan I'd say we transpile from the JSON spec to the target language (of our choice) directly in each case (be that java or kotlin). Right now you are saying `-g java` in the package.json scripts and I'm suggesting to keep with the pre-existing convention [1] and say `-g kotlin` unless there is a good reason to not to do that.
> 
> [1] => image

@petermetz I would prefer not to generate jars from kotlin (for the java code), because generating the jar artifact is sort of a confirmation that the generated code is correct.
So if I generate the code for java directly from openapi, but then I generate the relevant jar using kotlin (although both the jars from java and kotlin should be the same, but their source code can be different/wrong, given what the transpile logic is specified for it), it would be not appropriate

petermetz:
> > > Here we are generating the java client code and their corresponding artifacts, so
> > > 
> > > 1. For the java client code, we can transpile the code from kotlin to java (but then thats more like using a transpiler and not the raw openapi specs directly to generate the java code). Shall I look into generating the java code in this way then, if this reasoning is not right?
> > 
> > 
> > @jagpreetsinghsasan I'd say we transpile from the JSON spec to the target language (of our choice) directly in each case (be that java or kotlin). Right now you are saying `-g java` in the package.json scripts and I'm suggesting to keep with the pre-existing convention [1] and say `-g kotlin` unless there is a good reason to not to do that.
> > [1] => image
> 
> @petermetz I would prefer not to generate jars from kotlin (for the java code), because generating the jar artifact is sort of a confirmation that the generated code is correct. So if I generate the code for java directly from openapi, but then I generate the relevant jar using kotlin (although both the jars from java and kotlin should be the same, but their source code can be different/wrong, given what the transpile logic is specified for it), it would be not appropriate

@jagpreetsingsasan Yeah my point is to do kotlin everywhere. That's why I said `-g kotlin` above. E.g., don't generate java code, generate kotlin code.

jagpreetsinghsasan:
> > > > Here we are generating the java client code and their corresponding artifacts, so
> > > > 
> > > > 1. For the java client code, we can transpile the code from kotlin to java (but then thats more like using a transpiler and not the raw openapi specs directly to generate the java code). Shall I look into generating the java code in this way then, if this reasoning is not right?
> > > 
> > > 
> > > @jagpreetsinghsasan I'd say we transpile from the JSON spec to the target language (of our choice) directly in each case (be that java or kotlin). Right now you are saying `-g java` in the package.json scripts and I'm suggesting to keep with the pre-existing convention [1] and say `-g kotlin` unless there is a good reason to not to do that.
> > > [1] => image
> > 
> > 
> > @petermetz I would prefer not to generate jars from kotlin (for the java code), because generating the jar artifact is sort of a confirmation that the generated code is correct. So if I generate the code for java directly from openapi, but then I generate the relevant jar using kotlin (although both the jars from java and kotlin should be the same, but their source code can be different/wrong, given what the transpile logic is specified for it), it would be not appropriate
> 
> @jagpreetsingsasan Yeah my point is to do kotlin everywhere. That's why I said `-g kotlin` above. E.g., don't generate java code, generate kotlin code.

We are already generating kotlin code sepearetly with `-g kotlin` argument in the codegen scripts. So I shall I just remove the java codegen entirely?

petermetz:
> > > > > Here we are generating the java client code and their corresponding artifacts, so
> > > > > 
> > > > > 1. For the java client code, we can transpile the code from kotlin to java (but then thats more like using a transpiler and not the raw openapi specs directly to generate the java code). Shall I look into generating the java code in this way then, if this reasoning is not right?
> > > > 
> > > > 
> > > > @jagpreetsinghsasan I'd say we transpile from the JSON spec to the target language (of our choice) directly in each case (be that java or kotlin). Right now you are saying `-g java` in the package.json scripts and I'm suggesting to keep with the pre-existing convention [1] and say `-g kotlin` unless there is a good reason to not to do that.
> > > > [1] => image
> > > 
> > > 
> > > @petermetz I would prefer not to generate jars from kotlin (for the java code), because generating the jar artifact is sort of a confirmation that the generated code is correct. So if I generate the code for java directly from openapi, but then I generate the relevant jar using kotlin (although both the jars from java and kotlin should be the same, but their source code can be different/wrong, given what the transpile logic is specified for it), it would be not appropriate
> > 
> > 
> > @jagpreetsingsasan Yeah my point is to do kotlin everywhere. That's why I said `-g kotlin` above. E.g., don't generate java code, generate kotlin code.
> 
> We are already generating kotlin code sepearetly with `-g kotlin` argument in the codegen scripts. So I shall I just remove the java codegen entirely?

@jagpreetsinghsasan Yeah that's what I'd do. The kotlin code can be used to produce the JVM .jar files anyway so we don't need the java generated code IMO at all.

jagpreetsinghsasan:
Closing this PR. Will create a new one

PR #2977, opened 2024-01-15
ci(github): fix path filtering of required but skipped workflows

petermetz:
1. Migrated from the stock GitHub Actions yaml syntax to using this custom
action[1] that allows us to skip the running of jobs based on path filtering.
The difference compared to the old solution is that this one won't just
skip the checks that are marked as required but it will pass them instead.
2. This is a suboptimal solution because we have to run each job so that
the code can be checked out and then examined for changes. This uses
resources that we should not have to expend, but currently there's no way
to avoid this because of the way the GitHub handles required checks that
are skipped (which cause a deadlock on the CI at the currently).
3. There is a relevant discussion thread about the problem here [2] which
should be voted on heavily so that it gets the necessary attention from
the product teams at GitHub.

Quoting the relevant part of the GitHub documentation directly:

> Handling skipped but required checks
> Warning: If a workflow is skipped due to path filtering, branch
> filtering or a commit message, then checks associated with that workflow
> will remain in a "Pending" state. A pull request that requires those
> checks to be successful will be blocked from merging.
>
> For this reason you should not use path or branch filtering to skip
>  workflow runs if the workflow is required. For more information, see
>  "Skipping workflow runs" and "Required workflows."
>
> If, however, a job within a workflow is skipped due to a conditional,
> it will report its status as "Success". For more information, see
> "Using conditions to control job execution."

**Source** (link broken into multiple lines to avoid it being longer than
100 characters):

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/
collaborating-on-repositories-with-code-quality-features/
troubleshooting-required-status-checks#handling-skipped-but-required-checks

[1] https://github.com/dorny/paths-filter
[2] https://github.com/orgs/community/discussions/44490

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/.github/workflows/test_weaver-asset-exchange-besu.yaml
+++ b/.github/workflows/test_weaver-asset-exchange-besu.yaml
@@ -10,20 +10,8 @@ on:
   # Triggers the workflow on push or pull request events but only for the main branch
   push:
     branches: [ main ]
-    paths:
-        - '.github/workflows/test_weaver-**'
-        - 'weaver/**'
-        - '!weaver/docs/**'
-        - '!weaver/rfcs/**'
-        - '!weaver/resources/**'
   pull_request:
     branches: [ main ]
-    paths:
-        - '.github/workflows/test_weaver-**'
-        - 'weaver/**'
-        - '!weaver/docs/**'
-        - '!weaver/rfcs/**'
-        - '!weaver/resources/**'
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
@@ -34,9 +22,24 @@ concurrency:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
+  check_code_changed:
+    outputs:
+      status: ${{ steps.changes.outputs.weaver_code_changed }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.5.2
+
+      - uses: dorny/paths-filter@v2.11.1
+        id: changes
+        with:
+          filters: |
+            weaver_code_changed:
+              - './weaver/**!(*.md|*.css|*.html|*.jpg|*.jpeg|*.png)'
+              - '.github/workflows/test_weaver-asset-exchange-besu.yaml'
 
   asset-exchange-besu:
-    if: ${{ false }}
+    needs: check_code_changed
+    if: ${{ false && needs.check_code_changed.outputs.status == 'true' }}
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
 
@@ -216,6 +219,8 @@ jobs:
         working-directory: weaver/samples/besu/besu-cli
 
   asset-exchange-besu-local:
+    needs: check_code_changed
+    if: needs.check_code_changed.outputs.status == 'true'
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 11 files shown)


outSH (approved):

.github/workflows/test_weaver-asset-exchange-besu.yaml
```diff
@@ -50,31 +38,44 @@ jobs:
...
+  
+      - if: steps.changes.outputs.weaver_code_changed == 'true'
+        name: Set up JDK 11
         uses: actions/setup-java@v3.11.0
         with:
           java-version: '11'
           distribution: 'adopt'
 
-      - name: Use Node.js ${{ env.NODEJS_VERSION }}
+      - if: steps.changes.outputs.weaver_code_changed == 'true'
```
sandeepnRES, line 56: Instead of checking at every step, what we can do is create another job say `check_code_changed` just to compute above variable `weaver_code_changed` and put it into `GITHUB_OUTPUT`,
Then for each job we just need to check only once at the start of job
```
if: needs.check_code_changed.outputs.weaver_code_changed=='true'
```
rather than at all steps, would be efficient that way I believe.

Check this workflow file for example: https://github.com/hyperledger/cacti/blob/main/.github/workflows/test_weaver-pre-release.yaml, (the subsequent jobs depend upon output of `check_release` job)

.github/workflows/test_weaver-asset-transfer.yaml
```diff
@@ -48,24 +36,36 @@ jobs:
...
       - uses: actions/checkout@v3.5.2
 
-      - name: Set up JDK 8
+      - uses: dorny/paths-filter@v2.11.1
+        id: changes
+        with:
+          filters: |
+            weaver_code_changed:
+              - './weaver/**!(*.md|*.css|*.html|*.jpg|*.jpeg|*.png)'
+              - '.github/workflows/test_weaver-asset-exchange-fabric.yaml'
```
sandeepnRES, line 45: This should be `'.github/workflows/test_weaver-asset-transfer.yaml'`

.github/workflows/test_weaver-asset-transfer.yaml
```diff
@@ -682,153 +716,190 @@ jobs:
...
```
sandeepnRES, line 725: This should be `'.github/workflows/test_weaver-asset-transfer.yaml'`

.github/workflows/test_weaver-docker-build.yaml
```diff
@@ -37,7 +25,16 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v3.5.2
 
-      - name: Build Image
+      - uses: dorny/paths-filter@v2.11.1
+        id: changes
+        with:
+          filters: |
+            weaver_code_changed:
+              - './weaver/**!(*.md|*.css|*.html|*.jpg|*.jpeg|*.png)'
```
sandeepnRES, line 33: These are unit tests, we can be more specific to paths here.
Since `paths` were one time specified for each workflow I had to give whole weaver directory, but here we are specifying for each job, we can be more specific, do you want me to edit these so that they can be more efficient?

sandeepnRES (approved):

petermetz:
@sandeepnRES Thank you very much for the help and the optimizations!

PR #3007, opened 2024-01-30
perf(cmd-api-server): add demonstration of continuous benchmarking

petermetz:
Primary change:
---------------

This is the ice-breaker for some work that got stuck related to this issue:
https://github.com/hyperledger/cacti/issues/2672

The used benchamking library (benchmark.js) is old but solid and has
almost no dependencies which means that we'll be in the clear longer term
when it comes to CVEs popping up.

The benchmarks added here are very simple and measure the throughput of
the API server's Open API spec providing endpoints.

The GitHub action that we use is designed to do regression detection and
reporting but this is hard to verify before actually putting it in place
as we cannot simulate the CI environment's clone on a local environment.

The hope is that this change will make it so that if someone tries to
make a code change that will lower performance significantly, then we
can catch that at the review stage instead of having to find out later.

Secondary change:
-----------------

1. Started using tsx in favor of ts-node because it appers to be about
5% faster when looking at the benchmark execution. It also claims to have
less problems with ESM compared to ts-node so if this initial trial
goes well we could later on decide to swap out ts-node with it project-wide.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


packages/cactus-cmd-api-server/package.json
```diff
@@ -43,6 +43,7 @@
     "dist/*"
   ],
   "scripts": {
+    "benchmark": "tsx ./src/test/typescript/benchmark/run-cmd-api-server-benchmark.ts > ../../.tmp/benchmark-results/cmd-api-server/run-cmd-api-server-benchmark.ts.log",
     "codegen": "run-p 'codegen:*'",
```
outSH, line 47: Fails if `../../.tmp/benchmark-results/cmd-api-server/` was not created beforehand. Maybe add `mkdir` command here before? Or don't pipe the output anywhere (print to stdout) and let the user decide where to store the results. CI can pipe it to location it created before.

outSH (changes requested):

packages/cactus-cmd-api-server/package.json
petermetz: @outSH That was a rookie mistake on my side! Thanks for checking on it! 

I thought about just doing the quick and dirty `mkdir -p &&  ...` script but then the script wouldn't be able to run on windows (currently the project build doesn't work on windows anyway but at least I'm not getting us further away from it either)

So, I ended up refactoring the script in a way that it takes a path parameter and then does the mkdir -p from code which is cross-platform by default. I tested it by deleting the `.tmp` folder and then re-running the script and it works fine, please have a look when you get the chance!

outSH: @petermetz Thanks, looks great!

outSH (approved):

jagpreetsinghsasan (approved):
LGTM
@ruzell22 please refer to this implementation if this is what you are trying to work on, for Besu plugin.

PR #3011, opened 2024-02-02
feat(plugin-satp-hermes): replace IPFS dependency in SATP package

AndreAugusto11:
* implement the repository design pattern to make storage technology-agnostic
* due to the deprecation of the ipfs package, this allows one to choose another storage
* refactoring of the tests and the CBDC example that is based on the SATP
* implement the remote log storage as a SQLite database
* add post-build instruction to copy knex files to dist/

closes https://github.com/hyperledger/cacti/issues/2984

depends on #3006


petermetz (changes requested):
@AndreAugusto11 Please rebase onto upstream/main and resolve the conflicts and then pass it back for review with the 're-request review' button.

FYI: I just merged the parent PR so you should end up with only a single commit on this one after the rebase.

packages/cactus-plugin-satp-hermes/package.json
```diff
@@ -38,6 +38,7 @@
     "dist/*"
   ],
   "scripts": {
+    "build:dev:backend:postbuild": "mkdir -p ./dist/lib/main/knex && cp -r ./knex/* ./dist/lib/main/knex",
```
petermetz, line 41: @AndreAugusto11 
A quick note to self here (not a change request because I  can't provide specific guidance on the how, yet): We should do away with scripts like this and instead fold it back into the typscript compilation process. E.g., instead of manually copying around source code files we should let the compiler do that instead. 

In this case 
1. The `.js` file(s) in the ./knex/ dir should be rewritten into a `.ts` file that the compiler then picks up and generates the `.js` file for anyway (and places it in the dist folder but possibly under a different sub-path which is I'm guessing the real issue)

petermetz (approved):
@AndreAugusto11 LGTM, thank you very much for this contribution!

izuru0 (approved):
LGTM

packages/cactus-plugin-satp-hermes/package.json
AndreAugusto11: @petermetz I agree. Following that approach would be cleaner across the whole project. This change was due to this thread [2984](https://github.com/hyperledger/cacti/issues/2984#issuecomment-1898783774).

Nevertheless, if you'd like, I can revert the change and leave it like it was — i.e.,
```
    const configFile = require("../../../../../knex/knexfile.ts")[
      process.env.ENVIRONMENT || "development"
    ];
```
It is still not the best solution, but I'll leave it up to you to decide which approach to follow.

petermetz: @AndreAugusto11 Let's leave it as it is for now because I'm also not sure what my preference would be and I need to think on it but do not have the time unfortunately. ;-(

I opened an issue to keep track of it and CC'd you on it and we can figure it out later. This is good as is!

PR #3019, opened 2024-02-09
build(devcontainer): get git LTS from apt instead of compiling sources

petermetz:
1. Setting the git version to LTS makes it so that the VSCode dev container feature
that we use for installing git does not fetch the source code of a specific version
and then compiles it but instead of that it just does an apt install git which has
100x faster execution time compared to the full compilation.
2. The downside of this approach is that we no longer guaranteed a specific version
of git and this could cause issues because LTS is a moving target, e.g. a form of
auto-upgrade which is something Peter's been always on a crusade against, but
trade-offs have to be made sometimes.
3. The reasoning why trade-off is appropriate is that the dev container is one of
first things that new contributors encounter/use as they ramp up with the project
and it makes a big difference if the container can be built in a few minutes vs.
a few dozen minutes.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


izuru0 (approved):
LGTM

jagpreetsinghsasan (approved):

PR #3020, opened 2024-02-09
build(tools): codegen clean script now removes corda kotlin-spring src

petermetz:
Extended the existing globs in the clean script to include the back-end
code generated via the `kotlin-spring` template as well.

1. Previously only the generated client code was removed and regenerated
but with this change we now have the generated corda connector backend
API kotlin code removed as well.
2. The build files such as the build.gradle.kts and pom.xml files are not
deleted for now just the files under the `api` and the `model` directories
within the src/main and src/test code folders.
3. Running the delete script and then the code generation results in no
git index changes which is what we strive for.
4. The reason why we needed a slightly more complicated glob pattern
for achieving this is because the back-end code requires hand-written
implementation for the endpoints which are all placed within the `impl`
directory within the `src/main` and `src/test` folders and these need to
be ignored.
5. This change is setting the stage for an upcoming upgrade in the configuration
of the open api generator where we'll set it up to use spring boot 3 instead
of the current legacy version. The reason why this is needed is because
the legacy version has a large amount of security vulnerabilities that need
to be taken care of by upgrading our dependencies.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/tools/clear-openapi-codegen-folders.js
+++ b/tools/clear-openapi-codegen-folders.js
@@ -13,6 +13,8 @@ const PROJECT_DIR = path.join(SCRIPT_DIR, "../");
 const OPENAPI_CODEGEN_PATHS = [
   "src/main/kotlin/generated/openapi/**",
   "src/main/typescript/generated/openapi/typescript-axios/**",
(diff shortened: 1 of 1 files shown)


jagpreetsinghsasan (approved):
LGTM

PR #3021, opened 2024-02-12
ci(github): loosen up file matching when selecting tests to run

outSH:
- Current file match glob is incorrect and never matches a file.
- I didn't find a way to filter files by extension (as it was attempted now) so I propose to loose up the rules and run the test if any file in a package has changed.

## More context
- I've noticed in another PR (https://github.com/hyperledger/cacti/pull/3018) that most tests didn't run, even though this PR changes most ts files, from all packages I think.
- I've digged a little bit in dorny/paths-filter sources, found out the file matching string is passed to `picomatch` for actual filtering.
- `picomatch` is very simple (but fast), and does not support any advanced globbing techniques. I've done some tests locally, and the best solution I found is to use less strict matching rules.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,106 +40,106 @@ jobs:
         with:
           filters: |
             cmd-api-server-changed:
-              - './packages/cactus-cmd-api-server/**!(*.md|*.css|*.html|*.jpg|*.jpeg|*.png)'
-              - './packages/cactus-test-cmd-api-server/**!(*.md|*.css|*.html|*.jpg|*.jpeg|*.png)'
-              - './packages/cactus-common/**!(*.md|*.css|*.html|*.jpg|*.jpeg|*.png)'
-              - './packages/cactus-core/**!(*.md|*.css|*.html|*.jpg|*.jpeg|*.png)'
-              - './packages/cactus-core-api/**!(*.md|*.css|*.html|*.jpg|*.jpeg|*.png)'
-              - './packages/cactus-test-tooling/**!(*.md|*.css|*.html|*.jpg|*.jpeg|*.png)'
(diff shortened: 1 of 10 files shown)


izuru0 (approved):
LGTM

sandeepnRES:
Hi @outSH, I didn't get what you mean by `never matches a file`.
What I saw on PRs was, if there was a code change, the corresponding CI was running else it skipped.

outSH:
> Hi @outSH, I didn't get what you mean by `never matches a file`. What I saw on PRs was, if there was a code change, the corresponding CI was running else it skipped.

@sandeepnRES 
Yeah, I've changed the original PR so it may not be clear, sorry. I've opened another one on my fork to make it more persitent here - https://github.com/outSH/cactus/pull/11. As you can seet his PR changes all packages, but none are marked as changed by the CI:

image

This is because of invalid glob in config as I described above

outSH:
@petermetz The problem was spotted on PR opened here, you can see this happening here - https://github.com/hyperledger/cacti/pull/3018

sandeepnRES (approved):
LGTM

petermetz (commented):
@outSH My theory right now is that this won't fix the issue because of the underlying cause explained here: https://github.com/hyperledger/cacti/pull/3018#issuecomment-1957355218

Is there any chance that you verify it over there on that PR?
I'm thinking that if my research is correct than we don't need this PR at all.

petermetz (approved):
@outSH LGTM - later once the path-filter action has the ability to support our more fine-grained filtering use-case I'll recommend to change the glob patterns once again, but in the meantime this is definitely a net positive! Thank you very much for taking the time to send this in!

PR #3022, opened 2024-02-14
test(tools/docker): add new Corda v4.12-SNAPSHOT all-in-one image

petermetz:
Published on GHCR as
`ghcr.io/hyperledger/cactus-corda-4-12-all-in-one-negotiation:2024-02-13-dev`

1. It is a more bare bones flavor of the previous versions in the sense that it does not support
running the REST API web servers that come with the samples repository at all. It only supports
running the nodes A, B and Notary.
2. There is no Party C because we had to switch over to the negotiation-cordapp example.
3. This flavor of the image supports customizing the log levels of the Corda node through
the file at: tools/docker/corda-all-in-one/corda-v4_12/negotiation-cordapp/log4j2.xml
Note however that applying these changes requires a rebuild of the image so it's still less
convenient than idea lto use, but at least now we have an option to do so.
4. Another new feature of this image compared to the earlier Corda AIO images is that it
uses Ubuntu 22.04 LTS as it's base instead of alpine. This increases the image size a little
but it's well worth the trade-off of having a uniform image base (which is the long term plan)
5. There is a builder stage that can be used to clone an arbitrary branch + SHA combination of
the upstream corda git repository and then build the node .jar file from it which then will be
used by the image to run the nodes. The custom-built .jar file is placed under  /opt/bin/corda.jar
so do not mix that one up with the other one that is provided with the corda kotlin-samples repo.
6. The `run-notary-node.sh`, `run-party-a-node.sh` and `run-party-b-node.sh` scripts were altered
to use the custom-built .jar file mentioned above and this is the reason why this commit adds
a (seeming) clone of these scripts.
7. For now I did not add a CI task for building and publishing the image because even though
that would match the older images, I'm trying to save on CI resources and at the moment we
don't necessarily need a fresh build of this image on each new pull request.
8. Last but not least (first actually) this is also a security adjacent upgrade since using
this image is the precursor to a much bigger change that will see the Corda v4 connector's
dependencies upgraded to Spring Boot v3 and Corda v4.12-SNAPSHOT but we can't do that without
having this image published first if we want to have test-automation verifying those changes.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


jagpreetsinghsasan (approved):

izuru0 (approved):
LGTM

PR #3028, opened 2024-02-20
refactor(gui): gui framework change from solid to react

rwat17:
- changed gui framework to react because it is more common
  and more friendly for newcomers.
- routing moved from src/pages/Pages to src/main, beside that
  only refactoring were made

### DIFF
--- a/packages/cacti-ledger-browser-react/package.json
+++ b/packages/cacti-ledger-browser-react/package.json
@@ -1,0 +1,84 @@
+{
+  "name": "@hyperledger/cacti-ledger-browser-react",
+  "version": "2.0.0-alpha.2",
+  "description": "Cacti GUI for visualizing ledger data build on react.",
+  "keywords": [
+    "Hyperledger",
+    "Cacti",
+    "Integration",
+    "Blockchain",
+    "Distributed Ledger Technology"
+  ],
+  "homepage": "https://github.com/hyperledger/cacti#readme",
+  "bugs": {
+    "url": "https://github.com/hyperledger/cacti/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/hyperledger/cacti.git"
+  },
+  "license": "Apache-2.0",
+  "author": {
+    "name": "Hyperledger Cacti Contributors",
+    "email": "cacti@lists.hyperledger.org",
+    "url": "https://www.hyperledger.org/use/cacti"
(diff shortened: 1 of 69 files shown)


petermetz (commented):
@rwat17 LGTM with comments: I'd love some documentation why was this a net positive change. I'm guessing you find it easier to develop against react than solid, but still, if there were specific pain points that this addressed it would be good to list them both in the PR description and the **commit message as well**. 

The core idea behind this is that people who come on board to contribute to/maintain the project years from now will get useful context/knowledge from reading the wisdom we leave behind (and have none of that if we don't take time to document our choices and reasoning behind them)

rwat17:
@petermetz Yes, the point of changing is that React is more common among people. It will be easier to develop the package for newcomers too because imo React has that advantage over Solid it is more documented/supported and more people tend to know it.

packages/cacti-ledger-browser-react/package.json
```diff
@@ -0,0 +1,84 @@
...
+  ],
+  "scripts": {
+    "build": "yarn run build:prod:frontend",
+    "build:dev:frontend": "vite build --mode=development",
+    "build:prod:frontend": "vite build",
+    "serve": "vite preview",
+    "start": "vite"
+  },
+  "dependencies": {
+    "@emotion/react": "^11.11.3",
```
outSH, line 56: pin package versions

outSH: remove vite and react svg

outSH (changes requested):

packages/cacti-ledger-browser-react/src/pages/fabric/DashFabric/DashFabric.tsx
```diff
@@ -0,0 +1,104 @@
+import { supabase } from "../../../supabase-client";
+import CardWrapper from "../../../components/CardWrapper/CardWrapper";
+import { Transaction } from "../../../schema/supabase-types";
+import { Block } from "../../../schema/supabase-types";
+// @ts-expect-error
+import styles from "./DashFabric.module.css";
```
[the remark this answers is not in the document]
rwat17: Removed. They were used in Solid version, idk exactly for what reason, 0 errors after build.

petermetz (approved):
@rwat17 Thank you!

Issue #2811, opened 2023-10-18
feat(satp-hermes): rules orchestrator layer on SATP


RafaelAPB:
Create a rules orchestrator layer to fine-tune control over the execution of SATP, such that developers can easily create apps on top of SATP.

@petermetz please assign it to me. @AndreAugusto11 and I are managing these tasks.

petermetz:
@RafaelAPB Assigned, thank you again!

petermetz:
@AndreAugusto11 @RafaelAPB Please see this change of mine that is pre-emptively (on the main branch) fixing your issue that we talked about this morning: https://github.com/hyperledger/cacti/pull/3047

You can cherry pick it to your branch and then the build should start working (I tested and verified that it does work on your branch when this change is applied).

RafaelAPB:
Thank you for the help @petermetz .Additionally, we will be migrating the issues from my fork to Cacti in the upcoming weeks.

petermetz:
> Thank you for the help @petermetz .Additionally, we will be migrating the issues from my fork to Cacti in the upcoming weeks.

@RafaelAPB AWESOME news! Thank you very much for doing that!!!

AndreAugusto11:
> @AndreAugusto11 @RafaelAPB Please see this change of mine that is pre-emptively (on the main branch) fixing your issue that we talked about this morning: #3047
> 
> You can cherry pick it to your branch and then the build should start working (I tested and verified that it does work on your branch when this change is applied).

@petermetz thank you very much for the help!

I shared with you one branch where the error was visible, but that did not have the entire code. After cherry-picking your commit, there seems to be another error, unrelated to our changes, but also unrelated to the fix in #3047.

The new branch: https://github.com/RafaelAPB/blockchain-integration-framework/pull/73

Error: https://github.com/RafaelAPB/blockchain-integration-framework/actions/runs/8116917052/job/22187970415?pr=73#step:5:1046

Do you have any idea?

petermetz:
> > @AndreAugusto11 @RafaelAPB Please see this change of mine that is pre-emptively (on the main branch) fixing your issue that we talked about this morning: #3047
> > You can cherry pick it to your branch and then the build should start working (I tested and verified that it does work on your branch when this change is applied).
> 
> @petermetz thank you very much for the help!
> 
[...]

PR #3047, opened 2024-03-01
refactor(test-tooling): fix types of streams: use NodeJS.ReadableStream

petermetz:
1. The container management library that we use in the test infrastructure
(called dockerode) is expecting streams that are defined in the global
namespace of the `@types/node` library, e.g. the standard library of NodeJS
itself.
2. Previously we were using the "streams" package to provide type information
to the streams that we were passing around to dockerode and it was working
fine, but after some changes that seem unrelated this has broken the
compilation process.
3. The mentioned changes are not yet on the main branch, but we expect
them to be there soon and so this change is laying the groundwork for that
by pre-emptively fixing the broken build's root cause which is that the
test-tooling package does not declare it's typings related dependencies
correctly: It implicitly uses the NodeJS standard library's types but
so far had not declared them on the package level.
4. This change is therefore to rectify the issue of the `@types/node`
dependency missing from the test-tooling package and also the refactoring
of some of the test ledger classes which were relying on the `streams`
builtin package instead of correctly using the NodeJS.ReadableStream global.
5. Earlier the reasoning for this was that we try to avoid pulling in
types from the global scope because we try to avoid any sort of dependency
on the global scope in general. Once we have proof though that this is
causing issues with the build, then we must give up the principle for
practical reasons (and only in the minimum viable scope, e.g. this does
not change the fact that everywhere else in the codebase we should still
do our best to avoid using the global scoped classes, types, functions,
etc..).

Thank you to @AndreAugusto11 and @RafaelAPB for pointing out this issue
through the pull request of his that is currently being worked on at the
time of this writing:
https://github.com/RafaelAPB/blockchain-integration-framework/pull/72

Related to but does not address #2811

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-test-tooling/package.json
+++ b/packages/cactus-test-tooling/package.json
@@ -101,6 +101,7 @@
     "@types/fs-extra": "9.0.13",
     "@types/js-yaml": "4.0.3",
     "@types/lodash": "4.14.172",
+    "@types/node": "18.11.9",
     "@types/node-forge": "1.3.0",
     "@types/ssh2": "0.5.47",
     "@types/ssh2-streams": "0.1.9",
--- a/packages/cactus-test-tooling/src/main/typescript/common/containers.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/common/containers.ts
@@ -1,5 +1,5 @@
 import path from "path";
-import { Duplex, Stream } from "stream";
+import { Duplex } from "stream";
 import { IncomingMessage } from "http";
 import throttle from "lodash/throttle";
 import { Container, ContainerInfo } from "dockerode";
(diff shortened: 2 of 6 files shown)


RafaelAPB:
One comment: 🥵. In a more serious tone, great work @petermetz and @AndreAugusto11

jagpreetsinghsasan (approved):
LGTM

packages/cactus-test-tooling/package.json
```diff
@@ -101,6 +101,7 @@
     "@types/fs-extra": "9.0.13",
     "@types/js-yaml": "4.0.3",
     "@types/lodash": "4.14.172",
+    "@types/node": "20.11.24",
```
outSH, line 104: Can we use types for node V18 (minimal supported version) instead?

outSH (changes requested):

packages/cactus-test-tooling/package.json
petermetz: @outSH Oops, yes, of course! Sorry for accidentally trying to sneak Node v20 in again... :-)

outSH (approved):

PR #3049, opened 2024-03-03
build: upgrade Yarn from v3.6.0 to v4.1.0

petermetz:
**IMPORTANT**: The lock file format has been updated yet again so you
need to run a `yarn install` after pulling this update on the main branch.

This is not technically a breaking change in the software's APIs but it
is in terms of breaking developer's workflows unless a specific set of
instructions (see above) are performed in order to do the migration.

1. Upgraded Yarn to v4. Hopefully I didn't miss any of the places where
the version was/is declared/documented/etc.
2. Also added a new dependency vulnerability audit script which uses
npm's own audit script/mechanism under the hood. Previously this was
broken if you were using Yarn (e.g. pre-v4 releases of Yarn) but now
we can have the dependency auditing done locally which is a great resource
because we no longer have to depend only on GitHub's dependabot to tell
us when a dependency has some vulnerabilities associated with it.

Point 2 is also the motivation behind the upgrade apart from the usual
reason for trying to keep us on the latest and greatest when it comes
to dependencies so that we don't paint ourselves in a corner when old
dependencies start to have hard to fix vulnerabilities.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


jagpreetsinghsasan (approved):
LGTM, will inform about this to the team once this gets merged

izuru0 (approved):
LGTM

petermetz:
> LGTM, will inform about this to the team once this gets merged

@jagpreetsinghsasan Thank you! Please also note that if the `yarn install` itself fails, clearing all the node_modules folders might help with that.

PR #3054, opened 2024-03-04
test(test-tooling): migrate AIO image to Fabric v2.5.6, add constants

petermetz:
1. The Fabric v2 image has been migrated to the current LTS release
which has increased stability and of course adheres to general best
practices more thoroughly since now we are in sync with the Fabric
maintainers in terms of LTS.
2. The newer versions of the fabric-samples test-net containers have
some of the configuration files under different paths and this had to be
reverse engineered by manually inspecting the containers at runtime and
searching for the same files in different directories. To make this
easier in the future for people who are using the AIO image, we've added
a collection of constants to the test-tooling package that stores the
paths hardcoded that are exported via verbose variable names that pin these
to the specific Fabric version they are related to so that in the future
if these paths change again, we can accommodate the change in a way that
is not too confusing by exporting more variables with slightly different
names and values.
3. The image built from the updated `Dockerfile_v2` is accessible on GHCR
as: `ghcr.io/hyperledger/cactus-fabric2-all-in-one:2024-03-03--issue-2945-fabric-v2-5-6`
which has the v2.5.6 versioned container images pre-cached (embedded) so that
network transfer does not need to occur and rate limiting doesn't produce
CI flakes (DockerHub has rate limits for image downloading that we regularly
hit when we don't embed the Fabric container images this way...).
4. Why can't we just pull the values for these constants directly from the
container at runtime? 
4.1. The list of constants and their names/values change based on the
Fabric version that's being used by the AIO image.
4.2. The only ones that could be pulled are the ones that belong to the
first organization because this is the one that the CLI container uses.
4.3. Configuration files that store the constants' values for the second
organization are also located in different directories depending on the
Fabric version being used.
4.4. The fabric-samples repo have been known to make breaking changes
to older releases which would then make it even harder to debug if we
had logic for identifying the constants that suddenly broke (this specific
incident has happened in the past unfortunately)
4.5. In short, there is definitely a case for applying the DRY principle
here, but in this particular case it appears to be not worth it due to
the increased complexity that the convoluted additional logic would bring.

Related but does not fix https://github.com/hyperledger/cacti/issues/1899

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


izuru0 (approved):
LGTM

packages/cactus-test-tooling/src/main/typescript/fabric/fabric-samples-env-constants.ts
```diff
@@ -0,0 +1,125 @@
...
+  readonly CORE_PEER_LOCALMSPID: string;
+  readonly CORE_PEER_TLS_CERT_FILE: string;
+  readonly CORE_PEER_TLS_KEY_FILE: string;
+  readonly CORE_PEER_ADDRESS: string;
+  readonly CORE_PEER_MSPCONFIGPATH: string;
+  readonly CORE_PEER_TLS_ROOTCERT_FILE: string;
+  readonly ORDERER_TLS_ROOTCERT_FILE: string;
+}
+
+export const FABRIC_25_LTS_FABRIC_SAMPLES__CORE_PEER_TLS_ENABLED_ORG_1 = "true";
```
outSH, line 13: Can't we add a helper method in fabric-test-ledger-v1 class for reading env variables from container and use them instead? We could also add a method that would return ready configs for different orgs (i.e. multiple variables needed for by gateway). That way we're not hardcoding paths in code and are safe in case some of them changes in the future (very unlikely, hence this is just an optional comment, but still to consider).

jagpreetsinghsasan (approved):
LGTM

PR #3072, opened 2024-03-08
build(rust-compiler): retire the container image and the test cases

petermetz:
1. This started off with me trying to fix the CVEs in the rust-compiler image.
2. I was able to get rid of most of them by changing the base image to ubuntu-24.04
3. The remaining high and critical ones were due to wasm-pack so went to see
if we could upgrade that but we are already on the latest which is 8 months old.
The vulnerabilities were reported on the wasm-pack repo 6 months ago along with
a pull request that fixes them, neither the issue nor the pull request fixing it
received any attention from the wasm-pack maintainers which lead me to believe
that it is a liability to depend on it right now and we should instead look into
a different tooling where the maintenance happens to have a little more resources
dedicated to it. Java/Kotlin might be the way to go.
1. I've also looked into possible alternatives to wasm-pack but the only
other tool I found that does the same thing is cargo-web which hasn't had
a new release for 4 years and counting and has even more CVEs plaguing it
than wasm-pack.
1. The official web assembly site links to wasm-pack when it comes to
compiling to it from Rust so there's probably not a better maintained tool
out there, but if someone finds something I'd love to start using it.
In the meantime I'll just archive/retire/delete the rust compiler image
and the tests associated with it because it's a maintenance burden that
we don't need to carry.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


outSH (approved):

jagpreetsinghsasan (approved):
LGTM

PR #3128, opened 2024-03-19
feat(core-api): add IPluginGrpcService type & user-defined type guard

petermetz:
1. This will be used by the upcoming functionality of the API server that
allows all plugins to register their own gRPC services as part of the API
server's own gRPC service.
2. The above mechanism will largely be the same conceptually as the one
we have for HTTP and SocketIO endpoints already.
3. It is optional for plugins to implement gRPC services and therefore
the interface is a standalone one instead of being baked into the more
generic IPluginWebService interface for example.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-core-api/src/main/typescript/plugin/grpc-service/i-plugin-grpc-service.ts
+++ b/packages/cactus-core-api/src/main/typescript/plugin/grpc-service/i-plugin-grpc-service.ts
@@ -1,0 +1,62 @@
+import type {
+  ServiceDefinition,
+  UntypedServiceImplementation,
+} from "@grpc/grpc-js";
+
+/**
+ * Implementers of this interface are responsible for providing a gRPC service
+ * that can be dynamically registered at runtime by the API server.
+ *
+ * It describes what methods a class (plugin) needs to implement
+ * in order to be able to operate as a Cacti gRPC service (e.g. expose its
+ * functionality through gRPC not just HTTP or SocketIO)
+ *
+ * @see {IPluginWebService}
+ * @see {ApiServer}
+ */
+export interface IPluginGrpcService {
+  /**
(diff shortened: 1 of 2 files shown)


RafaelAPB:
@petermetz LGTM. If I may ask, would this service be extendable for protocols like Connect (https://connectrpc.com/docs/node/getting-started/) that are based on gRPC, or we should implement its own service?

petermetz:
> @petermetz LGTM. If I may ask, would this service be extendable for protocols like Connect (https://connectrpc.com/docs/node/getting-started/) that are based on gRPC, or we should implement its own service?

@RafaelAPB Answering in style, question with question... ;-)
Is "Connect" runtime compatible with vanilla gRPC services?
If yes, then "it should just work", otherwise no.

---

To actually provide some useful info: I went and checked their documentation (and the code) and it looks like it could work out of the box because their service registration is based on the same two parameters that vanilla gRPC does (service interface + service implementation goes to the addService method).

I also noticed that they have HTTP translation through an ExpressJS plugin too which is awesome and something that I've been looking into for a while (a few years now...) and it was always a little too complicated for my taste but with this one we could just have it added as a one-liner.

So, what I'll  look into is if we could either use both connect and vanilla, or if it even makes sense to just go with connect. The good thing is that I have integration tests that are already passing on my feature branch so I can very easily go and swap out the gRPC server implementation behind that and see if it would work and if it does then we'd get the HTTP translation for free (which is huge IMO)

outSH (approved):

RafaelAPB:
> ver implementation behind that and see if it wo

Thanks for the insightful response, as always ;) LGTM!

RafaelAPB (approved):

PR #3129, opened 2024-03-19
feat(common): add isGrpcStatusObjectWithCode user-defined type guard

petermetz:
1. User-defined Typescript type-guard function that asserts whether a value or object is a `@grpc/grpc-js` {Partial<StatusObject>} or not. The reason why it checks for {Partial} is because all of the properties of the {StatusObject} are defined as optional for some reason, hence we cannot assume anything about those being present or not by default.
2. Therefore this method will just check if the `code` property is set or not and return `true` or `false` based on that.
3. The above is also the reason why the name of the function is slightly more verbose than your average user-defined type-guard that could be named just "isGrpcStatusObject()" but we wanted to make sure that more specific type-guards can be added later that check for other optional properities or for the presence of all of them together.

Link to the status builder within grpc-js:
https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/status-builder.ts

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-common/package.json
+++ b/packages/cactus-common/package.json
@@ -66,6 +66,7 @@
     "sha3": "2.1.4"
   },
   "devDependencies": {
+    "@grpc/grpc-js": "1.10.3",
     "@types/json-stable-stringify": "1.0.33",
     "@types/sanitize-html": "2.9.5",
     "@types/secp256k1": "4.0.3",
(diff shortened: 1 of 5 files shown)


RafaelAPB (approved):
LGTM

jagpreetsinghsasan (approved):
LGTM

PR #3141, opened 2024-03-23
fix(indy-vdr-nodejs): update dependency version

RafaelAPB:
Previously breaking MacOS builds. See https://github.com/hyperledger/indy-vdr/issues/260

Also updated the copy-sql script at @hyperledger/cactus-plugin-persistence-ethereum@2.0.0-alpha.2 copy-sql due to breaking: https://gist.github.com/RafaelAPB/33f838f0f007770677e90723136f2308; replaced `-a`with `-p`, and additionally creating the `dist` folder if it does not exist. .
Although `-a` is a better technical solution, it breaks MacOS builds.

Source: https://unix.stackexchange.com/questions/43605/how-do-i-copy-a-folder-keeping-owners-and-permissions-intact


RafaelAPB:
cc @outSH

outSH (approved):
LGTM, thank you!

petermetz (approved):

PR #3169, opened 2024-04-03
fix(connector-ethereum): cannot serialize bigint

eduv09:
I came across an issue working with the  ledger-connector-ethereum.
Description: In the endpoint InvokeRawWeb3EthMethodV1 , some web3.eth functions are returning objects with BigInt attributes, which causes express Response.json() method to break when stringuifying those objects "TypeError: Do not know how to serialize a BigInt".

My fix was to include code in the endpoint file, to substitute BigInt attributes for their toString() value. 

How to reproduce:
 - checkout to this branch [connector-ethereum-bug](https://github.com/eduv09/blockchain-integration-framework/tree/connector-ethereum-bug);
 - run 'geth-contract-deploy-and-invoke-using-keychain-v1.test.ts'. I've added code (few lines, check last commit in that branch) in this test file that reproduces the error 

@petermetz @outSH would you have a look please? If you think there is a more suitable fix, let me know.

### DIFF
--- a/packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
+++ b/packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
@@ -304,6 +304,9 @@ export class PluginLedgerConnectorEthereum
     app: Express,
     wsApi: SocketIoServer,
   ): Promise<IWebServiceEndpoint[]> {
+    // Add custom replacer to handle bigint responses correctly
+    app.set("json replacer", this.stringifyBigIntReplacer);
+
     const { logLevel } = this.options;
     const webServices = await this.getOrCreateWebServices();
     await Promise.all(webServices.map((ws) => ws.registerExpress(app)));
@@ -1182,5 +1185,15 @@ export class PluginLedgerConnectorEthereum
     return methodRef(...contractMethodArgs)[args.invocationType](
       args.invocationParams,
     );
+  }
+
+  stringifyBigIntReplacer(
+    _key: string,
+    value: bigint | unknown,
+  ): string | unknown {
+    if (typeof value === "bigint") {
+      return value.toString();
+    }
+    return value;
   }
 }
--- a/packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/web-services/invoke-raw-web3eth-method-v1-endpoint.ts
+++ b/packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/web-services/invoke-raw-web3eth-method-v1-endpoint.ts
@@ -88,6 +88,7 @@ export class InvokeRawWeb3EthMethodEndpoint implements IWebServiceEndpoint {
     try {
       const methodResponse =
         await this.options.connector.invokeRawWeb3EthMethod(req.body);
+
       const response: InvokeRawWeb3EthMethodV1Response = {
         status: 200,
         data: methodResponse,


RafaelAPB:
LGTM

petermetz (commented):
@outSH I remember you added this to `api-server.ts` to handle the same exact scenario, do you think that it stopped working at some point? It seems like this PR shouldn't be necessary on account of what you wrote that should cover all endpoints:

```typescript
    // Add custom replacer to handle bigint responses correctly
    app.set("json replacer", this.stringifyBigIntReplacer);
```

petermetz (changes requested):
@eduv09 Please 
1. sync up the commit message and the PR description
2. use the conventional commit message format - I think `bugfix` is not a valid type but `fix` is.

RafaelAPB:
It seems that it would be better for the fix to be at the connector level because such a connector can be used as a dependency in other plugins, which would still have the serialization issue. cc @eduv09 

cc @eduv09

eduv09:
@petermetz @RafaelAPB @outSH I added the replacer in the connector. I saw that we are doing something simmillar in the iroha2 connector. I think this fixes the issue, without creating problems to other plugins.

Check if the commit message is adequate, or if something should be mentioned or removed. Thank you.

petermetz:
> > @outSH I remember you added this to `api-server.ts` to handle the same exact scenario, do you think that it stopped working at some point? It seems like this PR shouldn't be necessary on account of what you wrote that should cover all endpoints:
> > ```ts
> >     // Add custom replacer to handle bigint responses correctly
> >     app.set("json replacer", this.stringifyBigIntReplacer);
> > ```
> 
> @petermetz @eduv09 @RafaelAPB Yes, it should work but only when the connector is used as a api-server plugin. When it's used as standalone (like it's done in tests for simplicity), the fix does not apply. 

@outSH @eduv09 @RafaelAPB  Oh yeah I forgot about that! /facepalm Thanks for the reminder. There is a chicken-egg problem here for sure due to us wanting to avoid circular dependencies between plugins and the API server.
My suggestion to break it is like this:
1. Move the `stringifyBigIntReplacer` to a package designed to be shared by both the API server and the plugins: the `cactus-core` package.
2. Also add a new function to `cactus-core`, something called `configureExpressApp(app: Express): Promise<void>`. Then move all the common set-up logic (that is independent of specific plugins) of the API server to this new function and make api-server.ts import and use that function from `cactus-core`. 
3. Then the connector can import the exact same  function from `cactus-core` and have a one-liner to set up the express instance identical to the API server's, butt without having to duplicate any of that code.
4. Now we don't need the change that is in *this* pull request.

So the above steps is what I recommend for a long term solution. If you prefer to go fast I'm OK with merging this PR as-is and then sending a follow-up.

petermetz:
FYI: here's a suggestion on how to implement the common function outside of the API server package (in the core package)
https://github.com/hyperledger/cacti/pull/3186

If that linked PR gets merged then afterwards we can refactor both the connector plugin tests and the API server to use that implementation.

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
```diff
@@ -304,6 +304,9 @@ export class PluginLedgerConnectorEthereum
     app: Express,
     wsApi: SocketIoServer,
   ): Promise<IWebServiceEndpoint[]> {
+    // Add custom replacer to handle bigint responses correctly
+    app.set("json replacer", this.stringifyBigIntReplacer);
+
```
petermetz, line 309: @eduv09 This is one step in the right direction but now that I realize what the whole picture was: This is actually adding another issue, meaning that this change will make it work in the test environment but when it's used in production (e.g. through the API server) then it will conflict with the API server's own big int replacer implementation. So, the solution IMO is to move this json replacer registration code to the test cases themselves which then guarantees that it only happens where it's needed and doesn't cause trouble in the other scenario (API server usage)

petermetz (changes requested):

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
eduv09: Ok. In that case maybe we can close this PR, and leave this issue to be solved in the future the way you suggested in the previous comment (long term solution).
In the meantime I will add the replacer in my tests that use the connector

petermetz: @eduv09 OK, that works for me too, thanks and sorry for the back and forth!

PR #3183, opened 2024-04-04
feat(cmd-api-server): add ConnectRPC auto-registration for plugins

petermetz:
1. This is enabling plugins to expose their operations via ConnectRPC
services which is very similar to gRPC but it comes with a few extra
bells and whistles that can come in very handy.
2. There is an upcoming pull request that makes it so that the keychain
memory plugin implements and registers its services via this newly added
hook of the API server. The importance of this is that test coverage for
the code in this commit resides on another branch, meaning that even though
there are no new test cases on this branch, the feature has been extensively
tested and there is test-automation in place to continue verifying it
as well.
3. The main difference between the hook methods are that for CRPC the
API server expects an array of service definition+implementation pairs
instead of just a single one. This was a design decision forced by the
issues with implementing separate services in a single class: The compiler
was hard to appease in a way that kept the code clean. gRPC did not suffer
from this and therefore the registration methods defined for that only
return a single gRPC service defintion+implementation pair which can combine
any number of .proto services.

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

**Pull Request Requirements**
- [x] Rebased onto `upstream/main` branch and squashed into single commit to help maintainers review it more efficient and to avoid spaghetti git commit graphs that obfuscate which commit did exactly what change, when and, why.
- [x] Have git sign off at the end of commit message to avoid being marked red. You can add `-s` flag when using `git commit` command. You may refer to this [link](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) for more information.
- [x] Follow the Commit Linting specification. You may refer to this [link](https://www.conventionalcommits.org/en/v1.0.0-beta.4/#specification) for more information. 

**Character Limit**
- [x] Pull Request Title and Commit Subject must not exceed 72 characters (including spaces and special characters).
- [x] Commit Message per line must not exceed 80 characters (including spaces and special characters).

**A Must Read for Beginners**
For rebasing and squashing, here's a [must read guide](https://github.com/servo/servo/wiki/Beginner's-guide-to-rebasing-and-squashing) for beginners.

PR #3184, opened 2024-04-04
feat(plugin-keychain-memory): add ConnectRPC support

petermetz:
1. This leverages the newly introduced methods in core-api that the API
server is using to probe if a plugin has ConnectRPC support or not.
2. There is support for both HTTP 1.1 and HTTP 2. The caveat here is that
HTTP 2 is not supported by ExpressJS so we pulled in Fastify to handle
those type of requests and that means that HTTP 2 ConnectRPC traffic has
to go through a different port compared to the HTTP 1.1 ConnectRPC traffic.
3. The lesson here is that we probably need to migrate away from ExpressJS
longer term because it does not (and from the looks of it will not ever)
support HTTP 2 which is probably going to be a bit of technical debt/
limiting factor in architectural decisions going forward for both Cacti
maintainers and Cacti users.
4. A new code generator has been introduced by this commit as well which
is @buf/build - the tool where ConnectRPC originates from. The scripts are
structured in such a way that this should be seamlessly integrated into the
existing `codegen` root level script and therefore also the CI.
5. There is test coverage for both HTTP 1.1 and HTTP 2 traffic in the file at
```sh
packages/cactus-test-plugin-keychain-memory/src/test/typescript/integration/
test-keychain-memory-crpc-api-server.test.ts
```
6. The test case referenced above is also the example on how to use the
ConnectRPC client (very similar to the HTTP client we already had before)

Depends on #3183

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


jagpreetsinghsasan (approved):
LGTM

outSH (approved):

PR #3189, opened 2024-04-05
feat(plugin-keychain-memory): add observability via RxJS ReplaySubjects

petermetz:
1. This is an example of how to add observability to a plugin such as
if you had to somehow expose the stream of transaction execution requests
flowing through a connector plugin but did not feel like setting up Kafka
or RabbitMQ just for this and instead opted to do it with an in-process,
purely NodeJS/Javascript based solution.
2. The downside of this is of course that this doesn't work well in a
distributed computing environment just by itself, since if you were to
host a fleet of servers running the same connector plugin with horizontal
scaling, then this wouldn't be able to observe all the invocations across
the server fleet, but it would still make it easier to implement a functionality
like that.
3. The main purpose of this pull request is educational. The keychain memory
plugin is only used for testing and demonstration purposes and I wanted to
show to a few other contributors what I meant when I was explaining that
they could just use RxJS subjects to allow consumers of the connector plugins
to observe the stream of transactions flowing through said connector plugin
instance.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-plugin-keychain-memory/package.json
+++ b/packages/cactus-plugin-keychain-memory/package.json
@@ -64,6 +64,7 @@
     "axios": "1.6.0",
     "express": "4.19.2",
     "prom-client": "13.2.0",
+    "rxjs": "7.8.1",
     "uuid": "9.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


outSH (changes requested):
@petermetz 
It looks great, but I'm not sure if it's worth adding empty features that increases complexity of the code, even in a plugin that is used only for tests. Maybe we could add this as sample code in documentation? Like `docs/samples/internal/rxjs-observability.ts` or alike? If not, and you sure you want to introduce it here - I'll aprove it :)

RafaelAPB:
I think this is good because other plugins need this feature (namely a plugin to do cross-chain security being developed now). This way we can reduce more complex dependencies such as RabitMQ. However, @outSH might be right in the sense that documentation might suffice. 
Thanks for the code @petermetz.

petermetz:
@RafaelAPB @outSH I'd want to keep it in a code package instead of a docs folder because otherwise it's not getting compiled & tested with the rest of the project and as the code rusts it breaks over time and then the documentation becomes wrong. 
With that said, I agree with the larger point and what I'm thinking is that the entire package could be renamed to something else that better signifies that it's not for production, I just don't know what that name would be, but I'm open to trying to come up with something for sure.

outSH (approved):

PR #3196, opened 2024-04-08
feat(bungee-hermes): view proofs & ethereum strategy

eduv09:
This is an update for the bungee-hermes plugin. New features and changes:
   - Views have new attribute viewProof, with two attributes. One is the merkle root obtained by cunstructing a tree using all stateProofs in the view, and the other is the merkle root obtained by constructing a tree using all transactionProofs in the view
   -  Added new endpoint to verify integrity of view proofs (receives set of state/transaction proofs and checks if the provided root is valid).
   -  Created strategy for bungee-hermes to support cactus-ethereum-connector
   -  Created new test for the strategy-ethereum, and altered the API test to test the new viewProof endpoint
   -  Updated README.md documentation to reflect new changes.


eduv09:
@RafaelAPB I think this addresses all the improvements/requirements we discussed. Let me know any changes you see fit for this one, thank you !

packages/cactus-plugin-bungee-hermes/src/main/typescript/strategy/strategy-fabric.ts
```diff
@@ -217,7 +236,7 @@ export class StrategyFabric implements ObtainLedgerStrategy {
         new Transaction(
           txId,
           ts,
-          new TransactionProof(new Proof({ creator: "" })),
+          new TransactionProof(new Proof({ creator: "" }), txId),
```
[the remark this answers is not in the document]
eduv09: It is not clear, but in line 69 we set the creator, after we poll for the transaction receipt. I can leave a comment with this info.

packages/cactus-plugin-bungee-hermes/src/main/typescript/strategy/strategy-fabric.ts
```diff
@@ -65,11 +66,16 @@ export class StrategyFabric implements ObtainLedgerStrategy {
...
-
+        tx.getProof().setCreator(
+          new Proof({
+            creator: receipt.transactionCreator.creatorID,
+            mspid: receipt.transactionCreator.mspid,
+          }),
+        );
         assetValues.push(JSON.parse(receipt.rwsetWriteData).Value.toString());
         tx.setStateId(assetKey);
         //tx.setPayload(); //FIXME check what to assign here
```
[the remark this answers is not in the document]
eduv09: I didn't see it being relevant at the time, but i will set it to be the serialized transaction obtained through the contract (Context.stub).

packages/cactus-plugin-bungee-hermes/src/main/typescript/strategy/strategy-ethereum.ts
```diff
@@ -0,0 +1,228 @@
...
+    values: string[];
+    blocks: Map<string, EvmBlock>;
+  }> {
+    const filter = {
+      fromBlock: "earliest",
+      toBlock: "latest",
+      address: networkDetails.contractAddress,
+      topics: [null, Web3.utils.keccak256(key)], //filter logs by asset key
+    };
+    const getLogsReq: InvokeRawWeb3EthMethodV1Request = {
```
outSH, line 162: BTW - ethereum connector exposes JSON-RPC proxy so you can use web3 client directly instead of these endpoints (which also heavily depend on the web3js release used by the connector), see https://github.com/hyperledger/cacti/tree/main/packages/cactus-plugin-ledger-connector-ethereum#json-rpc-proxy

outSH: Using "raw" endpoints works as well but proxy might be a better alternative in your future contributions :)

outSH (approved):
LGTM, great work!

packages/cactus-plugin-bungee-hermes/src/main/typescript/strategy/strategy-ethereum.ts
eduv09: Thank you! I will look into this.

packages/cactus-plugin-bungee-hermes/src/main/typescript/strategy/strategy-ethereum.ts
```diff
@@ -0,0 +1,228 @@
...
+      signingCredential: networkDetails.signingCredential,
+    };
+    const response = await api.invokeContractV1(
+      parameters as InvokeContractV1Request,
+    );
+    if (response != undefined) {
+      return response.data.callOutput;
+    }
+
+    return "response undefined";
```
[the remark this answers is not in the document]
eduv09: changed it to empty array. Also changed the function return type.

The same code was present in the Besu strategy, so i changed that as well

RafaelAPB (approved):

packages/cactus-plugin-bungee-hermes/src/main/typescript/strategy/strategy-ethereum.ts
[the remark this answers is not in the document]
petermetz: @eduv09 Is there any version of events where the `callOutput` being falsy is a valid outcome? I'm guessing not and if that is the case then I recommend throwing an error instead. This prevents domino effect type of bugs that manifest somewhere else that we'll then have to trace back somehow to this point in the code (where it all started going wrong). If you just throw instead then it will fail at the earliest known point when something was going sideways. They call this principle "fail fast".

petermetz (changes requested):

packages/cactus-plugin-bungee-hermes/src/main/typescript/strategy/strategy-ethereum.ts
```diff
@@ -0,0 +1,223 @@
...
+      signingCredential: networkDetails.signingCredential,
+    };
+    const response = await api.invokeContractV1(
+      parameters as InvokeContractV1Request,
+    );
+    if (response != undefined && response.data.callOutput != undefined) {
+      return response.data.callOutput;
+    }
+
+    throw Error(
```
petermetz, line 140: @eduv09 Please instantiate with the `new` keyword.

packages/cactus-plugin-bungee-hermes/src/main/typescript/strategy/strategy-ethereum.ts
```diff
@@ -0,0 +1,223 @@
...
+    const getLogsReq: InvokeRawWeb3EthMethodV1Request = {
+      methodName: "getPastLogs",
+      params: [filter],
+    };
+    const response = await api.invokeWeb3EthMethodV1(getLogsReq);
+    if (
+      response == undefined ||
+      !(response.status >= 200 && response.status <= 299)
+    ) {
+      throw Error("Error on getting past logs");
```
petermetz, line 169: @eduv09 Looking a little deeper at the error handling: Please include the root cause here as well. `RuntimeError` can have a `cause` property which can be another `Error` or a `string`. The way you throw this error completely hides and swallows the original problem and will make people who have to debug this in production be disappointed when all they see in the error log is `Error on getting past logs` and literally nothing else. Ideally you'd break it down to multiple different checks like an if condition for the response being falsy (use the ! operator to check for falsyness instead of `== undefined` ) and another if condition for checking if the status code is in the valid range. Then you can throw different errors with d ifferent messages that are way more informative that way. 
Please also make sure to include the status code if it wasn't in the valid range. There could be valuable information hidden there about the nature of the problem that the code is having. 

For example if the return codes are 4xx then it's user error and you should throw an error accordingly. If it's 5xx, then it's supposedly a bug in our code.

Take a look at the besu connector's deploy contract and how I refactored it to use these principles a while back:

```typescript
public async deployContract(
    req: DeployContractSolidityBytecodeV1Request,
  ): Promise<DeployContractSolidityBytecodeV1Response> {
    const fnTag = `${this.className}#deployContract()`;
    Checks.truthy(req, `${fnTag} req`);
    if (isWeb3SigningCredentialNone(req.web3SigningCredential)) {
      throw createHttpError[400](
        `${fnTag} Cannot deploy contract with pre-signed TX`,
      );
    }
    const { keychainId, contractName } = req;
    if (!keychainId || !req.contractName) {
      const errorMessage = `${fnTag} Cannot deploy contract without keychainId and the contractName.`;
      throw createHttpError[400](errorMessage);
    }

    const keychainPlugin = this.pluginRegistry.findOneByKeychainId(keychainId);

    if (!keychainPlugin) {
      const errorMessage =
        `${fnTag} The plugin registry does not contain` +
        ` a keychain plugin for ID:"${req.keychainId}"`;
      throw createHttpError[400](errorMessage);
    }

    if (!keychainPlugin.has(contractName)) {
      const errorMessage =
        `${fnTag} Cannot create an instance of the contract instance because` +
        `the contractName in the request does not exist on the keychain`;
      throw new createHttpError[400](errorMessage);
    }
```

petermetz (changes requested):
@eduv09 Sorry for the multiple iterations, I keep finding things that I highly recommend you change because it will increase the stability of the code and the ease at which problems with it can be fixed in the future so I hope it's OK, these are long term investments that will pay lots of dividends in the future.

packages/cactus-plugin-bungee-hermes/src/main/typescript/strategy/strategy-ethereum.ts
eduv09: Thank you @petermetz ! I changed the error handling in all strategies taking into consideration this comment. Hopefully it is being accurately handled this time.
Let me know any improvements/corrections you see fit.

RafaelAPB:
LGTM, can we merge this?

RafaelAPB:
@petermetz is the PR ready to merge?

petermetz (approved):
LGTM - sorry for the long wait!

Issue #3106, opened 2024-03-15
Implement BUNGEE plugin


RafaelAPB:
Create a BUNGEE plugin (a base already exists). This includes the creation of views in Hyperledger Fabric and Hyperledger Besu for a certain state, and persisting the views. [Code is here](https://github.com/RafaelAPB/blockchain-integration-framework/blob/0c7d5dd87f3acca7db72b322674f8cab0ee5f8c5/packages/cactus-plugin-bungee/src/main/typescript/plugin-bungee.ts#L1) / under packages/cactus-plugin-bungee/src/dev/typescript/plugin-bungee.ts

See [paper/specification](https://dl.acm.org/doi/10.1145/3643689).

Requirements:
1. Plugin should be given a state ID and create a view in a certain timeframe, according to the BUNGEE paper (CREATE view).
2. Plugin should receive options to process/merge/integrate views (different modes) (MODIFY view). 
3.  The view is signed by the view generator and sent to the gateway, who then signs it. 
4. Add respective endpoints to create a view given the relevant input (transaction ids, state ids, time etc)
5. Update prometheus metrics (number of created, number of states processed, etc)

eduv09:
@RafaelAPB please assign me this one

RafaelAPB:
Done

PR #3250, opened 2024-05-08
feat(bungee-hermes): process & merge views

eduv09:
BUNGEE-Hermes is supposed to be able to process and merge views, following 'privacy-policies' (which are basicaly functions that alter the data included in a View/IntegratedView). 
The algorithm of Merging views is detailed  in the paper [BUNGEE](https://dl.acm.org/doi/pdf/10.1145/3643689).

Summary of Changes:
- Added new types for view merging
- Created some merge and process policies for demonstration
- new endpoints and plugin logic to merge and process views
- two new tests to test new functionality
- utility function to fully deserialize a view
- slightly changed View type, to keep record of its old versions metadata
- updated README.md

### DIFF
--- a/packages/cactus-plugin-bungee-hermes/src/main/typescript/plugin-bungee-hermes.ts
+++ b/packages/cactus-plugin-bungee-hermes/src/main/typescript/plugin-bungee-hermes.ts
@@ -24,9 +24,15 @@ import { Server as SecureServer } from "https";
 import {
   CreateViewRequest,
   CreateViewResponse,
+  MergePolicyOpts,
+  MergeViewsRequest,
+  MergeViewsResponse,
+  PrivacyPolicyOpts,
+  ProcessViewRequest,
 } from "./generated/openapi/typescript-axios";
 import { Snapshot } from "./view-creation/snapshot";
 import { View } from "./view-creation/view";
+import { IntegratedView } from "./view-merging/integrated-view";
 import {
   NetworkDetails,
   ObtainLedgerStrategy,
@@ -36,6 +42,12 @@ import { GetPublicKeyEndpointV1 } from "./web-services/get-public-key-endpoint";
 import { GetAvailableStrategiesEndpointV1 } from "./web-services/get-available-strategies-endpoint";
 import MerkleTree from "merkletreejs";
 import { VerifyMerkleRootEndpointV1 } from "./web-services/verify-merkle-root-endpoint";
+import { MergePolicies } from "./view-merging/merge-policies";
+import { deserializeView } from "./utils";
+import { MergeViewsEndpointV1 } from "./web-services/merge-views-endpoint";
+import { ProcessViewEndpointV1 } from "./web-services/process-view-endpoint";
+
+import { PrivacyPolicies } from "./view-creation/privacy-policies";
 
 export interface IKeyPair {
   publicKey: Uint8Array;
@@ -64,6 +76,10 @@ export class PluginBungeeHermes implements ICactusPlugin, IPluginWebService {
 
   private strategies: Map<string, ObtainLedgerStrategy>;
 
+  private mergePolicies: MergePolicies = new MergePolicies();
+
+  private viewPrivacyPolicies: PrivacyPolicies = new PrivacyPolicies();
+
   private level: LogLevelDesc;
   private endpoints: IWebServiceEndpoint[] | undefined;
 
@@ -74,6 +90,7 @@ export class PluginBungeeHermes implements ICactusPlugin, IPluginWebService {
 
     this.level = options.logLevel || "INFO";
     this.strategies = new Map<string, ObtainLedgerStrategy>();
+
     const label = this.className;
     const level = this.level;
 
@@ -156,12 +173,20 @@ export class PluginBungeeHermes implements ICactusPlugin, IPluginWebService {
     const verifyMerkleProofEndpoint = new VerifyMerkleRootEndpointV1({
       bungee: this,
     });
+    const mergeViewsEndpoint = new MergeViewsEndpointV1({
+      bungee: this,
+    });
+    const processViewEndpoint = new ProcessViewEndpointV1({
+      bungee: this,
+    });
 
     this.endpoints = [
       viewEndpoint,
       pubKeyEndpoint,
       availableStrategiesEndpoint,
       verifyMerkleProofEndpoint,
+      mergeViewsEndpoint,
+      processViewEndpoint,
     ];
     return this.endpoints;
   }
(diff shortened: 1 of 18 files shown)


packages/cactus-plugin-bungee-hermes/src/main/typescript/view-creation/view.ts
```diff
@@ -2,12 +2,29 @@ import { v4 as uuidV4 } from "uuid";
 import { Snapshot } from "./snapshot";
 import MerkleTree from "merkletreejs";
 import { Transaction } from "./transaction";
+import { ViewPrivacyPolicy } from "./privacy-policies";
+import * as CryptoJS from "crypto-js";
+
+export interface ViewMetadata {
```
petermetz, line 8: @eduv09 Please prefix interface names with `I`

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-creation/view.ts
```diff
@@ -85,4 +132,31 @@ export class View {
     });
     return transactions;
   }
+  public setPrivacyPolicyValue(
+    value:
+      | {
```
petermetz, line 137: @eduv09 I recommend extracting the inline type definition into it's own type definition, like an interface called `ISetPrivacyPolicyValue` or similar.

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-merging/integrated-view.ts
```diff
@@ -0,0 +1,162 @@
...
+    transactionsMerkleRoot: string;
+    statesMerkleRoot: string;
+    viewsMerkleRoot: string;
+  };
+  private privacyPolicy: { id: string; policyHash?: string };
+  constructor(privacyPolicyId: string, privacyPolicy: MergePolicy | undefined) {
+    this.stateList = new Map<string, ExtendedState>();
+    this.id = uuidV4();
+    //these are dummy values, they are changed in the process of merging Views
+    this.tI = "999999999999999";
```
petermetz, line 30: @eduv09 Question: Is there a non-zero probability that these initial values would ever be valid values? If yes, I'd go for something else that has zero probability of becoming a valid value, like a `-` character or negative integers, etc. (whichever it is)

petermetz (changes requested):

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-creation/privacy-policies.ts
```diff
@@ -0,0 +1,15 @@
+import { View } from "./view";
+
+export interface ViewPrivacyPolicy {
+  (view: View, ...args: string[]): View;
+}
+export class PrivacyPolicies {
+  constructor() {}
+
+  public hideState(view: View, stateId: string): View {
```
RafaelAPB, line 9: can you rename this to "prune" (like in the paper)

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-creation/view.ts
```diff
@@ -2,12 +2,29 @@ import { v4 as uuidV4 } from "uuid";
...
+  signature: string;
+}
+
 export class View {
   private key: string;
   private snapshot: Snapshot;
   private tI: string;
   private tF: string;
   private participant: string;
+  private creator: string = "";
```
RafaelAPB, line 25: this should be undefined at the beginning (we should not allow an empty creator)

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-creation/view.ts
RafaelAPB: +1

packages/cactus-plugin-bungee-hermes/src/main/typescript/plugin-bungee-hermes.ts
```diff
@@ -261,4 +354,128 @@
...
+        Object.getPrototypeOf(this.viewPrivacyPolicies as any),
+        policyId,
+      )
+    ) {
+      throw Error(
+        "The requested Privacy Policy " + policyId + "does not exist.",
+      );
+    }
+
+    view = (this.viewPrivacyPolicies as any)[policyId](view, ...args);
```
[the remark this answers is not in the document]
RafaelAPB: please add enum / switch that are called with users input instead of user calling method directly

packages/cactus-plugin-bungee-hermes/src/test/typescript/integration/bungee-merge-views.test.ts
```diff
@@ -0,0 +1,402 @@
+import {
+  IListenOptions,
+  LogLevelDesc,
+  LoggerProvider,
+  Secp256k1Keys,
+  Servers,
+} from "@hyperledger/cactus-common";
+import "jest-extended";
```
RafaelAPB, line 8: can you please add tests with the privacy policies?

packages/cactus-plugin-bungee-hermes/package.json
```diff
@@ -63,6 +63,7 @@
     "@hyperledger/cactus-plugin-ledger-connector-fabric": "2.0.0-alpha.2",
     "axios": "1.6.0",
     "body-parser": "1.20.2",
+    "crypto-js": "4.2.0",
```
RafaelAPB, line 66: if possible to use crypto functions from Cacti common use them instead of adding a new dep.

RafaelAPB (changes requested):

packages/cactus-plugin-bungee-hermes/package.json
eduv09: I think there is no utility function to hash in cactus common. I just use the module to hash the privacy policy.

I can change the dependency/function to the default one use in cactus-common perhaps? like 'sha3' module

packages/cactus-plugin-bungee-hermes/src/test/typescript/integration/bungee-merge-views.test.ts
eduv09: The test tests one of them. should I add the other ?

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-merging/integrated-view.ts
eduv09: I changes tF to a negative value, and tI to the max possible value. I also changed the function I was using parteInt by BigInt in timestamp handling, since most blockchains use bigints as timestamps

packages/cactus-plugin-bungee-hermes/package.json
eduv09: @RafaelAPB @petermetz I changed to a hashData function from private to public in the JsObjectSigner object, in the cactus-common package, so that i can use that instead of the extra dependency

petermetz: @eduv09 @RafaelAPB Works for me, thank you!

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-merging/integrated-view.ts
petermetz: @eduv09 The probability of "the max possible value" being a valid value is 100% so I highly recommend against using it as an initial value. Instead what I recommend is something intentionally invalid/tainted so that it can **always** immediately be distinguished from any possible valid values that were set during the lifetime of the object (post-initialization).

Here is an example:

0. Your code is running in production.
1. You create your object, values get initialized with the "9999..." thing.
2. Somebody somewhere sets the values to something  (you don't know what because the logs didn't say)
3. Now the object gets used to serve some request that crashes.
4. You get a stack trace and a message that says something was wrong and the `tI` variable was the "9999..." value (same as the init value)
5. You now have no way of knowing if something bad happened during the objects configuration (e.g. they failed to set the value to what they wanted and it was left on the initial value accidentally **OR** if something bad happened and they  did not set the value at all due to some failure of their own. In this context "them" refers to other developers who use your class for their own functionalities.
6. Not all of this edge case may be applicable to this code but I wanted to lay out the general principle why you want to set those to intentionally invalid values when possible (not always possible due to other things/circumstances so that's also a valid reason for not doing it if that is how it is in your case)

petermetz (approved):
@eduv09 LGTM - if you feel like the reasoning makes sense then please also apply this change if you have enough time: https://github.com/hyperledger/cacti/pull/3250#discussion_r1597125386

RafaelAPB (approved):

eduv09:
@petermetz I applied the changes. Both start as negative values now. Thank you!

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-merging/integrated-view.ts
petermetz: @eduv09 Thank you for the updates!

petermetz:
> @petermetz I applied the changes. Both start as negative values now. Thank you! I think we can merge it now.

@eduv09 Got it, thank you! I'll try to merge it today!

AndreAugusto11:
closes #3106

PR #3288, opened 2024-06-01
build: add OpenAPI spec bundling, stop using URL references

petermetz:
**IMPORTANT:** From now on, if you are changing the OpenAPI specification of any given
package within Cacti, please make sure to edit the template file instead of editing the
openapi.json specific file directly because changes in the openapi.json file will be
overwritten by the codegen script the next time you run it.
This slight alteration in the development flow is the least intrusive solution I could find
to resolving our issues with the release automation.

This change enables us to have our openapi.json files work without having remote and URL
references in them (which was a blocker issue for release automation).

1. The openapi.json files that we used to have are now called openapi.tpl.json where the
tpl stands for template. Their content is equivalent to what openapi.json files used to
have prior to this commit.
2. These template specs are fed into the bundler tool which then spits out the files which
then are saved as openapi.json files. The big change is that these bundled versions are
no longer containing any remote nor URL references, only local ones.
3. This means that we still get project-wide re-use of schema types from packages such as
cactus-core-api, but we no longer suffer from the additional complexities of having to deal
with remote and URL references.
4. The scirpt that performs the bundling is callable separately by executing this command
```sh
yarn tools:bundle-open-api-tpl-files
```
5. The `yarn tools:bundle-open-api-tpl-files` is also embedded as a warmup step of the
larger `codegen` script so there is no need usually to call the bundling script separately.
6. The heavylifting in terms of bundling is done by the tooling script that can be found
here: `tools/bundle-open-api-tpl-files.ts`. On a high level what it does is loop through
existing `openapi.tpl.json` files throughout the project and then renders their bundled
version next to it as `openapi.json` which then can be used by all of our tools as a self
contained version of the template file which *does* still have the remote and URL references
in it.

More information on what URL and remote references are can be read here on the official
OpenAPI website: https://swagger.io/docs/specification/using-ref/

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


jagpreetsinghsasan (approved):
LGTM. The local references are much better in my opinion too

petermetz:
@hyperledger/cacti-maintainers I've found a more efficient and less brittle way to do this type of vendoring automatically by way of this concept called bundling the open API specifications. It was originally a feature of the (now deprecated) swagger-cli tool, but the successor of it also implements it and it appears to work seamlessly based on my tests. I'll send an update to this pull request shortly with the refactored version which will be slightly less verbose but equivalent functinality.

petermetz:
@hyperledger/cacti-maintainers I've pushed the latest changes and now this is once again ready for your review. I made sure to update the pull request description with the new changes, please re-read carefully!

tools/bundle-open-api-tpl-files.ts
```diff
@@ -0,0 +1,262 @@
...
+const modulePath = path.resolve(fileURLToPath(import.meta.url));
+const isRunningDirectlyViaCLI = nodePath === modulePath;
+
+const main = async (argv: string[], env: NodeJS.ProcessEnv) => {
+  const req = await createRequest(argv, env);
+  await bundleOpenApiTplFiles(req);
+};
+
+if (isRunningDirectlyViaCLI) {
+  main(process.argv, process.env);
```
outSH, line 42: Could you catch and pretty print any uncaught exceptions here?

tools/bundle-open-api-tpl-files.ts
```diff
@@ -0,0 +1,262 @@
...
+  return req;
+}
+
+function log(
+  req: IBundleOpenApiTplFilesV1Request,
+  message: string,
+  ...args: unknown[]
+): void {
+  if (req.verbose) {
+    console.log(message, ...args);
```
outSH, line 139: Could you use some 3'rd party library for logging and set `DEBUG` when verbose flag was provided? This would allow more granular messaging in the future (in case script causes any troubles and need to be tampered with in the future).

packages/cactus-plugin-keychain-aws-sm/src/main/json/openapi.tpl.json
```diff
@@ -0,0 +1,179 @@
...
+          "http": {
+            "path": "/api/v1/plugins/@hyperledger/cactus-plugin-keychain-aws-sm/has-keychain-entry",
+            "verbLowerCase": "post"
+          }
+        },
+        "operationId": "hasKeychainEntryV1",
+        "summary": "Checks that an entry exists under a key on the keychain backend",
+        "parameters": [],
+        "requestBody": {
+          "$ref": "https://raw.githubusercontent.com/hyperledger/cactus/v2.0.0-alpha.2/packages/cactus-core-api/src/main/json/openapi.json#/components/requestBodies/keychain_has_entry_request_body"
```
outSH, line 135: Shouldn't theese be replaced with local reference? Is this an oversight or left by intention?

outSH (approved):
@petermetz Looks great, sorry for late review. I've left few comment but none of them are critical, you can look into them in future PR or here if there's no rush :)

packages/cactus-plugin-keychain-aws-sm/src/main/json/openapi.tpl.json
petermetz: > Shouldn't theese be replaced with local reference? Is this an oversight or left by intention?

@outSH I thought about doing that change at the same time, but it would've increased the diff even more so I decided to try and make it easier to review and broken it up to two separate RPs with the second one depending on this one getting merged first. Here's the link to that one (and I'd love a review on that one as well ASAP :-))

https://github.com/hyperledger/cacti/pull/3315

tools/bundle-open-api-tpl-files.ts
petermetz: @outSH I stoppped doing that as a pattern because it appears that the error message gets propagated corerctly regardless. With that said, I agree it's probably better to be extra cautious because it is really annoying when an error gets swallowed! I'll updated it in a second!

alt text

tools/bundle-open-api-tpl-files.ts
petermetz: @outSH Thank you for the great suggestion! I've made the changes just now, it uses the library called "debug"

tools/bundle-open-api-tpl-files.ts
petermetz: @outSH I updated the code, now the catch handler does log any uncaught exceptions before exiting with 1

PR #3291, opened 2024-06-01
build(yarn): add plugin-interactive-filter to manage deps per-package

petermetz:
This plugin allows to resolve some CVEs more surgically that are found in indirect
dependencies which are difficult to upgrade without triggering a large change
needed and potential migrations, breaking changes to the public APIs of packages.

The reason why the above problem happens is because `yarn up` and `yarn up -R`
are blunt instruments when it comes to managing a monorepo such as ours:
They do their upgrade all-or-nothing, e.g. you can't upgrade a single dependency
in a single monorepo package, you must upgrade the dependency project-wide
with the mentioned tools, but sometimes we need to perform the upgrade just in a
single monorepo package.

For example to the above, about 20 packages use web3 but only about 5 of those
are using v4.x versions of web3. A new CVE came out covering v4.1.x and so
I needed to upgrade web3 only in those packages where web3 was already above
v4.0.0 and leave the older ones alone (surgical upgrades).

To accomplish this I've found no way to do it with stock yarn CLI commands, but
someone who had the exact same problem had written a plugin for solving it.

The original issue reported to yarn with the same problem we are having:
https://github.com/yarnpkg/berry/issues/2591

The repository where the plugin resides that we are adding in this commit in
order to remediate the problem of lack of surgical (per-package) upgrades:
https://github.com/eyolas/yarn-plugin-interractive-filter

The original CVE that I was investigating as I stumbled upon the solution:
- https://github.com/hyperledger/cacti/pull/3264
- https://github.com/hyperledger/cacti/security/dependabot/987

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


jagpreetsinghsasan (approved):
LGTM

outSH (approved):

PR #3300, opened 2024-06-06
test(plugin-consortium-manual): fix API server CRPC port allocation

petermetz:
1. Also sneaking in a fix for a DCI-Lint failure that was introduced recently
    when we added a new Yarn plugin which then stored its install URL in the
    .yarnrc.yml file and it uses the old git default main branch name and does
    not support the new one so we had to exclude the config file from linting.
    2. Also ensured that the ConnectRPC ports are bound to zero in all tests where the API
    server is being used. This will prevent port conflicts randomly popping up across the
    test suite in the future.
    3. Also removed a few test cases from the taprc file because they were already migrated to
    Jest and therefore tap should not run them as they fail with the Jest syntax.
    4. Also fixing the lack of etherscan API key environment variable in the HTLC coordinator tests.
    
**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


simulified:
There is a problem
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 1000; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

petermetz:
@simulified What is the problem?

jagpreetsinghsasan (approved):

PR #3346, opened 2024-06-21
ci(custom-checks): fix depcheck not detecting missing dependencies

petermetz:
1. The depcheck tool we use have not correctly discovered some of the
missing dependencies that we have because it only verifies that the
imported dependency is present SOMEwhere in the package.json file, not that
it is specifically present in the production dependencies section which
leads to crashes and broken packages due to the API server not installing
dev dependencies when instantiating a plugin and therefore missing a few
of the dependencies that are otherwise very much needed at runtime in
production.
2. The solution to the problem was to implement our own typescript parsing
with babel and then double check the work of depcheck to make sure that
the dependencies that it marks as "no issues" are actually OK and have no
issues.
3. The hardest edge case was type imports e.g. `import type { Express } from "express";`
because the import was there, but we did not actually need that dependency
in the production dependencies as long as ALL of the imports to it in the
given package were type imports. To robustly verify this being the case or not
we had to pull out the big guns and parse all the typescript code per package
to make sure that we've looked at every single import of the dependency in
question at every single code file of the package in question.

Depends on #3345

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,8 @@
     "zod": ">=3.22.3"
   },
   "devDependencies": {
+    "@babel/parser": "7.24.7",
+    "@babel/types": "7.24.7",
     "@bufbuild/buf": "1.30.0",
     "@bufbuild/protobuf": "1.8.0",
     "@bufbuild/protoc-gen-es": "1.8.0",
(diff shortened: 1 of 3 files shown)


outSH (approved):

jagpreetsinghsasan (approved):
LGTM

Issue #3069, opened 2024-03-07
build(sync-ts-config): fix no such file or directory protos-js/tsconfig.json


petermetz:
### Description

The sync-ts-config script crashes when executed from the command line on my machine (Debian 12, KDE Plasma 5)

A `tsconfig.json` file missing is the most likely root cause, but that's just a guess for now.

Consistently reproduces if you execute `yarn run sync-ts-config` from the command line after having ran `yarn configure` successfully on the `main` branch.

### Logs

```sh
/workspaces/cacti/weaver/common/protos-js/tsconfig.json
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[Error: ENOENT: no such file or directory, open '/workspaces/cacti/weaver/common/protos-js/tsconfig.json'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/workspaces/cacti/weaver/common/protos-js/tsconfig.json'
}

Node.js v18.19.0

```

### Acceptance Criteria
1. Script in question runs fine
2. Tests were added so that if this breaks again the CI catches it

Issue #3366, opened 2024-06-28
build(sync-ts-config): integrating the weaver packages to the monorepo build


ruzell22:
### Description
The sync-ts-config script has ignore paths that are weaver related for missing `tsconfig.json` file when executing `yarn run sync-ts-config` from the command line after having `yarn run configure` successfully on the main branch.

The 4 paths are:
-**/packages/cacti-plugin-weaver-driver-fabric/**
-**/weaver/common/protos-js/**
-**/weaver/samples/besu/simpleasset/**
-**/weaver/samples/besu/simplestate/**

Related to: #3069 

### Acceptance Criteria
1. Script in question runs fine after integrating of packages and removing the added ignore paths.

ruzell22:
Hello @VRamakrishna , I would like to ask if there are tsconfig.json files for /common/protos-js, /samples/besu/simpleasset, and /samples/besu/simplestate that are not in that base folder but inside another folder? I am trying to do this solution https://github.com/hyperledger/cacti/pull/3367#issuecomment-2207777149 for them as well. I have tried it for /packages/cacti-plugin-weaver-driver-fabric and it worked when I ran the `yarn run sync-ts-config` without having it in ignore path. When I do it, however, a new path has missing tsconfig.json issue when running the script. /core/drivers/fabric-driver/

If there is no specific tsconfig.json file for the mentioned weaver related folders, is there something that I can do to resolve this issue? Thank you.

VRamakrishna:
@ruzell22 Sorry, your query passed under my radar. The answer to your question is: no, if the respective folders don't contain a `tsconfig.json` file, it doesn't exist anywhere else. You are welcome to try and add it (though #3400 didn't really solve the issue) to the various Weaver modules, and we will review it.

PR #3367, opened 2024-06-28
build(sync-ts-config): fix no such file or directory protos-js/tsconfig.json

ruzell22:
### **Commit** to be reviewed
---
build(sync-ts-config): fix no such file or directory protos-js/tsconfig.json
```
Primary Changes
---------------

1. Fix issue with the object key
2. Temporarily added ignore paths for the missing tsconfig.json so the script will run successfully
```

Fixes: #3069 
Related to: #3366

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/tools/sync-npm-deps-to-tsc-projects.ts
+++ b/tools/sync-npm-deps-to-tsc-projects.ts
@@ -40,6 +40,12 @@ const main = async (argv: string[], env: NodeJS.ProcessEnv) => {
   const globbyOptions: GlobbyOptions = {
     cwd: PROJECT_DIR,
     absolute: true,
+    ignore: [
+      "**/packages/cacti-plugin-weaver-driver-fabric/**",
+      "**/weaver/common/protos-js/**",
+      "**/weaver/samples/besu/simpleasset/**",
+      "**/weaver/samples/besu/simplestate/**",
+    ], // Follow-up issue regarding these hardcoded paths (https://github.com/hyperledger/cacti/issues/3366)
   };
   const pkgJsonPaths = await globby(pkgJsonGlobPatterns, globbyOptions);
   console.log(`Package paths (${pkgJsonPaths.length}): `, pkgJsonPaths);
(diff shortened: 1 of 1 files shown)


tools/sync-npm-deps-to-tsc-projects.ts
```diff
@@ -70,7 +76,7 @@ const main = async (argv: string[], env: NodeJS.ProcessEnv) => {
 
     const pkg = await fs.readJson(pkgJsonPath);
 
-    const deps = Object.keys(pkg.dependencies).filter((it) =>
+    const deps = Object.keys(pkg.dependencies || {}).filter((it) =>
```
petermetz, line 79: @ruzell22 If there is no `dependencies` object defined then log it out so that it's easier to discover later in the logs what exactly happened.  Some pseudo code below:

```typescript
if (!pkg.dependencies) {
   // log here the exact location of the package.json file that was being examined and the fact that it had no `dependencies` object and therefore we skipped that part.
}
```

tools/sync-npm-deps-to-tsc-projects.ts
```diff
@@ -40,6 +40,12 @@ const main = async (argv: string[], env: NodeJS.ProcessEnv) => {
   const globbyOptions: GlobbyOptions = {
     cwd: PROJECT_DIR,
     absolute: true,
+    ignore: [
```
petermetz, line 43: @ruzell22 Please place a direct link to the follow-up issue in a comment here in the code so that whoever comes around and looks at these hardcoded paths will know that there is a task in progress to remediate the problem.

petermetz (changes requested):
@ruzell22 Overall, looking good, but I have a few nits as usual. See the above comments and also one here: Please use a PR title and commit message that is an exact match to the issue title.

tools/sync-npm-deps-to-tsc-projects.ts
petermetz: @ruzell22 Actually, never mind, the fallback to the empty object is OK. I realized in the meantime that it's not something that we would want to debug too much because it's an expected scenario.

petermetz:
@ruzell22  I figured it out! Our sync-ts-config script assumes that the tsconfig is in the package's top level directory (e.g. `./packages/my-package/` but for this package the location of it's wrong so the fix is probably to relocate the tsconfig file to the expected path. So, move 
`packages/cacti-plugin-weaver-driver-fabric/src/main/typescript/tsconfig.json` to 
`packages/cacti-plugin-weaver-driver-fabric/tsconfig.json`

petermetz (approved):

outSH (approved):

PR #3375, opened 2024-07-03
fix(cmd-api-server): use ncc bundle in container image - CVE-2024-29415

petermetz:
Fixes the CVE mentioned and also improves our response time to future CVEs
by a very wide margin. Details below.

1. Fixing the mentioned vulnerability in the API server and doing so in
a way so that in the future our dependency upgrades automatically propagate
to the container builds as well.
2. The way we are achieving this is by making the container image build
use the pre-built bundle instead of pulling the package contents from npm.
3. This has the advantage of breaking the chicken egg problem with releases
to npm and container images, so from now on if we are adding a fix to the
API server in the code, the built container image will automatically contain
that fix when building on the CI for the pull request.
4. This is also a new pattern for how to create our container images that
has a couple more improvements to it:
4.1. The .dockerignore file is now specific to the particular package's
container image instead of the global one in the project root being used.
This was needed because we are copying files from the ./dist/ folder of the
package to the container image at build time but this was not possible while
the root dir .dockerignore file was in effect because it blanket ignores
the ./dist/ folders overall and so the image building was failing with errors
that it couldn't locate the bundle (which is inside the ./dist/ directory)
4.2. The healthcheck of the container is now 100% self-contained and needs
no external dependencies of any kind (neither npm nor operating system level ones)
This is beneficial because it reduces the attack-surface of the image and also
reduce the size of the image by at least a 100 MB.
4.3. With the introduction of the usage of the bundled version of the code
we have **dramatically** reduced the image size overall. The image built from
this revision of the code is 221 MB while the previous image versions were
hovering closer to a 0.5 GB.
5. Also updated the README of the package so that all the examples pertaining
to the container image are now fully functional once again.
6. Simplified the container image's definition: the custom docker entrypoint
script and the healthcheck bash script are no longer necessary.
7. Renamed the container image definition file from `Dockerfile` to
`cmd-api-server.Dockerfile` because this is mandated by Docker when building
images with custom .dockerignore files (it needs the custom filename to
disambiguate the .dockerignore files based on it)
8. Refactored how the CI executes the Trivy scan to reduce resource usage:
8.1. There is no separate image build job now. This was necessary because
with the new image definition we have to have the project compiled first
(since we no longer install directly from npm) so it would've been a lot of
duplicated compute time to recompile the project in yet another CI job for the
image build.

The image built from this revision is also published on the official repository
with the canary tag of:
`ghcr.io/hyperledger/cactus-cmd-api-server:2024-07-03T19-32-51-dev-3f5e97893`

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


jagpreetsinghsasan (approved):
LGTM

RafaelAPB (approved):
LGTM.
Why did we skip one fabric test in this PR?

petermetz:
> LGTM. Why did we skip one fabric test in this PR?

@RafaelAPB Sorry for the slow response! That test became broken due to it installing the latest version from the official npm repositories (which for v2.0.0-rc.1 was broken). So I figured for now it's pointless to have the test and always fail, but now that rc.2 is out I'm hoping when I try to reenable it will be good.

petermetz:
> LGTM. Why did we skip one fabric test in this PR?

@RafaelAPB Sorry for the slow response! That test became broken due to it installing the latest version from the official npm repositories (which for v2.0.0-rc.1 was broken). So I figured for now it's pointless to have the test and always fail, but now that rc.2 is out I'm hoping when I try to reenable it will be good.

PR #3418, opened 2024-07-17
test(test-tooling): add container image builder utilities

petermetz:
1. Currently our integration tests depend on pre-published container
images to be on the official registry (ghcr.io). This has pros and cons.
The pro is that we can pin the tests to a specific ledger version and
then have confidence that the test code works with that specific image.
On the other hand if the image itself has problems we won't know it until
after it was published and then tests were executed with it (unless we
perform manual testing which is a lot of effrot as it requires the
manual modification of the test cases).
2. In order to gives us the ability to test against the container image
definitions as they are in the current revision of the source code,
we are adding here a couple of utility functions to streamline writing
test cases that build the container images for themselves as part of the
test case.

An example of how to use it in a test case:

```typescript
const imgConnectorJvm = await buildImageConnectorCordaServer({
    logLevel,
});

// ...

connector = new CordaConnectorContainer({
    logLevel,
    imageName: imgConnectorJvm.imageName,
    imageVersion: imgConnectorJvm.imageVersion,
    envVars: [envVarSpringAppJson],
});

```

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-test-tooling/src/main/typescript/common/build-container-image.ts
+++ b/packages/cactus-test-tooling/src/main/typescript/common/build-container-image.ts
@@ -1,0 +1,64 @@
+import Docker, { ImageBuildContext, ImageBuildOptions } from "dockerode";
+
+import { LoggerProvider, LogLevelDesc } from "@hyperledger/cactus-common";
+
+export interface IBuildContainerImageRequest {
+  readonly logLevel: LogLevelDesc;
+  readonly buildDir: Readonly<string>;
+  readonly imageFile: Readonly<string>;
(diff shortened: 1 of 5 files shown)


jagpreetsinghsasan (approved):
LGTM.
This was a much needed feature! I will inform the team about the same

outSH (approved):
This is great, LGTM, thank you!

petermetz:
> This is great, LGTM, thank you!

@outSH Thank you! For now I only have the ready-made utility functions for the corda images, but we can add more, ideally all images would be possible to be built by the tests like this in the future.

PR #3448, opened 2024-07-30
feat(persistence-fabric): add sample setup scripts, improve documentation

outSH:
- Fix error when empty `transaction.actions` was received from a block.
- Copy schema SQL during build.
- Remove unique constraint on fabric transactions hash column (there are transactions with empty hash that would break it).
- Improve parsing of certificate subject/issuer attributes to handle more delimiters.
- Use new `install-fabric.sh` script (that is recommended for Fabric 2.5) instead of old bootstrap to fix some runtime issues that I've encountered.
- Add sample setup scripts. Simple can be used to run persistence against already running fabric ledger, complete will setup entire environment and run some basic operations to generate sample data.
- Improve documentation to include these new scripts and how to use them, fix smaller issues.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/get-block/cacti-block-formatters.ts
+++ b/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/get-block/cacti-block-formatters.ts
@@ -79,7 +79,7 @@ export function formatCactiFullBlockResponse(
       const transaction = payload.data;
 
       const transactionActions: FullBlockTransactionActionV1[] = [];
-      for (const action of transaction.actions) {
+      for (const action of transaction.actions ?? []) {
         const actionPayload = action.payload;
         const proposalPayload = actionPayload.chaincode_proposal_payload;
         const invocationSpec = proposalPayload.input;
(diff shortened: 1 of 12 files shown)


petermetz (approved):
@outSH LGTM, thank you very much!

RafaelAPB (approved):
This is a great PR, thanks!
I like very much the fact the examples are layed down as a tutorial. However, perhaps the best organization is instead of having scripts on `test/manual`, to have instead an example package under /examples?
Rest LGTM

outSH:
@RafaelAPB Thanks! I believe a similar idea was discussed before, but the challenges were:
- If `/examples` were not included in the CI build, it could lead to discrepancies between the examples and the actual code.
- If `/examples` were built as an actual package alongside other packages, it could quickly grow in complexity due to dependencies.

I opted for a `manual`  directory because I used a similar convention for other scripts. However, I'm open to adding /examples inside the persistence plugin folder if that would be clearer for users. :)

petermetz:
I agree with both of you. Pros and cons to both approaches unfortunately and it's hard to tell which one has worse cons and/or better pros.

PR #3456, opened 2024-08-03
ci(github): fix keychain-aws-sm no code coverage files/dir found

petermetz:
1. The Jest test execution was turned off for the AWS SM keychain plugin
for some reason so this PR enables it back and while at it
2. I also refactored one of the test cases to have less linter warnings
about us casting to `any`.
3. And on top of that I also migrated one of the test cases to Jest from
TAP/tape so that we are making some progress with the test coverage at
the same time.

One more notable thing is that I refactored the negative test assertions
while migrating to Jest so the test case now uses Jest native assertions
to check what exactly the error message returned by the API call for
deleting a non-existent item looks like. This is an example of the
refactoring that this issue is about:
https://github.com/hyperledger/cacti/issues/3455


jagpreetsinghsasan (approved):
LGTM

outSH (approved):

PR #3471, opened 2024-08-09
feat(consortium-static): new consortium plugin

eduv09:
New plugin consortium-static, based on consortium-manual. Long story short, the plugin allows for the addition of Cacti Nodes to consortium at runtime. New node entities must belong to one of the consortium entities (organizations) that were specified on consortium creation. So it allows to add new nodes 'CactusNode', but not new 'ConsortiumMember' entities.

The code is based on the plugin-consortium-manual, with the new features built on top of it. The process of adding a new node is conducted as follows: 
   1. The new node forges a request with information about itself and sends it to any Node in the consortium 
   2. The receiver verifies the request, and if it is approved, broadcasts it to the rest of the consortium Nodes.
   3. Each node verifies the same request, and adds the new node to the consortium database. At the same time, the receiver node from point 2 sends the consortium data to the new node, who updates it's database.

The package includes a policy model, which was developed based on the ideas of the Policy Core Information Model [RFC3060](https://www.rfc-editor.org/rfc/rfc3060) . The idea was to introduce the idea of a multi-purpose policy framework that can be leveraged by other packages in cacti.


petermetz (changes requested):
@eduv09 Please include all details in the commit message not just the PR description

packages/cacti-plugin-consortium-static/src/main/typescript/consortium/process-broadcast-endpoint-v1.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  async handleRequest(req: Request, res: Response): Promise<void> {
+    const fnTag = `${this.className}#handleRequest()`;
+    const reqTag = `${this.getVerbLowerCase()} - ${this.getPath()}`;
+
+    this.log.debug(`GET ${this.getPath()}`);
+    try {
+      await this.plugin.verifySignature(
+        req.body.signature,
+        safeStableStringify(req.body.message) as string,
```
petermetz, line 113: @eduv09 Please try to not nest function calls in each other, it makes the stack traces that much harder to read when you see that something crashed on this line but cannot be sure if it was something within `verifySignature` or something within a function that was computing a parameter of `verifySignature`. A separate line for the stringification and a dedicated variable to its output could then solve the issue and it would also allow you to validate the output with somehting like the snippet I'll put below. That would make it so that the casting to string can be removed (which in its current form is an unsafe cast because you are not really sure if its a string or not at runtime).

```typescript
if (typeof msgAsJson !== "string") { throw new InternalServerError("safeStableStringify() returned with non-string value.")}
```

packages/cacti-plugin-consortium-static/src/main/typescript/repository/new-consortium-provider.ts
```diff
@@ -0,0 +1,72 @@
...
+  DefaultApi as ConsortiumNewApi,
+  GetConsortiumJwsResponse,
+} from "../generated/openapi/typescript-axios/api";
+
+export interface INewConsortiumProviderOptions {
+  logLevel?: LogLevelDesc;
+  apiClient: ConsortiumNewApi;
+}
+
+export class NewConsortiumProvider
```
petermetz, line 20: @eduv09 Please rename this as well, I'm guessing it's a leftover from before we came up with the current name.

packages/cacti-plugin-consortium-static/src/main/typescript/repository/new-consortium-repository.ts
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Class responsible for making it convenient for developers to query the
+ * `ConsortiumDatabase` model type which is a flat data structure storing
+ * all the different types of entities for Consortium representation such as
+ * `CactusNode`, `ConsortiumMember`, `Ledger` etc..
+ */
+export class NewConsortiumRepository {
+  public static readonly CLASS_NAME = "NewConsortiumRepository";
+
```
petermetz, line 42: @eduv09 Please rename this as well, I'm guessing it's a leftover from before we came up with the current name.

packages/cacti-plugin-consortium-static/src/main/typescript/plugin-consortium-static.ts
```diff
@@ -0,0 +1,503 @@
...
+      // issues after the fact and therefore we must save the error response details somewhere (the logs)
+      this.log.debug(
+        "ConsortiumApi non-2xx HTTP response:",
+        data,
+        status,
+        config,
+      );
+
+      // For the caller/client we just send back a generic error admitting that we somehow messed up:
+      const errorMessage = `${fn} BesuAPI error status: ${status}: ${statusText}`;
```
petermetz, line 487: @eduv09 Please double check if logging `BesuAPI` here is accurate. I'm guessing it should be something else!

petermetz (changes requested):
@eduv09 Looks like a high quality PR, thank you very much! I have some nit-picks in the comments above and a couple requests here:
1. Remove the changelog file, it doesn't make sense  to have it in a new package - it gets auto-generated when we issue a release.
2. Add a CI job in ci.yaml that runs the Jest tests so that it can be verified to keep working after each PR. Note that the github action versions to use are changing so make sure to use the up to date versions as shown in this otherwise unrelated PR: https://github.com/hyperledger/cacti/pull/3478

petermetz:
@eduv09 One more thing: if you run `yarn lint` it will apply the automatic formatting to the openapi spec file (the CI will fail otherwise) so you need to run `yarn lint` and then commit the changes it makes. This is what the diff looks like on my machine:

image

packages/cacti-plugin-consortium-static/src/main/json/openapi.json
```diff
@@ -0,0 +1,498 @@
...
+    }
+  },
+  "components": {
+    "schemas": {
+      "Identity":{
+        "type": "object",
+        "required": ["pubKey", "memberId", "proof"],
+        "properties": {
+          "pubKey":{
+            "description": "TODO",
```
RafaelAPB, line 19: Please add a simple description to the properties of the new schemas. If schemas are already existing feel free to copy its description

packages/cacti-plugin-consortium-static/src/main/typescript/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export async function verifyOrganization(
+  token: string,
+  jwk: JWK,
+  expectedOrganization: string,
+): Promise<string | boolean> {
+  try {
+    // Import the JWK as a public key
+    const publicKey = await importJWK(jwk, "ES256K");
+
```
RafaelAPB, line 34: can we make the return type explicit? might help developers to add extra checks

packages/cacti-plugin-consortium-static/src/main/typescript/plugin-consortium-static.ts
```diff
@@ -0,0 +1,503 @@
...
+    const keyPair = await importPKCS8(keyPairPem, "ES256K");
+    const payloadObject = { consortiumDatabase: this.repo.consortiumDatabase };
+    const payloadJson = safeStableStringify(payloadObject);
+    const _protected = {
+      iat: Date.now(),
+      jti: uuidv4(),
+      iss: "Hyperledger Cactus",
+    };
+    // TODO: double check if this casting is safe (it is supposed to be)
+    const encoder = new TextEncoder();
```
RafaelAPB, line 266: please fix this. Installing the types for jose as a dev dependency might help. see https://www.npmjs.com/package/@types/node-jose

packages/cacti-plugin-consortium-static/src/main/typescript/plugin-consortium-static.ts
```diff
@@ -0,0 +1,503 @@
...
+    msg: string,
+    pubKey: string,
+  ): Promise<boolean> {
+    const publicKey = await importSPKI(pubKey, "ES256K");
+    const { payload } = await compactVerify(signature, publicKey);
+
+    // Extract the original message
+    const payload1 = JSON.parse(payload.toString());
+    const signedData = new Uint8Array(payload1.data);
+    const decodedMessage = new TextDecoder().decode(signedData);
```
RafaelAPB, line 426: nitpick
```suggestion
    const payload = JSON.parse(payload.toString());
    const signedData = new Uint8Array(payload.data);
    const decodedMessage = new TextDecoder().decode(signedData);
```

packages/cacti-plugin-consortium-static/src/main/typescript/policy-model/polity-item.ts
```diff
@@ -0,0 +1,47 @@
+import { IPolicy } from "./i-policy";
```
RafaelAPB, line 1: Please elaborate (can be on the README) on the relationship between conditions, groups, rules and items (this is already partially done, but would be good to make everything explicit)

RafaelAPB (commented):
This is very high-quality work, thanks for your contribution! I've left a few comments, most things should be relatively easy to fix. Please also refer to @petermz 's comments. Besides that, LGTM!

packages/cacti-plugin-consortium-static/src/main/json/openapi.json
petermetz: @RafaelAPB @eduv09 I would also like to spread the word about cross-spec references now being "easy" compared to how broken and complicated they were before (back when we used the HTTP URLs directly pointing at GitHub).

I don't want to derail this PR so this is not a change request, I'm just saying it  for future reference that since the introduction of the `openapi.tpl.json` mechanism you can directly reference other openapi.json files through relative paths on the local file-system ACROSS different packages. If you do a text search in the project for `"$ref": "../` then you can see examples of how it can be used (see screenshot below)

With this new mechanism we could move a lot of the centrally used types to the core-api package and all the other packages could re-use them from there instead of the endless copy-pasting (which then also leads to annoying bugs in the build and even at runtime occasionally).

Again, I'm not saying to change it right now, just something to keep in mind going forward as a helpful alternative for newer things or for refactors later as you see fit.

I also post this comment on the chat and the mailing list to make sure people are aware of this new technique we've enabled recently. 

image

packages/cacti-plugin-consortium-static/src/main/typescript/plugin-consortium-static.ts
eduv09: This piece of code is actually from the ConsortiumManual plugin, and the comment from peter so I am not sure what is the possible issue here. 
@petermetz any notes on this one? Maybe it's a comment left behind ...

eduv09:
@RafaelAPB @petermetz I think I addressed all of the changes you requested. Let me know if I missed something, or of any other recommendations. Thank you !

RafaelAPB:
@eduv09 please rebase and fix any conflicts or errors

packages/cacti-plugin-consortium-static/src/main/json/openapi.json
RafaelAPB: This looks great, thanks for sharing Peter. Some of these types are being re-declared consistently, and this fixes that!

eduv09:
@RafaelAPB Done

RafaelAPB:
LGTM, let's wait for @petermetz 's re-review. Cheers

packages/cacti-plugin-consortium-static/src/main/typescript/plugin-consortium-static.ts
petermetz: @eduv09 @RafaelAPB Sorry for the slow processing! 
Short answer: I think we are OK to leave this as-is for now and I'll help with some work on the longer term solution that guarantees correct validation and eliminates the need for casting altogether. 

Longer answer:
I looked into it in detail and the way I'd recommend handling it is writing some type-narrowing user-defined type guards (a Typescript language feature) within the core-api package itself and then use those to verify both at build and runtime that the object's are valid and are matching the schema of the openapi spec.

I'm pasting a rough example here which is not yet operational but is the draft idea I'll send a quick PR with soon.

While I do that PR this should get merged because it doesn't have a huge impact and I definitely don't want to hold up the PR with putting a dependency on it.

The draft code I'll use to get started with my improvements (no need to do anything with it I'm just taking it as notes for self mostly)

```typescript
import Ajv from 'ajv';

const ajv = new Ajv();

// JWSCompact type guard
function isJWSCompact(data: unknown): data is string {
  const schema = {
    type: 'string',
    minLength: 5,
    maxLength: 65535,
    pattern: /^[a-zA-Z0-9-_]+?.[a-zA-Z0-9-_]+?.([a-zA-Z0-9-_]+)?$/,
  };

  const validate = ajv.compile(schema);
  return validate(data);
}

// JWSRecipient type guard
function isJWSRecipient(data: unknown): data is JWSRecipient {
  const schema = {
    type: 'object',
    required: ['signature'],
    properties: {
      signature: { type: 'string' },
      protected: { type: 'string' },
      header: { type: 'object', additionalProperties: true },
    },
  };

  const validate = ajv.compile(schema);
  return validate(data);
}

// JWSGeneral type guard (interface for JWSRecipient)
interface JWSRecipient {
  signature: string;
  protected?: string;
  header?: Record<string, unknown>;
}

function isJWSGeneral(data: unknown): data is JWSGeneral {
  const schema = {
    type: 'object',
    required: ['payload', 'signatures'],
    properties: {
      payload: { type: 'string', minLength: 1, maxLength: 65535 },
      signatures: {
        type: 'array',
        items: { $ref: '#/components/schemas/JWSRecipient' },
      },
    },
  };

  const validate = ajv.compile(schema);
  return validate(data);
}

export { isJWSCompact, isJWSRecipient, isJWSGeneral };
```

petermetz (approved):
@eduv09 @RafaelAPB LGTM, thank you for addressing my nit-picks!

For the `JWSGeneral` casting question, please see my analysis here (short answer, let's merge this as is and I'll work on fixing the casting) => https://github.com/hyperledger/cacti/pull/3471#discussion_r1731894747

One more thing: If you hit the re-request review button it will automatically land back in my PR review queue (which I try to check daily even when I don't have time to process discord messages) so I recommend hitting that every time a PR I reviewed is ready for some more attention. Tagging me on Discord and on GitHub also helps so I recommend you also keep doing that of course!
image

eduv09:
@petermetz fixed the CI workflow issue. I think it's good to go. Thank you for the assist 🙏

RafaelAPB (approved):

petermetz (approved):

petermetz:
> @petermetz fixed the CI workflow issue. I think it's good to go. Thank you for the assist 🙏

@eduv09 You got it! Thank you very much for the contribution!

PR #3481, opened 2024-08-13
feat(satp-hermes): satp-manager and gol refactoring

LordKubaya:


packages/cactus-plugin-satp-hermes/src/main/typescript/gol/satp-manager.ts
```diff
@@ -297,10 +304,157 @@ export class SATPManager {
...
+    if (!channel) {
+      throw new Error(`${fnTag}, Channel not found`);
+    }
+
+    const sessionData: SessionData =
+      session.getClientSessionData() as SessionData;
+
+    const clientSatpStage0: PromiseConnectClient<typeof SatpStage0Service> =
+      channel.clients.get("0") as PromiseConnectClient<
+        typeof SatpStage0Service
```
RafaelAPB, line 337: Seems to be a problem here with the type narrowing

RafaelAPB (changes requested):

packages/cactus-plugin-satp-hermes/src/main/typescript/gol/satp-manager.ts
LordKubaya: this variables are just for each stage wouldnt it be better if the type is only of that service?

RafaelAPB (approved):

PR #3492, opened 2024-08-24
build(connector-corda): upgrade to Spring Boot v3.3.1

petermetz:
IMPORTANT: The project now needs JDK 17 and Gradle 8 for development.

1. The bean validation has been replaced to the jakarta flavor as mandated
by the spring boot upgrade.
2. Request bodies in the OpenAPI spec for Corda v4 were made mandatory
where applicable (which is most of the endpoints)
3. The Open API spec based request validation test case has been moved to
the new Corda v4.12 test case which now builds both the connector JVM app
and the AIO image at runtime so that we can guarantee that the latest
code changes are being verified.
4. Added quicker and easier steps in the readme of the corda connector
to run trivy scans on the .jar files instead of the container images.
The .jar files are 20x faster to build and scanning them instead of the
container images doesn't suffer from the problem that the dev build
dependencies are showing up in the scans (creating false positives)
5. Updated the CI to use the .jar file for scanning as well instead of
the container image.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


petermetz:
@hyperledger/cacti-maintainers Apologies for the massive PR - most of the diff is the generated code but there's also quite a few manual changes that were hard to separate out due to the build level dependencies (Spring 3, JDK 17, etc).
I already sent 5 or 6 smaller PRs leading up to this to try and make this one easier to review.

jagpreetsinghsasan (approved):
LGTM

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/build.gradle.kts
```diff
@@ -5,9 +5,10 @@ version = project.properties["version"]!!
 
 val corda_release_group = "net.corda"
 val corda_core_release_group =  "net.corda"
-val corda_release_version = "4.6"
-val corda_core_release_version = "4.6"
-val corda_platform_version = 5
+val corda_release_version = "4.12-HC02"
```
sandeepnRES, line 8: Why `4.12-HC02`, that's 7 months older? Why not stable `4.12`?

sandeepnRES (approved):
LGTM, just one query above. Still approving it, don't want to delay it.

packages/cactus-plugin-ledger-connector-corda/src/main-server/kotlin/gen/kotlin-spring/build.gradle.kts
petermetz: @sandeepnRES Thanks for catching that! This PR has been in the works for a very long time (about 7 months) so back then there was no 4.12 GA but now I agree it should be upgraded. It's not as trivial unfortunately as setting the dependency version here and be done with it because I also have to upgrade the kotlin samples, but it's important enough that I'll just do it now rather than have to come back to it later!

petermetz: @sandeepnRES Took me all day in the end but I made the changes and also a few bug-fixes along the way so now it's much more stable than it was, thank you very much once again for noticing this!

sandeepnRES: Thanks @petermetz. I will merge this now since all the tests are passing.

PR #3509, opened 2024-09-03
Draft PR - gateway runner and docker for SATP

brunoffmateus:


packages/cactus-plugin-satp-hermes/Dockerfile.gateway
```diff
@@ -0,0 +1,71 @@
+FROM ubuntu:22.04
```
RafaelAPB, line 1: Can you use a smaller / lighter image? ubuntu 22.04 is very heavy and we do not need most of it . I recommend the `NodeJS v18.19.0` which is the official node version we use in cacti (plus necessary dependencies)

packages/cactus-plugin-satp-hermes/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+      - SATP_COUNTER_PARTY_GATEWAYS=[{"id":"gateway2","version":"v02,v02,v02","supportedDLTs":"FabricSATPGateway,BesuSATPGateway"}]
+    volumes:
+      - gateway1_keys:/app/keys
+    deploy:
+      resources:
+        limits:
+          memory: 4G
+          cpus: '0.9'
+
+  # satp-hermes-gateway2:
```
RafaelAPB, line 32: why are those commented?

packages/cactus-plugin-satp-hermes/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+  #     - SATP_COUNTER_PARTY_GATEWAYS=[{"id":"gateway1","version":"v02,v02,v02","supportedDLTs":"FabricSATPGateway,BesuSATPGateway"}]
+  #   volumes:
+  #     - gateway2_keys:/app/keys
+  #   deploy:
+  #     resources:
+  #       limits:
+  #         memory: 2G
+  #         cpus: '0.5'
+
+volumes:
```
RafaelAPB, line 52: we need 1 volume to store the logs; other to store relevant data (such as keys). But its only for one gateway. (one docker container per gateway)

packages/cactus-plugin-satp-hermes/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
+version: '3.8'
+
+services:
+  ### To build new image version:
+  ### uncomment and run in the terminal "docker compose build"
+  # satp-hermes-gateway-base:
+  #   image: ghcr.io/brunoffmateus/cactus-plugin-satp-hermes-gateway:new-version
```
RafaelAPB, line 7: add service if needed. create a suitable tag (in accordance with package.json)

packages/cactus-plugin-satp-hermes/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+  #   build:
+  #     context: ./
+  #     dockerfile: Dockerfile.gateway
+
+  satp-hermes-gateway1:
```
RafaelAPB, line 12: why gateway1?

packages/cactus-plugin-satp-hermes/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
...
+      - gateway1_keys:/app/keys
+    deploy:
+      resources:
+        limits:
+          memory: 4G
+          cpus: '0.9'
+
+  # satp-hermes-gateway2:
+  #   image: ghcr.io/brunoffmateus/cactus-plugin-satp-hermes-gateway:2024-09-02-3117
+  #   ports:
```
RafaelAPB, line 34: also add the creation and deployment of a local database (see package.json and knex)

packages/cactus-plugin-satp-hermes/package.json
```diff
@@ -65,6 +65,12 @@
     "codegen:openapi": "npm run generate-sdk",
     "codegen:proto": "npm run generate-proto",
     "codegen:abi": "yarn forge:all && abi-types-generator './src/solidity/generated/satp-wrapper.sol/SATPWrapperContract.json' --output='./src/main/typescript/generated'",
+    "db:init": "run-s db:init:local db:init:remote",
```
RafaelAPB, line 68: this looks ok, but I thought there were some changes on these commands. make sure you have the latest rebased version of satp-dev

packages/cactus-plugin-satp-hermes/run-satp-gateway.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+generate_keys() {
```
RafaelAPB, line 3: please migrate this to the makefile

packages/cactus-plugin-satp-hermes/supervisord.conf
```diff
@@ -0,0 +1,19 @@
...
+logfile_maxbytes=50MB
+logfile_backups=10
+loglevel=info
+user=root
+
+[program:satp-gateway]
+command=/run-satp-gateway.sh
+autostart=true
+autorestart=unexpected
+exitcodes=0
```
RafaelAPB, line 12: make sure this is correct given the new dockerfile

RafaelAPB (changes requested):
Thank you for your contribution. Generally looks good. Please incorporate the feedback so we can merge

RafaelAPB:
Feel free to remove draft from the PR, so the lint passes

packages/cactus-plugin-satp-hermes/Dockerfile.gateway
VRamakrishna: Or Alpine?

RafaelAPB: or Alpine - probably even better

petermetz: @RafaelAPB @VRamakrishna Ubuntu is a little fat, I agree, but they have the best track record of fixing CVEs in a timely manner though (that's the reason why I advocate for us using ubuntu $LTS as the default (24.04 soon or already if I remember correctly).

I worked a lot in the past on trying to fix CVEs on alpine base images and also the thinner Debian based ones, but it usually ends with me not being able to install a combination of dependencies in the images that have no vulnerabilities and then I just end up going with Ubuntu.

When I say CVEs: I mean that the `trivy` scans we have for the images are failing with critical and high severity vulnerabilities in the images with the OS level dependencies (which are not in our code but if we are publishing images then we are on the hook for those as well in my opinion the same we as for our own code).

So in general I recommend keeping with ubuntu for the purpose of production readiness and security even if the slower image pulls are annoying. 

The other thing with alpine is that the packages break more oftne than on ubuntu. Me and @outSH and I think @sandeepnRES  too had to fix images several times where alpine broke something with their default openssh packages and then the images just would stop building overnight even with our best efforts to pin everything to a specific version.

petermetz (dismissed):
@brunoffmateus LGTM with comments => plus one for the things Rafael said, I just have a dissenting opinion on the base image ( https://github.com/hyperledger/cacti/pull/3509/files#r1744367913 )

packages/cactus-plugin-satp-hermes/Dockerfile.gateway
VRamakrishna: @petermetz Agree on the fact that Ubuntu images are kept more up-to-date w.r.t package vulnerabilities compared to Alpine. In other (non-Cacti) projects, I've seen vulnerability reports that prove hard to fix when the images are built on Alpine, and which then required me to manually replace and rebuild packages at build time. (E.g., `curl`)

So, on convenience (and arguably security) considerations, Ubuntu wins.
OTOH Alpine wins on performance.

petermetz: @VRamakrishna Agreed, 100%!

RafaelAPB:
@brunoffmateus can you please incorporate @petermetz 's feedback and my comments as well? With regard to the base image, in light of the arguments above, probably ubuntu is a better choice

brunoffmateus:
FYI - the latest force push simply adds the latest docker image as the default and in the tests

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-creation/privacy-policies.ts
```diff
@@ -8,6 +8,35 @@ export interface IPrivacyPolicyValue {
   policy: PrivacyPolicyOpts;
   policyHash: string;
 }
+
+// Type guard for PrivacyPolicyOpts
+export function isPrivacyPolicyOpts(
```
LordKubaya, line 13: Shouldn't this changes be in a different PR as they are done in the bungee package and the name of this pr is referent to the satp one?

brunoffmateus: I just had a talk with @petermetz (please correct me if I'm wrong), and we have these options:
- Merge everything together, since the future merge of `satp-dev` into main will be divided into multiple smaller PRs
- Merge these changes separately into `satp-dev`
- Merge these changes separately into `main` (they would be applied to `satp-dev` only after rebasing)

What do you think? (tagging @RafaelAPB so he sees this as well)

petermetz: @brunoffmateus @LordKubaya For reference, since this PR is against satp-dev, not main, I'm not too worried about keeping with the convention that @LordKubaya mentions in https://github.com/hyperledger-cacti/cacti/pull/3509#discussion_r1824588336 (if this was a PR against main then of course I'd ask for the same thing).

packages/cactus-plugin-satp-hermes/src/main/typescript/plugin-satp-hermes-gateway.ts
```diff
@@ -266,32 +263,32 @@ export class SATPGateway implements IPluginWebService, ICactusPlugin {
...
+        id: uuidv4(),
         pubKey: bufArray2HexStr(pluginOptions.keyPair.publicKey),
-        name: id,
+        name: uuidv4(),
         version: [
           {
-            Core: "v02",
-            Architecture: "v02",
-            Crash: "v02",
+            Core: SATP_VERSION,
```
LordKubaya, line 273: This cant be like, these are 3 kind of versions, they are not all the same, if they were we only whould have 1 version variable, please add 3 diferent variables for the versions. SATP_CORE_VERSION, SATP_ARCHITETURE_VERSION, SATP_CRASH_VERSION

packages/cactus-plugin-satp-hermes/src/main/typescript/plugin-satp-hermes-gateway.ts
```diff
@@ -266,32 +263,32 @@ export class SATPGateway implements IPluginWebService, ICactusPlugin {
...
     }
 
-    const id = uuidv4();
     if (!pluginOptions.gid) {
       pluginOptions.gid = {
-        id: id,
+        id: uuidv4(),
         pubKey: bufArray2HexStr(pluginOptions.keyPair.publicKey),
-        name: id,
+        name: uuidv4(),
```
LordKubaya, line 270: can't we give a name and not give and gid?
Shouldn't they be correlated?

packages/cactus-plugin-satp-hermes/src/main/typescript/plugin-satp-hermes-gateway.ts
```diff
@@ -266,32 +263,32 @@ export class SATPGateway implements IPluginWebService, ICactusPlugin {
...
       };
     } else {
       if (!pluginOptions.gid.id) {
-        pluginOptions.gid.id = id;
+        pluginOptions.gid.id = uuidv4();
       }
 
       if (!pluginOptions.gid.name) {
-        pluginOptions.gid.name = id;
+        pluginOptions.gid.name = uuidv4();
```
LordKubaya, line 291: The same here, if there is a case of the id and the name not existing they are created with diferent uuidv4, I think they should be correlated

packages/cactus-plugin-satp-hermes/src/main/typescript/plugin-satp-hermes-gateway.ts
```diff
@@ -303,18 +300,15 @@ export class SATPGateway implements IPluginWebService, ICactusPlugin {
       if (!pluginOptions.gid.version) {
         pluginOptions.gid.version = [
           {
-            Core: "v02",
-            Architecture: "v02",
-            Crash: "v02",
+            Core: SATP_VERSION,
+            Architecture: SATP_VERSION,
```
LordKubaya, line 304: Same as above 3 distinct variables

LordKubaya (changes requested):

packages/cactus-plugin-bungee-hermes/src/main/typescript/view-creation/privacy-policies.ts
LordKubaya: Okey @petermetz thank you.

packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-handlers/stage0-handler.ts
```diff
@@ -226,6 +226,7 @@ export class Stage0SATPHandler implements SATPHandler {
 
       return message;
     } catch (error) {
+      this.Log.debug(`Crash in ${fnTag}`, error);
```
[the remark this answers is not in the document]
petermetz: @LordKubaya Suggestion: I think it might be worthwhile to leave this in until the error handling is fixed. These log statements might be the only place these errors get logged with the current setup (if in the meantime it was fixed then please disregard this comment of mine).

packages/cactus-plugin-satp-hermes/src/main/typescript/plugin-satp-hermes-gateway.ts
brunoffmateus: sure, I'll add those

packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-handlers/stage0-handler.ts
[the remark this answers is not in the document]
brunoffmateus: @LordKubaya yeah as @petermetz mentioned I left that in because the error handling isn't 100% working, but since you are refactoring that already, I can remove these changes if you prefer it (and since the image was built with them, these logs will still appear in case something goes wrong until a image is pushed).

packages/cactus-plugin-satp-hermes/src/main/typescript/plugin-satp-hermes-gateway.ts
brunoffmateus: Sure, I can remove this

packages/cactus-plugin-satp-hermes/package.json
```diff
@@ -139,18 +150,24 @@
     "@types/fs-extra": "11.0.4",
     "@types/google-protobuf": "3.15.5",
     "@types/node": "18.18.2",
+    "@types/ssh2": "0.5.47",
```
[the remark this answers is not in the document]
RafaelAPB: @brunoffmateus have the same question as @LordKubaya

packages/cactus-plugin-satp-hermes/src/main/typescript/config-validating-functions/bridges-config-validating-functions/validateBridgesConfig.ts
```diff
@@ -0,0 +1,99 @@
+import { SupportedChain } from "../../core/types";
```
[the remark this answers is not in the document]
RafaelAPB: I'm not saying this is the wrong way, or that the one above is better. We can leave this as is, provided it works. But please open an issue in Cacti for refactoring this logic

packages/cactus-plugin-satp-hermes/run-gateway.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
RafaelAPB, line 1: please make this a package.json script and remove this file

RafaelAPB (changes requested):

packages/cactus-plugin-satp-hermes/src/main/typescript/.env.example
```diff
@@ -1,38 +0,0 @@
-# CREDENTIALs
```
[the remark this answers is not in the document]
petermetz: @RafaelAPB @brunoffmateus I'm too late with this info, but if you refactor the gateway to be a plugin that can be loaded through the API server then you get this functionality (of the API server) for free as well. 

Longer term I'd aim to refactor the gateway so that it doesn't need it's own container image (less maintenance burden) nor it's own CLI/env/file configuration parsing logic and of course a million other things like JWT basd authorization, SocketIO, gRPC, ConnectRPC, etc.

It's too late to redo the whole thing right now, but I suggest putting it on the road map for later. It's also a refactor that can be done in smaller pieces by way of implementing the necessary interfaces one by one in smaller PRs. Once all the necessary interfaces are implemented then the parts of the code that are redundant with the API server's functionality can be just deleted and the docs updated to show it using the API server.

I'm also happy to try and help with this in the future.

packages/cactus-plugin-satp-hermes/src/main/typescript/config-validating-functions/bridges-config-validating-functions/validateBridgesConfig.ts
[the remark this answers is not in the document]
petermetz: @RafaelAPB @brunoffmateus I'd vote  for the build-in language features over the `class-validator` library because it allows the compiler to have an awareness of the result of runtime checks. What I mean is that if you have a function input parameter that you are validating then with class-validator you know in your head that the execution won't get past the validation (it will crash which is the expected behavior) BUT the compiler won't know this so it will still treat a potentially falsy input parameter as potentially falsy, whereas if you validate with a type-guard (the built-in language feature) then the compiler will start treating the potentially falsy variable as definitely-not-falsy which means that you can eliminate unsafe looking casts from your code (which are a code smell)

packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-handlers/stage0-handler.ts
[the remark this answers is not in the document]
LordKubaya: @petermetz Yes this is being worked on now. But I was saying to add this log to all the handlers for now.

packages/cactus-plugin-satp-hermes/package.json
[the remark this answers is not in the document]
brunoffmateus: This was a consequence of adding `import { Config } from "node-ssh"` to some files; however, it turned out that it wasn't necessary, so I removed it. Thanks for the warning!

packages/cactus-plugin-satp-hermes/run-gateway.sh
brunoffmateus: done!

packages/cactus-plugin-satp-hermes/generate-keys.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -e
```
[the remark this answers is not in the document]
brunoffmateus: also done!

packages/cactus-plugin-satp-hermes/package.json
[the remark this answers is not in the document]
LordKubaya: Thanks Bruno

brunoffmateus:
still needs work due to the changes of commit 9f1b080dbcfd75b836fbdf71e011591e5d245937, doing them now

RafaelAPB (approved):

packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-handlers/stage0-handler.ts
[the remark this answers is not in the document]
petermetz: @LordKubaya Oh, facepalm. Sorry, I misread what you wrote. Then we are in agreement.

PR #3548, opened 2024-09-23
fix(weaver-corda): added duplicate handling strategy in build.gradle

sandeepnRES:
This bug was detected in release of `v2.0.0-rc.4`, by the deployment workflow

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/weaver/common/protos-java-kt/build.gradle
+++ b/weaver/common/protos-java-kt/build.gradle
@@ -76,7 +76,6 @@ dependencies {
     implementation "io.grpc:grpc-netty-shaded:$grpc_version"
     implementation "io.grpc:grpc-protobuf:$grpc_version"
     implementation "io.grpc:grpc-stub:$grpc_version"
-    implementation 'javax.annotation:javax.annotation-api:1.3.2'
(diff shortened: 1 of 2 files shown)


petermetz (approved):

petermetz:
@sandeepnRES Please include the same information in the commit message next time that you added in the PR description after the fact.

Context: The commit parity check is getting stabilized (it was failing with false negatives up until recently) and it failing on this PR seems like a legit problem because the body of the commit message is empty while the PR description has `This bug was detected in release of v2.0.0-rc.4, by the deployment workflow` in it. So the `This bug was detected in release of v2.0.0-rc.4, by the deployment workflow` string is only present in a private database (GitHub) and not accessible through the commit log.

We haven't turned on the PR commit parity check yet to be mandatory but eventually (hopefully in the near future) it would be great if we could do that and so I'm just setting the stage for that with this comment (and a few more where I'll let people know as well).

---

@jagpreetsinghsasan Please try to spread the word about the same in other PRs you review and then in some time from now we can propose to the maintainers to enable the check as a required one. I'm hoping that us talking about it first and calling it out, the transition will be smoother compared to if we just turned it on as mandatory and then everyone would be scrambling with their PRs.

Issue #3140, opened 2024-03-23
ci(github): add check to validate exported types being correct


petermetz:
### Description
As a maintainer/reviewer I want to have automation that checks if our exports are correct for both CJS/ESM so that we are less likely to accidentally merge something that would make the exports mess up and cause import issues for the downstream libraries/applications.

https://www.npmjs.com/package/@arethetypeswrong/cli

### Acceptance Criteria
1. Does not duplicate the entire build process needlessly.
2. Evaluate if we should do it for every commit or just prior to release issuance.
3. Be mindful of CI resources consumed by this in general.

petermetz:
@ruzell22 

```sh
yarn lerna exec 'npm pack'
```

```sh
$ attw ./packages/cactus-plugin-ledger-connector-besu/hyperledger-cactus-plugin-ledger-connector-besu-2.0.0-rc.3.tgz 

@hyperledger/cactus-plugin-ledger-connector-besu v2.0.0-rc.3

 No problems found 🌟


┌───────────────────┬────────────────────────────────────────────────────┐
│                   │ "@hyperledger/cactus-plugin-ledger-connector-besu" │
├───────────────────┼────────────────────────────────────────────────────┤
│ node10            │ 🟢                                                 │
├───────────────────┼────────────────────────────────────────────────────┤
│ node16 (from CJS) │ 🟢 (CJS)                                           │
├───────────────────┼────────────────────────────────────────────────────┤
│ node16 (from ESM) │ 🟢 (CJS)                                           │
├───────────────────┼────────────────────────────────────────────────────┤
│ bundler           │ 🟢                                                 │
└───────────────────┴────────────────────────────────────────────────────┘
```

petermetz:
@ruzell22 

1. The above example works for a single .tgz file
2. You  can write a globby pattern in one of the nodejs tooling scripts (the ones that are within the `custom-checks` area)
3. Capture a list of the file paths for all the `*.tgz` files and then run a shell command for each individually

petermetz:
https://github.com/arethetypeswrong/arethetypeswrong.github.io/tree/main/packages/core

@ruzell22 If possible use the library not the CLI (which would force you to run shell commands from the NodeJS script)

PR #3561, opened 2024-09-30
ci(github): add check to validate exported types being correct

ruzell22:
### **Commit** to be reviewed
---
ci(github): add check to validate exported types being correct
```
Primary Changes
---------------
1. Added get-all-tgz-path.ts to get all tgz files path
2. Added run-attw-on-tgz.ts to run attw on each tgz filepath
3. Added workflow to run the are-the-types-wrong script
```

Fixes: #3140

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


.github/workflows/are-the-types-wrong.yaml
```diff
@@ -0,0 +1,44 @@
+name: AreTheTypesWrong scan
```
outSH, line 1: Use dash-case for name to follow convention used by other workflows

.github/workflows/are-the-types-wrong.yaml
```diff
@@ -0,0 +1,44 @@
...
+        run: >
+          sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys CE7709D068DB5E88 \
+            && sudo add-apt-repository "deb https://repo.sovrin.org/sdk/deb bionic stable" \
+            && sudo apt-get update \
+            && sudo apt-get install -y \
+                libindy \
+                libnullpay \
+                libvcx \
+                indy-cli \
+            && sudo rm -f /etc/apt/sources.list.d/sovrin.list*
```
outSH, line 34: No need to install indy SDK anymore, if you encountered similar code in another workflow file remove it from there as well

tools/custom-checks/get-all-tgz-path.ts
```diff
@@ -0,0 +1,53 @@
+import path from "path";
```
outSH, line 1: I think functions from this file should be merged with the mains script, it's short and not self-contained. For me, custom-checks should contain only executables

tools/custom-checks/get-all-tgz-path.ts
```diff
@@ -0,0 +1,53 @@
...
+import lernaCfg from "../../lerna.json" assert { type: "json" };
+
+/**
+ * Interface for the response of the getAllTgzPath function.
+ * @property {string} gitRootDir - The root directory of the git project.
+ * @property {Array<string>} relativePaths - An array of relative paths to the
+ * package directories.
+ * @property {Array<string>} absolutePaths - An array of absolute paths to the
+ * package directories.
+ */
```
outSH, line 13: Interface comment does not match the interface

tools/custom-checks/run-attw-on-tgz.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  {
+    const { relativePaths: tgzFilesRelative } = await getAllTgzPath();
+    console.log(`${TAG} Found ${tgzFilesRelative.length} .tgz files.`);
+    for (const filePath of tgzFilesRelative) {
+      await execCommand("attw", filePath);
+    }
+  }
+
+  function execCommand(binaryName: string, filePath: string): Promise<void> {
```
outSH, line 20: You can use `promisify` from `node:util` to simplify this function (by awaiting the exec method). Or, you could just use execSync since we await for each file anyway

tools/custom-checks/run-attw-on-tgz.ts
```diff
@@ -0,0 +1,44 @@
...
+          resolve();
+        }
+      });
+    });
+  }
+}
+
+if (esMain(import.meta)) {
+  await runAttwOnTgz();
+  if (attwFailedPackages.length > 0) {
```
outSH, line 38: `attwFailedPackages` should be returned from the method itself and not updated by a function call as a side effect

outSH (changes requested):
Thanks for the PR! I've left few comments, plese re-request the review once they're resolved :)

tools/custom-checks/get-all-tgz-path.ts
ruzell22: as for this one, I got this from get-all-pkg-dirs.ts. I based the code there for get-all-tgz-path.ts so I don't know how to resolve this one

petermetz (changes requested):
@ruzell22 This is looking closer now, but what I need is that if we run `yarn custom-checks` then the attw check is executed as part of it. Right now it's a standalone script only (it's good that it can be executed by itself as well but it **also** needs to be a part of the custom-checks so that in ci.yaml when the custom-checks are executed this also runs)

outSH (approved):
LGTM, thank you

package.json
```diff
@@ -57,6 +57,7 @@
     "format:prettier": "prettier --write --config .prettierrc.js \"./**/{openapi.json,*.ts,*.js}\"",
     "spellcheck": "cspell lint --no-progress \"*/*/src/**/*.{js,ts}\"",
     "tsc": "NODE_OPTIONS=\"--max_old_space_size=3072\" tsc --build --verbose",
+    "are-the-types-wrong": "TS_NODE_PROJECT=./tools/tsconfig.json node --trace-deprecation --experimental-modules --abort-on-uncaught-exception --loader ts-node/esm --experimental-specifier-resolution=node ./tools/custom-checks/run-attw-on-tgz.ts",
```
petermetz, line 60: @ruzell22 Please prefix the script name with `tools:` to follow the convention of the other similar scripts

tools/custom-checks/run-custom-checks.ts
```diff
@@ -73,6 +75,18 @@ export async function runCustomChecks(
     overallSuccess = overallSuccess && success;
   }
 
+  {
+    const { relativePaths } = await getAllTgzPath();
+
+    const req: IGetAllTgzPathResponse = {
```
petermetz, line 81: @ruzell22 Please fix the linter error here regarding the unused variable `req`

.github/workflows/ci.yaml
```diff
@@ -359,7 +359,12 @@ jobs:
           path: ./.yarn/
           restore-keys: |
             ${{ runner.os }}-yarn-${{ hashFiles('./yarn.lock') }}
+      - name: Installing AreTheTypesWrong Library
+        run: npm i -g @arethetypeswrong/cli
```
petermetz, line 363: @ruzell22 Please declare dependencies in the package json file instead of manually installing them in a script.

.github/workflows/ci.yaml
```diff
@@ -359,7 +359,12 @@ jobs:
...
       - run: ./tools/ci.sh
+      - run: yarn lerna exec 'npm pack'
```
petermetz, line 365: @ruzell22 Please run this as part of the script so that the script is self contained. You can run shell commands from NodeJS the same way you already do when invoking attw

.github/workflows/ci.yaml
```diff
@@ -359,7 +359,12 @@ jobs:
...
+      - name: Running AreTheTypesWrong scan
+        run: yarn run are-the-types-wrong
```
petermetz, line 367: @ruzell22 This can be deleted because the custom-checks job will run the attw check as part of itself.

petermetz (changes requested):

petermetz (changes requested):
@ruzell22 We are getting closer! Now I tested it and found 2 more issues:
1. The error messages you return are just the paths of the .tgz files with issues but there's nothin in there about what the problem is. Instead of logging the execCommand output to stdout, return it as a string that can be appended to the custom-check  errors so that the error messages at the end contain the information we need.
2. Please also abide by the existing conventions of the error messages like prefixing them with ERROR if I remember correctly.
3. Right now the glob captures all .tgz files meaning that the check is failing on my machine  because I have some leftover .tgz files from an earlier release (2.0.0-rc.3). Please add an extra step that first wipes all the matching .tgz files BEFORE it runs attw and the packaging step (so that we always start with a clean slate with the only .tgz file being in there the one that we need)

petermetz (changes requested):
@ruzell22 One more thing: Please apply the automatic formatting to the .ts files that you've added. There are linter errors in there about let vs. const variable declarations for example.

petermetz:
@ruzell22 Are you still working on this?

petermetz:
> @petermetz yes, I am. I currently have the codes ready to be push with the added extra step to wipe all the existing .tgz files first and also the additional information on the .tgz files with error after running attw. However, I just noticed something when testing before pushing the code wherein on the previous test without the wiping, it patches and detects 71 tgz files. After I put the wipe before patching and detecting tgz files, it can only detect 46 file paths.

@ruzell22 Got it, thank you for the update! Take your time, I just wanted to make sure that the PR is not being held up by something.

petermetz (changes requested):
@ruzell22 Now it's crashing on my machine and the logging is unfortunately hiding the full output that you get from the child processes (attached log file 1).

I recommend refactoring the code in a way that streams the logs of the child processes directly to the stdout/stderr of the main process instead (attached code modified)

If you run it like that (with streamed output logs), then it at least gives you the full output which shows some logs that are just "undefined" which indicates that you still have some errors that aren't handled correctly.

The other thing I've seen is that there quite a few packages that are failing the ATTW checks. We don't have to fix these as part of this pull request, but we will need to fix them before this PR gets merged because otherwise the CI would break down because these new checks would make the mandatory checks fail for everyone in all PRs.

So, there might be still multiple issues but let's just take it one at a time, you wanna make sure that the error messages include the details (e.g. on the screenshot below you want to capture the information in red in the section of the logs where I marked it yellow).
This will ensure that the error details don't get lost when people have failures.

image


[2024-11-11-cacti-attw-custom-checks-streamed-output.crash.log](https://github.com/user-attachments/files/17706259/2024-11-11-cacti-attw-custom-checks-streamed-output.crash.log)
[2024-11-11-cacti-custom-check-attw.crash.log](https://github.com/user-attachments/files/17706261/2024-11-11-cacti-custom-check-attw.crash.log)

`tools/custom-checks/run-attw-on-tgz.ts`

```typescript
import esMain from "es-main";
import { getAllTgzPath } from "./get-all-tgz-path";
import { exit } from "process";
import { rm } from "fs";
import { spawn } from "child_process";

// New function to stream subprocess output in real-time
function spawnPromise(
  command: string,
  args: string[],
  options = {},
): Promise<void> {
  return new Promise((resolve, reject) => {
    const child = spawn(command, args, { shell: true, ...options });

    // Stream output to main process stdout and stderr
    child.stdout.pipe(process.stdout);
    child.stderr.pipe(process.stderr);

    child.on("close", (code) => {
      if (code === 0) {
        resolve();
      } else {
        reject(new Error(`Process exited with code ${code}`));
      }
    });

    child.on("error", (err) => reject(err));
  });
}

async function cleanUpTgzFiles(): Promise<void> {
  const TAG = "[tools/custom-checks/run-attw-on-tgz.ts]";
  console.log(`${TAG} Cleaning up existing .tgz files...`);

  const { relativePaths: tgzFilesRelative } = await getAllTgzPath();

  for (const filePath of tgzFilesRelative) {
    await rm(filePath, { recursive: true, force: true }, () => {});
    console.log(`${TAG} Deleted ${filePath}`);
  }
}

export async function runAttwOnTgz(): Promise<[boolean, string[]]> {
  await cleanUpTgzFiles();

  const TAG = "[tools/custom-checks/run-attw-on-tgz.ts]";
  await execCommand("yarn lerna exec 'npm pack'", true);
  console.log(`${TAG} Packaging .tgz files`);

  console.log(`${TAG} Fetching .tgz file paths.`);
  const { relativePaths: tgzFilesRelative } = await getAllTgzPath();

  const attwFailedPackages: string[] = [];

  for (const filePath of tgzFilesRelative) {
    try {
      const output = await execCommand("attw", filePath);
      console.log(output);
    } catch (error: any) {
      attwFailedPackages.push(
        `ERROR ${filePath}: ${error.message} ${error.output}`,
      );
    }
  }

  const success = attwFailedPackages.length === 0;
  return [success, attwFailedPackages];
}

async function execCommand(
  binaryName: string,
  argument: string | boolean,
): Promise<void> {
  let command;
  if (typeof argument === "boolean") {
    command = binaryName;
  } else if (typeof argument === "string") {
    command = `${binaryName} ./${argument}`;
  } else {
    throw new Error("Invalid arguments for execCommand");
  }

  await spawnPromise(command, []);
}

if (esMain(import.meta)) {
  const [success, attwFailedPackages] = await runAttwOnTgz();
  if (!success) {
    console.log("Types are wrong for these packages:");
    console.log(attwFailedPackages);
    exit(1);
  }
  exit(0);
}

```

ruzell22:
@petermetz on the latest push, there are no undefined output now. Also, I have created a new issue #3623 that relates to issue #3140 for fixing the packages that are failing with ATTW check in custom-checks.

petermetz (changes requested):
@ruzell22 Yes! We are looking good! Thank you!
Now for the final part, since we don't want to hold this PR up by waiting for the broken type exports to be fixed, what I'll ask just to resolve the circular dependency between the broken type exports and the passing CI checks is to:
1. Exclude the broken packages from the check for now, in a way that there is a single entry in the ignore patterns for each package AND
2. There is a comment above each ignore pattern referencing a GitHub issue that was created with the specific task of fixing the broken types of that specific package.

What this achieves is that we can merge this PR today without melting down the CI and yet for new packages we get the type export validation out of the box working.

We also get the benefit that from now on this check will guard against mistakes where someone breaks the otherwise currently correct type exports for packages where it's good.

I've collected the list of ignore patterns that you'll need but please create the issues for each separately and then add the comments for each line with the link to the specific issue. 
Also leave a comment explaining everything I just talked about. The guiding principle there should be that whoever stumbles upon those ignore patterns should be able to figure out what is happening without having to ask you or me for historical context. Ideally if they would like to work on fixing those type exports they can get cracking with it right after having seen the comments. This is the general principle behind it that gives us a better chance longer term of having a more maintainable codebase.

The ignore paths you'll need in the globby options:

`tools/custom-checks/get-all-tgz-path.ts`

```typescript
  const globbyOpts: GlobbyOptions = {
    cwd: PROJECT_DIR,
    onlyFiles: true,
    expandDirectories: false,
    ignore: [
      "**/node_modules",
      "weaver/core/drivers/fabric-driver/hyperledger-cacti-weaver-driver-fabric-*.tgz",
      "weaver/core/identity-management/iin-agent/hyperledger-cacti-weaver-iin-agent-*.tgz",
      "weaver/sdks/fabric/interoperation-node-sdk/hyperledger-cacti-weaver-sdk-fabric-*.tgz",
      "packages/cacti-plugin-weaver-driver-fabric/src/main/typescript/hyperledger-cacti-weaver-driver-fabric-*.tgz",
      "packages/cacti-plugin-copm-fabric/hyperledger-cacti-cacti-plugin-copm-fabric-*.tgz",
      "packages/cacti-ledger-browser/hyperledger-cacti-ledger-browser-*.tgz",
      "examples/cactus-example-cbdc-bridging-frontend/hyperledger-cacti-example-cbdc-bridging-frontend-*.tgz",
      "examples/cactus-common-example-server/hyperledger-cactus-common-example-server-*.tgz",
      "packages/cactus-verifier-client/hyperledger-cactus-verifier-client-*.tgz",
      "packages/cactus-plugin-ledger-connector-polkadot/hyperledger-cactus-plugin-ledger-connector-polkadot-*.tgz",
      "packages/cactus-common/hyperledger-cactus-common-*.tgz",
    ],
  };
```

Issue #3406, opened 2024-07-12
fix(connector-besu): do not crash if ledger unreachable - send HTTP 503


petermetz:
### Description

Right now the API server will exit with a crash if the Besu connector receives a request and the backing ledger is not accessible.
Instead of a full-on crash, we need to treat this as a recoverable error and send back `HTTP 503 - Service Unavailable`
with a  Retry-After header with exponential backoff where it starts from 1 second and the doubles every re-try.

The easiest way to reproduce is to follow this guide and remove the part of the command from step 2 that binds the API server to the host machine's network: `--network=host \`

https://github.com/hyperledger/cacti/tree/main/packages/cactus-plugin-ledger-connector-besu#testing-api-calls-with-the-container

### Acceptance Criteria
1. Test coverage included for this case.
2. Test verifies that API server does not crash in the mentioned use-case.

jagpreetsinghsasan:
Looks like Orion isn't configured correctly in the `cactus-besu-all-in-one:2024-07-04-8c030ae`.
@ruzell22 please check if the issue persists on the latest tag as well or not. If the error still persists, we will increase the scope of this task with a potential fix for the same.

petermetz:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After
Retry-After - HTTP | MDN
The Retry-After response HTTP header indicates how long
  the user agent should wait before making a follow-up request. There are three main cases
  this header is used:
 
Exponential back-off - the delay between retries is increasing exponentially with every retry.
So first you wait 2 seconds, 4, 8, etc.

petermetz:
Let's say start with the run-transaction endpoint of the besu connector and do these things:
1. Refactor the method within the connector so that it throw the specific kind of exception, let's say you create a new exception type called `BesuLedgerInaccessibleException` which is a sub-class of `Error` and you throw this when you detect the issue in the connector that a transaction was not possible to be executed because the network connection to the ledger is broken.
2. Then in the HTTP endpoint implementation in the handleRequest method you can write some logic in the `catch` block that sets the retry-after headers (linked above) but it only should do this if the exception you caught (that was thrown  by the connector) is an `instanceof  BesuLedgerInaccessibleException` so that way we know for sure that this was the problem.

petermetz:
https://github.com/web3/web3.js/issues/6900

petermetz:
[...]

PR #3573, opened 2024-10-04
fix(connector-besu): do not crash if ledger unreachable - send HTTP 503

ruzell22:
### Commit to be reviewed
---
fix(connector-besu): do not crash if ledger unreachable - send HTTP 503
```
Primary Changes
---------------
1. Yarn patch to web3-eth-accounts@1.6.1 so that it does not crash nodejs process
Without the catch block, the rejection is an unhandled rejection that bubbles
up to the top of the callstack where NodeJS itself catches it and then crashes
the entire process. (used to be that it just logged a warning but since some
of the newer versions it crashes which allows us to find these bugs in our
code / library's code)
2. It is returning a 503 instead of a 500
3. Added a static retry-after header value of 5 seconds
4. Added test case in test-ledger which has run-transaction at the end that is
expected to give error 503 when the backing ledger is unavailable
```

Fixes: #3406

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


.yarn/patches/web3-eth-accounts-npm-1.6.1-c95f31ca81.patch
```diff
@@ -0,0 +1,13 @@
...
++++ b/lib/index.js
+@@ -344,7 +344,7 @@ function _handleTxPricing(_this, tx) {
+                             throw Error("Network doesn't support eip-1559");
+                         resolve({ gasPrice });
+                     }
+-                });
++                }).catch((ex) => reject(ex));
+             }
+         }
+         catch (error) {
```
outSH, line 13: Please provide more details about this (in commit & PR message). In particular:
- Link to an issue reported on web3js repository
- What this patch does
- Why we need this, what will happen if we remove it (specific scenario).

outSH, line 13: Please provide more details about this (in commit & PR message). In particular:
- Link to an issue reported on web3js repository
- What this patch does
- Why we need this, what will happen if we remove it (specific scenario).

outSH (changes requested):

.yarn/patches/web3-eth-accounts-npm-1.6.1-c95f31ca81.patch
petermetz: @ruzell22 +1 to the above, this should be explained in detail.
@outSH I wrote the code, here's the answer that @ruzell22 can copy paste into the PR description and the commit message: Without the catch block the rejection is an unhandled rejection that bubbles up to the top of the callstack where NodeJS itself catches it and then crashes the entire process. (used to be that it just logged a warning but since some of the newer versions it crashes which allows us to find these bugs in our code / library's code)

petermetz (changes requested):
@ruzell22 Looking good, but please add a specific test case within the run-transaction endpoint's existing test suite so that we have a specific unit test verifying that this logic now works as intended.
Please put the test case in the test-plugin-ledger-connector-besu package because that way the API server can be used as well and so it's a more complete simulation.

.yarn/patches/web3-eth-accounts-npm-1.6.1-c95f31ca81.patch
jagpreetsinghsasan: @petermetz this is a nice learning! (Never knew that we can patch and include packages this way as well)

petermetz: @jagpreetsinghsasan Yeah, it's very useful when you need it. :-)

petermetz (changes requested):
@ruzell22 I can't find the relevant test case that creates the scenario in question (no ledger connectivity and then assertions to verify that the API server hasn't crashed and that the response was sent back as expected).

.yarn/patches/web3-eth-accounts-npm-1.6.1-c95f31ca81.patch
outSH: @ruzell22 Please update the PR message as well. Also, if you could, please put the description in the point where you describe the changes (`Yarn patch to web3-eth-acconts@1.6.1 so that it does not crash nodejs process`) to be sure that this description you've provided applies to the patching.

outSH (changes requested):

petermetz (changes requested):
@ruzell22 I don't see the API server being used in the tests, could you please double check?

ruzell22:
Hello @petermetz , the test is added at the end of packages/cactus-test-plugin-ledger-connector-besu/src/test/typescript/integration/plugin-ledger-connector-besu/openapi/openapi-validation.test.ts . The ledger was shutdown before it to force the ledger unavailable status when running the runTransactionV1.

image

petermetz (approved):

jagpreetsinghsasan:
Please fix `yarn-codegen` check as well @ruzell22

outSH (approved):

PR #3578, opened 2024-10-08
feat(fabric): serialization of ccp and sshconfig

### LINKED ISSUE #3577: feat(fabric): serialization of ccp and sshconfig
## Description
As a developer, I want to enable serialization of connection profile and sshconfig as base64 encoded strings, so that we can pass these complicated objects easily while initializing the fabric connector over api-server.
This task is related to #3408 (ease of doing api calls)

## Acceptance Criteria
The new feature shouldn't prevent the existing functionality to work (i.e. passing the `sshconfig` and `connectionProfile` in the usual way)

jagpreetsinghsasan:
## Commit to be reviewed

feat(fabric): serialization of ccp and sshconfig 
  
    Primary Changes
    ---------------
    1. sshConfig and connectionProfile can now be passed as base64 encoded strings

    2. Added sshConfigBase64Encoded and connectionProfileBase64encoded properties to IPluginLedgerConnectorFabricOptions
    3. Added necessary FabricConnector class variables
    4. Added a new testcase to demonstrate the same
    5. Updated an implementation which fetches sshConfig from class variable

Fixes #3577

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
+++ b/packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
@@ -189,9 +189,11 @@ export interface IPluginLedgerConnectorFabricOptions
   cliContainerGoPath?: string;
   cliContainerEnv: NodeJS.ProcessEnv;
   pluginRegistry: PluginRegistry;
-  sshConfig: SshConfig;
+  sshConfig?: SshConfig;
+  sshConfigB64?: string;
   readonly sshDebugOn?: boolean;
-  connectionProfile: ConnectionProfile;
+  connectionProfile?: ConnectionProfile;
+  connectionProfileB64?: string;
   prometheusExporter?: PrometheusExporter;
   discoveryOptions?: GatewayDiscoveryOptions;
   eventHandlerOptions?: GatewayEventHandlerOptions;
@@ -219,6 +221,8 @@ export class PluginLedgerConnectorFabric
   private readonly peerBinary: string;
   private readonly goBinary: string;
   private readonly cliContainerGoPath: string;
+  private readonly sshConfig: SshConfig;
+  private readonly connectionProfile: ConnectionProfile;
   public prometheusExporter: PrometheusExporter;
   private endpoints: IWebServiceEndpoint[] | undefined;
   private readonly secureIdentity: SecureIdentityProviders;
@@ -241,7 +245,6 @@ export class PluginLedgerConnectorFabric
     Checks.truthy(opts.instanceId, `${fnTag} options.instanceId`);
     Checks.truthy(opts.peerBinary, `${fnTag} options.peerBinary`);
     Checks.truthy(opts.pluginRegistry, `${fnTag} options.pluginRegistry`);
-    Checks.truthy(opts.connectionProfile, `${fnTag} options.connectionProfile`);
     this.prometheusExporter =
       opts.prometheusExporter ||
       new PrometheusExporter({ pollingIntervalInMin: 1 });
@@ -277,9 +280,33 @@ export class PluginLedgerConnectorFabric
     });
     this.certStore = new CertDatastore(opts.pluginRegistry);
 
+    if (this.opts.connectionProfile) {
+      this.connectionProfile = this.opts.connectionProfile;
+    } else if (this.opts.connectionProfileB64) {
+      const connectionProfileBuffer = Buffer.from(
+        this.opts.connectionProfileB64,
+        "base64",
+      );
+      const connectionProfileString = connectionProfileBuffer.toString("utf-8");
+      this.connectionProfile = JSON.parse(connectionProfileString);
+    } else {
+      throw new Error(
+        "Cannot instantiate Fabric connector without connection profile.",
+      );
+    }
+
     this.sshDebugOn = opts.sshDebugOn === true;
+    if (this.opts.sshConfig) {
+      this.sshConfig = this.opts.sshConfig;
+    } else if (this.opts.sshConfigB64) {
+      const sshConfigBuffer = Buffer.from(this.opts.sshConfigB64, "base64");
+      const sshConfigString = sshConfigBuffer.toString("utf-8");
+      this.sshConfig = JSON.parse(sshConfigString);
+    } else {
+      throw new Error("Cannot instantiate Fabric connector without SSH config");
+    }
     if (this.sshDebugOn) {
-      this.opts.sshConfig = this.enableSshDebugLogs(this.opts.sshConfig);
+      this.sshConfig = this.enableSshDebugLogs(this.sshConfig);
     }
 
     this.signCallback = opts.signCallback;
@@ -361,11 +388,11 @@ export class PluginLedgerConnectorFabric
     req: DeployContractV1Request,
   ): Promise<DeployContractV1Response> {
     const fnTag = `${this.className}#deployContract()`;
-    const { log, opts } = this;
+    const { log } = this;
 
     const ssh = new NodeSSH();
     this.log.debug(`${fnTag} Establishing SSH connection to peer...`);
-    await ssh.connect(opts.sshConfig);
+    await ssh.connect(this.sshConfig);
     this.log.debug(`${fnTag} Established SSH connection to peer OK.`);
 
     if (req.collectionsConfigFile) {
@@ -666,6 +693,7 @@ export class PluginLedgerConnectorFabric
       log,
       opts: this.opts,
       dockerBinary: this.dockerBinary,
+      sshConfig: this.sshConfig,
       className: this.className,
     };
     return deployContractGoSourceImplFabricV256(ctx, req);
@@ -910,7 +938,7 @@ export class PluginLedgerConnectorFabric
     return createGateway({
       logLevel: this.opts.logLevel,
       pluginRegistry: this.opts.pluginRegistry,
-      defaultConnectionProfile: this.opts.connectionProfile,
+      defaultConnectionProfile: this.connectionProfile,
       defaultDiscoveryOptions: this.opts.discoveryOptions || {
         enabled: true,
         asLocalhost: true,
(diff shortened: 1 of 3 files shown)


packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
```diff
@@ -189,9 +189,11 @@ export interface IPluginLedgerConnectorFabricOptions
   cliContainerGoPath?: string;
   cliContainerEnv: NodeJS.ProcessEnv;
   pluginRegistry: PluginRegistry;
-  sshConfig: SshConfig;
+  sshConfig?: SshConfig;
+  sshConfigBase64Encoded?: string;
   readonly sshDebugOn?: boolean;
-  connectionProfile: ConnectionProfile;
+  connectionProfile?: ConnectionProfile;
+  connectionProfileBase64Encoded?: string;
```
outSH, line 196: Can we accept `connectionProfile` as discriminated union? Something like: `{type: "base64", profile: "XXX"}`. I was thinking about adding a method for returning connection profile from a connector (as part of https://github.com/hyperledger-cacti/cacti/issues/3554) and this common union type could be used as a return value as well. Also, it will be more extensible if we decide to support more connection profile formats in the future (no need to bloat the options with more optional options). And finally this will simplify checking if connection profile was provided after all.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1-serialized.test.ts
```diff
@@ -0,0 +1,310 @@
+import http from "http";
```
outSH, line 1: I think peter was opting for minimizing number of fabric tests to speed up it's CI job, can we push this into already existing test file?

outSH (changes requested):
I've left two open questions, please have a look :)

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1-serialized.test.ts
petermetz: @jagpreetsinghsasan Plus one to the above. It's my fault for being inconsistent with the requirements over time, apologies. The performance of the CI is still very much below what we'd like to have in an ideal situation so some shortcuts have to be made and one of them is to accept that some test cases won't run in full isolation (with their dedicated pristine ledger)

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
jagpreetsinghsasan: @outSH an interesting suggestion in my opinion! If we see the functionality of returning connection profile from the connector itself, in a production grade fabric-network, then I think it will be very hard to create a connection profile as, creation of connection profile will make our connector tightly coupled with the infrastructure where HL Fabric is deployed. This is because, the main problem in publishing such network level configuration always end up in the question of "`who will be providing` `what part of the information`" and it then messes up with the overall governance.
Maybe we can instead request that information from the underlying ledger itself as a pre-requisite (in our case, straightaway fetch from the AIO itself without having another function for the same in our connector and we do have that function in the test tooling class for fabric aio).
My only point here being, whenever there is a requirement which deals with the entire ledger level information such as `who provides what part of that info`, `which channel is used to provide that info` and `where is such an info stored` like questions, it becomes very challenging to deal with. So integrating such functionality with our connector will pose future challenges to decouple it again.
This is just my point of view, what do you suggest @petermetz @outSH ?

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1-serialized.test.ts
jagpreetsinghsasan: @outSH @petermetz yes, a much better suggestion. I will add this to an existing test case itself and re-request again.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
jagpreetsinghsasan: I agree.

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
```diff
@@ -219,6 +221,8 @@ export class PluginLedgerConnectorFabric
   private readonly peerBinary: string;
   private readonly goBinary: string;
   private readonly cliContainerGoPath: string;
+  private readonly sshConfig!: SshConfig;
```
petermetz, line 224: @jagpreetsinghsasan Is there anything preventing us from definitively assigning these fields in the constructor so that we don't need the `!` syntax?

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
petermetz, line 196: @jagpreetsinghsasan I recommend calling it `connectionProfileB64` or `connectionProfileBase64` to be more concise but still carry the same information

petermetz: ^^ Agreed as well. @jagpreetsinghsasan @outSH

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1.test.ts
```diff
@@ -251,6 +251,29 @@ describe(testCase, () => {
     }
 
     {
+      const connectionProfileBase64Encoded = Buffer.from(
```
petermetz, line 254: @jagpreetsinghsasan Please try to avoid chaining method calls so that if there is a crash with a line number there is no chance of ambiguity where we would not be able to know for sure which method invocation crashed.

packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1.test.ts
```diff
@@ -267,10 +290,9 @@ describe(testCase, () => {
         endorsingOrgs: ["org1.example.com", "Org2MSP"],
       };
 
-      const res = await apiClient.runTransactionV1(req);
+      const res = await pluginWithSerializedInputs.transact(req);
```
petermetz, line 293: @jagpreetsinghsasan Please also leave in the test for the API client not just through direct invocation. Otherwise we might lose coverage. It's important to verify that the same operation works through the API too not just via direct invocation so that we can continue to be "API first"

petermetz (changes requested):
LGTM with comments

packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts
jagpreetsinghsasan: I declared that this field cannot be null via `!` syntax, then assigned this field in the constuctor from the 2 possible options, either from the usual `sshConfig` or from the base64 encoded version of it and finally controlled the flow using `Checks.truthy(this.sshConfig != null, `${fnTag}:sshConfig`);`
I did it in this way as the other way would be to declare the var without `!` and then do some workaround to address the issue of `variable is used before assigning`
Screenshot (56)
In short, leveraged the use of `Checks.truthy( )` alongwith `var!` to have precise declaration.
Please let me know if this has some flaws/vulnerability or there exists a better way to do the same.

petermetz: @jagpreetsinghsasan The if conditions can be refactored this way and then the compiler no longer has an issue with us dropping the `!` (which should only be used if there is no way to avoid it due it turning off a safety feature of the compiler):

```typescript
if (this.opts.sshConfig) {
    this.sshConfig = this.opts.sshConfig;
} else if (this.opts.sshConfigB64) {
    const sshConfigBuffer = Buffer.from(this.opts.sshConfigB64, "base64");
    const sshConfigString = sshConfigBuffer.toString("utf-8");
    this.sshConfig = JSON.parse(sshConfigString);
} else {
    throw new Error("Cannot instantiate Fabric connector without SSH config");
}

```

I'm guessing the other class member could be refactored the same way so that it does not need the `!`

petermetz (changes requested):

petermetz (approved):
LGTM, thank you for the updates!

petermetz (approved):

outSH (approved):

petermetz (changes requested):
@jagpreetsinghsasan We are almost there! Please fix the tests that started failing. I saw this in the logs:

```sh
AIL packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1.test.ts (177.176 s, 1102 MB heap size)
  ● runs tx on a Fabric v2.5.6 ledger › runs tx on a Fabric v2.5.6 ledger

    Cannot instantiate Fabric connector without connection profile.

      292 |       this.connectionProfile = JSON.parse(connectionProfileString);
      293 |     } else {
    > 294 |       throw new Error(
          |             ^
      295 |         "Cannot instantiate Fabric connector without connection profile.",
      296 |       );
      297 |     }

      at new PluginLedgerConnectorFabric (packages/cactus-plugin-ledger-connector-fabric/src/main/typescript/plugin-ledger-connector-fabric.ts:294:13)
      at Object.<anonymous> (packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/run-transaction-endpoint-v1.test.ts:158:20)

```

petermetz (approved):
@jagpreetsinghsasan Thank you for the updates, looking good now. Please fix the linter errors and then we are good to go from my side.

petermetz (commented):
@jagpreetsinghsasan Please fix the linter errors

petermetz (approved):

PR #3600, opened 2024-10-23
fix(security): prevent XSS in wallet deletion route

bhaskar-allam:
Replaced the string concatenation with a template literal (), which automatically escapes the interpolated values.This change ensures that the walletIdEsc value is properly escaped when inserted into the output string, preventing potential XSS attacks while maintaining the core functionality of the route handler.

### DIFF
--- a/examples/cactus-common-example-server/src/main/typescript/routing-interface/routes/index.ts
+++ b/examples/cactus-common-example-server/src/main/typescript/routing-interface/routes/index.ts
@@ -102,7 +102,7 @@ router.delete(
   (req: Request, res: Response, next: NextFunction) => {
     try {
       const walletIdEsc = escapeHtml(req.params.id);
-      const out = "Not Implemented (Delete a Wallet, id=" + walletIdEsc + ")\n";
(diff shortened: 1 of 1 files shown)


outSH (changes requested):
I don't really see added value here:
- The value of `walletIdEsc` is already escaped by `escape-html`
- Template literals doesn't escape the string AFAIK, can you show small example of what you mean?

petermetz (changes requested):
> I don't really see added value here:
> - The value of `walletIdEsc` is already escaped by `escape-html`
> - Template literals doesn't escape the string AFAIK, can you show small example of what you mean?
>  

+1 - I'm also not aware of this behavior. The closest thing I can think of is tagged templates [1] but you aren't using that in your change so please do explain the details a little further.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates

petermetz:
@bhaskar-allam Closing due to inactivity. Please feel free to re-open anytime in the future if you decide to come back to this.

PR #3625, opened 2024-11-12
feat(copm): add Corda COPM implementation

jenniferlianne:
Primary change:

Implements COPM primitives for Corda as a cacti plugin.  The implementation follows the model of
the Corda ledger connector plugin, where a typescript pass-through implementation is registered on the plugin server, and commands are implemented on a separate grpc server in the Kotlin Spring framework.

Secondary changes:

- The lock claim protobuf structure was updated to reduce the number of parameters.
- A no-op endpoint was added, per project specification
- The format of the inter-network command for requesting the status of a pledge varies by remote network type and asset.  Adds a function to the typescript interop configuration  interface to supply the appropriate command given the remote network and asset.
- Adds test package to test combinations of network types
- Updates CI script to use new testing framework
- Updates linter ignore to include weaver build directories, allowing yarn lint to pass on a system where weaver libraries have been built. Also excludes docs directory to avoid linting minified js from documentation packages.

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


.github/actions/copm_test/action.yaml
```diff
@@ -1,8 +1,22 @@
...
+  github-token:
+    description: the github secret token variable
+    required: true
 runs:
   using: "composite"
   steps:
+    - name: free github runner disk space
+      shell: bash
+      run: |
+          sudo rm -rf /usr/share/dotnet
```
petermetz, line 16: @jenniferlianne Suggestion: Consider re-using the logic from ci.sh that does the same kind of space clearing. If it's not flexible enough and/or too heavy / too slow then please make a note in the comments here or open an issue on GH with suggestions (for exmaple we could pull the disk space freeing logic to a separate bash script that you can source and then re-use easily)

packages/cacti-copm-core/src/main/json/openapi.tpl.json
```diff
@@ -302,6 +291,24 @@
     }
   },
   "paths": {
+    "/api/v1/plugins/@hyperledger/cacti-plugin-copm/no-op": {
```
petermetz, line 294: @jenniferlianne I'm totally fine with naming it no-op but it also sounds like from the description that it could also just be called healthcheck. Not a change request. Just saying.

packages/cacti-copm-core/src/main/protos-kt/out/org/hyperledger/cacti/plugin/cacti/plugin/copm/core/services/defaultservice/DefaultServiceOuterClassGrpcKt.kt
```diff
@@ -0,0 +1,314 @@
+package org.hyperledger.cacti.plugin.cacti.plugin.copm.core.services.defaultservice
```
petermetz, line 1: @jenniferlianne If this is generated code then please make sure to have that represented somewhere along the directory path with a path element of `.../generated/...` so that it's always clearly distinguishable from hand-maintained code. It helps during reviews and also whenever we add more automation it's extremely helpful while constructing globs/regexes that find or exclude generated code from automation such as linting.

packages/cacti-copm-test/config/corda_rpc.json
```diff
@@ -0,0 +1 @@
+{"O=PartyA, L=London, C=GB@Corda_Network":{"host":"127.0.0.1","username":"clientUser1","password":"test","port":10006},"O=PartyB, L=London, C=GB@Corda_Network":{"host":"127.0.0.1","username":"clientUser1","password":"test","port":10009},"O=PartyA, L=London, C=GB@Corda_Network2":{"host":"127.0.0.1","username":"clientUser1","password":"test","port":30006},"O=PartyB, L=London, C=GB@Corda_Network2":{"host":"127.0.0.1","username":"clientUser1","password":"test","port":30009}}
```
petermetz, line 1: @jenniferlianne Is this something that has to be shipped to production or just a testing aritfact? If it's only for testing then I'd recommend placing it somewhere under src/test/json/resources or similar like src/test/json/fixtures/

packages/cacti-plugin-copm-corda/start-app.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Without these we get crashes on JDK 17 an above since the introduction of the
+# Java Modules system.
+
+for i in 1 2 3; do java -jar ${APP}/kotlin-spring/build/libs/copm-corda-0.4.jar && break || sleep 5; done
```
petermetz, line 6: @jenniferlianne Please update the version to the current latest release if possible

packages/cacti-plugin-copm-corda/src/main/typescript/plugin-copm-corda.ts
```diff
@@ -0,0 +1,77 @@
...
+import { CopmCordaImpl, DefaultApi } from "./service-implementation";
+import { createGrpcTransport } from "@connectrpc/connect-node";
+import { createPromiseClient } from "@connectrpc/connect";
+
+export interface IPluginCopmCordaOptions extends ICactusPluginOptions {
+  logLevel?: LogLevelDesc;
+  copmKotlinServerBaseUrl: string;
+}
+
+export class PluginCopmCorda implements IPluginCrpcService {
```
petermetz, line 22: @jenniferlianne Please also implement the `ICactusPlugin` interface

packages/cacti-plugin-copm-corda/src/main/kotlin/org/hyperledger/cacti/plugin/copm/corda/CordaAssetClaim.kt
```diff
@@ -0,0 +1,26 @@
+package org.hyperledger.cacti.plugin.copm.corda
+import org.hyperledger.cacti.plugin.copm.validators.ValidatedClaimPledgedAssetV1Request
+import net.corda.core.contracts.CommandData
+import org.hyperledger.cacti.weaver.imodule.corda.states.AssetClaimParameters
+
+class CordaAssetClaim(private var data: ValidatedClaimPledgedAssetV1Request,
```
petermetz, line 6: @jenniferlianne This looks like it might also be generated code but I'm not sure. If it is, please move it under a directory path that contains `generated` as a path element so that it's easier to distinguish. There might be other pockets of generated code that I didn't find but please apply this change to all other generated code as well.

petermetz (changes requested):
@jenniferlianne Looking pretty high quality in general, but as usual I have some requests. On top of the comments I left inline, please make sure to include a test case which verifies the plugin being functional while used through the API server (you'll most likely need a separate package because this one cannot depend on the API server due to circular dependency issues)
If you need examples just let me know.

petermetz (approved):
@jenniferlianne LGTM, thank you!

packages/cacti-copm-core/src/main/json/openapi.tpl.json
VRamakrishna: In our grand architecture diagram, the `no-op` function was meant to serve as a pass-through to the target ledger (Fabric, Corda, etc.) That way, when the COPM is integrated with the pipelines of all other packages (as envisioned in the architecture), it offers direct access to the ledger without performing any of the operations offered by the COPM (proof generation, locks, claims, etc.)

Currently, this function seems to be just returning without doing anything, which is fine. Later, when we integrate this module with other pipelines, we can change the logic to do a pass-through (we will need to design this in such a way that allows existing pipelines to work as intended with the COPM mediating interactions with the ledger.)

.github/actions/copm_test/action.yaml
jenniferlianne: I took a few runs at this but:
- the space must be cleared before building and deploying weaver, and
- as is, the command in ci.sh does not clear up enough space.  
 
I tried adding additional directories to the remove section, however that could impact other jobs -- and the copm_jobs failed regardless on the second call to ci.sh to run the tests. https://github.com/hyperledger-cacti/cacti/actions/runs/11838070296/job/32986450316

packages/cacti-copm-core/src/main/protos-kt/out/org/hyperledger/cacti/plugin/cacti/plugin/copm/core/services/defaultservice/DefaultServiceOuterClassGrpcKt.kt
jenniferlianne: This was committed by mistake.  Thanks.

packages/cacti-copm-test/config/corda_rpc.json
jenniferlianne: The file has been moved to /src/test/json

packages/cacti-plugin-copm-corda/start-app.sh
jenniferlianne: Done

packages/cacti-plugin-copm-corda/src/main/typescript/plugin-copm-corda.ts
jenniferlianne: Done.

packages/cacti-plugin-copm-corda/src/main/kotlin/org/hyperledger/cacti/plugin/copm/corda/CordaAssetClaim.kt
jenniferlianne: It's actually not generated.

packages/cacti-copm-core/src/main/proto/generated/openapi/models/asset_lock_claim_v1_pb.proto
```diff
@@ -22,16 +22,10 @@ message AssetLockClaimV1PB {
 
   string lock_id = 1;
 
-  AssetAccountV1PB source = 359634918;
-
   AssetAccountV1PB destination = 356105204;
```
sandeepnRES, line 25: I think I forgot to ask this in previous PR? or sorry if I'm asking again, but why such high numbers here? Usually we start from 1,2,3... and so on, contiguous small numbers.
Any reason for this?

packages/cacti-copm-test/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@hyperledger-cacti/cacti-copm-test",
```
sandeepnRES, line 2: I believe this should also be in examples. Considering this is only testing by calling client commands.

packages/cacti-copm-core/src/main/proto/generated/openapi/models/asset_lock_claim_v1_pb.proto
jenniferlianne: This was discussed here: https://github.com/hyperledger-cacti/cacti/pull/3546#discussion_r1766255442 and also on the project channel on October 24. 

The proto files are generated from the open api specification. The generator makes the numbers from the hash of the variable name, so it is guaranteed to be unique within the proto message, serving its intended purpose.  As requested by @VRamakrishna I tried setting the 'x-protobuf-index' in the openapi spec, however the variable is ignored for members of type 'ref'.

packages/cacti-copm-test/package.json
jenniferlianne: The PR follows the example of `cactus-plugin-ledger-connector-corda`, which is tested by `cactus-test-tooling' in the packages directory.  Other packages, such as the besu connector, have their own test directory under packages.  Please correct me if I'm wrong, @petermetz.

packages/cacti-copm-test/src/main/typescript/lib/weaver-interop-configuration.ts
```diff
@@ -0,0 +1,144 @@
...
+  Interfaces as CopmIF,
+  DLTransactionParams,
+  Validators,
+} from "@hyperledger-cacti/cacti-copm-core";
+import path from "path";
+import fs from "fs-extra";
+
+export class WeaverInteropConfiguration implements CopmIF.InteropConfiguration {
+  private log: Logger;
+  private weaverRelativePath = "../../../../../../weaver/";
```
sandeepnRES, line 14: There are hardcoded paths in this file, since this is for testing purpose, may be make it more configurable, instead of hardcoded values?
Also this is also why I believe this package should be in examples and not in packages directory...

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/SampleInteropConfig.kt
```diff
@@ -0,0 +1,119 @@
+package com.copmCorda
```
sandeepnRES, line 1: Why is this file is a sample?

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/interop/corda/RemoteCopmContractCordaFungible.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.hyperledger.cacti.plugin.copm.types.DLTransactionParams
+import org.hyperledger.cacti.plugin.copm.validators.ValidatedClaimPledgedAssetV1Request
+
+class RemoteCopmContractCordaFungible : RemoteCopmContract {
+    override fun matchesNetworkTypeAndAsset(networkType: String, asset: DLAsset): Boolean {
+        return networkType == "corda" && ! asset.isNFT
+    }
+
+    override fun getPledgeInfoCmd(claim: ValidatedClaimPledgedAssetV1Request): DLTransactionParams {
+        return DLTransactionParams("com.cordaSimpleApplication.flow",
```
sandeepnRES, line 14: hardcoded value here. I'll suggest take it from parameters, or configurations.

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/interop/corda/RemoteCopmContractCordaNFT.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.hyperledger.cacti.plugin.copm.validators.ValidatedClaimPledgedAssetV1Request
+
+class RemoteCopmContractCordaNFT : RemoteCopmContract {
+
+    override fun matchesNetworkTypeAndAsset(networkType: String, asset: DLAsset): Boolean {
+        return networkType == "corda" && asset.isNFT
+    }
+
+    override fun getPledgeInfoCmd(claim: ValidatedClaimPledgedAssetV1Request): DLTransactionParams {
+        return DLTransactionParams("com.cordaSimpleApplication.flow",
```
sandeepnRES, line 15: same here... and below line 16 too

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/interop/fabric/RemoteCopmContractFabricFungible.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    override fun matchesNetworkTypeAndAsset(networkType: String, asset: DLAsset): Boolean {
+        return networkType == "fabric" && ! asset.isNFT
+    }
+    override fun getPledgeInfoCmd(claim: ValidatedClaimPledgedAssetV1Request): DLTransactionParams {
+        val params  = listOf<Any>(claim.pledgeId,
+                                  claim.sourceCert,
+                                  claim.destinationAccount.organization,
+                                  claim.destCert)
+        return DLTransactionParams("simpleassettransfer","GetTokenAssetPledgeStatus", params )
```
sandeepnRES, line 18: Here too hardcoded...

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/interop/fabric/RemoteCopmContractFabricNFT.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.hyperledger.cacti.plugin.copm.validators.ValidatedClaimPledgedAssetV1Request
+
+class RemoteCopmContractFabricNFT : RemoteCopmContract {
+
+    override fun matchesNetworkTypeAndAsset(networkType: String, asset: DLAsset): Boolean {
+        return networkType == "fabric" && asset.isNFT
+    }
+
+    override fun getPledgeInfoCmd(claim: ValidatedClaimPledgedAssetV1Request): DLTransactionParams {
+        return DLTransactionParams("simpleassettransfer",
```
sandeepnRES, line 15: same here.

packages/cacti-plugin-copm-corda/Makefile
```diff
@@ -0,0 +1,39 @@
...
+	cp $(WORKSPACE)/weaver/samples/corda/corda-simple-application/contracts-kotlin/build/libs/contracts-kotlin-*.jar localdeps
+	cp $(WORKSPACE)/weaver/samples/corda/corda-simple-application/workflows-kotlin/build/libs/workflows-kotlin-*.jar localdeps
+
+server-docker-image:
+	DOCKER_BUILDKIT=1 docker build . --tag copm-corda 
+
+setup: build-local-deps server-docker-image
+
+
+pledge-network:
```
sandeepnRES, line 23: Would suggest moving these makefile targets, like starting network and other components to tests, and those should be ideally in examples directory

packages/cacti-copm-test/package.json
sandeepnRES: Oh okay, if you have discussed with @petermetz and he is fine with it then I'm also fine.

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/interop/corda/RemoteCopmContractCordaFungible.kt
sandeepnRES: Sorry I just saw that this is inside `sample` directory, so for every application someone needs to write code similar to these files?

sandeepnRES:
@jenniferlianne Thank you for the PR, overall looks good to me. Just some suggestions above I've shared.

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/interop/corda/RemoteCopmContractCordaFungible.kt
jenniferlianne: Correct.  This is left up to the app for now (as opposed to json files) as in larger applications configuration info could be stored in, for instance, a database to make it easier to administer.  

As I have mentioned elsewhere, I also think some of this application configuration could be pushed down to the contracts.  As requested by @VRamakrishna , have a working draft where asset type registration is done in the interop contract: https://github.com/jenniferlianne/cacti/tree/copm_register_assets_in_chaincode

packages/cacti-plugin-copm-corda/src/main/typescript/plugin-copm-corda.ts
```diff
@@ -0,0 +1,78 @@
...
+    out.push(crpcSvcRegistration);
+    return out;
+  }
+
+  public getInstanceId(): string {
+    return this.instanceId;
+  }
+
+  public getPackageName(): string {
+    return `@hyperledger-cacti/cacti-plugin-copm-corda`;
```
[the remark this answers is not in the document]
jenniferlianne: This follows the example of other cactus packages.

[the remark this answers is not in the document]
VRamakrishna: I'd prefer to get this from a configuration file or an environmental variable. Personally, I'd suggest that other Cactus packages do something similar too, @petermetz, rather than hardcode such values.

packages/cacti-copm-core/src/main/proto/generated/openapi/models/asset_lock_claim_v1_pb.proto
VRamakrishna: If there aren't too many such proto files, can we manually edit them to assign readable (i.e, falling in a sequence 1..2..3..) numbers? I was going to suggest this in the previous PR too, but thought it was better to merge it first as I didn't want to hold it up longer than it already had been.

jest.config.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
   moduleNameMapper: {
     "^(\\.\\.?\\/.+)\\.jsx?$": "$1",
   },
+  modulePathIgnorePatterns: ["./weaver/core/identity-management"],
```
VRamakrishna, line 9: This module does get exercised in Weaver transaction pipelines, though currently only the `iin-agent` subfolder (other subfolders are placeholders for mechanisms that are specified in the Weaver RFCs but haven't been implemented yet).

Is this module causing any issues that necessitates its exclusion?

packages/cacti-copm-core/Makefile
```diff
@@ -0,0 +1,194 @@
+WORKSPACE ?= $(shell realpath `pwd`/../../)
```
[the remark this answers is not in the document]
VRamakrishna: I'd prefer it be moved to cacti-copm-test then. Let's completely separate the testing setup and logic from the generic components.

Ideally, I'd separate the test setups from the testing logic too, so that the testing logic could get a "spec" for the test setup and run generic tests on the network it's been given, but I think the Cacti package structure doesn't make this easy.

Moving this logic to the cacti-copm-test package seems to be the low-hanging-fruit option now, so let's do that. Perhaps later, we can refactor this to move the testnet launch logic to `examples` (as @sandeepnRES suggests), thereby making cacti-copm-test also oblivious to the network structure.

packages/cacti-copm-core/Makefile
```diff
@@ -0,0 +1,194 @@
...
+fabric1-relay:
+	cd $(WORKSPACE)/weaver/core/relay && \
+	make start-server COMPOSE_ARG='--env-file docker/testnet-envs/.env.n1' 
+
+fabric2-relay:
+	cd $(WORKSPACE)/weaver/core/relay && \
+	make convert-compose-method2 && \
+	make start-server COMPOSE_ARG='--env-file docker/testnet-envs/.env.n2' 
+
+# IIN --------------------------------------------------------------------------------------
```
VRamakrishna, line 77: @sandeepnRES I think we don't need the `iin` agents for scenarios involving Corda, right? They won't do anything useful.

VRamakrishna: @jenniferlianne Is this used for Fabric-Fabric testing too, or just ones that involve Corda networks?

packages/cacti-copm-core/src/main/json/openapi.json
```diff
@@ -160,28 +165,16 @@
...
         "description": "Claim a previously locked asset on the same network.\n\"lock_id\" is required for fungible assets, \"asset\" is required for non-fungible assets.\nhttps://github.com/hyperledger-cacti/cacti/blob/main/weaver/rfcs/protocols/asset-exchange/generic-htlc.md",
-        "required": [
-          "lock_id",
-          "source",
-          "destination",
-          "asset",
-          "source_certificate",
-          "hash_info"
-        ],
+        "required": ["lock_id", "destination", "asset", "hash_info"],
```
VRamakrishna, line 168: Why remove `source` (or `source_certificate`)?

packages/cacti-copm-core/src/main/json/openapi.json
```diff
@@ -301,6 +286,24 @@
     }
   },
   "paths": {
+    "/api/v1/plugins/@hyperledger/cacti-plugin-copm/no-op": {
+      "post": {
+        "x-hyperledger-cacti": {
+          "http": {
+            "path": "/api/v1/plugins/@hyperledger/cacti-plugin-copm/no-op",
```
VRamakrishna, line 293: Change `@hyperledger` --> `@hyperledger-cacti` here and in other locations.

packages/cacti-copm-core/src/main/typescript/lib/types.ts
```diff
@@ -15,7 +16,8 @@ export type CopmContractNames = {
 };
 
 export type RemoteNetworkConfig = {
-  channelName: string;
+  channelName: string; // fabric-specific
+  chaincode: string; // fabric-specific
```
VRamakrishna, line 20: Please rename this to `chaincodeName`.

packages/cacti-copm-test/package.json
```diff
@@ -0,0 +1,74 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/hyperledger-cacti/cactus.git"
+  },
+  "license": "Apache-2.0",
+  "author": {
+    "name": "Hyperledger Cacti Contributors",
+    "email": "cactus@lists.hyperledger.org",
+    "url": "https://www.hyperledger.org/use/cacti"
```
VRamakrishna, line 24: This URL now automatically redirects to https://www.lfdecentralizedtrust.org/projects/cacti, so perhaps you can change it.

packages/cacti-copm-test/src/main/typescript/corda/copm-tester-corda.ts
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  private async createTestAssets(
+    account: DLAccount,
+  ): Promise<TestAssetsCordaCli> {
+    return new TestAssetsCordaCli(account, this.log);
+  }
+
+  public networkNames(): string[] {
+    return ["Corda_Network", "Corda_Network2"];
```
VRamakrishna, line 78: I wonder if this should be hardcoded or lifted from a configuration file. @sandeepnRES ?

VRamakrishna: I see the user identities and org names are also hardcoded below. I looked at the older code, where such values were hardcoded in the `test` folders of the `corda-interop-app` module, so if this package is the equivalent of that test suite, then I'm fine with the hardcoding.

packages/cacti-copm-test/src/main/typescript/lib/weaver-interop-configuration.ts
VRamakrishna: I see this in several places above (and probably below) too. Given that this is a test package, I suppose hardcoding these paths is OK for now. But I wonder if you can allow such values to be passed through constructors instead, with the hardcoded paths then being moved to the commands that start and run the tests. That will make the code base more maintainable.

jest.config.js
jenniferlianne: This file is specific to the jest testing system, and there are no jest tests in weaver.  The weaver convention of copying packages when building causes collisions in the haste map, so jest tests are unable to run.  In my branch where I am building the dependencies as part of running the test, I have had to make additional exceptions.

packages/cacti-copm-core/Makefile
jenniferlianne: This file is used for testing all combinations.  I will move it from cacti-copm-core to cacti-copm-test as it references the weaver samples.

packages/cacti-copm-core/src/main/proto/generated/openapi/models/asset_lock_claim_v1_pb.proto
jenniferlianne: This is a non-trivial request, as there are a number of files and the manual edits would need to be re-done when the openapi spec changes.  Please note that other protobuf files in the project generated with this toolset also have large numbers, for instance the directory here: 

https://github.com/jenniferlianne/cacti/blob/copm_register_assets_in_chaincode/packages/cactus-plugin-keychain-memory/src/main/proto/generated/openapi/models/has_keychain_entry_response_v1_pb.proto   

In some ways, generating a hash from the variable name is a better approach as it removes human error.

packages/cacti-copm-test/src/main/typescript/lib/weaver-interop-configuration.ts
sandeepnRES: okay yeah agree with Rama.

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/interop/corda/RemoteCopmContractCordaFungible.kt
sandeepnRES: okay makes sense.

packages/cacti-plugin-copm-corda/src/main/typescript/plugin-copm-corda.ts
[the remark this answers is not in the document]
sandeepnRES: Check if we can get it from `package.json`, I think that would be better.

packages/cacti-plugin-copm-corda/Makefile
```diff
@@ -0,0 +1,39 @@
...
+	cd $(WORKSPACE)/packages/cacti-copm-core && \
+	make corda-setup
+	mkdir -p localdeps && \
+	cp $(WORKSPACE)/packages/cacti-copm-core/src/main/protos-kt/build/libs/copm-corda-*jar localdeps 
+	cp $(WORKSPACE)/weaver/common/protos-java-kt/build/libs/protos-java-kt-*.jar localdeps
+	cp $(WORKSPACE)/weaver/sdks/corda/build/libs/weaver-sdk-corda-*.jar localdeps
+	cp $(WORKSPACE)/weaver/core/network/corda-interop-app/interop-workflows/build/libs/interop-workflows-*.jar localdeps
+	cp $(WORKSPACE)/weaver/core/network/corda-interop-app/interop-contracts/build/libs/interop-contracts-*.jar localdeps
+	cp $(WORKSPACE)/weaver/samples/corda/corda-simple-application/contracts-kotlin/build/libs/contracts-kotlin-*.jar localdeps
+	cp $(WORKSPACE)/weaver/samples/corda/corda-simple-application/workflows-kotlin/build/libs/workflows-kotlin-*.jar localdeps
```
VRamakrishna, line 15: I can understand the rest, but why do you need `weaver/samples/...` as dependencies?

packages/cacti-plugin-copm-corda/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "dependencies": {
+    "@bufbuild/buf": "1.30.0",
+    "@bufbuild/protobuf": "1.10.0",
+    "@connectrpc/connect": "1.4.0",
+    "@connectrpc/connect-node": "1.4.0",
+    "@connectrpc/connect-web": "1.4.0",
+    "@hyperledger-cacti/cacti-copm-core": "2.0.0",
+    "@hyperledger-cacti/cacti-copm-test": "2.0.0",
+    "@hyperledger/cacti-weaver-protos-js": "2.0.0",
+    "@hyperledger/cacti-weaver-sdk-fabric": "2.0.0",
```
VRamakrishna, line 34: `2.0.0` version doesn't exist for the two packages above. Because of the organization change, the last version tagged under `@hyperledger` was `[2.0.0-rc.7](https://github.com/orgs/hyperledger/packages/container/cacti-weaver-driver-fabric/283162801?tag=2.0.0-rc.7)`. Weaver publish actions failed during the `2.0.0` release because paths were not updated. These actions should succeed when the next release (`2.0.1`) is done, hopefully in the next few days. So you have two options: (1) wait for the next release and then update these version numbers here and elsewhere in your code, or (2) use the last release version `2.0.0-rc.7`.

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/SampleInteropConfig.kt
VRamakrishna: This is in the `sample` subfolder within the Corda plugin folder.

packages/cacti-plugin-copm-corda/package.json
jenniferlianne: These refer to the local monorepo versions, which is why they are able to compile.  The packages are built in the cactus way, with `npm run configure` at the top level.  This uses the tsconfig file at at the root level which builds the packages in a specific order and keeps a reference of local packages.  One of the challenges of this project was getting everything to build successfully given the differences in build philosophy between cactus and weaver.

packages/cacti-plugin-copm-corda/Makefile
jenniferlianne: The current corda interop chaincode require passing down the asset-specific CommandData instances for issue and burn, as well as getting the pledgeId.   This is something that could be avoided with asset type registration. 

Here are the asset-specific commands being passed to the chaincode in the corda CLI:  https://github.com/hyperledger-cacti/cacti/blob/27a24dd4579ef241df8d6bddf1cef30ee499b91d/weaver/samples/corda/corda-simple-application/clients/src/main/kotlin/com/cordaSimpleApplication/client/AssetTransferManager.kt#L116

packages/cacti-copm-core/src/main/json/openapi.json
jenniferlianne: I changed the lock claim for the fabric NFT case to use the lock id instead of the asset agreement to claim the lock, to be consistent with the Corda implementation.  After making this change you don't need information about source any more.

packages/cacti-copm-core/Makefile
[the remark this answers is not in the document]
jenniferlianne: I have moved all the Makefiles to cacti-copm-test.  The Makefile_corda and Makefile_fabric files have 3 key targes: setup,  lock-network, and pledge-network.  These can be referenced either directly:
 
`make -f Makefile_corda setup`

or via wildcard targets in the main makefile:

`make corda-setup`

Future network types can follow this pattern.

packages/cacti-copm-test/src/main/typescript/lib/weaver-interop-configuration.ts
jenniferlianne: This file is an example implementation of an abstract interface, and not meant to be re-used.  In addition to the hard-coded location of the weaver fabric-cli json settings (stored in a file), the class also contains hard-coded reference to the weaver contracts.

In general, I would advise against forcing use of a local json file for data storage.  While it's nice for getting an example system going, it scales poorly and precludes other options that would be easier for end-users to maintain.

packages/cacti-copm-core/src/main/json/openapi.json
jenniferlianne: Done

packages/cacti-copm-core/src/main/typescript/lib/types.ts
sandeepnRES: Probably `networkName` instead of `network` in the below line 21?

packages/cacti-copm-test/package.json
jenniferlianne: This has been updated.

packages/cacti-plugin-copm-corda/Makefile
jenniferlianne: Makefiles have been moved to cacti-copm-test and the cacti-plugin-copm-corda/src/test/kotlin directories.

packages/cacti-copm-core/src/main/typescript/lib/types.ts
```diff
@@ -1,7 +1,8 @@
 export type DLTransactionParams = {
   contract: string;
```
VRamakrishna, line 2: Sorry, I didn't notice this earlier. Can you rename this to `contractId`? That's just more meaningful.

packages/cacti-copm-test/package.json
```diff
@@ -0,0 +1,74 @@
...
+  "version": "2.0.0",
+  "description": "COPM testing.",
+  "keywords": [
+    "Hyperledger",
+    "Cacti",
+    "Integration",
+    "Blockchain",
+    "Distributed Ledger Technology"
+  ],
+  "homepage": "https://github.com/hyperledger-cacti/cactus#readme",
```
VRamakrishna, line 12: URL is incorrect. Should be https://github.com/hyperledger-cacti/cacti#readme

packages/cacti-copm-test/package.json
```diff
@@ -0,0 +1,74 @@
...
+  "bugs": {
+    "url": "https://github.com/hyperledger-cacti/cactus/issues"
```
VRamakrishna, line 14: `cactus` --> `cacti`

packages/cacti-copm-test/package.json
```diff
@@ -0,0 +1,74 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/hyperledger-cacti/cactus.git"
```
VRamakrishna, line 18: `cactus` --> `cacti`.

packages/cacti-plugin-copm-corda/Makefile
sandeepnRES: Corda interop app doesn't depend upon asset specific cordapp, it is designed to be general.
It just takes that command as a parameter, not needed to import or bundle the app cordapp with interop cordapp.

This makefile is moved to `src/test` folder, so I believe this is only for testing purpose and not a production based deployment? @jenniferlianne

packages/cacti-plugin-copm-corda/package.json
VRamakrishna: OK, I guess this works for a completely local build, i.e., using nothing outside the source code. It's not that the Weaver build build philosophy was different, but that we want the option to publish reusable assets (packages, containers) and then allow clients and third-party apps to depend on them rather than have to build everything locally. So we used actions to ensure that the flows worked both with locally-built code and published packages.

Anyway, this works for now. The Cacti maintainers can have a discussion later about building packages with external dependencies.

packages/cacti-copm-core/src/main/typescript/lib/types.ts
sandeepnRES: Hi @jenniferlianne I'm not sure about the purpose of this `type RemoteNetworkConfig`, but even the channelName and corda party endpoints could also be different for every remote call for the same network, as there can be multiple channels (as discussed on discord).

Its okay if you can just open an issue to track this.

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/SampleInteropConfig.kt
sandeepnRES: This needs to be reviewed whether a `SampleInteropConfig` is required as the code seems general. Please open an issue to track this?

packages/cacti-copm-test/src/main/typescript/lib/weaver-interop-configuration.ts
sandeepnRES: The problem with hardcoded paths like that it makes the code less manageable, for example if there is some code restructure, that path may change, and its not easy to scan files for such paths..

VRamakrishna: I'll echo Sandeep's concern about hardcoded paths. They have a habit of springing nasty surprises later and hence are hard to maintain. While json files are not a panacea either, updating their contents impose considerably less overhead than rebuilding source code (at least in my experience).

When code has to be used in production, we may have to make different choices (and there, I take your point about hardcoding some things that won't change unless a system upgrade is required), but this code is targeted primarily at developers, who need all the knobs that are useful to tweak.

Anyway, what you have is good for now, especially in test code. TODOs should work as reminders for others who will touch your code later. Thanks!

packages/cacti-copm-core/src/main/typescript/lib/types.ts
jenniferlianne: As discussed with @VRamakrishna 2 weeks ago, the naming, which is consistent with a similar structure in the fabric implementation submitted 3 months ago, reflects the data structures in the weaver test network json configuration files. While named 'network', they seem to talk about organizations, not networks.   While I may be wrong, I expect this information will not change on the fly for a given organization.  Can rename to RemoteOrganizationConfig.

packages/cacti-copm-test/src/main/typescript/lib/weaver-interop-configuration.ts
jenniferlianne: To reiterate: this is a test implementation. If the paths change the tests will fail in a very obvious way.  

Providing an abstract interface for the info was intentional -- it allows the production customer to make a local json implementation if they want, or an implementation that fetches the data with a REST call to a configuration system.

packages/cacti-plugin-copm-corda/sample/src/main/kotlin/com/copmCorda/SampleInteropConfig.kt
jenniferlianne: The application-specific code is in the companion object.  As with [above](https://github.com/hyperledger-cacti/cacti/pull/3625#discussion_r1860754213), this is a test implementation for running on the test docker image which has this data in the envrionment.  

It is deliberately set up as an abstract interface so that a production customer can do what they want -- with the info coming from json files or a future configuration system.

packages/cacti-copm-core/src/main/typescript/lib/types.ts
sandeepnRES: Lets it be as it is for now, just open an issue to track this, so that we can think about it more later, if we actually need restructuring.

jenniferlianne: Too late - I updated the naming to be RemoteOrganization, as well as using 'organization' instead of 'network' for AssetAccounts in the OpenApi spec.  The code is more consistent now.  I opened an issue here for revisiting in the future: https://github.com/hyperledger-cacti/cacti/issues/3656

packages/cacti-copm-core/src/main/typescript/lib/types.ts
jenniferlianne: This has been renamed.

packages/cacti-copm-test/package.json
jenniferlianne: Corrected.

packages/cacti-copm-test/package.json
jenniferlianne: Corrected.

jenniferlianne:
> @jenniferlianne LGTM now. Please resolve the merge conflicts.

Done.

petermetz:
@jenniferlianne FYI the bundle name validation CI check is failing: https://github.com/hyperledger-cacti/cacti/actions/runs/12129347453/job/33817605282?pr=3625

sandeepnRES (approved):

VRamakrishna:
> @jenniferlianne FYI the bundle name validation CI check is failing: https://github.com/hyperledger-cacti/cacti/actions/runs/12129347453/job/33817605282?pr=3625

@jenniferlianne This is the offending change: https://github.com/hyperledger-cacti/cacti/pull/3625/files#diff-de7fa4d1015b8595a94d0ba37b1fa72ddea3875aab9ce169b08d386691fa547c. Every `package.json` in a package requires a `scripts` section, otherwise the bundle validation action will crash.

I think you can just leave an empty `scripts` to pass the check, @petermetz ?

petermetz:
> > @jenniferlianne FYI the bundle name validation CI check is failing: https://github.com/hyperledger-cacti/cacti/actions/runs/12129347453/job/33817605282?pr=3625
> 
> @jenniferlianne This is the offending change: https://github.com/hyperledger-cacti/cacti/pull/3625/files#diff-de7fa4d1015b8595a94d0ba37b1fa72ddea3875aab9ce169b08d386691fa547c. Every `package.json` in a package requires a `scripts` section, otherwise the bundle validation action will crash.
> 
> I think you can just leave an empty `scripts` to pass the check, @petermetz ? But Jennifer, don't you need a build action there at least? I see there's a `yarn` command mentioned in the package's README. Perhaps that could be added to `scripts:build`?

@VRamakrishna There are additional requirements that could be the issue, but I haven't checked which ones are being violated exactly. It appears to be passing now though so I'm good.

VRamakrishna:
@petermetz Can you approve the PR and merge if you are happy with it? We just have approvals from one organization currently.

petermetz:
> @petermetz Can you approve the PR and merge if you are happy with it? We just have approvals from one organization currently.

@VRamakrishna Weird, I think my approval from 3 weeks [1] ago got dismissed by somebody. Approving it again now.

image

petermetz (approved):

PR #3641, opened 2024-11-15
feat(ccmodel-hepheastus): add Hepheastus plugin

RafaelAPB:
Implementation  of the paper Hepheastus https://ieeexplore.ieee.org/document/10363680

Authored-by: Bruno Mateus <brumat315@gmail.com>
Co-authored-by: Rafael Belchior <rafael.belchior@tecnico.ulisboa.pt>

fix(ccmodel-hepheastus): incorporate PR feedback

Authored-by: Bruno Mateus <brumat315@gmail.com>

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/pm4py-adapter/check_conformance.py
```diff
@@ -0,0 +1,226 @@
+import sys
```
petermetz, line 1: @RafaelAPB I recommend storing Python source code under `./src/main/python` instead of `./src/main/typescript/`

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
```diff
@@ -0,0 +1,911 @@
...
+
+export class CcModelHephaestus implements ICactusPlugin, IPluginWebService {
+  private readonly log: Logger;
+  private readonly instanceId: string;
+  private endpoints: IWebServiceEndpoint[] | undefined;
+  private httpServer: Server | SecureServer | null = null;
+  private crossChainLog: CrossChainEventLog;
+  private unmodeledEventLog: CrossChainEventLog;
+  private nonConformedCrossChainLog: CrossChainEventLog;
+  private nonConformedCCTxs: string[];
```
petermetz, line 83: @RafaelAPB If these are not getting re-assigned anywhere outside of the constructor then you can also mark them `readonly` which doesn't change anything at runtime but it's a good practice to have as much of the state of your class being readonly as possible which helps lower bugs in this regard.

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
```diff
@@ -0,0 +1,911 @@
...
+    if (serverMaybe.isPresent()) {
+      this.log.info(`Awaiting server.close() ...`);
+      const server = serverMaybe.get();
+      await promisify(server.close.bind(server))();
+      this.log.info(`server.close() OK`);
+    } else {
+      this.log.info(`No HTTP server found, skipping...`);
+    }
+  }
+
```
petermetz, line 202: @RafaelAPB This is no longer a needed method of the plugin interfaces and you can also delete the `getHttpServer` method and the `httpServer` field itself. Makes no difference but I figured it helps to tidy up a little.

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
```diff
@@ -0,0 +1,911 @@
...
+      from: data.response.transactionReceipt.from,
+      invocationType: data.request.invocationType,
+      methodName: data.request.methodName,
+      parameters: data.request.params,
+      gasUsed: data.response.transactionReceipt.gasUsed,
+      gasPrice: data.request.gasPrice as number,
+    };
+  }
+
+  private pollTxReceiptsBesu(
```
petermetz, line 254: @RafaelAPB Does this do any actual polling? It looks like it just massages the  data. If I'm understanding it correctly then I'd recommend renaming it to something expressing that instead of polling for tx receipts.

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
```diff
@@ -0,0 +1,911 @@
...
+      from: data.response.transactionReceipt.from,
+      invocationType: data.request.invocationType,
+      methodName: data.request.methodName,
+      parameters: data.request.params,
+      gasUsed: data.response.transactionReceipt.gasUsed,
+      effectiveGasPrice: data.response.transactionReceipt.effectiveGasPrice,
+    };
+  }
+
+  private pollTxReceiptsEth(
```
petermetz, line 285: @RafaelAPB Same question+conditional suggestion as above

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
```diff
@@ -0,0 +1,911 @@
...
+  }
+
+  public async persistCrossChainLogCsv(name?: string): Promise<string> {
+    const startTime = new Date();
+    const columns =
+      this.crossChainLog.getCrossChainLogAttributes() as (keyof CrossChainEvent)[];
+    const logName = name
+      ? `${name}.csv`
+      : `hephaestus_log_${startTime.getTime()}.csv`;
+    const csvFolder = path.join(__dirname, "../", "../", "test", "csv");
```
petermetz, line 723: @RafaelAPB Depending on this will break the code in production when it is under a different sub-directory path (`./dist/lib/.../`) I recommend making it configurable through constructor parameters so that this issue can be avoided

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
```diff
@@ -0,0 +1,911 @@
...
+      throw new RuntimeError(errorMessage, error);
+    }
+  }
+
+  public async persistCrossChainLogJson(name?: string): Promise<string> {
+    const startTime = new Date();
+    const logName = name
+      ? `${name}.json`
+      : `hephaestus_log_${startTime.getTime()}.json`;
+    const jsonFolder = path.join(__dirname, "../", "../", "test", "json");
```
petermetz, line 767: @RafaelAPB Depending on this will break the code in production when it is under a different sub-directory path (./dist/lib/.../) I recommend making it configurable through constructor parameters so that this issue can be avoided

petermetz (approved):
@RafaelAPB I left some suggestions in the code, but it's fine by me if you'd prefer to address those later or never. I want to make sure I'm not slowing down the development.

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
brunoffmateus: okay, then the 3 Observables and `sourceLedger` and `targetLedger` should also be `readonly`, right?

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
brunoffmateus: if i remove the `shutdown()` method i get an error saying `Property 'shutdown' is missing in type 'CcModelHephaestus' but required in type 'IPluginWebService'`, should i keep it with just the log message then?

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/pm4py-adapter/check_conformance.py
RafaelAPB: @brunoffmateus please move the python code to this other folder please

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
RafaelAPB: @brunoffmateus can you please address the path issue

brunoffmateus: @RafaelAPB I forgot to mention, but I added the parameter `ccLogsDir: string` that is set in the constructor. When not provided in the plugin options a default path is used.

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/pm4py-adapter/check_conformance.py
brunoffmateus: @RafaelAPB I forgot to mention that this was done as well.

brunoffmateus:
@petermetz @RafaelAPB I also added a quick message about the python dependency in the prerequisites of `README.md`. If you don't find it necessary I can remove it.
Otherwise I think everything is done in this commit: 89bf3affb6387a8ae1f61d89ccc2cb8d1935acdd. You should be able to cherry-pick and squash it.

petermetz:
@brunoffmateus Dependencies should always be documented as explicitly as possible (better yet via automation that fails with a helpful error message). So please do not remove it. It's great that you did what you did. :-)

petermetz:
@brunoffmateus LGTM but please fix the linter

RafaelAPB:
@petermetz i believe your concerns are addressed. Can we use the squash and merge functionality (there are 2 PRs), or should they be squashed?

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
petermetz: @brunoffmateus Great, thank you!

petermetz:
> @petermetz i believe your concerns are addressed. Can we use the squash and merge functionality (there are 2 PRs), or should they be squashed?

@RafaelAPB @brunoffmateus Yes, they were, thank you! I'd still recommend using the "manual" method of rebasing onto main interactively and squash them together that way just to be sure. The github GUI can have negative & unintended side effects sometimes.

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
```diff
@@ -0,0 +1,844 @@
...
+      name: "HEPHAESTUS_EVENT_LOGS",
+    });
+    this.unmodeledEventLog = new CrossChainEventLog({
+      name: "HEPHAESTUS_UNMODELED_LOGS",
+    });
+    this.nonConformedCrossChainLog = new CrossChainEventLog({
+      name: "HEPHAESTUS_NON_CONFORMANCE_LOGS",
+    });
+
+    this.caseID = "UNDEFINED_CASE_ID";
```
outSH, line 113: Is it safe to use no-unique string as a default caseID? Wouldn't it collide if we don't set the caseID ourselves? Maybe we could set this as some uuid by default?

packages/cactus-plugin-ccmodel-hephaestus/package.json
```diff
@@ -0,0 +1,99 @@
...
+    "Blockchain",
+    "Distributed Ledger Technology"
+  ],
+  "homepage": "https://github.com/hyperledger/cactus#readme",
+  "bugs": {
+    "url": "https://github.com/hyperledger/cactus/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/hyperledger/cactus.git"
```
outSH, line 18: This should point to the current URL in separate org (https://github.com/hyperledger-cacti/cacti/) - applies to all URLs in here.

I think we will have to fix these paths in other packages as well in one separate PR, so this is not critical, nevertheless it's better to change it now in case that other PR is merged before this.

packages/cactus-plugin-ccmodel-hephaestus/package.json
```diff
@@ -0,0 +1,99 @@
...
+    "body-parser": "1.20.3",
+    "fabric-network": "2.2.20",
+    "socket.io": "4.6.2",
+    "uuid": "10.0.0",
+    "web3": "1.6.1",
+    "web3-core": "1.6.1"
+  },
+  "engines": {
+    "node": ">=10",
+    "npm": ">=6"
```
outSH, line 92: Cacti in general requires `node >= 18`, is there a reason why this package loosens that restriction? I'd prefer to have it set to common value because of the script in tools that does it automatically for all package.json files in the repository.

packages/cactus-plugin-ccmodel-hephaestus/src/test/typescript/fabric-contracts/lock-asset/chaincode-typescript/package.json
```diff
@@ -0,0 +1,62 @@
+{
+    "name": "asset-transfer-basic",
```
outSH, line 2: I think this is copied from https://github.com/hyperledger/fabric-samples/, could you please add a comment in each copied file with a link to the original sources (and possibly a date when we copied it)?

outSH (changes requested):
Looks awesome, thank you for this contribution! I've left a few comments / questions :)

packages/cactus-plugin-ccmodel-hephaestus/package.json
outSH: @petermetz Is there already a task for fixing url's in cacti packages or can I create one?

packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
RafaelAPB: Agreed

packages/cactus-plugin-ccmodel-hephaestus/package.json
RafaelAPB: Updated

packages/cactus-plugin-ccmodel-hephaestus/src/test/typescript/fabric-contracts/lock-asset/chaincode-typescript/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "version": "1.0.0",
+    "description": "Asset Transfer Basic contract implemented in TypeScript",
+    "main": "dist/index.js",
+    "typings": "dist/index.d.ts",
+    "engines": {
+        "node": ">=12",
+        "npm": ">=5"
```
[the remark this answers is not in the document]
RafaelAPB: This is not our code, so I would not change the dependencies

outSH (approved):

RafaelAPB:
@outSH and @petermetz thanks a lot for the reviews.

petermetz:
> @outSH and @petermetz thanks a lot for the reviews.

@RafaelAPB My pleasure! :-)

petermetz:
@RafaelAPB Took a quick look and found a few low hanging  fruit fixes (easy to spot bugs that were breaking the build). I'm fixing them but I'll also post some comments here for awareness:

The root tsconfig.json file that ties all the packages together for the typescript compiler was missing the hephaestus plugin's reference so the plugin code was not part of the compilation when you did a `yarn tsc` from the project root.

The indirect way this manifested as an issue was that the build was seemingly good (it compiled successfully) but despite that when you tried to run one of the testcases of the package, Jest would crash with compilation errors within the test case itself.

Adding this snippet below to the root tsconfig.json file fixed the root cause but then it highlighted the other issues that were otherwise hidden until now, more on those later.

```json
 {
    "path": "./packages/cactus-plugin-ccmodel-hephaestus/tsconfig.json"
  },
```

petermetz:
@RafaelAPB Afterwards, I moved my attention to the compilation issues (see screenshots for full picture). 
1. It seems like the `IRunTransactionV1Exchange` type is inconsistently named, because in the besu connector it is called `IRunTransactionV1Exchange` but in the ethereum connector it is missing the `I` (probably my fault for missing it in one of my reviews that the `I` was missing, sorry about that)
2. The `IRunTxReqWithTxId` from the fabric connector was being imported as `RunTxReqWithTxId`, again I remember asking for changes on the interfaces so that they are prefixed with `I` and this is most likely what caused the issues down the line in the dependent PRs so sorry about the extra overhead here as well.
3. The third issue was a more boring one, the versions of the package were set to `2.0.0` but since we've released `2.1.0` and so this again caused a bunch of trouble with the build, but that's also fixed now.
4. So after applying all of those fixes and me trying to run one of the test cases ( `yarn test:jest:all 'packages/cactus-plugin-ccmodel-hephaestus/src/test/typescript/integration/besu-ethereum-asset-transfer.test.ts'`) it still fails unfortunately with some python dependency issues which are either because I haven't read the documentation on how to set them up or because they are missing in some other way. The logs for that crash are below just in case.
5. I know that these python dependencies are required right now with good reason, but for the longer term, if possible, it would be nice to keep it NodeJS only just so that it's simpler to set up the development environment (though I have given up on trying to achieve this goal, it shouldn't stop from us for at least trying to reach it and pick the low hanging fruit wherever possible). So again, not a change request, just a note for ourselves for an ideal future when we have infinite time and resources to tidy everything up... :-)

```sh
 Command failed: python3 /home/peter/a/cacti-upstream/packages/cactus-plugin-ccmodel-hephaestus/src/main/python/create_model.py /home/peter/a/cacti-upstream/packages/cactus-plugin-ccmodel-hephaestus/src/test/ccLogs/json/hephaestus_log_1734293360776.json
    Traceback (most recent call last):
      File "/home/peter/a/cacti-upstream/packages/cactus-plugin-ccmodel-hephaestus/src/main/python/create_model.py", line 6, in <module>
        import pm4py
    ModuleNotFoundError: No module named 'pm4py'

       9 |   try {
      10 |     const startTime = new Date();
    > 11 |     const serializedCCModel = execSync(command).toString("utf-8");
         |                                       ^
      12 |     const finalTime = new Date();
      13 |     console.log(
      14 |       `CREATE-MODEL-PM4PY:${finalTime.getTime() - startTime.getTime()}`,
```

Screenshot from 2024-12-15 12-08-00
Screenshot from 2024-12-15 12-05-59
Screenshot from 2024-12-15 12-04-54

petermetz:
@RafaelAPB Finally, I'm going to force push now onto your branch with the fixes mentioned above which I'm hoping won't be too confusing and are steps in the right direction. If the CI has the correct setup steps for the python dependencies then it should actually be able to run the tests and have them succeed and if that is the case then I'm good to merge this right after that. If the CI ends up having issues with the tests too then I'll ask that you tweak the CI job to have the python dependencies set up as needed just so that we can see for now at least on the CI passing the tests and I'll make some effort in the future to set up the python dependencies on my dev machine(s) as well.

RafaelAPB:
@petermetz, I appreciate the help and your time fixing the issues. I take all responsibility for overlooking those issues. 

Regarding the Python dependencies, that's a good point. This does not work without that specific dependency, but there is no JS/TS alternative (to the best of my knowledge). I propose that future developments for this plugin can  migrate to a JS solution, or we remove it if there is no broader interest in it. 
I will check the CI.

RafaelAPB:
@petermetz seems we do not have the CI configured for this specific plugin. I'd say I add to the documentation the need for this dependency and (optionally) add the CI configuration for running this package tests (however I'd rather not before it is updated/worked upon). What is your opinion?

RafaelAPB:
@petermetz I've noticed you already added it. Thanks!

petermetz:
> seems we do not have the CI configured for this specific plugin

@RafaelAPB Oh, /facepalm, sorry I didn't even look. We should definitely add a new issue to track adding those CI steps.

> I'd say I add to the documentation the need for this dependency and (optionally) add the CI configuration for running this package tests (however I'd rather not before it is updated/worked upon). What is your opinion?

@RafaelAPB I'm OK with that schedule, as long as we track the steps to be done via issues (which guarantees that we don't forget/lose track/falls through the cracks/etc.)

petermetz:
> @petermetz, I appreciate the help and your time fixing the issues. I take all responsibility for overlooking those issues.
> 
> Regarding the Python dependencies, that's a good point. This does not work without that specific dependency, but there is no JS/TS alternative (to the best of my knowledge). I propose that future developments for this plugin can migrate to a JS solution, or we remove it if there is no broader interest in it. I will check the CI.

@RafaelAPB I would recommend two ideas for consideration (no idea which one is better or if any of them would work well in practice for now)
1. Use Docker at runtime to run a container containing the python deps for process mining. This might feel like a "from the pan to the fire" type of solution but I think depending on Docker is better than depending on Python because through Docker we get at least the stable environment setup question solved (there won't be any python version issues, we don't have to teach people how to install and configure multiple python environments, solve those dreaded module version constraint hells, etc). The other upside would be that if we depend on Docker, later we can use that as a stepping stone to depend on other non-NodeJS modules as well (Rust/Go for example) and we wouldn't have to worry about environment setup either also due to containerization. 
2. Second idea: Try to use something like this which appears to be a direct javascript port of the python library we are using currently (which implies nothing about whether it's maintained as well, bugs, etc.) https://github.com/pm4js/pm4js-core

RafaelAPB:
Thanks for your suggestions @petermetz. We will commit to implement 1. (or to remove the package altogether because as is it breaks at least one of the principles of the project)

PR #3645, opened 2024-11-19
ci(github): add dependency version consistency check

### LINKED ISSUE #3612: ci(github): add dependency version consistency custom-check
### Problem Statement

In our monorepository code reviewers have to manually keep track of contributors
using the correct version string for packages and their dependencies.

#### Types of issues that arise (not a complete list)
1. We are on 2.0.0 and people send in a PR with a package with version 0.0.1 or 2.0.0-rc.1 or something else that's outdated.
2. We are on 2.0.0 and people send in a PR where some of their local package dependencies use outdated versions of other packages. For example the `core` package declares that it depends on the `common@1.0.0` package or for another exmaple, the `core-api` package declares that it depends on the common@2.0.0-rc.1 package. Both of these cases are wrong becaus they should only be using the current latest relesae version which in this case is `2.0.0`
3. Third party dependencies are being inconsistently dependend on. For examples 5 different packages end up using 5 different versions of `express` or `@grpc-js/grpc` which causes compilation issues across the board.
4. Sometimes people make the effort to be consistent but then versions change on main and they don't realize that their pull request now needs updating just because of this. The issue is that git won't mark this as a merge conflict if it can auto-resolve the merge by using the wrong version so we need tooling to makes this type of mistake impossible to help both code reviewers and contributors alike (less manual effort for everyone)

### Acceptance Criteria

1. Use the software provided at
https://github.com/bmish/check-dependency-version-consistency
2. Ensure that there is a custom-check step implemented which uses that package.
3. If you run `yarn custom-check` and any of the problems are present from the `Types of issues that arise (not a complete list)` section then it needs to fail.
4. It needs to report failures using the existing conventions for all the other custom-checks we have in place.
5. There has to be a separate npm script in the root package.json which allows contributors to run just this one check by itself as well. Call it `tools:check-dependency-version-consistency`

jagpreetsinghsasan:
## Commit to be reviewed

ci(github): add dependency version consistency check

    Primary Changes
    ---------------
    1. Added a new custom check which checks if there are any inconsistency with the dependency version, which is already being used with an exsisting package.

    Changes needed to incorporate 1)
    2. Added a new dependency in the root package.json
    3. Added a new npm script in the root package.json
    4. Updated the lock file for the dependency inclusion

Fixes #3612

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


petermetz (changes requested):
> @petermetz shall I create a new task to update the existing inconsistencies in the dependency versions in our current codebase? I tried to resolve them as well in this PR, but it seems that for some version mismatches, it might not work with just the version updation (for example, web3 related dependencies need to be updated from 1.x -> 4.x)

@jagpreetsinghsasan Yes please, separate tasks for the easy low hanging fruit and then the difficult stuff (web3 1.x vs 4.x) so that we can chip away at the task in smaller pieces of work.

In the meantime please also add changes to this PR that will exclude the difficult to update dependencies from being checked and then add a comment to those exclusion declarations referencing the issues you've created for their fixing. This way we can kick off this check in the codebase with it passing while we work our way through the backlog of issues that would otherwise make it fail on the CI for everyone on every pull request.

LGTM otherwise in general.

jagpreetsinghsasan:
> > @petermetz shall I create a new task to update the existing inconsistencies in the dependency versions in our current codebase? I tried to resolve them as well in this PR, but it seems that for some version mismatches, it might not work with just the version updation (for example, web3 related dependencies need to be updated from 1.x -> 4.x)
> 
> @jagpreetsinghsasan Yes please, separate tasks for the easy low hanging fruit and then the difficult stuff (web3 1.x vs 4.x) so that we can chip away at the task in smaller pieces of work.
> 
> In the meantime please also add changes to this PR that will exclude the difficult to update dependencies from being checked and then add a comment to those exclusion declarations referencing the issues you've created for their fixing. This way we can kick off this check in the codebase with it passing while we work our way through the backlog of issues that would otherwise make it fail on the CI for everyone on every pull request.
> 
> LGTM otherwise in general.

Sure, I will fix the inconsistencies with the easier ones in this PR itself and reference the tougher ones via comments here to their respective tasks.

tools/custom-checks/check-dependency-version-consistency.ts
```diff
@@ -0,0 +1,15 @@
...
+  const errors: string[] = [];
+  const cdvc = new CDVC(process.cwd(), {
+    fix: false,
+  });
+  errors.push(cdvc.toMismatchSummary());
+  return [errors.length === 0, errors];
+}
+
+const dependencyMismatchSummary = await checkDependencyVersionConsistency();
+console.log(dependencyMismatchSummary[1][0]);
```
petermetz, line 15: @jagpreetsinghsasan This code will run every time the file gets imported. Please use the pattern established in the other similar scripts instead where we have a main function for logic like this and the main function gets invoked on condition of the script not having been imported by another script but instead having been ran from a shell.

tools/custom-checks/check-dependency-version-consistency.ts
```diff
@@ -0,0 +1,15 @@
+import { CDVC } from "check-dependency-version-consistency";
+
+export async function checkDependencyVersionConsistency(): Promise<
+  [boolean, string[]]
+> {
+  const errors: string[] = [];
+  const cdvc = new CDVC(process.cwd(), {
+    fix: false,
+  });
+  errors.push(cdvc.toMismatchSummary());
```
petermetz, line 10: @jagpreetsinghsasan If there is a way to extract the list of issues one by one that would work best (but I don't know if the library itself supports this or not).
If we can only get the summary then it's fine by me like this, BUT if we can get an array or other collection of individual issues instead of just a summary, then I'd loop through those and add a separate error for each. 
This will (hopefully) aid us in interpreting the results more easily later on and it could also allow us the chance to add extra details for each error (I don't know what the output of the summary looks like, I'm just speculating that it may not contain all the relevant details needed to fix the problems such as the exact location of the package.json file where the problem dependency is declared, or the version it is declared with ,etc.)

petermetz (changes requested):
@jagpreetsinghsasan Sorry, I've noticed a couple more little things, but still looking great in general!

tools/custom-checks/check-dependency-version-consistency.ts
jagpreetsinghsasan: Yes, you are right @petermetz. I straightaway pushed the summary because of the nature of the result it provides.
The library outputs this as a single, highly formatted string with tables and color encoding.
image

jagpreetsinghsasan:
@hyperledger-cacti/cacti-maintainers I am currently pushing some dependency version upgrades (in stages) alongside adding of the logic. So please do not review this PR until I reply back here asking for a review (The idea is, as what @petermetz suggested, to upgrade the easier dependencies alongside this PR itself and have a new issue fixing the harder ones).

tools/custom-checks/check-dependency-version-consistency.ts
petermetz: @jagpreetsinghsasan OK, thank you for confirming, in that case, I'm OK with just leaving it as is.

tools/custom-checks/check-dependency-version-consistency.ts
petermetz: @jagpreetsinghsasan Thank you!

jagpreetsinghsasan:
I am closing this PR.
Will create another one with relevant code changes.

PR #3646, opened 2024-11-19
feat: add knex database integration

Yogesh01000100:
## Add Knex Database Integration

- added **knex database integration** for production and test environments.


packages/cactus-plugin-satp-hermes/src/knex/knexfile.ts
```diff
@@ -1,16 +1,34 @@
 import path from "path";
 import { v4 as uuidv4 } from "uuid";
+import dotenv from "dotenv";
+
+dotenv.config({ path: path.resolve(__dirname, "../../.env") });
```
petermetz, line 5: @Yogesh01000100 @RafaelAPB I recommend making this path configurable because once you compile the TS sources the depth of the directory structure will be different (plus there's a million other scenarios where users would want to customize where their .env file is located)

petermetz (commented):
@Yogesh01000100 Just a small suggestion from me, please see above

packages/cactus-plugin-satp-hermes/src/knex/knexfile.ts
RafaelAPB: @petermetz thank you for your review. @Yogesh01000100 please address @petermetz 's suggestion. Rest LGTM

RafaelAPB (changes requested):
LGTM please review Peter's suggestion

PR #3649, opened 2024-11-21
ci(github): fix type exports in packages/cactus-common

ruzell22:
### Commit to be reviewed
---
ci(github): fix type exports in packages/cactus-common
```
Primary Changes
---------------
1. Removed packages/cactus-common/hyperledger-cactus-common-*.tgz in ignore 
paths in get-all-tgz-path.ts file
2. Added the log-level.ts for the missing exports detected by attw -f json
3. Added exports in public-api.ts from log-level.ts
```

Fixes: #3635

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-common/src/main/typescript/log-level.ts
+++ b/packages/cactus-common/src/main/typescript/log-level.ts
@@ -1,0 +1,18 @@
+export const LogLevel = {
+  TRACE: 0,
+  DEBUG: 1,
+  INFO: 2,
+  WARN: 3,
+  ERROR: 4,
+  SILENT: 5,
+} as const;
+export type LogLevelNumbers = (typeof LogLevel)[keyof typeof LogLevel];
+export type LogLevelDesc =
+  | LogLevelNumbers
+  | "trace"
+  | "debug"
+  | "info"
+  | "warn"
+  | "error"
+  | "silent"
+  | keyof typeof LogLevel;
--- a/packages/cactus-common/src/main/typescript/public-api.ts
+++ b/packages/cactus-common/src/main/typescript/public-api.ts
@@ -1,6 +1,7 @@
 export { LoggerProvider } from "./logging/logger-provider";
 export { Logger, ILoggerOptions } from "./logging/logger";
-export { LogLevelDesc } from "loglevel";
+export { LogLevel, LogLevelNumbers, LogLevelDesc } from "./log-level";
+
 export { Objects } from "./objects";
 export { Strings } from "./strings";
 export { Bools } from "./bools";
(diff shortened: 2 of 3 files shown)


ruzell22:
Before fix:
image
adding `-f json` to attw command, it shows which file and which problem with export it is having.

After fix:
image
`yarn run custom-checks` and attw with and without `-f json` flag are all passing

petermetz (changes requested):
@ruzell22 Thank you! For your changes to be version controlled you cannot change index.js, you must change index.ts (which is the one that gets compiled to index.js)

outSH (approved):
LGTM, I've re-run failed yarn codegen check to see if it's caused by the PR or not

packages/cactus-common/src/main/typescript/index.ts
```diff
@@ -1,2 +1,17 @@
...
+  WARN: "WARN",
+  warn: "warn",
+  INFO: "INFO",
+  info: "info",
+  DEBUG: "DEBUG",
+  debug: "debug",
+  SILENT: "SILENT",
+  silent: "silent",
+} as const;
+export type LogLevelDesc = (typeof LogLevel)[keyof typeof LogLevel];
```
petermetz, line 17: @ruzell22 
1. Please do not write code directly in index.ts, it is meant to collect the public API surface exports and that will very noisy if you start writing code directly in it.
2. Could you make the LogLevelDesc type a match to what is defined in `packages/cactus-common/node_modules/loglevel/index.d.ts` ? There's nothing wrong by the looks of it with what you have here, but it is the safest if we use the exact same definition that the library uses internally to reduce the risk of type compatibility issues later on.

```typescript
   /**
     * Log levels
     */
    interface LogLevel {
        TRACE: 0;
        DEBUG: 1;
        INFO: 2;
        WARN: 3;
        ERROR: 4;
        SILENT: 5;
    }

    /**
     * Possible log level numbers.
     */
    type LogLevelNumbers = LogLevel[keyof LogLevel];

    /**
     * Possible log level descriptors, may be string, lower or upper case, or number.
     */
    type LogLevelDesc = LogLevelNumbers
        | 'trace'
        | 'debug'
        | 'info'
        | 'warn'
        | 'error'
        | 'silent'
        | keyof LogLevel;
```

petermetz (changes requested):

petermetz (approved):
@ruzell22 LGTM, thank you for the updates!

packages/cactus-common/src/main/typescript/index.ts
petermetz: @ruzell22 It checks out, thank you.

PR #3654, opened 2024-11-26
ci(github): fix type exports in packages/cactus-verifier-client

ruzell22:
### Commit to be reviewed
---
ci(github): fix type exports in packages/cactus-verifier-client
```
Primary Changes
---------------
1. Removed packages/cactus-verifier-client/hyperledger-cactus-verifier-
client-*.tgz in ignore paths in get-all-tgz-path.ts file
2. Removed the export of IVerifierEventListener and LedgerEvent from
./verifier and changed it to source from @hyperledger/cactus-core-api to
resolve the type export issue encountered in this package
```

Fixes: #3633

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-verifier-client/src/main/typescript/public-api.ts
+++ b/packages/cactus-verifier-client/src/main/typescript/public-api.ts
@@ -1,4 +1,4 @@
-export { Verifier, IVerifierEventListener, LedgerEvent } from "./verifier";
+export { Verifier } from "./verifier";
 export { getValidatorApiClient } from "./get-validator-api-client";
 export {
   VerifierFactory,
--- a/packages/cactus-verifier-client/src/main/typescript/verifier.ts
+++ b/packages/cactus-verifier-client/src/main/typescript/verifier.ts
@@ -18,11 +18,6 @@ import {
   IVerifierEventListener,
 } from "@hyperledger/cactus-core-api";
 
-export {
-  IVerifierEventListener,
-  LedgerEvent,
-} from "@hyperledger/cactus-core-api";
-
 /**
  * Utility type for retrieving monitoring event / new block type from generic ISocketApiClient interface.
  */
--- a/packages/cactus-verifier-client/src/test/typescript/unit/verifier.test.ts
+++ b/packages/cactus-verifier-client/src/test/typescript/unit/verifier.test.ts
@@ -18,12 +18,12 @@ const log: Logger = LoggerProvider.getOrCreate({
   level: testLogLevel,
 });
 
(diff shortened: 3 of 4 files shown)


petermetz (changes requested):
@ruzell22 Unfortunately if we remove the `VerifierEventListener,
  LedgerEvent,` types from the public API of the verifier-client package then that would be a breaking change and that can only happen when we release v3.0.0.

Before we continue the conversation, could you please double check if there are any other ways to fix the type exports that doesn't involve us not re-exporting the mentioned types from the core-api package?

It's not ideal but we must abide by the conventions to make sure people can safely upgrade minor versions such as from 2.0.0 to 2.0.1 or 2.1.0 when the time comes.

If you can't find any other workaround to make the type exports look good and also doesn't involve breaking changes then just respond back with a list of things/ideas/approaches that you have tried and I'll take a look at it as well. We can decide on how to proceed once we have more data about whether this is possible at all or not.

ruzell22:
@petermetz here are different ways that I did in index.ts back then and now tried it to public-api.ts but it is still not fixing the type exports error.

Test 1:
export { IVerifierEventListener, LedgerEvent } from "./verifier";
 
Test 2:
import { IVerifierEventListener, LedgerEvent } from "./verifier";
export { IVerifierEventListener, LedgerEvent };
 
Test 3:
export { IVerifierEventListener, LedgerEvent } from "@hyperledger/cactus-core-api";
 
Test 4:
export { Verifier, IVerifierEventListener, LedgerEvent } from "./verifier";
 
etc....

I ran `yarn run configure` for each test and it reflects the exports in .d.ts file but not on the .js file so maybe that is one finding that can help. On one test (Test 4), I have tried adding Verifier to see if it will reflect in .d.ts and .js and it did, but the two exports IVerifierEventListener, LedgerEvent is not reflecting. 

I also tried tweaking the compilerOptions in tsconfig.json but did not do any changes.

petermetz:
@ruzell22 OK, thank you for going the extra mile with the investigations. In that case unfortunately our only resort is to close this PR, temporarily until release v3.0.0 is due for issuance next Spring.
I've created an epic here to collect links to breaking changes we need to make before v3.0.0 and there is also a label applied.
Sorry that we couldn't merge this just now but we must not make breaking changes. 
I'm approving and closing it now and we can re-open + merge it as-is when the time comes next year.

petermetz (approved):
LGTM

petermetz:
Temporarily closing, please see above for explanation.

PR #3728, opened 2025-01-21
fix(satp-hermes): when a db is not provided it creates a local repository, remote db is not required anymore

LordKubaya:
- Fixed logging issues.
- If no configuration for the local database is provided, the system will default to using a local SQLite3 database and perform a migration upon the first login.

### DIFF
--- a/package.json
+++ b/package.json
@@ -335,5 +335,9 @@
       "built": false
     }
   },
-  "packageManager": "yarn@4.3.1"
+  "packageManager": "yarn@4.3.1",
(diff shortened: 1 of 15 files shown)


RafaelAPB (approved):
LGTM. A suggestion: "fix(satp-hermes): when a db is not provided it creates a local repository, remote db is not required anymore" should state the code changes, not why.
E.g., a more suitable title could be "fix(satp-hermes): make remoteRepository optional", with your description in the commit message

PR #3791, opened 2025-03-02
fix(satp): check gas usage on satp bridge manager

JJSantos22:
Added abstract function for merging receipts and implemented it on each bridge;
Logged new receipts for user visualization;
Added e2e test for Fabric to Besu transactions;
Fixed some path mistakes;

### DIFF
--- a/packages/cactus-plugin-satp-hermes/src/main/typescript/cross-chain-mechanisms/satp-bridge/besu-bridge.ts
+++ b/packages/cactus-plugin-satp-hermes/src/main/typescript/cross-chain-mechanisms/satp-bridge/besu-bridge.ts
@@ -72,7 +72,7 @@ export class BesuBridge implements NetworkBridge {
   }
 
   public async wrapAsset(asset: EvmAsset): Promise<TransactionResponse> {
-    const fnTag = `${BesuBridge.CLASS_NAME}}#wrapAsset`;
+    const fnTag = `${BesuBridge.CLASS_NAME}#wrapAsset`;
     this.log.debug(
       `${fnTag}, Wrapping Asset: {${asset.tokenId}, ${asset.owner}, ${asset.contractAddress}, ${asset.tokenType}}`,
     );
@@ -110,7 +110,7 @@ export class BesuBridge implements NetworkBridge {
     };
   }
   public async unwrapAsset(assetId: string): Promise<TransactionResponse> {
-    const fnTag = `${BesuBridge.CLASS_NAME}}#unwrapAsset`;
+    const fnTag = `${BesuBridge.CLASS_NAME}#unwrapAsset`;
     this.log.debug(`${fnTag}, Unwrapping Asset: ${assetId}`);
     const response = (await this.connector.invokeContract({
       contractName: this.config.contractName,
@@ -134,7 +134,7 @@ export class BesuBridge implements NetworkBridge {
     assetId: string,
     amount: number,
   ): Promise<TransactionResponse> {
-    const fnTag = `${BesuBridge.CLASS_NAME}}#lockAsset`;
+    const fnTag = `${BesuBridge.CLASS_NAME}#lockAsset`;
     this.log.debug(`${fnTag}, Locking Asset: ${assetId} amount: ${amount}`);
     const response = (await this.connector.invokeContract({
       contractName: this.config.contractName,
@@ -158,7 +158,7 @@ export class BesuBridge implements NetworkBridge {
     assetId: string,
     amount: number,
   ): Promise<TransactionResponse> {
-    const fnTag = `${BesuBridge.CLASS_NAME}}#unlockAsset`;
+    const fnTag = `${BesuBridge.CLASS_NAME}#unlockAsset`;
     this.log.debug(`${fnTag}, Unlocking Asset: ${assetId} amount: ${amount}`);
     const response = (await this.connector.invokeContract({
       contractName: this.config.contractName,
@@ -182,7 +182,7 @@ export class BesuBridge implements NetworkBridge {
     assetId: string,
     amount: number,
   ): Promise<TransactionResponse> {
-    const fnTag = `${BesuBridge.CLASS_NAME}}#mintAsset`;
+    const fnTag = `${BesuBridge.CLASS_NAME}#mintAsset`;
     this.log.debug(`${fnTag}, Minting Asset: ${assetId} amount: ${amount}`);
     const response = (await this.connector.invokeContract({
       contractName: this.config.contractName,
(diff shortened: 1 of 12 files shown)


packages/cactus-plugin-satp-hermes/src/main/typescript/core/satp-utils.ts
```diff
@@ -11,3 +11,25 @@ export function getMessageTypeName(
...
+  let _receipt1 = receipt1;
+  let _receipt2 = receipt2;
+
+  if (!receipt1) {
+    _receipt1 = "{}";
+  }
+
+  if (!receipt2) {
+    _receipt2 = "{}";
+  }
```
LordKubaya, line 28: If there is one receipt missing return the other as stringify, these conditions are unnecessary,

packages/cactus-plugin-satp-hermes/src/main/typescript/core/satp-utils.ts
```diff
@@ -11,3 +11,25 @@ export function getMessageTypeName(
     "UNSPECIFIED"
   );
 }
+
+export function merge_receipt(
```
LordKubaya, line 15: Please comment on this method, what I do, and what it receives.

Also, if the receipts contain multiple identical variables, which one would you choose?

Please add some algorithms for that.

packages/cactus-plugin-satp-hermes/src/main/typescript/cross-chain-mechanisms/satp-bridge/satp-bridge-manager.ts
```diff
@@ -96,13 +97,13 @@ export class SATPBridgeManager implements BridgeManager {
...
 
-    return receipt;
+    return merge_receipt(response.transactionReceipt, receipt);
   }
 
   public async mintAsset(assetId: string, amount: number): Promise<string> {
     const fnTag = `${this.className}#mintAsset()`;
 
-    const transaction = await this.config.network.mintAsset(assetId, amount);
+    const transaction = await this.config.network.mintAsset(assetId, amount); // Why different name?
```
LordKubaya, line 106: Why this comment?

packages/cactus-plugin-satp-hermes/src/main/typescript/core/satp-utils.ts
LordKubaya: Also, move this method to the bridge path, this is not part of the SATP, or shouldn't be.

LordKubaya (changes requested):
Please take a look at the review and change the code accordingly.

packages/cactus-plugin-satp-hermes/src/main/typescript/cross-chain-mechanisms/satp-bridge/satp-bridge-manager.ts
JJSantos22: Forgot about it, but basically every method calls this interaction response instead of transaction, with the exception of the mint and burn function. I wanted to ask if it was on purpose or something else.

packages/cactus-plugin-satp-hermes/src/main/typescript/core/satp-utils.ts
JJSantos22: When you mention bridge path, you mean the cross-chain-mechanisms folder?

LordKubaya: yes

packages/cactus-plugin-satp-hermes/src/main/typescript/cross-chain-mechanisms/satp-bridge/besu-bridge.ts
```diff
@@ -406,4 +406,31 @@ export class BesuBridge implements NetworkBridge {
...
+      ...JSON.parse(receipt1 as string),
+      ...JSON.parse(receipt2 as string),
+    };
+
+    // Remove unwanted fields
+    delete mergedReceipt.effectiveGasPrice;
+    delete mergedReceipt.hash;
+
+    return JSON.stringify(mergedReceipt, null, 2);
+  }
```
LordKubaya, line 435: I think doing this will be prone to cause tech debt, maybe we should just leave these and not delete them,

packages/cactus-plugin-satp-hermes/src/main/typescript/cross-chain-mechanisms/satp-bridge/besu-bridge.ts
```diff
@@ -406,4 +406,31 @@ export class BesuBridge implements NetworkBridge {
...
+    const mergedReceipt = {
+      ...JSON.parse(receipt1 as string),
+      ...JSON.parse(receipt2 as string),
+    };
+
+    // Remove unwanted fields
+    delete mergedReceipt.effectiveGasPrice;
+    delete mergedReceipt.hash;
+
+    return JSON.stringify(mergedReceipt, null, 2);
```
LordKubaya, line 434: Use `safeStringify`.
Why the other variables in the input? Why are they necessary? If so why are those,

packages/cactus-plugin-satp-hermes/src/main/typescript/cross-chain-mechanisms/satp-bridge/ethereum-bridge.ts
```diff
@@ -419,4 +419,31 @@ export class EthereumBridge implements NetworkBridge {
...
+      ...JSON.parse(receipt1 as string),
+      ...JSON.parse(receipt2 as string),
+    };
+
+    // Remove unwanted fields
+    delete mergedReceipt.effectiveGasPrice;
+    delete mergedReceipt.hash;
+
+    return JSON.stringify(mergedReceipt, null, 2);
+  }
```
LordKubaya, line 448: Same as the before

LordKubaya (changes requested):

packages/cactus-plugin-satp-hermes/src/main/typescript/cross-chain-mechanisms/satp-bridge/besu-bridge.ts
JJSantos22: I'll switch.
The other variables were related to indentation

packages/cactus-plugin-satp-hermes/src/main/typescript/cross-chain-mechanisms/satp-bridge/besu-bridge.ts
```diff
@@ -406,4 +406,31 @@ export class BesuBridge implements NetworkBridge {
...
+    receipt2: string | undefined,
+  ): string {
+    if (!receipt1 && !receipt2) return JSON.stringify({}, null, 2);
+    if (!receipt1) return receipt2 as string;
+    if (!receipt2) return receipt1 as string;
+
+    const mergedReceipt = {
+      ...JSON.parse(receipt1 as string),
+      ...JSON.parse(receipt2 as string),
+    };
```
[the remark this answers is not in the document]
JJSantos22: Would it be better to separate the fields in case of a field with the same name, meaning keep both but distinguish the name?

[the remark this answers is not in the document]
LordKubaya: or you can create a list in that field with the 2 values.

LordKubaya (approved):
LGTM

PR #3800, opened 2025-03-10
feat(satp): Implement timeout and retry logic for SATP step messages

### LINKED ISSUE #3744: fix(satp): Implement timeout and max retries of the SATP steps messages
When a session is created a number of retries that a gateway should try is given, the usage of those variables is not implemented.

sfilangio01:
Introduces a new utility function, executeOperationWithRetriesAndTimeout, which encapsulates retry and timeout logic for SATP step message processing. This update aims to enhance reliability by handling transient failures and timeouts effectively.

Additionally, two new error types are introduced in satp-errors.ts:
1. GatewayRetryOperationError: Thrown when retries are exhausted without success.
2. InvalidStageError: Thrown when an invalid stage is encountered during the SATP process.

Resolves #3744.

**Pull Request Requirements**
- [ X] Rebased onto `upstream/main` branch and squashed into single commit to help maintainers review it more efficient and to avoid spaghetti git commit graphs that obfuscate which commit did exactly what change, when and, why.
- [X ] Have git sign off at the end of commit message to avoid being marked red. You can add `-s` flag when using `git commit` command. You may refer to this [link](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) for more information.
- [ X] Follow the Commit Linting specification. You may refer to this [link](https://www.conventionalcommits.org/en/v1.0.0-beta.4/#specification) for more information. 

**Character Limit**
- [X ] Pull Request Title and Commit Subject must not exceed 72 characters (including spaces and special characters).
- [ X] Commit Message per line must not exceed 80 characters (including spaces and special characters).

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-plugin-satp-hermes/src/main/typescript/core/errors/satp-errors.ts
+++ b/packages/cactus-plugin-satp-hermes/src/main/typescript/core/errors/satp-errors.ts
@@ -101,3 +101,24 @@ export class RecoverMessageError extends SATPInternalError {
 }
 // TODO client-facing error logic, maps SATPInternalErrors to user friendly errors
 export class SATPError extends Error {}
+
+export class GatewayRetryOperationError extends SATPInternalError {
+  constructor(message: string, cause?: string | Error | null) {
+    super(message, cause ?? null, 500);
+  }
+}
+
+export class InvalidStageError extends SATPInternalError {
+  constructor(
+    fnTag: string,
+    stage: string,
+    sessionID: string,
+    cause?: string | Error | null,
+  ) {
+    super(
+      `${fnTag}, Invalid stage: ${stage} for session: ${sessionID}`,
+      cause ?? null,
+      400,
+    );
+  }
+}
--- a/packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
+++ b/packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
@@ -74,6 +74,8 @@ import {
   RecoverMessageError,
   RetrieveSATPMessageError,
   TransactError,
+  GatewayRetryOperationError,
+  InvalidStageError,
 } from "../../core/errors/satp-errors";
 import { getMessageTypeName } from "../../core/satp-utils";
 import { HealthCheckResponseStatusEnum } from "../../generated/gateway-client/typescript-axios";
@@ -110,6 +112,7 @@ export class SATPManager {
   public connectedDLTs: NetworkId[] = [];
   private supportedDLTs: LedgerType[] = [];
   private sessions: Map<string, SATPSession>;
+  private sessionConfig: any;
   // maps stage to client/service and service class
   private readonly satpServices: Map<
     string,
@@ -307,6 +310,16 @@ export class SATPManager {
   public getSATPHandler(type: SATPHandlerType): SATPHandler | undefined {
     return this.satpHandlers.get(type);
   }
+  private updateSessionConfig(sessionData: {
+    maxRetries: string;
+    maxTimeout: string;
+  }) {
+    this.sessionConfig = {
+      retries: parseInt(sessionData.maxRetries, 10),
+      timeoutMs: parseInt(sessionData.maxTimeout, 10),
+      delayMs: 500,
+    };
+  }
 
   public getOrCreateSession(
     sessionId?: string,
@@ -483,6 +496,51 @@ export class SATPManager {
     );
   }
 
+  /**
+   * Executes an asynchronous operation with retries and a timeout.
+   * - The function attempts to perform the 'operation' up to 'retries' times until it succeeds.
+   * - Each attempt must complete within 'timeoutMs', or it is considered failed.
+   * - Waits 'delayMs' milliseconds between retries to avoid immediate retries.
+   * @param operation The asynchronous operation to execute, returning a Promise.
+   * @param error The error to throw if the operation fails after all retries/timeout.
+   * @returns A Promise resolved with the operation's result or rejected after all retries fail.
+   * @throws Error if the operation fails after all retries or if the operation times out.
+   */
+  private async executeOperationWithRetriesAndTimeout<T>(
+    operation: () => Promise<T>,
+    error: Error,
+  ): Promise<T> {
+    let { retries, timeoutMs } = this.sessionConfig;
+    const { delayMs } = this.sessionConfig;
+    retries = isNaN(retries) ? 3 : retries;
+    timeoutMs = isNaN(timeoutMs) ? 5000 : timeoutMs;
+    for (let attempt = 1; attempt <= retries; attempt++) {
+      try {
+        this.logger.debug(`Attempt ${attempt}: Starting operation`);
+        return await Promise.race([
+          operation(),
+          new Promise<T>((_, reject) =>
+            setTimeout(
+              () => () =>
+                reject(
+                  new GatewayRetryOperationError(
+                    `Operation ${operation} timed out`,
+                  ),
+                ),
+              timeoutMs,
+            ),
+          ),
+        ]);
+      } catch (err) {
+        this.logger.error(`Attempt ${attempt} failed: ${err.message}`);
+        if (attempt < retries) {
+          await new Promise((resolve) => setTimeout(resolve, delayMs));
+        }
+      }
+    }
+    throw error;
+  }
+
   public async transfer(
     session: SATPSession,
     stage?: MessageType,
@@ -526,6 +584,9 @@ export class SATPManager {
       }
 
       const sessionData: SessionData = clientSessionData;
+
+      // Updating retry and timeout settings for the session
+      this.updateSessionConfig(sessionData);
 
       sessionData.receiverGatewayOwnerId = channel.toGatewayID;
 
@@ -585,16 +646,17 @@ export class SATPManager {
         case undefined:
         case MessageType.NEW_SESSION_REQUEST:
           this.logger.debug(`${fnTag}, Initiating Stage 0`);
-          newSessionRequest = await (
-            this.getSATPHandler(SATPHandlerType.STAGE0) as Stage0SATPHandler
-          ).NewSessionRequest(session.getSessionId());
-
-          if (!newSessionRequest) {
-            throw new CreateSATPRequestError(
+          this.logger.debug("CREATING A NEW REQUEST");
+          newSessionRequest = await this.executeOperationWithRetriesAndTimeout(
+            async () =>
+              await (
+                this.getSATPHandler(SATPHandlerType.STAGE0) as Stage0SATPHandler
+              ).NewSessionRequest(session.getSessionId()),
+            new CreateSATPRequestError(
               fnTag,
               getMessageTypeName(MessageType.NEW_SESSION_REQUEST),
-            );
-          }
+            ),
+          );
 
         case MessageType.NEW_SESSION_RESPONSE:
           if (!newSessionRequest) {
@@ -613,21 +675,18 @@ export class SATPManager {
               );
             }
           }
-
-          newSessionResponse =
-            await clientSatpStage0.newSession(newSessionRequest);
-
+          newSessionResponse = await this.executeOperationWithRetriesAndTimeout(
+            async () => {
+              return await clientSatpStage0.newSession(newSessionRequest!);
+            },
+            new RetrieveSATPMessageError(
+              fnTag,
+              getMessageTypeName(MessageType.NEW_SESSION_RESPONSE),
+            ),
+          );
           this.logger.debug(
             `${fnTag}, newSessionResponse: ${safeStableStringify(newSessionResponse)}`,
           );
-
-          if (!newSessionResponse) {
-            throw new RetrieveSATPMessageError(
-              fnTag,
-              getMessageTypeName(MessageType.NEW_SESSION_RESPONSE),
-            );
-          }
-
         case MessageType.PRE_SATP_TRANSFER_REQUEST:
           if (!newSessionResponse) {
             this.logger.debug(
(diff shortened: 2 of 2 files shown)


LordKubaya:
Could you resolve DCO errors, your commit is not signed. 
Set up GPG keys and use `git commit -s`.

Be careful with the CI errors.

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
```diff
@@ -483,6 +496,51 @@ export class SATPManager {
...
+   * @returns A Promise resolved with the operation's result or rejected after all retries fail.
+   * @throws Error if the operation fails after all retries or if the operation times out.
+   */
+  private async executeOperationWithRetriesAndTimeout<T>(
+    operation: () => Promise<T>,
+    error: Error,
+  ): Promise<T> {
+    let { retries, timeoutMs } = this.sessionConfig;
+    const { delayMs } = this.sessionConfig;
+    retries = isNaN(retries) ? 3 : retries;
```
LordKubaya, line 515: I recommend you to read this[https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/isNaN] and be careful, that `isNaN()` does not resolve everything.

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
```diff
@@ -734,28 +800,29 @@ export class SATPManager {
               MessageType.INIT_PROPOSAL,
             ) as TransferProposalRequest;
 
-            if (!transferProposalRequest) {
-              throw new RecoverMessageError(
-                fnTag,
-                getMessageTypeName(MessageType.INIT_PROPOSAL),
+                new RecoverMessageError(
```
LordKubaya, line 803: Where is the throw? This is not throwing errors.

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
```diff
@@ -709,20 +773,22 @@ export class SATPManager {
...
+              async () =>
+                await (
+                  this.getSATPHandler(
+                    SATPHandlerType.STAGE1,
+                  ) as Stage1SATPHandler
+                ).TransferProposalRequest(
+                  session.getSessionId(),
+                  preSATPTransferResponse!,
+                ),
+              new CreateSATPRequestError(
```
LordKubaya, line 787: same here

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
```diff
@@ -774,16 +841,21 @@ export class SATPManager {
...
+                  this.getSATPHandler(
+                    SATPHandlerType.STAGE1,
+                  ) as Stage1SATPHandler
+                ).TransferCommenceRequest(transferProposalResponse!),
 
-          if (!transferCommenceRequest) {
-            throw new CreateSATPRequestError(
-              fnTag,
-              getMessageTypeName(MessageType.TRANSFER_COMMENCE_REQUEST),
+              new CreateSATPRequestError(
```
LordKubaya, line 853: same here

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
```diff
@@ -802,20 +874,22 @@ export class SATPManager {
...
-            transferCommenceRequest,
-          );
+          transferCommenceResponse =
+            await this.executeOperationWithRetriesAndTimeout(
+              async () =>
+                await clientSatpStage1.transferCommence(
+                  transferCommenceRequest!,
+                ),
 
+              new RetrieveSATPMessageError(
```
LordKubaya, line 884: here

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
```diff
@@ -836,21 +910,27 @@ export class SATPManager {
...
+                  this.getSATPHandler(
+                    SATPHandlerType.STAGE2,
+                  ) as Stage2SATPHandler
+                ).LockAssertionRequest(transferCommenceResponse!),
 
-          if (!lockAssertionRequest) {
-            throw new CreateSATPRequestError(
-              fnTag,
-              getMessageTypeName(MessageType.LOCK_ASSERT),
+              new CreateSATPRequestError(
```
LordKubaya, line 922: here

packages/cactus-plugin-satp-hermes/src/main/typescript/core/errors/satp-errors.ts
```diff
@@ -101,3 +101,24 @@ export class RecoverMessageError extends SATPInternalError {
 }
 // TODO client-facing error logic, maps SATPInternalErrors to user friendly errors
 export class SATPError extends Error {}
+
+export class GatewayRetryOperationError extends SATPInternalError {
+  constructor(message: string, cause?: string | Error | null) {
+    super(message, cause ?? null, 500);
+  }
+}
```
LordKubaya, line 109: should this be of type SATP Error?

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
```diff
@@ -613,21 +675,18 @@ export class SATPManager {
...
             `${fnTag}, newSessionResponse: ${safeStableStringify(newSessionResponse)}`,
           );
-
-          if (!newSessionResponse) {
-            throw new RetrieveSATPMessageError(
-              fnTag,
-              getMessageTypeName(MessageType.NEW_SESSION_RESPONSE),
-            );
-          }
-
```
LordKubaya, line 630: Why did you delete this?

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
```diff
@@ -865,23 +945,24 @@ export class SATPManager {
...
           this.logger.debug(
             `${fnTag}, lockAssertionResponse: ${safeStableStringify(lockAssertionResponse)}`,
           );
 
-          if (!lockAssertionResponse) {
-            throw new RetrieveSATPMessageError(
-              fnTag,
-              getMessageTypeName(MessageType.ASSERTION_RECEIPT),
-            );
-          }
```
LordKubaya, line 879: This too, and more...

LordKubaya (changes requested):
Please do not change how the transfer is executed, errors are not throwing, be careful when we are starting in the middle of a transfer(recovering) because there can be messages already received or created in the session... That code was already done and it should not be deleted.

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
sfilangio01: The throw is inside the executeOperationWithRetriesAndTimeout() function at line 541. The error here (and in all the other instances where the executeOperationWithRetriesAndTimeout function is called) is passed as a parameter and it's thrown if the timeout expires a "retries" amount of time. By passing the error as a parameter, we avoid repeating writing the same code to check if it's undefined and then throwing the error

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
sfilangio01: I fixed this now

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
sfilangio01: here as well the error is passed as a parameter and it will be thrown (in case of errors) later inside the auxiliar function

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
sfilangio01: Same as before, the error is passed as a parameter

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
sfilangio01: Same as before, parameter

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
sfilangio01: same here, it's a parameter of the executeOperationWithRetriesAndTimeout function

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
sfilangio01: Ok, I will check!

packages/cactus-plugin-satp-hermes/src/main/typescript/core/errors/satp-errors.ts
sfilangio01: I don't know if I am wrong but I remember you saying that the error I needed to create should have been extending RuntimeError. As I can see, SatpError is extending directly Error and not RuntimeError, so in this case I would say that it's correct the way it is now. But if instead I am being mistaken, then yes, it should be extending the SatpError. 
What do you think?

sfilangio01:
First of all, thank you for all of your feedbacks and for your review.

Regarding your comments:
1)
> Could you resolve DCO errors, your commit is not signed. Set up GPG keys and use `git commit -s`.
> 
> Be careful with the CI errors.

Yes, I already set up the key and signed my first commit, but then there was the problem with the conflicts and that's why we talked yesterday. When I did the commit with you I think forgot to add the signature again

2) 
Please do not change how the transfer is executed, errors are not throwing, be careful when we are starting in the middle of a transfer(recovering) because there can be messages already received or created in the session... That code was already done and it should not be deleted.

As i said before, the errors are indeed thrown (but inside the executeOperationWithRetriesAndTimeout instead of the previous position). I don't know if this was not clear or maybe the code is not super clean, but I also added in the documentation of the code the way it was functioning. If you think it will be better, i can revert to the previous stage and instead of throwing the error in the auxiliary function i can throw it like it was in the previous code.

Thank you again!

packages/cactus-plugin-satp-hermes/src/main/typescript/core/errors/satp-errors.ts
LordKubaya: Gateway communication is not SATP Protocol, you should create another type of error that extends RuntimeError.

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
LordKubaya: Right.

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
LordKubaya: okay

packages/cactus-plugin-satp-hermes/src/main/typescript/core/errors/satp-errors.ts
sfilangio01: Ok, will work on this today

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
```diff
@@ -485,6 +487,100 @@ export class SATPManager {
...
+   * - Each attempt must complete within 'timeoutMs', or it is considered failed.
+   * - Waits 'delayMs' milliseconds between retries to avoid immediate retries.
+   * @param operation The asynchronous operation to execute, returning a Promise.
+   * @param error The error to throw if the operation fails after all retries/timeout.
+   * @param retries The number of times to retry the operation if it fails.
+   * @param timeoutMs The maximum time in milliseconds to wait for the operation to complete.
+   * @returns A Promise resolved with the operation's result or rejected after all retries fail.
+   * @throws Error if the operation fails after all retries or if the operation times out.
+   */
+  private async executeOperationWithRetriesAndTimeout<T>(
```
petermetz, line 502: @sfilangio01 2 things from my side:

**First** I echo what @LordKubaya said, I personally prefer to check 
1. if a parameter is falsy, then throw with a specific message to that if it was indeed falsy
2. The check if it's isNan - throw specific message if it is
3. Check with `isFinite()`
4. Check if it's negative for things that cannot be negative (for example I'd recommend the number of retries and the timeout to never be negative)

**Second** This function looks re-usable enough (which is a great thing so kudos for that) that you can use it in other files as well so I suggest you move it to its own file and import it where needed. 
The other additional benefit of this will be the reduction of the size of the satp-manager file which is already a little oversized.
The thing that will give you some trouble will the be local state which needs to be passed around in function parameters instead. My favored pattern for that is to have a single parameter which is an object that can then have arbitrary properties carrying the information/context that you need to pass around

I hope this helps!

petermetz (changes requested):
@sfilangio01 It looks like there might be a file with special characters in its name added to the diff. Unless there is a very specific purpose and need for it, I'd like to ask that you remove the `:monitor:` file.

image

packages/cactus-plugin-satp-hermes/src/main/typescript/services/gateway/satp-manager.ts
sfilangio01: Hello Peter!
First of all, thank you so much for the time spent on giving the feedback and for the suggestions you gave me.

I followed your advice and I created another file containing the function and added a check for every parameter the functions receives.

Regarding the :monitor: file, I have no idea how it was added in the commit and I think it was an auto generated file, I will remove it in my next commit.


Thanks again

petermetz: @sfilangio01 My pleasure, you are welcome and thank you again for the contributions!!

petermetz (approved):

RafaelAPB:
@sfilangio01 could you please rebase your branch with the latest satp-dev, and re-test? We would like to incorporate your changes soon

RafaelAPB:
cc @LordKubaya @AndreAugusto11

RafaelAPB:
@sfilangio01 could you please rebase your branch with the latest satp-stg, and re-test?

PR #3830, opened 2025-03-29
feat(common): add BigInt JSON replacer (number -> string)

petermetz:
This can be used by code on the back-end to be able to safely send
BigInt values through the  various transports (HTTP, gRPC, CRPC, etc.)

By default JSON.stringify will convert BigInt values to numbers which
can cause data loss if the number is too large. By using this replacer
function, we can ensure that BigInt values are always sent as strings.

It adds overhead on the client side to have to manually parse the BigInt
values, but it is a necessary trade-off for preserving data integrity.

Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>

**Pull Request Requirements**
- [x] Rebased onto `upstream/main` branch and squashed into single commit to help maintainers review it more efficient and to avoid spaghetti git commit graphs that obfuscate which commit did exactly what change, when and, why.
- [x] Have git sign off at the end of commit message to avoid being marked red. You can add `-s` flag when using `git commit` command. You may refer to this [link](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) for more information.
- [x] Follow the Commit Linting specification. You may refer to this [link](https://www.conventionalcommits.org/en/v1.0.0-beta.4/#specification) for more information. 

**Character Limit**
- [x] Pull Request Title and Commit Subject must not exceed 72 characters (including spaces and special characters).
- [x] Commit Message per line must not exceed 80 characters (including spaces and special characters).

**A Must Read for Beginners**
For rebasing and squashing, here's a [must read guide](https://github.com/servo/servo/wiki/Beginner's-guide-to-rebasing-and-squashing) for beginners.

PR #3844, opened 2025-04-09
feat(connector-besu): add WatchSolidityEventV1 endpoint

### LINKED ISSUE #3825: feat(connector-besu): add WatchSolidityEventV1 endpoint

Currently the Besu connector only supports watching entire blocks which makes it much more cumbersome to watch for specific events.

A dedicated endpoint that is purpose-built for solidity events would add a lot of convenience to users.

None, I'm confident we need this.

petermetz:
A dedicated endpoint that is purpose-built for solidity events
would add a lot of convenience to users.

Depends on #3830

Fixes #3825

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


packages/cactus-plugin-ledger-connector-besu/src/main/json/openapi.json
```diff
@@ -249,6 +249,147 @@
...
+          },
+          "logs": {
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/ViemV2_24_2_EventLog"
+            }
+          }
+        }
+      },
+      "ViemV2_24_2_EventLog": {
```
outSH, line 321: I wouldn't use Viem/Web3js name in our return specifications. In case something changes in View output in the future we will have to transform it to match our (older) response specification for API stability. This may be confusing for users who may expect that this structure is pinned to actual View output

outSH (approved):
I have only one small comment, otherwise it looks great

packages/cactus-plugin-ledger-connector-besu/src/main/json/openapi.json
jagpreetsinghsasan: I think @petermetz did this intentionally so that we can replace web3 with viem gradually over time, without breaking functionality of the existing codebase (@petermetz please correct me if I am wrong here). 
But yes, if we call it `Viem_EventLog` instead of `ViemV2_24_2_EventLog` , it would be cleaner in my opinion.

petermetz: @outSH @jagpreetsinghsasan sorry for the slow response! I'll note down here my thoughts, please let me know if based on these it makes sense to keep it as-is or not. If not, I'm fine with changing it, but I wanted to make sure we are on the same page first:
1. The probability of breaking changes in the Viem types is 100% as time goes on.
2. The probability of security vulnerabilities in Viem that will force us to upgrade is also a 100% as time approaches infinity.
3. Based on 1 and 2 it is 100% that we'll have to adjust our types.
4. When we are at that juncture, we'll have the choices of a) make a breaking change in the `Viem_EventLog` type which will be more confusing to users or b) avoid a breaking change by having `ViemV2_24_2_EventLog` now and add a new `ViemV3_0_0_EventLog` and let it be immediately obvious to anyone what exactly happened and why (breaking change in the upstream library and in which specific version, etc.)

I'm not married to option b) so just let me know what do you prefer and I'm happy to update it as such.

jagpreetsinghsasan: @petermetz true, I just checked the viem gh repo and it seems like they have pretty frequent updates, breaking changes and dependency upgrades. I am thus fine with the current state of the code in this PR.
@outSH what are your thoughts about this?

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/common/ensure-0x-prefix.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ *
+ * @param maybePrefixed
+ * @returns The string itself if it was 0x prefixed or a new 0x prefixed version of if it wasn't.
+ */
+export function ensure0xPrefix(maybePrefixed: string): string & `0x${string}` {
+  const fn = "ensure-0x-prefix.ts";
```
AndreAugusto11, line 7: Is the `.ts` there by mistake?

packages/cactus-plugin-ledger-connector-besu/src/main/typescript/web-services/watch-events-v1-endpoint.ts
```diff
@@ -0,0 +1,149 @@
...
+import { watchEventsV1Impl } from "../impl/watch-events-v1/watch-events-v1-impl";
+import { ensure0xPrefix } from "../common/ensure-0x-prefix";
+
+export interface IWatchEventsV1EndpointOptions {
+  readonly logLevel?: LogLevelDesc;
+  readonly socket: SocketIoSocket;
+  readonly viemClient: ViemPublicClient;
+}
+
+export class WatchEventsV1Endpoint {
```
AndreAugusto11, line 30: @petermetz, I would just suggest renaming classes/types to something like SolidityEvent or ContractEvent, just like the PR title. This may be a bit misleading since there is usually the possibility of listening for multiple events beyond the ones emitted by contracts (e.g., pending Transactions, new Heads, etc.)

packages/cactus-plugin-ledger-connector-besu/src/main/json/openapi.json
outSH: @petermetz @jagpreetsinghsasan 

> When we are at that juncture, we'll have the choices of a) make a breaking change in the Viem_EventLog type which will be more confusing to users or b) avoid a breaking change by having ViemV2_24_2_EventLog now and add a new ViemV3_0_0_EventLog

I'm thinking about a case where we don't actualy change `Viem_EventLog`, but write an adapter code that will transform the Viem output to backward compatible format, so that our user can use latest library release without security issues and not modify his code at all, but only rely on Cacti API stability. We can of course introduce breaking changes in `Viem_EventLog:V2` and remove `Viem_EventLog:V1` in next major Cacti release to follow viem implementation in a long run. If we use `ViemV2_24_2_EventLog` then we're strictly pinned to Viem and user has to update the code every time something in Viem changes, or use older Cacti/Viem version.

I think head/block subscription structure is likely linked to ethereum JSON-RPC and not Viem itself, so the changes we can expect in the future are likely in types or some fields being named differently / marked optional, but the data will be the same between libraries, so I don't expect much work to be done here.

AndreAugusto11:
@petermetz, we are working on the `satp-hermes` plugin, which is dependent on this PR. Can I help accelerate acceptance and merge in any way?

RafaelAPB:
closing as tracking issue was created https://github.com/hyperledger-cacti/cacti/issues/3980

petermetz:
@RafaelAPB Of course, no worries at all and sorry again for the slow responses, I'm just catching up on my GitHub notifications!

PR #3865, opened 2025-04-16
feat(connector-ethereum): add event listener for smart contracts

AndreAugusto11:
* Added new function to perform event listening in smart contracts
* Added new test file

### DIFF
--- a/packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
+++ b/packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/plugin-ledger-connector-ethereum.ts
@@ -11,6 +11,7 @@ import Web3, {
   TransactionReceiptBase,
   WebSocketProvider,
 } from "web3";
+
 import { PayableMethodObject } from "web3-eth-contract";
 
 import OAS from "../json/openapi.json";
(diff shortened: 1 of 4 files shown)


petermetz (approved):

AndreAugusto11:
> LGTM, but can't this be implemented and used as a connector service, like it was done here: #3844?

@outSH yes, definitely! I was not aware of that.

I guess the implementation will be very similar (if not equal). Do you guys (cc @petermetz) have any suggestions on how to proceed? Do you see any alternative to avoid a huge code duplication, and possibly having to change two packages if Viem introduces some breaking changes?

petermetz:
> > LGTM, but can't this be implemented and used as a connector service, like it was done here: #3844?
> 
> @outSH yes, definitely! I was not aware of that.
> 
> I guess the implementation will be very similar (if not equal). Do you guys (cc @petermetz) have any suggestions on how to proceed? Do you see any alternative to avoid a huge code duplication, and possibly having to change two packages if Viem introduces some breaking changes?

@AndreAugusto11 Sorry, I'm late for the discussion but wanted to make sure I respond anyway: You could pull the code into it's own package or to an existing cross-functional package such as `cactus-core` and that way both connector's can just import the re-usable code from there instead of defining it themselves.

PR #3868, opened 2025-04-17
test(besu-all-in-one): add support for >=v6.12 Linux kernels (OpenJDK)

### LINKED ISSUE #3867: test(besu-all-in-one): add support for >=v6.12 Linux kernels (OpenJDK)
On newer Linux kernel versions the Besu AIO image is broken due to an issue with cgroupv2.
Upgrading the image was long overdue anyway so this task will be about swapping out the custom config we were using so far with an image definition that is mostly just inheriting the existing besu container image (which already has a JDK version that works with the newer Linux kernel versions mentioned).

This will fix the problem and also reduce our technical debt.

Reading list:
- https://bugs.openjdk.org/browse/JDK-8346874
- https://bugs.openjdk.org/browse/JDK-8348566

petermetz:
Primary change:
---------------

The upgraded image will now work with Linux kernels equal to and newer than v6.12.

Secondary changes:
------------------

1. Besu AIO is now based much more closely on the vanilla Besu image itself.
Previously we couldn't do this because way back when the Besu AIO Cacti image
was created we needed a lot more customization to get it working for our tests.
2. Also migrating the image to the new namespace and name as we were planning
on it for a while. At this point we can drop the `cacti` prefix from the image
name because the same prefix is now present in the repo name. Leaving the cacti
prefix in would make it so that it's in the fully qualified image tag twice but
our image names are already way too long so this is a nice quality of life improvement.
3. Deleted a lot of the files that are no longer needed for the Besu AIO
image's functioning.

Fixes #3867

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


tools/docker/besu-all-in-one/Dockerfile
```diff
@@ -1,29 +1,21 @@
-ARG BESU_VERSION=22.7.4
-ARG TESSERA_VERSION=22.1.7
+FROM hyperledger/besu:25.4.0
 
-FROM hyperledger/besu:$BESU_VERSION AS besu
-FROM quorumengineering/tessera:$TESSERA_VERSION AS tessera
+LABEL org.opencontainers.image.source="https://github.com/hyperledger-cacti/cacti"
 
-COPY --from=besu /opt/besu/ /opt/besu/
+RUN apt update -y && apt install -y curl
```
outSH, line 5: Please add  `&& rm -rf /var/lib/apt/lists/*` and optionally `--no-install-recommends` to minimize image size

outSH (changes requested):

tools/docker/besu-all-in-one/Dockerfile
petermetz: @outSH Oops, thank you for catching that! I'll update it in a sec!

petermetz: @outSH Please re-review when you get the chance and let me know! Heads up: I've also fixed some of the failing tests that were dependent on the key files we were adding to the image earlier (now they just pull the key data directly from the genesis.json instead)

jagpreetsinghsasan (approved):
LGTM

outSH (approved):
Looks great, thanks!

PR #3870, opened 2025-04-18
fix(satp): fix log level consistency

JJSantos22:
Connected log levels between files to avoid inconsistencies;


AndreAugusto11 (approved):
LGTM

Issue #3507, opened 2024-09-03
test(ethereum): additional properties to deployContract does not throw error


ashnashahgrover:
**Describe the bug**

The deployContract function in cactus-plugin-ledger-connector-ethereum takes DeployContractV1Request as an argument, which already declares additional properties as forbidden in the Open API spec as can be seen [here](https://github.com/hyperledger/cacti/blob/957da7c3e1d80068391485a825ba6bb1e68333ac/packages/cactus-plugin-ledger-connector-ethereum/src/main/json/openapi.json#L435): 

```
"DeployContractV1Request": {
        "type": "object",
        "required": ["web3SigningCredential", "contract"],
        **"additionalProperties": false,**
        "properties": {
          "web3SigningCredential": {
            "$ref": "#/components/schemas/Web3SigningCredential"
          },
          "contract": {
            "oneOf": [
              {
                "$ref": "#/components/schemas/ContractJsonDefinition"
              },
              {
                "$ref": "#/components/schemas/ContractKeychainDefinition"
              }
            ],
            "nullable": false
          },
          "constructorArgs": {
            "description": "The list of arguments to pass in to the constructor of the contract being deployed.",
            "type": "array",
            "default": [],
            "items": {}
          },
          "gasConfig": {
            "$ref": "#/components/schemas/GasTransactionConfig"
          },
          "value": {
            "type": "string",
            "description": "Ether balance to send on deployment.",
            "nullable": false
          }
        }
      }
```

However in various tests that input additional properties, and expect an error to be thrown, the test actually passes. 

**To Reproduce**

Run the test beginning on [this line](https://github.com/hyperledger/cacti/blob/957da7c3e1d80068391485a825ba6bb1e68333ac/packages/cactus-plugin-ledger-connector-ethereum/src/test/typescript/integration/geth-contract-deploy-and-invoke-using-json-object-v1.test.ts#L245).

The test should throw an error, but it passes without an error. 

**Expected behavior**

The test should throw an error saying additional properties are forbidden. 

**Logs/Stack traces**

Can help maintainers identify root causes

**Hyperledger Cactus Plugins/Connectors Used**

- Ethereum 

**Additional context**

- This issue was derived from inconsistencies found in [issue 3487](https://github.com/hyperledger/cacti/issues/3487) and [issue 3475](https://github.com/hyperledger/cacti/issues/3475)

petermetz:
@ashnashahgrover Thank you very much, this will help us keep track of the bug!

[...]

PR #3885, opened 2025-04-30
fix: Fixed deployContract to throw error if get additional property

Harshdev098:
Fix: #3507 
- Fixed the deployContract to throw error if get any additional property 
- Used `ajv` and to validate the requests 
- Used `json-schema-ref-parser`  to handle the nullable property in the openAPI as `ajv` was throwing errors becuase it require a type keyword when using nullable, while openapi.json uses nullable in contexts. So preferred to go with ref-parser instead of updating the openAPI


packages/cactus-plugin-ledger-connector-ethereum/package.json
```diff
@@ -65,11 +65,13 @@
     "webpack:dev:web": "webpack --env=dev --target=web --config ../../webpack.config.js"
   },
   "dependencies": {
+    "@apidevtools/json-schema-ref-parser": "^12.0.1",
```
petermetz, line 68: @Harshdev098 Please pin to exact versions instead of having auto-upgrade enabled. It's not a surefire way to protect against supply-chain attacks but better than nothing.

packages/cactus-plugin-ledger-connector-ethereum/package.json
```diff
@@ -65,11 +65,13 @@
...
     "@ethereumjs/common": "4.0.0",
     "@ethereumjs/tx": "5.0.0",
     "@hyperledger/cactus-common": "2.1.0",
     "@hyperledger/cactus-core": "2.1.0",
     "@hyperledger/cactus-core-api": "2.1.0",
+    "ajv": "^8.17.1",
```
petermetz, line 74: @Harshdev098 Please pin to exact versions instead of having auto-upgrade enabled. It's not a surefire way to protect against supply-chain attacks but better than nothing.

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/web-services/deploy-contract-v1-endpoint.ts
```diff
@@ -88,6 +128,17 @@ export class DeployContractEndpoint implements IWebServiceEndpoint {
   public async handleRequest(req: Request, res: Response): Promise<void> {
     const reqTag = `${this.getVerbLowerCase()} - ${this.getPath()}`;
     this.log.debug(reqTag);
+    const isValid = validateDeployContract(req.body);
+    if (!isValid) {
+      this.log.warn(
```
petermetz, line 133: @Harshdev098 Please update this to log on the debug level instead. While in a low traffic/dev environment the warn log level might help bringing this to the attention of the user quickly, in production it will just spam the logs with warnings that are not actionable to an ops person because an invalid incoming request doesn't imply that there is a bug to be fixed at all.

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/web-services/deploy-contract-v1-endpoint.ts
```diff
@@ -88,6 +128,17 @@ export class DeployContractEndpoint implements IWebServiceEndpoint {
...
+        `Validation failed: ${JSON.stringify(validateDeployContract.errors, null, 2)}`,
```
petermetz, line 134: @Harshdev098 Please extract the serialization code into a separate line and don't pretty print the JSON. Also make sure to include the req tag so that it's immediately obvious which endpoint encountered the invalid request.

packages/cactus-plugin-ledger-connector-ethereum/src/test/typescript/integration/geth-contract-deploy-and-invoke-using-json-object-v1.test.ts
```diff
@@ -256,7 +256,6 @@ describe("Ethereum contract deploy and invoke using keychain tests", () => {
         gas: 1000000,
         fake: 4,
       } as DeployContractV1Request);
-      fail("Expected deployContract call to fail but it succeeded.");
     } catch (error) {
       console.log("deployContract failed as expected");
     }
```
petermetz, line 261: @Harshdev098 Please modify this test case in a way that it asserts that the returned status code is HTTP 400 (bad request). I think that `fail("...");` line of code was a no-op to begin with so it was good that you removed it but it doesn't necessarily change the outcome of the test case unfortunately.

If you'd like an example you can see in this file how to use Jest matchers declaratively for this kind of assertion:

`./packages/cactus-cmd-api-server/src/test/typescript/unit/get-open-api-spec-v1-oauth2-scopes.test.ts`

```ts
 await expect(
      apiClient.getOpenApiSpecV1({
        headers: { Authorization: validBearerToken },
      }),
    ).rejects.toMatchObject({
      response: {
        status: 403,
        statusText: expect.stringContaining("Forbidden"),
      },
    });
```

petermetz (changes requested):
@Harshdev098 Thank you for your contribution! Overall, looking pretty good, but I did have a few minor change requests that you can see above. On top of those, I'll also ask that you amend your commit message to match the text that you've put in the PR description. 
If you need help with any of these tasks/updates just let me know I'm happy to help out. If you want to take advantage of that help you can also join the daily pair programming sessions on Discord at 10 AM Pacific Time (runs most weekdays).

Harshdev098:
@petermetz Thanks for the detailed feedback! I have updated the PR with required changes, can you please review it

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/web-services/deploy-contract-v1-endpoint.ts
```diff
@@ -22,6 +24,44 @@ import OAS from "../../json/openapi.json";
 import { ERR_INVALID_RESPONSE } from "web3";
 import { isWeb3Error } from "../public-api";
 
+async function preprocessSchema() {
+  const resolvedSchema = await RefParser.dereference(OAS);
+  function adjustNullable(obj: any) {
```
outSH, line 29: Instead of any you can accept `Record<string, unknown>`

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/web-services/deploy-contract-v1-endpoint.ts
```diff
@@ -22,6 +24,44 @@ import OAS from "../../json/openapi.json";
...
+    strict: false,
+    allowMatchingProperties: true,
+  });
+  const processedSchema = await preprocessSchema();
+  ajv.addSchema(processedSchema, "openapi.json");
+
+  validateDeployContract = ajv.compile({
+    $ref: "openapi.json#/components/schemas/DeployContractV1Request",
+  });
+})();
```
outSH, line 63: I don't think IIFE is safe here, it may not finish until `handleRequest` is first called. Please use memoization instead (e.g. add function `getValidateDeployContract` that will compile the schema or return already compiled schema if available.

outSH (changes requested):
Looks great, thanks you! I've left few comments

packages/cactus-plugin-ledger-connector-ethereum/src/main/typescript/web-services/deploy-contract-v1-endpoint.ts
Harshdev098: @outSH, Done with the changes!

petermetz (approved):

petermetz:
@Harshdev098 Sorry for the slow response, LGMT!

RafaelAPB:
Thank you for your contribution. @Harshdev098 please update the MR, rebasing it as per your availability. Please note that we will close stale MRs to cleanup in a close future.

PR #3914, opened 2025-05-20
feat: Supply-chain with wallet authentication

Eajunnn:
**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**


Eajunnn:
I created a whole new supply-chain workflow by adding wallet authentication, using sepolia testnet as ethereum network for testing. All endpoints now are in fabric. Public data stored in sepolia while private are in fabric. User can see certain information based on their connected wallet role. I also created a payment endpoint where it uses sepolia to pay the shipment. If you guys wanted this as additional 3 examples of supply chain in that examples directory can just put all these as new 3 folders in examples directory instead of merging into the original ones. The original ones are using besu and xdai preset testing account, but this is more real world approach. Thanks

package.json
```diff
@@ -326,5 +326,20 @@
...
+    "@angular/forms": "17.3.11",
+    "@angular/material": "17.3.0",
+    "@angular/platform-browser": "17.3.11",
+    "@angular/platform-browser-dynamic": "17.3.11",
+    "@angular/router": "17.3.11",
+    "@hyperledger/cactus-plugin-ledger-connector-ethereum": "workspace:^",
+    "@metamask/detect-provider": "^2.0.0",
+    "ether": "^0.0.9",
+    "ethers": "6.13.5"
+  }
```
outSH, line 344: There should be no dependencies in root package.json file

examples/cactus-example-supply-chain-backend/package.json
```diff
@@ -80,7 +85,11 @@
   },
   "devDependencies": {
     "@types/express": "5.0.1",
-    "@types/uuid": "10.0.0"
+    "@types/jest": "^29.5.14",
```
outSH, line 88: Please pin the dependencies version (i.e. `29.5.14`). We use pinned versions to prevent supply chain attacks.

examples/cactus-example-supply-chain-backend/src/main/typescript/infrastructure/supply-chain-app-dummy-infrastructure.ts
```diff
@@ -166,6 +218,30 @@ export class SupplyChainAppDummyInfrastructure {
...
+      this.log.info("Initializing Ethereum connector for Sepolia...");
+      const pluginRegistry = new PluginRegistry();
+      pluginRegistry.add(this.keychain);
+
+      const ethereumConnector = new PluginLedgerConnectorEthereum({
+        instanceId: "PluginLedgerConnectorEthereum_Sepolia",
+        rpcApiHttpHost: SEPOLIA_RPC_ENDPOINT,
+        logLevel: "DEBUG",
+        pluginRegistry,
+      });
```
outSH, line 232: We don't want to force users to have access to sepolia node, through Alchemy or any other provider, to run the test application. Is it possible to make the sample app run on local ethereum network by default? Running on Sepolia could be an optional way of testing the application (must be well documented), but the default should work without much additional setup.

I think you can run geth-all-in-one (tools/docker/geth-all-in-one/README.md) the same way we run besu right now, and deploy and use all the necessary contracts there, or even use besu as sepolia alternative for local run. Please investigate which option suits you better :)

outSH (changes requested):
It all looks great, although I've left some high level comments to be addressed first.

Additionally, please update the PR and commit to match our contributors guide (https://github.com/hyperledger-cacti/cacti/blob/main/CONTRIBUTING.md). In particular:
- Name of the commit and PR should be something like (feat(supply-chain): add wallet authentication (the key thing is to include in paranthesis after feat).
- Squash all commits in this PR into single one.
- Include a signoff in your commit.
- Add description to the commit and PR body.
- (optional) Create an issue where you describe and justify feature shortly, then link to it in PR body by including `Closes: #XXXX` where XXXX is number of your issue.

examples/cactus-example-supply-chain-backend/package.json
outSH: This applies to all package.json affected by this PR

examples/cactus-example-supply-chain-backend/src/main/typescript/infrastructure/supply-chain-app-dummy-infrastructure.ts
Eajunnn: Maybe I can make the sepolia approach optional by checking if they inserted in the process.env? If not it will just go through the original process (Besu and Xdai)? Because now I changed all to Sepolia and Fabric only. Or you guys wanted this to be another directory in examples?

outSH: That works :) For me it's good to keep it single place, unless other maintainers have other opinion on it

RafaelAPB (changes requested):
Thank you for your contribution. I agree with all that outSH said, rest LGTM. 
Please make sure not to add dependencies from example to the root package

RafaelAPB:
@Eajunnn any updates on this PR?

RafaelAPB:
@Eajunnn are you still working on this?

RafaelAPB:
Closing for inactivity

PR #3938, opened 2025-06-04
fix(ccmodel): unsafe shell command constructed from library input

ptrgits:
https://github.com/hyperledger-cacti/cacti/blob/8beaa46d07fa32624ba5f8e1ee340a2959b69d65/packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts#L711-L715

Fix the issue need to sanitize the `name` parameter to ensure that it does not contain any special characters or sequences that could be interpreted as part of a shell command. This can be achieved by:
1. Validating the `name` parameter to allow only safe characters (e.g., alphanumeric characters, underscores, and hyphens).
2. Escaping any special characters in the `name` parameter to prevent command injection.

In this case, we will implement a validation step to ensure that `name` contains only safe characters. If `name` is invalid, we will throw an error to prevent further processing.

---

Dynamically constructing a shell command with inputs from exported functions may inadvertently change the meaning of the shell command. Clients using the exported function may use inputs containing characters that the shell interprets in a special way, for instance quotes and spaces. This can result in the shell command misbehaving, or even allowing a malicious user to execute arbitrary commands on the system.

[shell-quote](https://www.npmjs.com/package/shell-quote)

**Pull Request Requirements**

**Character Limit**

**A Must Read for Beginners**

### DIFF
--- a/packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
+++ b/packages/cactus-plugin-ccmodel-hephaestus/src/main/typescript/plugin-ccmodel-hephaestus.ts
@@ -667,6 +667,9 @@ export class CcModelHephaestus implements ICactusPlugin, IPluginWebService {
 
   public async persistCrossChainLogCsv(name?: string): Promise<string> {
     const startTime = new Date();
+    if (name && !/^[a-zA-Z0-9_-]+$/.test(name)) {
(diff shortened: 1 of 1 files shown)


outSH (approved):
LGTM, thank you for your contribution!

RafaelAPB (approved):

PR #4039, opened 2025-10-20
fix(satp-hermes): logs in oracle operations

edsonCVN:
Introduce logging in the Oracle operations, similar to what we already
have in the SATP operations. Each operation include the following log
entries:

1. Log "init" before starting the operation.
2. Log "exec" immediately before executing the operation.
3. After the operation is executed:
    - Log "done" if it was successful.
    - Log "fail" if it failed.

### DIFF
--- a/packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-services/client/stage0-client-service.ts
+++ b/packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-services/client/stage0-client-service.ts
@@ -248,7 +248,7 @@ export class Stage0ClientService extends SATPService {
         const sessionData = session.getClientSessionData();
 
         await this.dbLogger.persistLogEntry({
-          sessionId: sessionData.id,
+          sessionID: sessionData.id,
           type: messageType,
           operation: "init",
           data: safeStableStringify(sessionData),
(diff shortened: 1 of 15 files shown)


AndreAugusto11 (changes requested):
@edsonCVN could you please add some very simple tests to make sure everything is working as expected?

edsonCVN:
> @edsonCVN, also merge both commits into a single one, update the description of the commit (and the PR) mentioning the linked issue that will be closed once this is merged

Thank you, I will do it

packages/cactus-plugin-satp-hermes/src/test/typescript/unit/oracle-logging.test.ts
```diff
@@ -0,0 +1,276 @@
+import "jest-extended";
```
RafaelAPB, line 1: These tests are insufficient. We need to assert that, running a gateway with an oracle component, logs are sent to the stdout. The current test uses mocks and therefore does not give us the confidence that the current functionality works

RafaelAPB (changes requested):
Please make sure to add robust tests

RafaelAPB:
Closing as no updates were provided

PR #4094, opened 2026-01-21
ci: fix compute packages changes to avoid enobufs by streaming git diff output to disk

LordKubaya:
### Problem

CI was failing with:

spawnSync /bin/sh ENOBUFS

The failure was caused by `tools/compute-affected-packages.cjs` using `execSync()` to capture the full output of `git diff` into memory.
For large diffs, this exceeds Node’s internal buffer limits and causes the process to be terminated.

### Solution

- Redirect `git diff` output to temporary files instead of buffering stdout in Node.
- Read diff contents from disk only when needed.
- Keep existing affected-package detection logic unchanged.

This eliminates the ENOBUFS failure mode and makes the script safe for large PRs and monorepo-scale diffs.

### Impact

- CI stability improvement
- No behavioral changes to package detection
- No performance regression (disk I/O replaces large in-memory buffers)

### DIFF
--- a/tools/compute-affected-packages.cjs
+++ b/tools/compute-affected-packages.cjs
@@ -9,6 +9,7 @@
  * 4. Finds directly changed packages and recursively adds their dependents.
  * 5. Outputs a JSON array of affected package directories (for CI matrix).
  */
+const os = require("os");
 
 const fs = require("fs");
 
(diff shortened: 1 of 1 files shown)


AndreAugusto11 (approved):
LGTM

PR #4166, opened 2026-03-31
fix(fabric-connector): close gateway and channel on block subscription failure

mn-ram:
## Summary

- Add gateway.disconnect() in WatchBlocksV1Endpoint.subscribe() error path to prevent gRPC connection leak on subscription failure
- Add channel.close() and channel.client.close() in SubscribeDelegatedSign error path to prevent connection leak

## Problem

When getNetwork() or addBlockListener() fails during block subscription, the gateway is never disconnected. Similarly, when eventService.send() fails during delegated-sign subscription, the channel and client connections are never closed.

## Test plan

- [ ] Deploy with a Fabric peer, subscribe to block events via Socket.IO
- [ ] Stop the peer to trigger subscription failures
- [ ] Monitor file descriptors to confirm no connection growth during repeated failures

### DIFF
--- a/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
+++ b/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
@@ -222,7 +222,6 @@ export class ApiServer {
         this.shutdown()
           .catch((ex: unknown) => {
             this.log.warn("Failed async-exit-hook for cmd-api-server", ex);
-            throw ex;
           })
           .finally(() => {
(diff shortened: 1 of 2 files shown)


mn-ram:
Closing this in favor of #4187, which covers all 5 gateway-using API methods (getBlock, getChainInfo, getTransactionReceiptByTxID, getDiscoveryResults, transact) with try/finally blocks — not just the subscription paths. That PR is a more comprehensive fix for the same gateway/gRPC connection leak issue.

PR #4191, opened 2026-04-06
fix(cmd-api-server): await gRPC plugin registration before server bind

### LINKED ISSUE #4190: bug: gRPC server starts before plugin services are registered due to async forEach race condition
## Summary

The `startGrpcServer()` method in `packages/cactus-cmd-api-server/src/main/typescript/api-server.ts` (line 832) uses `Array.forEach()` with an `async` callback to register gRPC services from plugins. Since `forEach` does not await async callbacks, the `grpcServer.bindAsync()` call on line 856 executes **immediately** — before any plugin service has been registered.

This means the gRPC server starts and accepts connections with **zero plugin services available**. All plugin gRPC calls return `UNIMPLEMENTED`.

## Affected Code

```typescript
// Line 832 — forEach does NOT await async callbacks
pluginRegistry.getPlugins().forEach(async (x: ICactusPlugin) => {
  if (!isIPluginGrpcService(x)) { return; }
  const svcPairs = await x.createGrpcSvcDefAndImplPairs(opts);
  svcPairs.forEach(({ definition, implementation }) => {
    this.grpcServer.addService(definition, implementation);
  });
});

// Line 854 — runs IMMEDIATELY, before async forEach callbacks resolve
log.debug("Installed all IPluginGrpcService instances OK");

// Line 856 — server binds with no plugin services registered
this.grpcServer.bindAsync(grpcHostAndPort, grpcTlsCredentials, callback);
```

## Impact

- **Every gRPC-enabled plugin is silently unavailable** after server startup.
- Currently affects `PluginLedgerConnectorBesu`, which registers both its OpenAPI gRPC service and its block-streaming gRPC service via `createGrpcSvcDefAndImplPairs()`.
- No error is thrown or logged — the server reports "Installed all IPluginGrpcService instances OK" before the async work has even started.
- REST endpoints on the same server work fine, so REST-based health checks pass while gRPC is completely broken.
- gRPC clients receive opaque `UNIMPLEMENTED` status codes with no useful diagnostics.
- Any downstream system relying on `WatchBlocksV1` gRPC streaming (e.g., persistence plugins, dashboards) silently receives nothing.

## Steps to Reproduce

1. Start the Cacti API server with the Besu connector plugin configured (Docker Compose or standalone).
2. Wait for the server to log that it's listening on the gRPC port.
3. Call any Besu gRPC service:
   ```bash
   grpcurl -plaintext localhost:4000 \
     org.hyperledger.cactus.plugin.ledger.connector.besu.services.besuservice.BesuGrpcSvcOpenApi/GetBlockV1
   ```
4. **Expected:** Block data returned.  
   **Actual:** `UNIMPLEMENTED` error.
5. Check server logs — "Installed all IPluginGrpcService instances OK" appears **before** "Creating gRPC service of: @hyperledger/cactus-plugin-ledger-connector-besu".

This is **deterministic** — happens on every startup, not a timing-dependent race.

## Root Cause

`Array.prototype.forEach()` ignores the Promise returned by an `async` callback. Each `await` inside the callback only suspends that callback's own microtask — it does not block the outer function. The code after `forEach` runs synchronously and immediately.

## Suggested Fix

Replace `forEach` with a `for...of` loop that properly awaits each plugin registration before binding the server. Since the code is inside a `new Promise()` constructor, wrap in an async IIFE:

```typescript
(async () => {
  for (const x of pluginRegistry.getPlugins()) {
    if (!isIPluginGrpcService(x)) {
      this.log.debug("%s skipping %s instance", fnTag, x.getPackageName());
      continue;
    }
    const opts = { logLevel };
    log.info("%s Creating gRPC service of: %s", fnTag, x.getPackageName());

    const svcPairs = await x.createGrpcSvcDefAndImplPairs(opts);
    log.debug("%s Obtained %o gRPC svc pairs OK", fnTag, svcPairs.length);

    for (const { definition, implementation } of svcPairs) {
      this.grpcServer.addService(definition, implementation);
    }
    log.info("%s Added gRPC service of: %s OK", fnTag, x.getPackageName());
  }

  log.debug("%s Installed all IPluginGrpcService instances OK", fnTag);
  this.grpcServer.bindAsync(grpcHostAndPort, grpcTlsCredentials, callback);
})().catch(reject);
```

This ensures all plugin services are registered before the server starts accepting connections, and errors during registration properly reject the promise.

## Environment

- Package: `@hyperledger/cactus-cmd-api-server`
- File: `src/main/typescript/api-server.ts`, lines 832–875
- Affects any deployment using gRPC-enabled plugins (currently Besu connector)

mn-ram:
● Summary

* startGrpcServer() used Array.forEach() with an async callback to register plugin gRPC services
* forEach does not await async callbacks, so bindAsync() ran before any plugin services were registered
* All plugin gRPC endpoints (e.g., Besu connector) returned UNIMPLEMENTED on every startup
* Replaced forEach with for...of loop to properly await registration before binding

Fixes #4190

### DIFF
--- a/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
+++ b/packages/cactus-cmd-api-server/src/main/typescript/api-server.ts
@@ -829,49 +829,67 @@ export class ApiServer {
       );
 
       log.debug("Installing gRPC services of IPluginGrpcService instances...");
-      pluginRegistry.getPlugins().forEach(async (x: ICactusPlugin) => {
-        if (!isIPluginGrpcService(x)) {
-          this.log.debug("%s skipping %s instance", fnTag, x.getPackageName());
-          return;
-        }
-        const opts = { logLevel };
-        log.info("%s Creating gRPC service of: %s", fnTag, x.getPackageName());
+      (async () => {
+        for (const x of pluginRegistry.getPlugins()) {
+          if (!isIPluginGrpcService(x)) {
+            this.log.debug(
+              "%s skipping %s instance",
+              fnTag,
+              x.getPackageName(),
+            );
+            continue;
+          }
+          const opts = { logLevel };
+          log.info(
+            "%s Creating gRPC service of: %s",
+            fnTag,
+            x.getPackageName(),
+          );
 
-        const svcPairs = await x.createGrpcSvcDefAndImplPairs(opts);
-        log.debug("%s Obtained %o gRPC svc pairs OK", fnTag, svcPairs.length);
+          const svcPairs = await x.createGrpcSvcDefAndImplPairs(opts);
+          log.debug("%s Obtained %o gRPC svc pairs OK", fnTag, svcPairs.length);
 
-        svcPairs.forEach(({ definition, implementation }) => {
-          const svcNames = Object.values(definition).map((x) => x.originalName);
-          const svcPaths = Object.values(definition).map((x) => x.path);
-          log.debug("%s Adding gRPC svc names %o ...", fnTag, svcNames);
-          log.debug("%s Adding gRPC svc paths %o ...", fnTag, svcPaths);
-          this.grpcServer.addService(definition, implementation);
-          log.debug("%s Added gRPC svc OK ...", fnTag);
-        });
-
-        log.info("%s Added gRPC service of: %s OK", fnTag, x.getPackageName());
-      });
-      log.debug("%s Installed all IPluginGrpcService instances OK", fnTag);
-
-      this.grpcServer.bindAsync(
-        grpcHostAndPort,
-        grpcTlsCredentials,
-        (error: Error | null, port: number) => {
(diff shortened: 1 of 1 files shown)


sandeepnRES:
Please squash the commits into single commit

mn-ram:
@sandeepnRES Done  Squashed all commits into one. Let me know if anything else is needed.

RafaelAPB:
@mn-ram thank you for your contributions. Do you think it makes sense to decompose the solution you provided into two functions, one to obtain the services, and another to do the bind?

mn-ram:
Thanks, @RafaelAPB  - after reviewing again, I found that this PR duplicates #4163, which already fixes the same gRPC registration bug in a cleaner way using `Promise.all(map)`, so we are closing this PR and keeping #4163 to avoid merge conflicts.

PR #4205, opened 2026-04-07
fix(connector-sawtooth): pass through all txs when no family_name filter

### LINKED ISSUE #4204: bug(connector-sawtooth): WatchBlocksV1Endpoint CactiTransactions filter silently drops every transaction when no family_name filter is supplied
## Summary

`WatchBlocksV1Endpoint.sendCactiTransactions()` in the Sawtooth connector contains a `filter()` predicate with no default return. When the (optional) `txFilterBy.family_name` is not configured — i.e. the default usage of `WatchBlocksV1ListenerType.CactiTransactions` — the predicate falls off the end of the function and returns `undefined`, which `Array.prototype.filter` treats as falsy. As a result **every transaction in every block is dropped** and the client receives nothing, indefinitely, with no error and no log.

## Affected file

`packages/cactus-plugin-ledger-connector-sawtooth/src/main/typescript/web-services/watch-blocks-v1-endpoint.ts`

## Offending code

```ts
private async sendCactiTransactions(
  block: Block,
  txFilterBy?: WatchBlocksV1TransactionFilter,
) {
  const cactiTransactionsEvents = block.batches
    .flatMap((block) => block.transactions)
    .filter((tx) => {
      if (txFilterBy && txFilterBy.family_name) {
        return tx.header.family_name === txFilterBy.family_name;
      }
      // ← no else / no default return → returns undefined → falsy
    })
    .map((tx) => {
      return {
        ...tx,
        payload_decoded: cbor.decodeAllSync(
          Buffer.from(tx.payload, "base64"),
        ),
      };
    });

  if (cactiTransactionsEvents.length > 0) {
    this.socket.emit(WatchBlocksV1.Next, {
      cactiTransactionsEvents,
    });
  }
}
```

## Why this is broken

1. `txFilterBy` is **optional** in the OpenAPI spec (`WatchBlocksV1Options.txFilterBy?: WatchBlocksV1TransactionFilter`). Subscribing with `type: CactiTransactions` and no filter is the documented "give me all Cacti transactions" path.
2. With `txFilterBy` (or `txFilterBy.family_name`) undefined, the `if` body is skipped and the arrow function returns `undefined`. `filter(() => undefined)` removes every element.
3. Because `cactiTransactionsEvents.length` is then always `0`, the guarded `socket.emit(WatchBlocksV1.Next, …)` is never reached. The client receives **zero** events.

## Why it is silent

- No exception is thrown.
- The Sawtooth REST poll (`blocksGet`) succeeds normally every 5s.
- `currentBlockHeight` advances correctly.
- The Socket.IO connection stays open.
- Nothing is logged at WARN/ERROR.

From the operator's perspective the connector \"works\". From the consumer's perspective the chain looks frozen.

## Reproduction

1. Start the Sawtooth connector with the standard config.
2. Open a Socket.IO client and emit `WatchBlocksV1.Subscribe` with:
   ```json
   { \"type\": \"CactiTransactions\" }
   ```
   (no `txFilterBy`).
3. Submit any Sawtooth transaction in any transaction family.
4. Observe: blocks are produced, the connector polls them, but the client never receives a `WatchBlocksV1.Next` event with `cactiTransactionsEvents`.

If the same client instead subscribes with `type: Full`, blocks come through fine — confirming the bug is isolated to the `CactiTransactions` filter path.

## Impact

- **Severity: high / silent correctness.** Any downstream consumer (indexer, oracle, SATP-Hermes pipeline, dashboard, alerting) that relies on `CactiTransactions` without an explicit `family_name` filter is effectively offline against the Sawtooth ledger and has no signal that anything is wrong.
- The bug affects the **default** and most common configuration of the high-level subscription type, not an edge case.
- TypeScript would normally catch this with `noImplicitReturns` — its presence implies that strict-return checking is not enabled for this package, so similar latent bugs may exist nearby.

## Proposed fix

Restore the missing default-true branch so that when no filter is configured, all transactions pass through:

```ts
.filter((tx) => {
  if (txFilterBy?.family_name) {
    return tx.header.family_name === txFilterBy.family_name;
  }
  return true;
})
```

I will follow up with a PR.

mn-ram:
## Summary

Fixes #4204.

`WatchBlocksV1Endpoint.sendCactiTransactions()` in the Sawtooth connector had a `filter()` predicate with no default return. When the optional `txFilterBy.family_name` was not configured — i.e. the default usage of `WatchBlocksV1ListenerType.CactiTransactions` — the arrow function fell off the end and returned `undefined`, which `Array.prototype.filter` treats as falsy. Every transaction was silently dropped, the guarded `socket.emit(WatchBlocksV1.Next, …)` was never reached, and clients received nothing forever with no error and no log.

This PR restores the missing default-true branch so that when no filter is configured all transactions pass through.

## Change

`packages/cactus-plugin-ledger-connector-sawtooth/src/main/typescript/web-services/watch-blocks-v1-endpoint.ts`

```ts
.filter((tx) => {
  if (txFilterBy && txFilterBy.family_name) {
    return tx.header.family_name === txFilterBy.family_name;
  }
  return true;
})
```

One-line fix, no behavior change for the filtered path.

## Test plan

- [ ] Subscribe with `{ type: "CactiTransactions" }` (no `txFilterBy`) and confirm `WatchBlocksV1.Next` events arrive with `cactiTransactionsEvents` populated.
- [ ] Subscribe with `{ type: "CactiTransactions", txFilterBy: { family_name: "intkey" } }` and confirm only matching-family transactions are emitted.
- [ ] Subscribe with `{ type: "Full" }` and confirm full-block path is unaffected.

### DIFF
--- a/packages/cactus-plugin-ledger-connector-sawtooth/src/main/typescript/web-services/watch-blocks-v1-endpoint.ts
+++ b/packages/cactus-plugin-ledger-connector-sawtooth/src/main/typescript/web-services/watch-blocks-v1-endpoint.ts
@@ -83,6 +83,7 @@ export class WatchBlocksV1Endpoint {
         if (txFilterBy && txFilterBy.family_name) {
           return tx.header.family_name === txFilterBy.family_name;
         }
+        return true;
       })
       .map((tx) => {
         return {


RafaelAPB:
Closing as this connector is archived

PR #4207, opened 2026-04-07
fix(plugin-keychain-vault): return boolean from has() to stop leaking secret value

### LINKED ISSUE #4206: bug(plugin-keychain-vault): has() returns full Vault read response instead of boolean, leaking secret values through has-keychain-entry endpoint
## Summary

`PluginKeychainVault.has()` is declared as `Promise<boolean>` but actually returns the entire `node-vault` `read()` response object (which contains the secret value under `data.data.value`). The HTTP `has-keychain-entry` endpoint puts that object straight into the `isPresent` field of its JSON response. As a result, **anyone authorized to call the presence-check endpoint can read the plaintext value of any key**, without ever calling `get-keychain-entry`. This is a confidentiality bug in a security-critical plugin.

## Affected files

- `packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts`
- `packages/cactus-plugin-keychain-vault/src/main/typescript/web-services/has-keychain-entry-endpoint-v1.ts`

## Offending code

`plugin-keychain-vault.ts` (lines 247-266):

```ts
async has(key: string): Promise<boolean> {
  const path = this.pathFor(key);
  try {
    const res = await this.backend.read(path);
    return res;          // ← node-vault read() response object, NOT a boolean
  } catch (ex) {
    if (ex?.response?.statusCode === HttpStatus.NOT_FOUND) {
      return false;
    } else {
      this.log.error(`Presence check of \"${key}\" crashed:`, ex);
      throw ex;
    }
  }
}
```

`has-keychain-entry-endpoint-v1.ts` (lines 92-108):

```ts
const isPresent = await this.plugin.has(req.body.key);
const resBody: HasKeychainEntryResponseV1 = {
  isPresent,                   // ← typed boolean, actually a Vault response
  checkedAt: new Date().toJSON(),
  key: req.body.key,
};
res.status(200);
res.json(resBody);             // ← serialized to the wire as-is
```

## What the wire response actually looks like

For an existing key, the endpoint returns something like:

```json
{
  \"isPresent\": {
    \"request_id\": \"…\",
    \"lease_id\": \"\",
    \"renewable\": false,
    \"data\": {
      \"data\": { \"value\": \"PLAINTEXT_SECRET_HERE\" },
      \"metadata\": { … }
    },
    \"lease_duration\": 0,
    \"wrap_info\": null,
    \"warnings\": null,
    \"auth\": null
  },
  \"checkedAt\": \"...\",
  \"key\": \"...\"
}
```

instead of the documented `{ \"isPresent\": true, … }`.

## Reproduction

1. Stand up the Vault keychain plugin against any Vault dev server.
2. Store a secret:
   ```
   POST /api/v1/plugins/@hyperledger/cactus-plugin-keychain-vault/set-keychain-entry
   { \"key\": \"my-key\", \"value\": \"super-secret-value\" }
   ```
3. Call the presence-check endpoint:
   ```
   POST /api/v1/plugins/@hyperledger/cactus-plugin-keychain-vault/has-keychain-entry
   { \"key\": \"my-key\" }
   ```
4. Observe: the response body's `isPresent` field is the full Vault read object containing `data.data.value === \"super-secret-value\"`. The plaintext secret has been returned through an endpoint whose entire purpose is supposed to be a yes/no presence check.

## Impact

- **Severity: high (confidentiality / information disclosure).** The plugin under discussion is the trust anchor for credential material used by the connectors above it (Fabric/Besu/Ethereum signing keys, SATP-Hermes oracle keys, etc.). A leak here is a credential-exfiltration primitive against every higher layer that stores its private keys in this keychain.
- **RBAC bypass.** Any deployment that hardens `get-keychain-entry` (treating reads as more sensitive than presence checks) and leaves `has-keychain-entry` more broadly accessible — the obvious reading of the API surface — has just published its secrets to every caller of the presence endpoint.
- **Audit gap.** SIEM/audit pipelines that count secret reads by counting `get-keychain-entry` calls under-report by design: every `has-keychain-entry` call is also a covert read of the same data.
- **OpenAPI contract violation.** `HasKeychainEntryResponseV1.isPresent` is typed `boolean` in the generated TypeScript/Java/Go/Rust clients. JS clients doing `if (resp.isPresent)` happen to work because objects are truthy, but `=== true` checks, schema validators, and strongly-typed clients all break.
- **Default code path.** This fires on every successful `has()` call against an existing key — not an edge case. The only correct branch today is the `NOT_FOUND → false` negative path.

## Proposed fix

Discard the response object entirely; presence is determined purely by whether `read()` resolved or rejected with `404`. No secret material ever escapes the `has()` boundary.

```ts
async has(key: string): Promise<boolean> {
  const path = this.pathFor(key);
  try {
    await this.backend.read(path);
    return true;
  } catch (ex) {
    if (ex?.response?.statusCode === HttpStatus.NOT_FOUND) {
      return false;
    }
    this.log.error(`Presence check of \"${key}\" crashed:`, ex);
    throw ex;
  }
}
```

I will follow up with a PR.

mn-ram:
## Summary

Fixes #4206.

`PluginKeychainVault.has()` was declared as `Promise<boolean>` but actually returned the entire `node-vault` `read()` response object (which contains the secret value under `data.data.value`). The HTTP `has-keychain-entry` endpoint then put that object straight into the `isPresent` field of its JSON response, so any caller of the presence-check endpoint received the plaintext secret of the queried key — without ever calling `get-keychain-entry`.

This PR makes `has()` discard the response and return a real boolean. Presence is now determined purely by whether `read()` resolved or rejected with `404`. No secret material escapes the `has()` boundary.

## Change

`packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts`

```ts
async has(key: string): Promise<boolean> {
  const path = this.pathFor(key);
  try {
    await this.backend.read(path);
    return true;
  } catch (ex) {
    if (ex?.response?.statusCode === HttpStatus.NOT_FOUND) {
      return false;
    }
    this.log.error(`Presence check of "${key}" crashed:`, ex);
    throw ex;
  }
}
```

Two-line behavioural change. The `404 → false` and rethrow branches are unchanged.

## Why this matters

- **Confidentiality / information disclosure** in a security-critical plugin. The keychain is the trust anchor for credential material used by the connectors above it (Fabric/Besu/Ethereum signing keys, SATP-Hermes oracle keys, etc.).
- **RBAC bypass.** Deployments that hardened `get-keychain-entry` and left `has-keychain-entry` more broadly accessible were silently exposing every secret to every caller of the presence endpoint.
- **Audit gap.** SIEM rules counting secret reads via `get-keychain-entry` under-reported by design — every `has-keychain-entry` call was a covert read of the same data.
- **OpenAPI contract violation.** `HasKeychainEntryResponseV1.isPresent` is typed `boolean` in the generated clients. JS clients doing `if (resp.isPresent)` happened to work because objects are truthy, but `=== true` checks, schema validators, and strongly-typed clients (Java/Go/Rust) were all broken.
- **Default code path.** This fired on every successful `has()` call against an existing key — not an edge case.

## Test plan

- [ ] Store a secret via `set-keychain-entry`, then call `has-keychain-entry` for the same key. Confirm the response body is `{ \"isPresent\": true, \"checkedAt\": \"…\", \"key\": \"…\" }` and that **no** Vault response object / secret value appears anywhere in the body.
- [ ] Call `has-keychain-entry` for a key that does not exist. Confirm `{ \"isPresent\": false, … }`.
- [ ] Trigger an unrelated Vault failure (e.g. wrong token) and confirm the endpoint surfaces the error rather than masking it.
- [ ] Run the existing keychain-vault test suite and confirm it still passes (and ideally extend it with a `has()` regression test that asserts the return type is a primitive boolean).

### DIFF
--- a/packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
+++ b/packages/cactus-plugin-keychain-vault/src/main/typescript/plugin-keychain-vault.ts
@@ -247,8 +247,8 @@ export class PluginKeychainVault implements IPluginWebService, IPluginKeychain {
   async has(key: string): Promise<boolean> {
     const path = this.pathFor(key);
     try {
-      const res = await this.backend.read(path);
-      return res;
+      await this.backend.read(path);
+      return true;
     } catch (ex) {
       // We have to make sure that the exception is either an expected
       // or an unexpected one where the expected exception is what we


RafaelAPB (approved):

mn-ram:
@LordKubaya  Could you please take a look and share your review on this PR when you get a chance? Thanks!

RafaelAPB:
Closing as this plugin is archived

PR #4232, opened 2026-04-13
fix(connector-corda): prevent re-entrant poll in watchBlocksAsyncV1

### LINKED ISSUE #4231: bug(connector-corda): re-entrant poll in watchBlocksAsyncV1 delivers duplicate transactions to subscribers
## Summary

`CordaApiClient.watchBlocksAsyncV1` uses `setInterval` to periodically call `pollTransactionsLogin`, but the callback does **not** await the async function. When the Corda REST API response takes longer than `pollRate` (default 5 s), two poll iterations run concurrently against the same uncleared monitoring queue, causing every subscriber to receive each transaction **twice**.

## Affected file

`packages/cactus-plugin-ledger-connector-corda/src/main/typescript/api-client/corda-api-client.ts`

## Root cause

```typescript
// lines 216-224
const monitoringInterval = setInterval(
  () =>
    this.pollTransactionsLogin(   // async — Promise silently dropped
      subject,
      options.clientAppId,
      options.stateFullClassName,
    ),
  pollRate,
);
```

`pollTransactionsLogin` is `async` and performs two sequential HTTP calls:

1. `getMonitorTransactionsV1` — reads pending transactions
2. `clearMonitorTransactionsV1` — marks them as read
3. `subject.next(tx)` — pushes each transaction to subscribers

When two concurrent invocations overlap (because the first hasn't finished before the next tick fires), both calls see the same unacknowledged transactions, both clear them, and both call `subject.next` — resulting in duplicate delivery.

## How to reproduce

1. Configure the Corda REST API to respond slowly (> `pollRate`, e.g. 6 s latency via a proxy or `tc netem`)
2. Call `watchBlocksAsyncV1({ stateFullClassName: "...", clientAppId: "test", pollRate: 5000 })`
3. Submit one Corda transaction on the monitored state
4. Observe: the RxJS subscriber receives the same transaction **2 or more times** within a single poll cycle

Deterministic reproduction: set `pollRate: 100` and patch `getMonitorTransactionsV1` to add a 200 ms delay — duplicate emission is guaranteed.

## Impact

- Any consumer of `watchBlocksAsyncV1` (including `cactus-verifier-client`) processes the same state-change event twice under load
- Silent failure — no error is thrown, no warning logged
- Worsens under exactly the conditions where correctness matters most (slow or loaded Corda node)

## Proposed fix

Replace `setInterval` with a self-rescheduling `setTimeout` chain that awaits each poll before scheduling the next, identical in pattern to the fix already applied to the Ethereum connector's `WatchBlocksV1HttpPollEndpoint`:

```typescript
let cancelled = false;

const schedulePoll = () => {
  if (cancelled) return;
  setTimeout(async () => {
    try {
      await this.pollTransactionsLogin(
        subject,
        options.clientAppId,
        options.stateFullClassName,
      );
    } catch (err) {
      this.log.warn("Unexpected error in poll loop:", err);
    }
    schedulePoll();
  }, pollRate);
};

schedulePoll();
```

In `finalizeMonitoring`, set `cancelled = true` instead of calling `clearInterval`.

This ensures the next poll only starts after the current one completes, eliminating the overlap window entirely.

mn-ram:
## Summary

Fixes #4231

`watchBlocksAsyncV1` used `setInterval` to call `pollTransactionsLogin` (an `async` function) without `await`, silently dropping the returned Promise. When the Corda REST API response took longer than `pollRate` (default 5 s), two concurrent poll iterations ran against the same uncleared monitoring queue — causing every subscriber to receive each transaction twice.

## Root cause

```typescript
// before: Promise silently dropped, re-entrancy possible
const monitoringInterval = setInterval(
  () => this.pollTransactionsLogin(subject, clientAppId, stateName),
  pollRate,
);
```

## Fix

Replace `setInterval` with a self-rescheduling `setTimeout` chain. Each iteration awaits `pollTransactionsLogin` before scheduling the next one, so two poll calls can never overlap:

```typescript
// after: next poll only starts after current one completes
let cancelled = false;

const schedulePoll = () => {
  if (cancelled) return;
  setTimeout(async () => {
    try {
      await this.pollTransactionsLogin(subject, clientAppId, stateName);
    } catch (err) {
      this.log.warn("Unexpected error in poll scheduler:", err);
    }
    schedulePoll();
  }, pollRate);
};

schedulePoll();
```

A `cancelled` flag (set by `finalizeMonitoring` on unsubscribe) stops the loop cleanly without needing a timer handle. `finalizeMonitoring` now accepts a `cancelPoll: () => void` callback instead of a timer id.

## Testing

The existing integration test `monitor-transactions-v4.8.test.ts` covers the happy path. The duplicate-delivery scenario is reproducible by setting `pollRate: 100` and patching `getMonitorTransactionsV1` with a 200 ms artificial delay — after this fix the subscriber receives each transaction exactly once.

### DIFF
--- a/packages/cactus-plugin-ledger-connector-corda/src/main/typescript/api-client/corda-api-client.ts
+++ b/packages/cactus-plugin-ledger-connector-corda/src/main/typescript/api-client/corda-api-client.ts
@@ -147,18 +147,18 @@ export class CordaApiClient
    * Should be called to stop monitoring on the connector.
    * Calling this will remove all pending transactions (that were not read yet)!
    *
-   * @param monitor Monitoring interval set with `setTimeout`.
+   * @param cancelPoll Function that stops the self-rescheduling poll loop.
    * @param clientAppId Client application ID to identify monitoring queue on the connector.
    * @param stateName Corda state to monitor.
    */
   private finalizeMonitoring(
-    monitor: ReturnType<typeof setTimeout>,
+    cancelPoll: () => void,
     clientAppId: string,
     stateName: string,
   ) {
     this.log.info("Unsubscribe from the monitoring...");
 
-    clearInterval(monitor);
+    cancelPoll();
 
     this.stopMonitorV1({
       clientAppId: clientAppId,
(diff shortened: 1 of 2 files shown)


RafaelAPB:
@mn-ram are you still working on this?

RafaelAPB:
closed for inactivity

PR #4245, opened 2026-04-15
fix(besu): bind aio rpc endpoints to docker network

### LINKED ISSUE #4065: bug(besu-AIO): Cannot communicate with Besu when using Docker networks
## Describe the bug
When deploying a Hyperledger Cactus Besu AIO environment using Docker, the Besu server is configured to listen on `127.0.0.1`.  
This causes communication failures because other containers cannot access services bound to localhost inside another container.  
The expected configuration should bind Besu to `0.0.0.0` so that it becomes reachable within the Docker network.

---

## To Reproduce
Steps to reproduce the behavior on a successfully deployed Hyperledger Cactus cluster:

1. Deploy the Besu AIO container using Docker Compose or similar.
2. Attempt to connect to the Besu RPC endpoint from another container.
3. Requests fail because Besu is only listening on `127.0.0.1`.

---

## Expected behavior
Besu should bind to `0.0.0.0` so that RPC and WebSocket endpoints are accessible to other containers in the same Docker network.  
This would allow plugins, connectors, and validators to communicate with Besu normally.

malsomesh9:
## Summary

- Set the Besu all-in-one Docker image HTTP RPC host to `0.0.0.0`.
- Set the WebSocket RPC host to `0.0.0.0` so other containers can reach Besu over Docker networks.

## Root Cause

Besu enables HTTP and WebSocket RPC in the AIO image, but without explicit host overrides Besu binds those endpoints to loopback. That makes the endpoints unreachable from other containers in the same Docker network.

## Impact

Plugins, connectors, and validators running in sibling containers can connect to the Besu AIO RPC and WebSocket endpoints via the compose service name.

Fixes #4065

## Validation

- `git diff --check -- tools/docker/besu-all-in-one/Dockerfile`
- Verified the resulting Dockerfile contains `BESU_RPC_HTTP_HOST="0.0.0.0"` and `BESU_RPC_WS_HOST="0.0.0.0"`.


.github/workflows/dependent-issues.yaml
```diff
@@ -6,7 +6,7 @@ on:
       - opened
       - edited
       - reopened
-  pull_request:
+  pull_request_target:
```
VRamakrishna, line 9: Why do we need this change exactly? I believe this will reduce security.

RafaelAPB: Agreed - even if we are to do that change, it is out of the scope of this PR. Please remove @malsomesh9

malsomesh9:
Removed the workflow-trigger change from this PR to keep the scope limited to the Besu AIO networking fix. I’ll handle any dependent-issues workflow changes separately if still needed.

malsomesh9:
The intent here is only to make Besu RPC reachable from other containers on the same Docker network, which is required for the AIO use case in #4065. This does not publish ports by itself, but I agree exposure becomes broader if users also publish those ports on the host, so I can add a short note clarifying that risk.

RafaelAPB:
@malsomesh9 can you please address the comments suggested by the reviewers and copilot?

RafaelAPB:
Closing due to inactivity

PR #4247, opened 2026-04-16
test(satp-hermes): retry fabric endorsement failures

### LINKED ISSUE #3978: test(satp-hermes): flaky tests caused by ENDORSEMENT_POLICY_FAILURE
From time to time, the CI fails with the error ENDORSEMENT_POLICY_FAILURE from a fabric transaction.

What can cause this?
- Endorsement policy mismatch: The chaincode requires endorsements from certain peers/orgs, but the proposal isn’t endorsed by all required peers. Example: Policy requires AND('Org1.peer','Org2.peer'), but only Org1 endorsed.

- Peer/container performance issues: Under heavier load, peers or orderers may be slower to respond. Endorsements may timeout, leaving the transaction short of required signatures.

- Chaincode not instantiated everywhere: If the chaincode isn’t instantiated on all peers required by the policy, some endorsements fail.

- Ledger state mismatch: If peers have diverged state (due to missed blocks, crashes, or DB issues), they may reject the proposal differently.

- Transient network issues: In test setups (especially on Docker), slower responses or dropped requests can cause intermittent failures.

- Non-deterministic chaincode: When a peer simulates a transaction proposal and sends a peer response, it has to be the same as other peer responses. So using a random number in the chaincode logic will most likely result in different peer responses, which consequently lands you with an "ENDORSEMENT POLICY FAILURE".

Image

It happens with more frequency in tests with more load, more containers running... So we can conclude that the problem here is about performance.

So, things to try to make the test more resilient:
- Increase the fabric timeout
- Make an X retry for the fabric transactions

malsomesh9:
## Summary

- Add a narrow retry wrapper for SATP Hermes Fabric test transactions.
- Retry only transient `ENDORSEMENT_POLICY_FAILURE` responses, with three attempts and a short delay.
- Route the shared Fabric test-environment transaction helpers through the retry wrapper so flaky CI failures are handled in one place.

## Notes

The retry is intentionally limited to the Fabric endorsement-policy failure described in #3978. Other transaction errors still fail immediately.

Fixes #3978

## Validation

- Verified the edited helper routes direct `connector.transact()` calls through `transactWithFabricRetry()`.
- Could not run the SATP Hermes TypeScript/Jest tests locally because this checkout does not have `node_modules` installed.

### DIFF
--- a/packages/cactus-plugin-satp-hermes/src/test/typescript/environments/fabric-test-environment.ts
+++ b/packages/cactus-plugin-satp-hermes/src/test/typescript/environments/fabric-test-environment.ts
@@ -57,6 +57,11 @@ export interface IFabricTestEnvironment {
   claimFormat?: ClaimFormat;
   network?: string;
 }
+
+const FABRIC_TRANSACTION_RETRY_ATTEMPTS = 3;
+const FABRIC_TRANSACTION_RETRY_DELAY_MS = 1000;
+const FABRIC_ENDORSEMENT_POLICY_FAILURE = "ENDORSEMENT_POLICY_FAILURE";
+
 export class FabricTestEnvironment {
   public static readonly FABRIC_ASSET_ID: string = "FabricExampleAsset";
   public static readonly FABRIC_REFERENCE_ID: string = ExampleOntology.id;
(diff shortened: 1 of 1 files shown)


RafaelAPB:
@malsomesh9 thank you for your contribution. Unfortunately, we verified that simply adding a retrying does not address the root cause of the problem specified in #3978 . The problem appears intermittently. We have a suspicion that the Fabric AIO is the culprit, but a longer-term solution would be the create a new Fabric 3 connector and respective test tooling cc @LordKubaya  @VRamakrishna

RafaelAPB:
Closing for now as it remains unadressed

PR #4255, opened 2026-04-16
[codex] fix node engine range validation

### LINKED ISSUE #3963: ci(custom-checks): node engine validation should not be hard coded
**Describe the bug**

This is about the script `tools/custom-checks/check-package-json-fields.ts`, which validates lots of fields in `package.json`, one of them is `engine`. The current behaviour is not correct and should be fixed as per expected behaviour below.
This check is commented in PR #3962 (Line 60-65).
But once the expected behaviour is implemented, this should be uncommented.

**To Reproduce**

Change `engine` field in any package.json, to not equal `>=18`. And the this check will fail.

**Expected behavior**

This check should understand the pattern and verify if the required version is greater than our desired number or not, instead of fixed hardcoded string match to `>=18`.

For e.g. If check says version should be >=18,
then all the below patterns should pass as well:
* `>=20`
* `>21`
* `=23`

and following should fail:
* `>=17`

Other (weaker) acceptable solution is to have list of supported node engine versions, and any match in that list should be valid.

malsomesh9:
## Summary

- Re-enable `package.json` `engines` validation in `tools/custom-checks/check-package-json-fields.ts`.
- Replace the hard-coded exact match with a custom major-version range validator.
- Allow acceptable stronger ranges such as `>=20`, `>21`, and `=23` while rejecting weaker ranges such as `>=17`.
- Add a GitHub-hosted proof screenshot to the PR branch.

Fixes #3963

## Validation

- `node --check tools/custom-checks/check-package-json-fields.ts`
- Engine validation matrix:
  - `>=18`, `>=20`, `>21`, `=23` pass for Node minimum 18
  - `>=17`, `=17`, `>16`, `^18` fail for Node minimum 18

## Proof Screenshot

Issue 3963 validation proof

## Note

I could not run the full custom-check entrypoint in the local checkout because `node_modules` are not installed and a full monorepo install previously ran out of disk space. The edited TypeScript syntax and the new engine-range logic were validated directly.


RafaelAPB:
Closing as there is no approved issue for this

PR #4286, opened 2026-04-24
fix(verifier): stopMonitor throws for connectors using watchBlocksAsyncV1

### LINKED ISSUE #4250: bug(verifier): stopMonitor fails for connectors that only implement watchBlocksAsyncV1
## Describe the bug

`Verifier.startMonitor()` supports both synchronous and async watch APIs:

```ts
this.ledgerApi.watchBlocksV1
  ? this.ledgerApi.watchBlocksV1(options)
  : await this.ledgerApi.watchBlocksAsyncV1?.(options)
```

However, `Verifier.stopMonitor()` only checks `watchBlocksV1` before unsubscribing:

```ts
if (!this.ledgerApi.watchBlocksV1) {
  throw new Error("stopMonitor not supported on this ledger");
}
```

As a result, a connector that only implements `watchBlocksAsyncV1` can start a monitor but cannot stop it through `stopMonitor()`.

## To Reproduce

1. Use a ledger API client that implements `watchBlocksAsyncV1` but not `watchBlocksV1`.
2. Call `Verifier.startMonitor(appId, options, listener)`.
3. After the monitor starts, call `Verifier.stopMonitor(appId)`.

## Expected behavior

`stopMonitor()` should unsubscribe and remove the monitor if a subscription exists, regardless of whether the connector uses `watchBlocksV1` or `watchBlocksAsyncV1`.

## Actual behavior

`stopMonitor()` throws `stopMonitor not supported on this ledger` before checking the stored subscription.

## Impact

Applications using async-only watch connectors can leak active subscriptions and cannot cleanly stop monitors through the verifier API.

## Suggested fix

Remove the `watchBlocksV1`-only guard in `stopMonitor()` and rely on the presence of the stored subscription for `appId`, or update the guard to account for both watch APIs.

ArshLabs:
stopMonitor() contained a guard that checked for the presence of `watchBlocksV1` and threw `stopMonitor not supported on this ledger` before inspecting the stored subscription. 
connectors that only implement `watchBlocksAsyncV1` can start a monitor via `startMonitor()` (which selects the async path correctly) but cannot stop it, leaking active subscriptions.

this removes the API-specific guard in `stopMonitor()` and relies on the stored subscription for `appId` instead.

Fixes #4250

### DIFF
--- a/packages/cactus-verifier-client/src/main/typescript/verifier.ts
+++ b/packages/cactus-verifier-client/src/main/typescript/verifier.ts
@@ -133,10 +133,6 @@ export class Verifier<LedgerApiType extends ISocketApiClient<unknown>>
    * @param appId - ID of application that requested the monitoring.
    */
   stopMonitor(appId: string): void {
-    if (!this.ledgerApi.watchBlocksV1) {
-      throw new Error("stopMonitor not supported on this ledger");
-    }
-
     const watchBlocksSub = this.runningMonitors.get(appId);
     if (!watchBlocksSub) {
       throw new Error("No monitor running with appId: " + appId);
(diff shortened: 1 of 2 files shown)


RafaelAPB:
closing as there is no triage ready issue for this

PR #4297, opened 2026-04-30
fix(verifier): propagate startMonitor startup errors to callers

### LINKED ISSUE #4251: bug(verifier): startMonitor resolves successfully after watchBlocks startup failure
## Describe the bug

`Verifier.startMonitor()` catches errors thrown while creating/subscribing to the watch blocks observable, logs them, deletes the monitor entry, but does not rethrow or reject:

```ts
} catch (err) {
  this.log.error(`##Error: startMonitor, ${err}`);
  this.runningMonitors.delete(appId);
}
```

Because the method returns `Promise<void>`, callers can receive a resolved promise even though monitoring failed to start.

## To Reproduce

1. Use a ledger API client whose `watchBlocksV1()` or `watchBlocksAsyncV1()` throws, rejects, or returns no observable.
2. Call `await verifier.startMonitor(appId, options, listener)`.
3. Observe that the call resolves after logging the error instead of rejecting.

## Expected behavior

`startMonitor()` should reject when the monitor cannot be started, so callers can handle the failure and avoid assuming monitoring is active.

## Actual behavior

`startMonitor()` swallows the startup error and resolves successfully after logging.

## Impact

Applications can believe block/event monitoring is active when it failed to start. This can make monitoring failures silent unless logs are inspected.

## Suggested fix

Rethrow the caught error after cleanup, or return a status/result that explicitly tells callers the monitor was not started.

ArshLabs:
`Verifier.startMonitor()` catches errors thrown by `watchBlocksV1()` or `watchBlocksAsyncV1()` during subscription setup, logs them, cleans up `runningMonitors`, but never rethrows. Because the method returns `Promise<void>`, the returned promise resolves successfully even when monitoring failed to start.

Callers that `await startMonitor()` receive no indication that monitoring is inactive. Applications can silently believe block monitoring is running when it is not, which is detectable only by inspecting logs.

## Root cause

```ts
} catch (err) {
  this.log.error(`##Error: startMonitor, ${err}`);
  this.runningMonitors.delete(appId);
  // error swallowed here - promise resolves
}
```

`packages/cactus-verifier-client/src/main/typescript/verifier.ts`, lines 124-127.

## Fix

Add `throw err;` after the cleanup so the error propagates to the caller:

```ts
} catch (err) {
  this.log.error(`##Error: startMonitor, ${err}`);
  this.runningMonitors.delete(appId);
  throw err;
}
```

The cleanup (deleting from `runningMonitors`) still happens before rethrowing, so the map stays consistent.

## Test

The existing test `In case of ApiClient exception runningMonitors is not updated` was asserting the broken behaviour (promise resolves on error). Updated it to:

- make the test `async` so the `await` actually waits for the promise
- assert `toReject()` instead of `toResolve()` to match the corrected behaviour
- rename to `In case of ApiClient exception startMonitor rejects and runningMonitors is not updated`

Fixes #4251

### DIFF
--- a/packages/cactus-verifier-client/src/main/typescript/verifier.ts
+++ b/packages/cactus-verifier-client/src/main/typescript/verifier.ts
@@ -124,6 +124,7 @@ export class Verifier<LedgerApiType extends ISocketApiClient<unknown>>
     } catch (err) {
       this.log.error(`##Error: startMonitor, ${err}`);
       this.runningMonitors.delete(appId);
+      throw err;
     }
   }
 
--- a/packages/cactus-verifier-client/src/test/typescript/unit/verifier.test.ts
+++ b/packages/cactus-verifier-client/src/test/typescript/unit/verifier.test.ts
@@ -99,7 +99,7 @@ describe("Monitoring Tests", () => {
     expect(sut.startMonitor("someId", {}, eventListenerMock)).toReject();
   });
 
-  test("In case of ApiClient exception runningMonitors is not updated", () => {
+  test("In case of ApiClient exception startMonitor rejects and runningMonitors is not updated", async () => {
     apiClientMock = new MockApiClient();
     apiClientMock.watchBlocksV1.mockImplementation(() => {
       throw Error("Some mock error in watchBlocks");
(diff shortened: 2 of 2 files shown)


RafaelAPB:
Closing as there is no approved issue for this

Issue #4269, opened 2026-04-19
## Bug: Inverted Undefined Check in Stage3 Server Service Causes Verification Failure.


oindrilakha12-ui:
### Describe the bug

In `stage3-server-service.ts`, the `checkCommitPreparationRequest` method contains an inverted conditional check. The code invokes `commonBodyVerifier` when `request.common` is **undefined** instead of when it is **defined**, causing verification to fail for valid requests.

**Affected File:**
`packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-services/server/stage3-server-service.ts`
**Lines:** 457–464

---

### Steps to Reproduce

1. Initiate a SATP asset transfer reaching Stage 3 (Commit Preparation)
2. Send a valid `CommitPreparationRequest` with a populated `request.common` field
3. Observe that the request returns a verification error

---

### Expected Behavior

When `request.common` is provided, `commonBodyVerifier` should validate the request. If it is missing, an error should be thrown early.

---

### Actual Behavior

```ts id="9i4yhs"
if (request.common == undefined) {
  commonBodyVerifier(
    fnTag,
    request.common,
    sessionData,
    MessageType.COMMIT_PREPARE,
  );
}
```

This invokes verification only when `request.common` is undefined:

* Invalid requests trigger verification with `undefined`
* Valid requests skip verification entirely

---

### Impact

* Valid requests are not properly verified
* `COMMIT_PREPARE` stage fails for legitimate inputs
* SATP flows may not complete as expected

---

### Suggested Fix

```ts id="zsj9ds"
if (request.common == undefined) {
  throw new Error(`${fnTag} request.common is undefined`);
}

commonBodyVerifier(
  fnTag,
  request.common,
  sessionData,
  MessageType.COMMIT_PREPARE,
);
```

---

### Additional Context

Similar inverted checks may exist in:

* `checkCommitFinalAssertionRequest` (lines 516–523)
* `checkTransferCompleteRequest` (lines 592–599)

These should be reviewed for consistency.

oindrilakha12-ui:
I’d like to work on this issue.

ShauryaChauhan1411:
I have submitted a PR to resolve this issue. Although the literal inverted logic was not present in the current version of the code, the underlying bug existed: the common field was being passed to the verifier without ensuring it was defined first.

[...]

PR #4307, opened 2026-05-02
fix(satp): ensure request.common is defined in stage3 server service

ShauryaChauhan1411:
This PR addresses issue #4269 by implementing required validation checks for the common field in several request types within the Stage 3 Server Service.

Root Cause Analysis
The original bug report indicated an inverted logic check; however, the current version of the code lacked any explicit undefined check before passing request.common to the commonBodyVerifier. This could lead to runtime errors or silent verification failures if a request is received with a missing common block.

Changes Made
Modified packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-services/server/stage3-server-service.ts to include explicit undefined checks and throw descriptive errors in the following methods:

checkCommitPreparationRequest: Added check to ensure request.common is defined before calling the verifier.

checkCommitFinalAssertionRequest: Added check to ensure request.common is defined before calling the verifier.

checkTransferCompleteRequest: Added check to ensure request.common is defined before calling the verifier.

Impact
Stability: Prevents the gateway from processing malformed requests that lack essential common metadata.

Error Reporting: Provides clear, logged error messages when a request fails validation, making debugging easier for SATP flows.

Consistency: Aligns the Stage 3 server logic with the project's expected validation standards.

### DIFF
--- a/packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-services/server/stage3-server-service.ts
+++ b/packages/cactus-plugin-satp-hermes/src/main/typescript/core/stage-services/server/stage3-server-service.ts
@@ -454,6 +454,10 @@ export class Stage3ServerService extends SATPService {
 
         const sessionData = session.getServerSessionData();
 
+        if (request.common == undefined) {
+          throw new Error(`${fnTag} request.common is undefined`);
+        }
+
         commonBodyVerifier(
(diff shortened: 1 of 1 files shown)


RafaelAPB:
Closing as there is no approved issue for this. More, the request should be verified in a common function (eg body checker)

PR #4318, opened 2026-05-05
fix(satp-hermes): correct early return behavior in config validation

### LINKED ISSUE #4317: Validation incorrectly passes for invalid nested array elements due to forEach early return
There is an issue in `checkConfigElementFormat` where validation of nested array elements may incorrectly pass.

The current implementation uses `Array.forEach` with `return false` inside the callback. However, `return` inside a `forEach` callback does not exit the parent function. As a result, even if an invalid element is encountered, the outer validation function may still return `true`.

## Example

```ts
names: ["alice", 123, "charlie"]
````

Expected: validation should return `false`
Actual: validation may return `true`

## Root Cause

`return false` inside `forEach` only exits the callback, not the enclosing function, preventing early termination of validation.

## Proposed Fix

Replace `forEach` with a `for...of` loop to allow early return when an invalid element is found.

## Impact

* Invalid configurations can pass validation silently
* Affects correctness of config validation logic

### PR DESCRIPTION
**Pull request by ZainabTravadi**
## Summary

Fixes an issue in `checkConfigElementFormat` where validation of nested array elements could incorrectly pass.

## Problem

The implementation previously used `Array.forEach()` with `return false` inside the callback.

However, returning from a `forEach` callback does not exit the parent function, which could allow invalid nested array elements to pass validation silently.

## Solution

Replaced the affected `forEach` loops with `for...of` loops so validation failures correctly return from the parent function immediately.

## Tests

Added concise regression coverage validating:

* one valid nested-array case
* one invalid nested-array case exercising the early-return bug path

## Impact

* Improves correctness of config validation
* No API changes
* No new dependencies
* Minimal isolated fix

Fixes #4317

### CODE REVIEW

File: packages/cactus-plugin-satp-hermes/src/main/typescript/services/utils.ts [hunk 1]
```diff
@@ -161,7 +161,10 @@ export function checkConfigElementFormat<T>(
       return false;
     } else {
       if (ccElement.configSubElementType) {
-        obj[ccElement.configElement].forEach((subEl: unknown) => {
+        // Use for...of instead of forEach to enable early return on validation failure.
```
RafaelAPB: Please delete this comment and place it in the commit message / PR description instead

packages/cactus-plugin-satp-hermes/src/main/typescript/services/utils.ts
```diff
@@ -171,9 +174,10 @@ export function checkConfigElementFormat<T>(
             );
             return false;
           }
-        });
+        }
       } else if (ccElement.configSubElementFunctionTypeguard) {
-        obj[ccElement.configElement].forEach((subEl: unknown) => {
+        // Use for...of instead of forEach to enable early return on validation failure.
```
RafaelAPB: Please delete this comment and place it in the commit message / PR description instead

packages/cactus-plugin-satp-hermes/src/test/typescript/unit/config-validating-functions/utils.test.ts
```diff
@@ -0,0 +1,443 @@
...
+        configElementType: "object",
+        configSubElementType: "string",
+      };
+
+      const result = checkConfigElementFormat(element, config, log, "test");
+      expect(result).toBe(false);
+    });
+
+    it("REGRESSION: should return false for array with invalid element type (last element wrong)", () => {
+      // Before fix: this would return true (bug)
```
RafaelAPB: please remove

packages/cactus-plugin-satp-hermes/src/test/typescript/unit/config-validating-functions/utils.test.ts
```diff
@@ -0,0 +1,443 @@
...
+        configElementType: "object",
+        configSubElementType: "string",
+      };
+
+      const result = checkConfigElementFormat(element, config, log, "test");
+      expect(result).toBe(true);
+    });
+
+    it("REGRESSION: should return false for array with invalid element type (first element wrong)", () => {
+      // Before fix: this would return true (bug)
```
RafaelAPB: Please remove the generated comments.

RafaelAPB (changes requested):
Thank you for your contribution.
Generally looks good; can you please make testing more succint - one positive and one negative cases, that asserts that the `for...of` refactor allow  early validation returns?

RafaelAPB:
@ZainabTravadi it seems you have done a few pushes. Have you incorporated the feedback? If yes, how?
Thank you again for your contribution!

ZainabTravadi:
Hi @RafaelAPB, yes — I incorporated all requested feedback in the latest force-push.

Changes made:

* Reduced the test file to only two concise regression cases:

  * one positive validation case
  * one negative case exercising the early-return bug path
* Removed the inline explanatory comments from both `utils.ts` and the regression tests
* Kept the implementation scoped strictly to the `forEach` → `for...of` early-return fix without unrelated test coverage

The current test file is now focused specifically on validating the regression scenario discussed in the review.

Thank you again for the feedback!

RafaelAPB:
@ZainabTravadi are you still working on this?

RafaelAPB:
closing per inactivity. @ZainabTravadi please check the ci jobs that are failing. if you have availability, please let us know so we can merge it.

PR #4409, opened 2026-06-15
fix(weaver): zero-pad ECDH shared secret in eciesEncryptMessage

mskrzypkows:
The ECDH shared secret Z was passed to the KDF without zero-padding in eciesEncryptMessage, while eciesDecryptMessage always padded it to the full field size. When a random ephemeral key
produced a shared secret with leading zero bytes, the two sides derived different AES keys, causing decryption to silently produce garbage.

Fix by applying the same zero-padding in the encrypt path, and replace the hardcoded 32 in the decrypt path with Math.ceil(level / 8) to correctly handle both P-256 and P-384 curves.

Regression test added with a pre-found ephemeral key that deterministically
produces a 31-byte shared secret with the test certificate's public key.

Assisted-by: anthropic:claude-sonnet-4.6


sandeepnRES:
@mskrzypkows Please update your PR and check if this is already implemented now?

mskrzypkows:
Yes, it's implemented, we can close it or if we need additional test then it can be merged.

sandeepnRES (approved):

sandeepnRES:
Closing this PR as it has merge commits and 11 commits. Please @mskrzypkows  open new PR if you want to merge the test, I'm good to add the unit test.

PR #4197, opened 2026-04-07
docs: update outdated pre-merger references in contributor documentation

abhigyan1102:
## Description

Audits and updates outdated pre-merger references in contributor-facing documentation.

This is part of the **Cacti Cleanup Initiative** (goal #4: Improve documentation).

### Changes (5 files, 30 lines)

- **`CONTRIBUTING.md`** — Fixed fork URL (`hyperledger/cacti` → `hyperledger-cacti/cacti`), upstream remote URL, and Code of Conduct link
- **`docs/docs/cactus/contributing.md`** — Updated 17 outdated references: repo URLs, issue template links, fork/clone instructions, and project naming (Cactus → Cacti)
- **`docs/docs/cactus/build.md`** — Fixed clone URL, directory name, and 11 Cactus → Cacti naming references
- **`CODE_OF_CONDUCT.md`** & **`docs/docs/cactus/code-of-conduct.md`** — Updated Code of Conduct links from defunct `wiki.hyperledger.org` to current LF Decentralized Trust governance docs

### Audit Summary

Found **689 references** to the old `hyperledger/cactus` repository URL across markdown files. This PR addresses the **contributor-facing subset** — the files new contributors read first. Additional cleanup PRs can follow for package-level and Weaver docs.

### Not Changed (intentionally)
- `@hyperledger/cactus-*` npm package names (still valid published package names)
- `CHANGELOG.md` (historical record)
- Blog posts (historical context)

### Related
- Cacti Cleanup Initiative: https://github.com/orgs/hyperledger-cacti/projects/2
- Mentorship Issue: LF-Decentralized-Trust-Mentorships/mentorship-program#62

PR #4467, opened 2026-06-25
docs: update root references and remove archived docs

ParthSinghPS:
This PR supersedes and replaces #4197.

As discussed in the original PR's review by @sandeepnRES and @RafaelAPB, updating the documentation inside docs/docs/cactus/ is unnecessary as it is archived and outdated.

To resolve this cleanly, I have:
1. Retained @abhigyan1102's URL and naming fixes for the root level CONTRIBUTING.md and CODE_OF_CONDUCT.md.
2. Completely removed the deprecated docs/docs/cactus/ directory to prevent future contributor confusion.

The original author has been credited via the Co-authored-by tag for their fixes.

Fixes #4197


sandeepnRES (approved):

sandeepnRES (changes requested):
After looking at this PR: https://github.com/hyperledger-cacti/cacti/pull/4332/changes
It seems there are references at other places to archived docs.

Would suggest to resolve these before deleting them. Probably moving these commonly referred files into different appropriately named directories, and update references. When nothing else points to the `docs/docs/cactus` dir being deleted, then it would be fine to delete.

ParthSinghPS:
that is right, i'll work on that and open a separate PR for the links migration soon

AndreAugusto11:
> that is right, i'll work on that and open a separate PR for the links migration soon

@ParthSinghPS does that mean that we can close this one? Or is it still needed?

ParthSinghPS:
I think it can be closed, as this is decided to be done in a more phased approach

PR #4542, opened 2026-07-06
ci(weaver): fix actionlint warnings and normalize disabled-job guards

ryjones:
Address the higher-signal actionlint findings in the weaver workflows (which the actionlint CI job currently excludes via the `! -name "*weaver*"` filter). Quoting-only findings (SC2086/SC2046/SC2206) are intentionally left for a separate pass.

- SC2193 (7) + SC2001 (7): rewrite the "Update version" step to derive the tag from `$GITHUB_REF` and use parameter expansion instead of piping through `sed`. Using the env var (not the `${{ github.ref }}` expression) is also the recommended, injection-safe form and is what silenced the shellcheck false-positive that flagged the comparison as never-equal. VERSION="${GITHUB_REF##*/}" [[ "$GITHUB_REF" == refs/tags/* ]] && VERSION="${VERSION#v}"
- SC2001 (2 more): same `sed -e 's/^v//'` -> `${VERSION#v}` in test_weaver-pre-release.yaml.
- SC2129 (14): group consecutive `echo ... >> github[.main].properties` redirects into a single `{ ... } >> file` block.
- if-cond: fix test_weaver-go.yaml `unit_test_sdk` — its `if:` mixed a bare expression with a `${{ }}`-wrapped one, which GitHub evaluates as a non-empty (always-true) string. Wrap the whole condition so the three-way OR gate actually applies, matching the sibling jobs.

Left as-is: the `if: ${{ false }}` on test_weaver-asset-exchange-besu.yaml's `asset-exchange-besu` job. actionlint suggests removing the `if:`, but that job was deliberately disabled in Jan 2024 and removing the guard would re-enable it; that is a behavior change, not a lint fix.

Verified with shellcheck-enabled actionlint (Docker): the 31 targeted findings are gone and no new findings were introduced.

Assisted-by: anthropic:claude-opus-4-8 actionlint

**A Must Read for Beginners**
Please read [the contribution guides](..CONTRIBUTING.md) prior to your first contribution.

**Pull Request Requirements**
- [x] Read and followed the [Pull Request Guidelines](../PULL.md).
- [x] If AI tools were used, include an `Assisted-by` tag in the commit message per the [AI Guidelines §2.2](./AI_GUIDELINES.md#22-disclosure) (e.g., `Assisted-by: GitHub-Copilot:claude-opus-4`). All AI-generated code must be human-reviewed before submission.

### DIFF
--- a/.github/workflows/test_weaver-asset-exchange-besu.yaml
+++ b/.github/workflows/test_weaver-asset-exchange-besu.yaml
@@ -36,7 +36,7 @@ jobs:
 
   asset-exchange-besu:
     needs: check_code_changed
-    if: ${{ false }}
+    if: ${{ vars.ENABLE_PACKAGES_CI == 'true' }}
     # The type of runner that the job will run on
     runs-on: ubuntu-22.04
 
--- a/.github/workflows/test_weaver-asset-exchange-corda.yaml
+++ b/.github/workflows/test_weaver-asset-exchange-corda.yaml
@@ -38,7 +38,7 @@ jobs:
 
   asset-exchange-corda:
     needs: check_code_changed
-    if: ${{ false && needs.check_code_changed.outputs.status == 'true' }}
+    if: ${{ vars.ENABLE_PACKAGES_CI == 'true' && needs.check_code_changed.outputs.status == 'true' }}
     # The type of runner that the job will run on
     runs-on: ubuntu-22.04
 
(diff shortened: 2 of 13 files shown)


sandeepnRES (approved):
Looks good

sandeepnRES:
@ryjones about this

> Left as-is: the if: ${{ false }} on test_weaver-asset-exchange-besu.yaml's asset-exchange-besu job. actionlint suggests removing the if:, but that job was deliberately disabled in Jan 2024 and removing the guard would re-enable it; that is a behavior change, not a lint fix.

This might be present at multiple jobs in weaver CI, these were disabled because we couldn't test and fix if these were working with cacti published packages after cactus-weaver merger.
If we completely want to remove warnings by actionlint,
We can do so by replacing:
```
if: ${{ false }}
```
with
```
if: ${{ vars.ENABLE_WEAVER_PACKAGES_CI == 'true' }}
```
since `ENABLE_WEAVER_PACKAGES_CI` is not declared in github settings, it will always evaluate to false, and once these jobs are fixed and tested, we can actually declare the variable and enable them.
Let me know if it sounds good.

ryjones:
@sandeepnRES I think this is updated now?

sandeepnRES:
> @sandeepnRES I think this is updated now?

It is there in some other places too, which has multiple conditions, not sure if actionlint picks them up:
`test_weaver-asset-exchange-corda.yaml`: Line 41 and 234
`test_weaver-asset-exchange-fabric.yaml`: Line 44
`test_weaver-asset-transfer.yaml`: Line 44
`test_weaver-data-sharing.yaml`: Line 44
`test_weaver-docker-build`: Line 111, 160
`test_weaver-relay.yaml`: Line 144, 184

sandeepnRES:
Looks good now. Thank you @ryjones

ryjones:
OK @sandeepnRES I've rebased on main

RafaelAPB (approved):

Issue #4472, opened 2026-06-26
ci(cpl-connector-fabric): Jest tests being timed out


sandeepnRES:
**Describe the bug**

The `cpl-connector-fabric-v2-2-x` job in the connector-packages workflow
consistently hits the 60-minute job timeout at the Jest step. Tests appear
to complete but the job never exits cleanly.

**Expected behavior**

Either tests should succeed or fail before timing out.

**Logs**

A log from `cpl-connector-fabric-v2-2-x`: [logs.txt](https://github.com/user-attachments/files/29382441/logs.txt)

This is the log from v3.0.0-beta.1 release commit CI run: https://github.com/hyperledger-cacti/cacti/actions/runs/28219046808/job/83602585179

PR #4682, opened 2026-08-14
fix(connector-fabric): send full/private blocks as JSON not binary

ryjones:
## Description

Step one: shard Fabric CI so it fails faster
Step two: fix the serialization issue

## Related issue

Refs #4472

## PR author checklist

- [x] I read and followed the [Pull Request Guidelines](../PULL.md), including linking a `Triage_Ready` issue (§9).
- [x] If AI tools were used, the commit includes an `Assisted-by` tag per [AI Guidelines §2.2](../AI_GUIDELINES.md#22-disclosure). All AI-generated code has been human-reviewed before submission.

### DIFF
--- a/.github/workflows/connector-packages-workflow.yaml
+++ b/.github/workflows/connector-packages-workflow.yaml
@@ -239,17 +239,44 @@ jobs:
           path: ./code-coverage-ts/**/
 
   cpl-connector-fabric-v2-2-x:
-    # Heavy suite: Fabric v2.2.x AIO integration tests. Runs on its own VM
-    # so each AIO container start is the first on the host (no port/state
-    # leakage from prior Jest files). The three deploy-cc-from-* tests are
-    # skipped via describe.skip pending FabricTestLedgerV1 teardown fixes.
-    # TODO: Set continue-on-error back to false once FabricTestLedgerV1
-    # teardown is fixed (see cpl-connector-fabric note above).
+    # Heavy suite: Fabric v2.2.x AIO integration tests, one shard per test file.
+    #
+    # Previously this ran all 13 files in a single job under `maxWorkers: 1`,
+    # and never finished: it completed 2 suites, then spent the rest of the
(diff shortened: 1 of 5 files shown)


sandeepnRES (approved):

.github/workflows/connector-packages-workflow.yaml
```diff
@@ -239,17 +239,44 @@ jobs:
...
+          - deploy-cc-from-typescript-source
+          - fabric-watch-blocks-delegated-sign-v1-endpoint
+          - fabric-watch-blocks-v1-endpoint
+          - obtain-profiles
+          - run-transaction-endpoint-v1
+          - run-transaction-with-identities
+          - run-transaction-with-ws-ids
     env:
-      JEST_TEST_PATTERN: packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/.*\.test\.ts
+      JEST_TEST_PATTERN: packages/cactus-plugin-ledger-connector-fabric/src/test/typescript/integration/fabric-v2-2-x/${{ matrix.suite }}\.test\.ts
```
RafaelAPB, line 279: why are we executing only the hardcoded tests above (instead of all tests?

RafaelAPB: @sandeepnRES @ryjones , does look good to me but I want to undersatnd why we are moving away from the used pattern

sandeepnRES: @RafaelAPB Currently this job used to timeout after an hour (or 2). I'm guessing Ry's fix will now split the job into multiple jobs, and fix the issue or atleast catch the culprit.
But your confirmation is required that its not missing any test-case. Please check.

ryjones: correct. by splitting into a matrix with a faster fails, we avoid the timeouts and capture data on which tests are working that usually are not run due to timeouts.

RafaelAPB (approved):

PR #4713, opened 2026-09-01
ci: fix documentation publishing

RafaelAPB:
Publishes for each commit to main, all end-user
documentation (root level docs, package docs, and
mkdocs)

## PR author checklist

- [X] I read and followed the [Pull Request Guidelines](../PULL.md), including linking a `Triage_Ready` issue (§9).
- [X] If AI tools were used, the commit includes an `Assisted-by` tag per [AI Guidelines §2.2](../AI_GUIDELINES.md#22-disclosure). All AI-generated code has been human-reviewed before submission.

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -6,13 +6,16 @@ name: Deploy Docs (Github Pages)
 
 on:
   push:
+    branches:
+      - main
+    paths:
+      - '*.md'
+      - 'docs/**'
+      - 'packages/**/*.md'
+      - 'weaver/**/*.md'
     tags:
       # only tags with stable versions a.b.c i.e. without -alpha or -dev or -beta etc
       - v[0-9]+.[0-9]+.[0-9]+
-
-    paths:
-      - 'docs/**'
-      - 'packages/cactus-plugin-satp-hermes/docs/**'
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:


.github/workflows/deploy_docs.yml
```diff
@@ -6,14 +6,17 @@ name: Deploy Docs (Github Pages)
 
 on:
   push:
+    branches:
+      - main
+    paths:
+      - '*.md'
+      - 'docs/**'
+      - 'packages/**/*.md'
+      - 'weaver/**/*.md'
```
sandeepnRES, line 15: I want to ask: why any markdown change in whole repo should trigger this workflow?
My understanding is, all docs related markdown files are under `docs` directory, and it publishes from there only. Please let me know if markdown files across repo also now contribute to the final render of docs.

.github/workflows/deploy_docs.yml
```diff
@@ -6,14 +6,17 @@ name: Deploy Docs (Github Pages)
 
 on:
   push:
+    branches:
+      - main
```
sandeepnRES, line 10: Since we are publishing from `main` branch for every commit, the packages published might be outdated w.r.t. docs (that's why I made the change to publish the docs together with a new release only).

One thing I'd want to propose: shall we maintain version for docs, and keep a `dev` tag for these publish triggers for each commit from main? I'd created similar thing in other project using mkdocs, I can work on it here too.

.github/workflows/deploy_docs.yml
RafaelAPB: Not any markdown change in the whole repo; any markdown in the root (README, BUID, CHANGELOG, etc). Furthermore, one of the ideas would be to publish all package READMEs to mkdocs as well. Does that make sense?

.github/workflows/deploy_docs.yml
RafaelAPB: Agreed!

.github/workflows/deploy_docs.yml
sandeepnRES: Okay yeah if we want to publish all package READMEs to mkdocs, then this makes sense. But is it feasible without duplicating READMEs or creating symlinks?

.github/workflows/deploy_docs.yml
sandeepnRES: Let me re-do this, I'll create another PR after testing from my fork, is that okay?

RafaelAPB: Sure thing, I will close this one then.

.github/workflows/deploy_docs.yml
RafaelAPB: Ideally symlinks, if not possible then we will have to leave a reference instead

